PR #35, opened 2026-04-16
fix(mem): fix various issues in mem

krokoko:
- Fix critical design flaw in memory integrity hashing: The original implementation hashed raw content on write but verified against sanitized content on read — hashes would never
  match for records containing sanitizable content. Additionally, AgentCore's extraction pipeline transforms content via LLM summarization/consolidation, meaning extracted records will
  always differ from write-time content. Changed integrity verification from fail-closed (discard records) to audit-only (log at WARN, keep records). Read-path sanitization
  (sanitizeExternalContent) is the real defense against content tampering.
  - Harden memory pipeline with sanitization, provenance, and cross-language parity: Extract shared sanitization.py module (Python mirror of TS sanitizeExternalContent), add
  MemorySourceType provenance tagging on all memory writes, sanitize taskDescription before prompt injection in context hydration, add severity-aware error handling (programming errors
  re-thrown/logged at ERROR, infra failures fail-open at WARN), and add cross-language SHA-256 test fixtures.
  - Improve test coverage: Add tests for semantic + episodic hash mismatch paths, v3-missing-hash detection, backward compatibility (v2 records), hash-matches-sanitized-content
  correctness, negative-log assertions, cross-language hash parity (shared JSON fixture consumed by both Jest and pytest), and context hydration sanitization.

#### Area

#### Acknowledgment


leandrodamascena:
I did some analysis with the automated review agents and I think this closes the loop on the hash concern you raised @krokoko. Hashing sanitized content fix this and also the shared fixture for cross-language parity is something I wanted to add but didn't get to. 

I'm pulling it down to test now in my local.

PR #39, opened 2026-04-17
docs(governance-receipts): add guide for signed decision receipts on ABCA runs

tomjwxf:
## Summary

Adds `docs/guides/GOVERNANCE_RECEIPTS.md` showing how to wire
cryptographically signed decision receipts into the ABCA agent pipeline.
Each autonomous agent run then produces a tamper-evident, offline-verifiable
chain of receipts that PR reviewers and regulators can verify without
trusting the orchestrator.

No changes to ABCA code. Purely additive documentation.

## Motivation

ABCA agents run autonomously in isolated cloud environments, clone repos,
write code, run tests, and open PRs. The default evidence trail lives in
CloudWatch logs, which are operator-controlled. For regulated deployments
(finance, healthcare, public sector) teams need a stronger evidence model:
decisions signed at the moment they are made, with a tamper-evident chain
across the full run.

The pattern in the guide is already used by:

- **Microsoft Agent Governance Toolkit** ([PR #667 merged](https://github.com/microsoft/agent-governance-toolkit/pull/667), [`examples/protect-mcp-governed/`](https://github.com/microsoft/agent-governance-toolkit/tree/main/examples/protect-mcp-governed))
- **AWS Cedar for Agents** ([PR #64 merged](https://github.com/cedar-policy/cedar-for-agents/pull/64), WASM bindings for Cedar policy evaluation in JS/TS/Python agents)
- The [protect-mcp npm package](https://www.npmjs.com/package/protect-mcp) (10K+ monthly downloads)

## Why this fits ABCA specifically

1. **Cedar is AWS's own open policy language.** Used in Amazon Verified
   Permissions, IAM Access Analyzer, and the CDK policy constructs. ABCA
   teams adopting Cedar for agent authorization reuse tooling they already
   know.
2. **ABCA produces PRs.** The guide shows archiving receipts to S3 and
   referencing them from the PR body so reviewers see the verification
   instructions inline.
3. **The `agent/` Dockerfile is the natural install point.** Adding Node.js
   18 (for `npx`) is one line, and the `.claude/settings.json` hook
   configuration is a drop-in if the agent uses Claude Code.
4. **No code changes.** The pattern uses the existing hook architecture in
   the Claude Code runtime ABCA agents typically embed. The guide is a
   recipe, not a new agent mode.

## What the guide covers

- Motivation (default vs signed evidence trail)
- Architecture diagram showing PreToolUse + PostToolUse hooks in the agent
- Setup steps: Dockerfile line for Node.js, Cedar policy example pinned to
  `/workspace/`, `.claude/settings.json` hook block, Python snippet for
  archiving receipts to S3 and referencing from the PR body
- Receipt format with the three cryptographic invariants (JCS, Ed25519,
  hash chain)
- Why Cedar specifically for ABCA (AWS-native fit)
- SLSA provenance composition via ResourceDescriptor byproducts (tracking
  the [agent-commit build type](https://refs.arewm.com/agent-commit/v0.2))
- External-reviewer verification flow (no AWS account required)
- Cross-implementation interop table (four independent implementations of
  the receipt format)
- What the guide intentionally does not cover (policy authoring at scale,
  key management in Secrets Manager/KMS/CloudHSM, Rekor transparency-log
  anchoring)

## Testing

- [x] Guide is referenced only from this file; no existing docs or index
      needs updating
- [x] All external links resolve (IETF draft, RFCs, Cedar docs, npm
      packages, related GitHub issues and PRs)
- [x] Code snippets are syntactically valid (`python3 -m py_compile` on
      the Python block; `cedar validate` on the policy block would pass
      given the schema in `cedar-for-agents`)
- [x] No changes to ABCA source code, CDK, or CLI

## Related

- IETF: [`draft-farley-acta-signed-receipts`](https://datatracker.ietf.org/doc/draft-farley-acta-signed-receipts/)
- SLSA for agents discussion: [slsa-framework/slsa#1594](https://github.com/slsa-framework/slsa/issues/1594)
- In-toto predicate proposal: [in-toto/attestation#549](https://github.com/in-toto/attestation/pull/549)
- AWS Cedar-for-agents WASM bindings: [cedar-policy/cedar-for-agents#64](https://github.com/cedar-policy/cedar-for-agents/pull/64)

## Notes

- Author: Tom Farley (independent capacity; IETF draft author). Filed under
  the aws-samples CLA per CONTRIBUTING.md.
- The guide deliberately keeps the protect-mcp dependency optional: all hook
  commands use `npx protect-mcp@latest` so nothing needs pre-installation
  in the agent container beyond Node.js.
- Happy to iterate on scope, structure, or additional ABCA-specific details
  if reviewers want anything adjusted.

**Change type**: `docs(governance-receipts)` -- pure additive documentation
under `docs/guides/`.

### DIFF
--- a/docs/guides/GOVERNANCE_RECEIPTS.md
+++ b/docs/guides/GOVERNANCE_RECEIPTS.md
@@ -1,0 +1,338 @@
+# Signed Decision Receipts for ABCA Agent Runs
+
+**Add cryptographically verifiable evidence of every tool call an ABCA agent makes, so PR reviewers and regulators can verify the agent's behavior without trusting the orchestrator.**
+
+---
+
+## Why this matters for ABCA
+
+ABCA agents run autonomously in isolated cloud environments, clone repos,
+write code, run tests, and open pull requests. Each step is a tool call
+(`git clone`, `npm install`, `npm test`, shell commands, file edits). By
+default, the evidence that each step happened and was authorized lives in
+CloudWatch logs, which are operator-controlled and trust the orchestrator
+not to redact or edit entries after the fact.
+
+For regulated environments (financial services, healthcare, public sector)
+or for teams that want to hand a PR to a regulator or counterparty, a
+stronger evidence model is needed: each policy decision, signed at the
+moment it is made, with a tamper-evident chain across the full agent run.
+
+Signed decision receipts provide this. A receipt is a small JSON object,
+signed with Ed25519, JCS-canonicalized, and hash-chained to its
+predecessor. Anyone with the public key can verify the whole chain offline.
+
+## What this adds to the default ABCA pipeline
+
+The ABCA agent runs inside the Dockerfile in `agent/`. Three additions:
+
+1. **PreToolUse hook**: before every tool call, evaluate the call against a
+   Cedar policy. Cedar denial blocks the tool call.
+2. **PostToolUse hook**: after every tool call, sign an Ed25519 receipt
+   describing the decision, the policy digest, the inputs, and a hash link
+   to the previous receipt.
+3. **Receipt artifact**: the receipt directory (`/tmp/receipts/`) is
+   uploaded as a build artifact alongside the PR, and the PR body includes
+   a one-line verification instruction.
+
+Result: a reviewer or auditor running `npx @veritasacta/verify
+./receipts/*.json` gets an offline proof that every step of the agent run
+was authorized under the declared policy, with exit code 0 for valid, 1 for
+tampered, or 2 for malformed.
+
+## The pattern
+
+```
+┌──────────────────────────────────────────────────────────────────────────┐
+│                   ABCA Autonomous Agent Runtime                          │
+│                                                                          │
+│  ┌────────────────┐                                                      │
+│  │ Orchestrator   │                                                      │
+│  └───────┬────────┘                                                      │
+│          │ invokes                                                       │
+│          ▼                                                               │
+│  ┌────────────────┐    PreToolUse hook    ┌───────────────────────────┐  │
+│  │ Agent Runtime  │ ───────────────────▶  │ Cedar Policy Evaluator    │  │
+│  │ (Claude Code / │   (before each call)  │ ./protect.cedar           │  │
+│  │  Amazon Q /    │                       │ allow → continue          │  │
+│  │  similar)      │                       │ deny  → exit 2, block     │  │
+│  └───────┬────────┘                       └───────────────────────────┘  │
+│          │ post-execution                                                │
+│          ▼                                                               │
+│  ┌────────────────┐    PostToolUse hook    ┌──────────────────────────┐  │
+│  │ Tool output    │ ────────────────────▶  │ Ed25519 Receipt Signer   │  │
+│  └────────────────┘   (after each call)    │ JCS canonical + chain    │  │
+│                                            │ /tmp/receipts/*.json     │  │
+│                                            └──────────────────────────┘  │
+└──────────────────────────────────────────────────────────────────────────┘
+```
+
+## Setup
+
+### 1. Add protect-mcp to the agent container
+
+In `agent/Dockerfile`, ensure Node.js 18+ is installed (for `npx`):
+
+```dockerfile
+RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash - \
+    && apt-get install -y nodejs
+```
+
+### 2. Ship a Cedar policy with the agent image
+
+Create `agent/protect.cedar` with allow/deny rules appropriate to the agent's
+scope. Example for a coding agent that runs in an isolated environment:
+
+```cedar
+// Allow read-oriented tools on source files.
+permit (
+    principal,
+    action in [Action::"Read", Action::"Glob", Action::"Grep"],
+    resource
+);
+
+// Allow the build/test commands the agent needs.
+permit (
+    principal,
+    action == Action::"Bash",
+    resource
+) when {
+    context.command_pattern in [
+        "git", "npm", "pnpm", "yarn", "uv", "python",
+        "pytest", "cargo", "go", "make"
+    ]
+};
+
+// Deny destructive commands even in the isolated environment.
+forbid (
+    principal,
+    action == Action::"Bash",
+    resource
+) when {
+    context.command_pattern in ["rm -rf", "dd", "mkfs", "shred"]
+};
+
+// Writes only to the agent's working directory.
+permit (
+    principal,
+    action in [Action::"Write", Action::"Edit"],
+    resource
+) when {
+    context.path_starts_with == "/workspace/"
+};
+```
+
+Policy authoring tips:
+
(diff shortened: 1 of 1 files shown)


krokoko:
Hi @tomjwxf , thank you for this PR !

Some questions/notes:

- Just to confirm, you are the specific author of the different PRs and tools mentioned in this PR, correct ? So I think this guide is not a neutral documentation but an integration recipe for your specific toolchain packaged as a guidance. Just want to clarify
- I think the "no code changes" is misleading since it provides specific instructions/guidance to modify the code and create infra, so this looks more like an architectural proposal ?
- Correct me if I misunderstand anything here, but the signing key lives inside the agent container, same as the hooks. If the agent or container is compromised, then the attacker controls both the tool execution and the receipt signing. I think this gives a false sense of security to the audience it's targeting. The signer should be outside the operator boundary

I think this is a valid requirement for regulated environments though. I would prefer having an RFC/design doc proposing the governance model with the trust boundaries explicitly drawn and a tool agnostic design (like the receipt contract), CDK managed infra (S3 bucket, IAM policies, KMS,...), and from there see if the implementation can use the tools mentioned.

One option: we will be adding soon Amazon Bedrock AgentCore Gateway for the tool access layer (which is mentioned in the docs). That would enable decision/signing at the Gateway interceptor layer, something like:

```
Agent (inside MicroVM)
    │
    │  tool call request
    ▼
  AgentCore Gateway
    │
    ├─► REQUEST Interceptor Lambda ◄── Cedar policy evaluation + receipt signing
    │     • Evaluate Cedar policy
    │     • Sign a "decision" receipt (allow/deny) with a key the agent NEVER sees
    │     • Deny → return error, tool never executes
    │
    ├─► Tool Target (Lambda, OpenAPI, MCP server, etc.)
    │     • Actual execution
    │
    ├─► RESPONSE Interceptor Lambda ◄── Sign an "outcome" receipt
    │     • Hash the output, chain to the decision receipt
    │     • Redact secrets/PII before returning to agent
    │
    ▼
  Agent receives sanitized output
```

That way the key lives in the interceptor lambda so it's never exposed to the agent (the agent will never see not forge the key). The signing key would live in KMS or Secrets Manager, accessible only to the interceptor Lambda's IAM role (not the Runtime's role). The receipt chain gets written to S3 by the interceptor, not by the agent. The agent literally cannot interfere.

What do you think ?

Thank you !

tomjwxf:
@krokoko thank you, all four points land and the architectural critique is correct. Working through each:

**1. Yes, I am the author of the tools referenced (protect-mcp, @veritasacta/verify, the IETF draft). I should have made that transparency explicit in the PR body rather than only in the sign-off; that is on me.** You are right to frame this as an integration recipe for my own toolchain rather than neutral documentation. I will make that explicit in whatever form this eventually takes.

**2. "No code changes" was wrong framing.** The guide prescribes infrastructure changes (Dockerfile, settings, S3 layout) even if they sit outside `agent/src`. I should have called it what it is: an architecture proposal. Accepting that rename.

**3. The trust-boundary critique is the most important one and you are right.** As the guide is written, the signing key lives inside the agent container alongside the hooks. A container compromise compromises both the tool execution AND the receipts signed over it. That is exactly the kind of "false sense of security" a guide like this should not produce, and I should have drawn the trust boundary explicitly. This is a real weakness in the current draft, not a framing issue.

**4. An RFC / design doc first is the right sequencing.** Tool-agnostic design, receipt contract as a neutral interface, CDK-managed infra (S3 + IAM + KMS), trust boundaries drawn explicitly, and then implementations mapped into it. Let me accept this as the path forward.

**On the AgentCore Gateway model you proposed.** This is the right architecture. The interceptor Lambda has its own IAM role, the signing key lives in KMS scoped to that role (not the Runtime role), receipts write to S3 from the interceptor (not the agent), and the agent literally cannot interfere with the signing path. Two additional properties that fall out of your design:

- **The REQUEST interceptor signs the decision receipt before the tool runs.** That makes the receipt the authority for the decision, not a post-hoc recording of it.
- **The RESPONSE interceptor signs the outcome receipt after the tool runs.** Chaining decision → outcome gives you the full call boundary. If the output is redacted (secrets / PII) before return-to-agent, the outcome receipt signs the redacted form the agent saw, which is a property the in-container pattern cannot provide.

I would like to reframe this contribution around your architecture specifically. Two options for the mechanics, your call:

**Option A: Close this PR and open a fresh RFC**. New branch, new PR titled something like `rfc(governance): decision-receipt contract for ABCA + AgentCore Gateway interceptor pattern`. Structure:

1. Problem statement (agent-produced artifacts need verifiable authorization evidence; CloudWatch is operator-trust)
2. Trust model and threat model (explicit boundaries around Runtime, Gateway, KMS, S3)
3. The receipt contract, tool-agnostic (fields, signing algorithm, chain semantics, NOT specific to protect-mcp)
4. CDK-managed infra (S3 bucket w/ bucket policy, IAM roles for interceptor Lambda vs Runtime, KMS key policy, CloudTrail for signing-key access audit)
5. Reference implementation path (AgentCore Gateway REQUEST + RESPONSE interceptors)
6. What compromises what (container compromise, Runtime role compromise, interceptor compromise, KMS access compromise)
7. Non-goals and open questions

**Option B: Transform this PR into the RFC** instead of opening fresh. Same structure, same content, just rebased from the guide into the design-doc form. Might keep the review history in one place.

I lean slightly toward **Option A** because the Git trail is cleaner and the reviewers who saw the guide version should not have to unwind their impression of a different artifact. But if you want the conversation to stay on #39, B works too.

**On the implementation-specific sections**. Once the RFC is approved, the receipt contract is tool-agnostic, and any implementation that produces conformant receipts can be dropped into the reference-implementation section. That includes my own (protect-mcp, sb-runtime) and explicitly excludes single-sourcing: happy to link to at least three independent implementations (protect-mcp in TypeScript, protect-mcp-adk in Python, sb-runtime in Rust) so the recipe is not vendor-locked.

**On timing**. I will draft the RFC over the next several days. Before I push, would it be useful to sync on the AgentCore Gateway details you have in mind (expected interceptor Lambda surface, KMS key-usage policy shape, S3 bucket layout)? Either a written doc or async here works for me; I do not want to get ahead of a service announcement if AgentCore Gateway is not yet public.

Thank you again for engaging substantively. This is exactly the kind of critique the guide needed; you have pulled it toward a materially better design.

Tom

krokoko:
Thank you Tom ! Yes "Option A: Close this PR and open a fresh RFC" works for me!

"Before I push, would it be useful to sync on the AgentCore Gateway details you have in mind (expected interceptor Lambda surface, KMS key-usage policy shape, S3 bucket layout)? Either a written doc or async here works for me; I do not want to get ahead of a service announcement if AgentCore Gateway is not yet public." -> the service is publicly available and you can get more details here: https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/policy-getting-started.html, we can discuss that in the RFC as well !

Thank you !

krokoko:
Closing this PR following the discussion and feel free to open the RFC, thanks !

PR #52, opened 2026-04-29
feat(interactive-agents): async-only background task UX + Cedar HITL design

scoropeza:
## Summary

Introduces **interactive background agents** — users can watch, steer, trace, and review agent runs through a unified async-only architecture. Ships five user-facing surfaces (`watch`, `nudge`, `status`, `trace`, webhook) plus the first real notification-plane dispatcher (GitHub edit-in-place). Includes the **locked design for Phase 3 Cedar-driven HITL approval gates** — no implementation; design only.

All shipped features are deployed to a dev stack and E2E-validated against `scoropeza/agent-plugins` PRs #32/#33/#43. Three runtime bugs were caught and fixed during deploy-validation — see [Deploy validation](#deploy-validation) below.

Branch state: ~82 commits. Will squash on merge.

### Polling / observability

- **`bgagent watch <task-id>`** — adaptive polling (500 ms active; 1 / 2 / 5 s idle backoff; resets on event arrival), cold-start retry wrapper, clean terminal exit. Streams turn-by-turn events from `TaskEventsTable` via `GET /v1/tasks/{id}/events?after=<cursor>`.
- **`bgagent status <task-id>`** — deterministic snapshot (AD-6): Type, Channel, Description, Turn, Last milestone, Current, Cost, Reason, Trace S3, Last event. `--wait` is a pure blocking flag; same layout in both modes.
- **`bgagent events <task-id>`** — full event log (paged).
- **`--verbose`** — HTTP debug output on stderr.

### Interactive steering

- **`bgagent nudge <task-id> "<message>"`** — mid-run steering via `POST /v1/tasks/{id}/nudges` → `TaskNudgesTable` → agent's between-turns hook. Combined-turn acknowledgement per **AD-5** — the agent emits a `nudge_acknowledged` milestone before incorporating the nudge, giving users a visible handshake without the complexity of mid-turn interrupt.
- **`bgagent cancel <task-id>`** — terminates the running agent, persists `task_cancelled`, prevents PR push.

### Trace / debug

- **`--trace`** on submit — raises progress-writer preview cap from 200 chars to 4 KB and uploads a full gzipped NDJSON trajectory to S3 on terminal.
- **`bgagent trace download <task-id>`** — streams gzipped NDJSON to stdout (pipe-friendly: `| gunzip | jq -s .`) or writes raw gzip to `-o <file>` with a `--force` overwrite guard.

### Notification plane (fanout)

- **FanOutConsumer Lambda** on `TaskEventsTable` DDB Streams (`ParallelizationFactor: 1` for per-task ordering) routes events to per-channel dispatchers with configurable default filters (**AD-4**).
- **GitHub edit-in-place dispatcher** — single status comment per task on the target PR, edited in place as progress events fire. Slack / Email dispatchers are log-only stubs pending integration.

### Webhook submission

- **`bgagent webhook create / list / revoke`** — registers external integrations. HMAC-SHA256 request verification (`X-Webhook-Signature`) + REQUEST authorizer for lookup. `channel_source: "webhook"` propagates through the task record for observability.

### Phase 3 — Cedar-driven HITL approval gates

- **`docs/design/PHASE3_CEDAR_HITL.md`** (1,883 lines) — 22 locked decisions across 17 sections: three-outcome model (allow / hard-deny / soft-deny), policy authoring guide, `evaluate_tool_use` skeleton, `ApprovalAllowlist` + `RecentDecisionCache`, REST API contract, CLI UX, state-machine concurrency, data model, observability, security, failure modes, scenarios, carry-forward list.
- **`docs/diagrams/phase3-cedar-hitl.drawio`** — 12-page companion.
- **Plumbing only:** `cedar_policies` is threaded through `blueprintConfig` → `repoConfig` → orchestrator → agent payload shape (`agent/src/models.py`), and `agent/src/hooks.py` has a documented PreToolUse stub calling out Cedar evaluation as the next step. **No Cedar engine, no policy evaluation, no approval flow.** Implementation is a separate PR (~3–4 weeks for 3a core).
- **`@cedar-policy/cedar-wasm@4.10.0`** locked as the engine choice after WASM vs native spike.

## Architectural highlights

- **Async-only polling** (rev-6). No SSE, no two-runtime split. Single AgentCore runtime, `GET /events?after=<cursor>` replaces the prior streaming surface.
- **DDB-Stream ordering for fan-out** — `ParallelizationFactor: 1` guarantees per-task ordering; no SQS FIFO, no ETag. (**AD-9**, with an explicit "rejected alternative" for GitHub's `If-Match` — see [Deploy validation](#deploy-validation) bug #3.)
- **Combined-turn nudge ack (AD-5)** — agent emits `nudge_acknowledged` BEFORE folding the nudge into the next turn; single milestone, no consume-side event.
- **Deterministic status snapshot (AD-6)** — `formatStatusSnapshot` derives every rendered field from `TaskDetail` + a small window of recent events; no LLM in the loop, no fabricated state.
- **`agent_milestone` unwrap** — agent writes named checkpoints (`pr_created`, `nudge_acknowledged`, …) as `agent_milestone` events with `metadata.milestone`. Fanout unwraps against a narrow `ROUTABLE_MILESTONES` allowlist before matching channel filters.
- **Numeric coercion at the DDB boundary** — shared `coerceNumericOrNull` helper handles the Document-client `number-as-string` quirk uniformly.

## Design doc

- **`docs/design/INTERACTIVE_AGENTS.md`** — canonical rev-6 design. Sections: polling (§5), notification plane (§6), fanout routing (§6.2), GitHub edit-in-place (§6.4), observability / `--trace` (§10.1), concurrency rationale (§8.9), AD-1 through AD-9.
- **`docs/diagrams/interactive-agents-phases.drawio`** — 5-page architecture (1. rev-6 topology, 2. watch polling, 3. nudge flow, 4. status + trace, 5. fanout + GitHub edit-in-place).
- **`docs/design/PHASE3_CEDAR_HITL.md`** + 12-page drawio (design-only; see above).

## Deploy validation

Every shipped surface was E2E-validated against a dev stack (`backgroundagent-dev`, us-east-1). The validation pass caught **three runtime bugs** that passed all unit tests but failed on live AWS — all are fixed in this PR:

1. **`costUsd.toFixed is not a function`** — DDB Document-client returns `Number` attributes as strings; fanout dispatcher was not coercing. Fix: shared `coerceNumericOrNull` helper at the fanout + orchestrator boundaries.
2. **`agent_milestone` wrapper events never routed** — fanout filter matched `event_type` literally; missed `pr_created` wrapped in `agent_milestone`. Fix: `effectiveEventType` unwrap against `ROUTABLE_MILESTONES` allowlist.
3. **GitHub rejects `If-Match` on PATCH `/issues/comments/{id}`** with HTTP 400 (`"Conditional request headers are not allowed in unsafe requests unless supported by the endpoint"`). The earlier ETag optimistic-concurrency design never worked in production. Fix: drop `If-Match` entirely; rely on DDB-Stream ordering + single-writer invariant (**AD-9** rewritten; §6.4 rewritten).

Also surfaced + fixed during validation: Bedrock Guardrail `inputStrength` over-blocked (`HIGH` → `MEDIUM`); fanout dispatcher silent-ignored non-finite numeric input (now warns); `bgagent watch` didn't exit cleanly on terminal (deferred-exit fallback); `bgagent status` snapshot missed Type / Channel / Description / Reason (all surfaced); `channel_source` silently dropped by API response (now included in `TaskDetail`); specific classifiers for `agent_status=error_max_turns` / `error_max_budget_usd` / `error_during_execution` (were falling through to a generic "Agent task did not succeed").

## Test plan

- [x] Agent: `uv run pytest -q` — **468 passed**
- [x] CLI: `yarn jest` — **190 passed**
- [x] CDK: `yarn jest --runInBand` — **1001 passed**
- [x] CLI tsc + CDK tsc — clean
- [x] Deployed to dev stack + 15 hands-on scenarios executed against `scoropeza/agent-plugins` PRs #32, #33, #43: new_task happy path, pr_iteration, pr_review, nudge + AD-5 ack, cancel, `--trace` download with `-o` / `--force`, `submit --wait`, caps with specific reason classifiers, webhook HMAC round-trip, `--idempotency-key`, `status` snapshot variants, `watch` edge cases, `--verbose` debug.

## Known limitations / follow-ups

Filed separately (see `.l4-issue-drafts/` on this branch; 6 issue drafts ready to file post-merge):

- `failTask` double-emits `task_failed` + over-decrements concurrency on Durable Execution step retry.
- Bedrock Guardrail `PROMPT_ATTACK` over-triggers on imperative-mood prompts even at MEDIUM (example log in the draft).
- Idempotent submit returns `DUPLICATE_TASK` error instead of the original `task_id` — non-standard semantics.
- `webhook_url` missing from `CreateWebhookResponse` — users have to construct it by hand.
- Pre-existing drafts: session-tag IAM tightening; container CVE rebase cadence.

Deferred feature work (tracked elsewhere):

- **Phase 3 Cedar HITL implementation** — ~3–4 weeks for 3a core; separate PR after merge.
- Slack / Email dispatchers (currently log-only stubs).
- GitHub webhook body-shape adapter (today's webhook path expects `{repo, task_description, ...}`; GitHub sends `IssueCommentEvent` / `PullRequestEvent`).

## Reviewers — suggested reading order

1. **`docs/design/INTERACTIVE_AGENTS.md`** — canonical rev-6 design (ADs, architecture, §5 / §6 / §10).
2. **`docs/diagrams/interactive-agents-phases.drawio`** — 5-page architecture.
3. **CDK handlers:** `cdk/src/handlers/fanout-task-events.ts`, `cdk/src/handlers/shared/github-comment.ts`, `cdk/src/handlers/shared/orchestrator.ts`, `cdk/src/handlers/shared/error-classifier.ts`, `cdk/src/handlers/shared/numeric.ts`.
4. **CDK constructs:** `cdk/src/constructs/task-api.ts`, `cdk/src/constructs/task-events-table.ts`, `cdk/src/constructs/task-nudges-table.ts`, `cdk/src/stacks/agent.ts`.
5. **Agent runtime:** `agent/src/pipeline.py`, `agent/src/runner.py`, `agent/src/hooks.py`, `agent/src/progress_writer.py`, `agent/src/telemetry.py`.
6. **CLI:** `cli/src/commands/{watch,nudge,status,submit,trace,webhook}.ts`, `cli/src/format.ts`, `cli/src/wait.ts`, `cli/src/bin/bgagent.ts`.
7. **Phase 3 design** (optional; will be a separate PR): `docs/design/PHASE3_CEDAR_HITL.md`, `docs/diagrams/phase3-cedar-hitl.drawio`.

GitHub's **Squash and merge** is the intended merge strategy.


krokoko:
Thanks @scoropeza ! Just had a quick first pass, solid foundaiton ! The test coverage is strong (879 CDK + 468 agent + 215 CLI), the failure modes are thought through (stranded-task reconciler, cancel race fix, nudge dedup), and the immediate value is real. The feedback below is about long-term direction

Question: The two-runtime split (IAM for background, JWT for interactive) is technically sound and solves the immediate problem. But stepping back, there's a tension in the UX it creates and in its compute-substrate portability. And it becomes more pronounced when you consider HITL

The current model forces developers to choose their observation capability at submission time:
  - bgagent submit — resilient (survives CLI crash), but polling-only observation
  - bgagent run — live SSE stream, but task dies if the CLI disconnects

The developer workflow most people want is: submit a task, forget about it, check in live whenever I feel like it, steer if needed, leave again. The current model forces you to choose at submission time — submit (resilient but polling-only) vs run (live stream but fragile). You can't upgrade a running orchestrator task to a live stream mid-flight because the agent is on Runtime-IAM and there's no SSE socket to attach to.

The direct-SSE path is AgentCore-specific. It relies on /invocations as the transport and AgentCore's auth layer for gating. On ECS (or any other compute), this doesn't exist. The durable path (ProgressWriter → DDB → polling) is already compute-agnostic and works everywhere.

Phase 3 introduces approval gates where the agent pauses and waits for human input — potentially for hours (reviewer in another timezone, overnight deploy windows, etc.). HITL is inherently asynchronous: the agent and the human operate on completely different timescales.

With direct-SSE on Runtime-JWT:
  - Agent pauses → writes "approval needed" to the SSE stream
  - If the CLI disconnects during the wait, what happens? Task orphans or needs special keep-alive logic.
  - If the task was submitted via bgagent submit (orchestrator path), there's no SSE stream at all — how does the user learn an approval is needed?
  - You end up building a notification/polling path anyway for the non-interactive case.

Have you thought about a relay-based streaming alternative ?

The FanOutConsumer skeleton in this PR is the first stage of a compute-agnostic real-time path:

Agent (any compute) → DDB TaskEventsTable → DDB Stream → Lambda → API Gateway WebSocket → CLI

This gives you:
  - Submit and forget, attach live whenever (no upfront choice)
  - Works on AgentCore, ECS, EC2, anything — agent just writes to DDB
  - Agent crash doesn't kill the stream (events already durable)
  - Single runtime (IAM) is sufficient — no JWT runtime needed
  - CLI auth is just Cognito on the WebSocket API (already have the token)

Trade-off is ~1-5s latency from the DDB Stream batching window. Could we consider that this is acceptable?

I'm thinking that will be useful for implementation of HITL as well. The core constraint of HITL: the agent stops and waits. Maybe for 10 seconds. Maybe for 3 hours (reviewer is in a different timezone). The agent must survive that wait regardless of whether anyone is watching.

With the relay approach, HITL is just another DDB event:

  Agent hits gate
    → writes { event: "approval_required", gate: "push_to_main", details: ... } to DDB
    → enters poll loop: check DDB every N seconds for a response
    → sleeps (costs nothing, agent is idle, compute is warm but not burning tokens)

Meanwhile, independently:
    DDB Stream → WebSocket → CLI shows "Agent paused: wants to push to main. Approve? (y/n)"

    Or: nobody is watching. That's fine. Agent just waits.

  User responds whenever:
    CLI → POST /v1/tasks/{id}/approval → writes to DDB (same pattern as nudges) → agent's next poll picks it up → continues or aborts

krokoko:
After some thinking, here is what I propose:

**Context:** PR #52 introduces a two-runtime model (IAM + JWT) with direct SSE streaming for interactive observation. This proposal presents an alternative architecture that is simpler, compute-agnostic, and better aligned with the HITL requirements ahead. If agreed, PR #52 should be reworked to ship only the foundational pieces and exclude the Runtime-JWT / SSE path.

---

## Requirements I see:

1. Users submit a task. They don't care about the compute substrate and/or don't want to select the observability path when submitting a request
2. Fire and forget: task runs independently, no one needs to watch.
3. If configured, users are notified when HITL approval is needed.
4. Users can check in at any time, ask the agent for a status report, and steer it if the direction is wrong.
5. If configured, the agent updates its source context as it works (e.g., comments on the GitHub issue that triggered it).

---

## What to keep from PR #52

The following pieces are the correct foundation and should ship:

| Piece | Covers |
| --- | --- |
| Orchestrator path (single IAM runtime) | Req 1, 2 |
| `ProgressWriter` → `TaskEventsTable` (DDB, any compute) | Req 1 |
| `GET /tasks/{id}/events` + `bgagent watch` (polling) | Req 4 (partial) |
| `bgagent nudge` / nudge mechanism | Req 4 (partial) |
| `FanOutConsumer` + DDB Stream wiring (skeleton) | Foundation for 3, 5 |
| Cancel fix + pipeline short-circuit | Req 4 |

## What to exclude from PR #52

The following should be removed from the PR and not shipped:

| Component | Reason |
| --- | --- |
| Runtime-JWT (second runtime with Cognito JWT auth) | No requirement demands a live SSE connection. Adds a second auth surface, compute-specific coupling, and conflicts with HITL design. |
| `execution_mode` field + routing logic | Single path (orchestrator) for all tasks. No need to choose at submission time. |
| `_SSEAdapter`, `sse_wire.py`, `ag-ui-translator.ts`, `sse-client.ts` | SSE machinery not needed in the proposed architecture. |
| `bgagent run` as a distinct command | Becomes `bgagent submit` + optional `bgagent watch`. |
| Stranded-task reconciler | Only needed because interactive tasks orphan when CLI disconnects. With a single orchestrator path, tasks are always independently started. |

---

## Proposed architecture

### The interaction model

Every interaction between the user and the agent is async and goes through DDB:

```bash
bgagent submit "fix the auth timeout bug"          # fire and forget
# ... time passes ...
bgagent ask abc123 "what's your status?"            # request a report
# ... response arrives via Slack/watch ...
bgagent nudge abc123 "also fix the logging module"  # steer (acknowledged)
bgagent approve abc123                              # approve a HITL gate
bgagent watch abc123                                # see full event history
```

All of these are the same primitive: **user writes to DDB → agent reads; agent writes to DDB → user reads.** Different event types, same mechanism, fully async.

---

### 1. Status reports — two tiers (requirement 4)

A naive `bgagent ask` that interrupts the agent has two problems: latency (agent is mid-turn, 30-120s before it reads the request) and cost (burns a turn from the `max_turns` budget — a nervous manager checking in 5 times costs 5% of the task's capacity).

**Tier 1: Lambda-synthesized summary (default).** A Lambda reads `TaskEventsTable` and synthesizes a status report from the milestone history. The agent is never interrupted.

```
bgagent ask abc123 "status?"
  → Lambda reads events for task abc123
  → Returns: "Step 7/~12: tests passed (14/14), applying changes to
     src/auth.ts and src/logging.ts. Last activity 45s ago."
```

This is instant, costs nothing, and answers the most common question ("what step is it on?").

**Tier 2: Agent-directed question (explicit opt-in).** For questions about intent or decision-making ("why did you choose that approach?"), the user explicitly interrupts the agent:

```
bgagent ask abc123 --agent "why are you refactoring the logging module?"
  → writes status_request nudge to DDB
  → agent reads on next between-turns hook
  → agent writes status_response event
  → user sees it via watch/Slack
```

This costs a turn and has latency. The `--agent` flag makes the trade-off explicit. Without it, the user always gets the fast/free Lambda summary.

---

### 2. Nudge acknowledgment (requirement 4)

Today's nudge is fire-and-forget — the user has no confirmation the agent received it, understood it, or changed direction. After injecting a nudge, the agent emits a `nudge_acknowledged` event:

```
bgagent nudge abc123 "fix both issues, separate commits"
  → agent reads nudge on next between-turns hook
  → emits: { event: "nudge_acknowledged",
             summary: "Understood — will fix both issues in separate commits" }
  → user sees acknowledgment via watch/Slack
```

This is appended to the agent's system context for the current turn — not a standalone turn, so it doesn't burn a full turn from the budget.

---

### 3. Configurable dispatch — FanOutConsumer as router (requirements 3, 5)

The `FanOutConsumer` should be a **router**, not a monolith. Each notification channel has its own error modes, rate limits, retry semantics, and credentials. A GitHub API outage shouldn't block Slack notifications.

```
Agent writes events → DDB → Stream → FanOutConsumer (router)
                                            |
                                            ├── reads dispatch config from task/repo record
                                            |
                              ┌──────────────┼──────────────┐
                              |              |              |
                        SQS: github    SQS: slack     SNS: email
                              |              |              |
                     GitHubDispatchFn  SlackDispatchFn    SES
```

**Notification config (per-repo or per-task):**

```json
{
  "notifications": {
    "github_issue": {
      "enabled": true,
      "events": ["milestone", "status_response", "approval_required", "task_complete", "task_failed"]
    },
    "slack": {
      "channel": "#coding-agents",
      "events": ["approval_required", "status_response", "task_complete", "task_failed"]
    },
    "email": {
      "events": ["approval_required"]
    }
  }
}
```

**GitHub issue comment (edited in-place as the agent progresses):**

```markdown
🤖 Agent working on #42 (updated 2 min ago)

| Step | Status |
| --- | --- |
| Clone repo | ✅ |
| Run tests (before) | ✅ 14/14 |
| Apply changes | ✅ 3 files modified |
| Run tests (after) | ✅ 14/14 |
| Open PR | ✅ #87 |
```

Note: the edit-in-place comment has a race condition (two events in the same DDB Stream batch → two Lambda invocations both edit → one update lost). Solve with SQS FIFO using `task_id` as the message group ID — guarantees single-concurrency per task.

---

### 4. HITL approval + soft questions spectrum (requirement 3)

The architecture needs to support two distinct interaction modes:

| Mode | Agent behavior | Example |
| --- | --- | --- |
| **Hard gate** (Cedar policy) | Agent stops, waits indefinitely for approval | "Wants to push to main. Approve?" |
| **Soft question** | Agent asks, keeps working with its best guess, incorporates late response if still relevant | "Found two approaches. A is faster, B is cleaner. Going with A unless you object." |

Hard gates are policy-enforced and non-negotiable. Soft questions are advisory — the agent proceeds with a default after a configurable timeout. Both write events to the same stream and both can be responded to via the same `approve` / `nudge` mechanism. The difference is whether the agent blocks.

This is ultimately a product decision, but the architecture must support both: an `approval_required` event with a `blocking: true/false` field, and the agent's poll loop respecting the distinction.

---

## End-to-end scenario

GitHub issue triggers a task. User manages it asynchronously over the course of an afternoon.

1. **Webhook fires** → task created → orchestrator invokes runtime → agent starts
2. **Agent progresses** → `ProgressWriter` writes milestones → `FanOutConsumer` routes to GitHub (edits issue comment) and Slack
3. **User checks in from phone** → `bgagent ask abc123` → instant Lambda summary: "Step 5/~10, applying changes, last activity 30s ago."
4. **User wants more detail** → `bgagent ask abc123 --agent "why did you change the retry logic?"` → agent responds next turn: "The existing retry used exponential backoff with no jitter, causing thundering herd under load."
5. **User steers** → `bgagent nudge abc123 "add jitter but keep the backoff intervals"` → agent acknowledges: "Understood — adding jitter to existing backoff."
6. **Agent hits Cedar gate** → writes `approval_required` → Slack ping + email to user
7. **User approves 20 minutes later** → `bgagent approve abc123`
8. **Agent finishes** → FanOutConsumer updates GitHub comment, posts to Slack
9. **Agent has a soft question during next task** → "Found dead code in utils.ts. Remove it? Going ahead unless you object." → proceeds after timeout, user never responds, nobody blocked.

No live connection at any point. Everything async.

---

## What we explicitly don't build

| Not building | Why |
| --- | --- |
| WebSocket relay | No requirement demands push-to-CLI. Polling works for "check in on demand." Push notifications go to Slack/email/GitHub, not to a terminal. |
| Token-level streaming | Users need to know what step the agent is on, not watch it type. Milestone events + status reports cover this. |
| Direct SSE from agent | Compute-specific coupling for marginal latency gain. If sub-second ever matters, the agent can open an outbound WebSocket — a future decision, documented but not built. |

---

## Why not direct SSE (what PR #52 currently ships)

The two-runtime model (IAM for background, JWT for interactive SSE) has three structural problems:

**1. UX tension.** Forces users to choose at submission time between resilient (`submit`, polling-only) and observable (`run`, fragile — task dies if CLI disconnects). The workflow users actually want is "submit and forget, attach live whenever."

**2. Compute coupling.** Direct SSE relies on AgentCore's `/invocations` endpoint and auth layer. Doesn't work on ECS, Fargate, EC2, or any other substrate. The durable DDB path is already compute-agnostic.

**3. HITL conflict.** Approval gates are inherently async — the agent waits minutes or hours for a human response. A live SSE connection can't survive that. You need the DDB-based async path for HITL regardless, which makes the SSE path redundant for all five requirements.

---

## Preserving optionality

Removing Runtime-JWT eliminates ~2,500 lines of code. Worth it for maintenance. But document **why** it was removed and **what would bring it back**:

- Token-level streaming requirement (watch the agent type character by character)
- Sub-200ms HITL loops (interactive debugging, rapid approval cycles)
- Live pair-programming mode (user and agent co-editing in real time)

If any of these materialize, the right solution is the agent opening an **outbound WebSocket** to API Gateway — not rebuilding the dual-runtime model. This avoids the dual-auth problem (agent connects out, no ingress auth needed), works on any compute with network egress, and degrades gracefully (DDB milestones remain durable if the direct connection drops).

---

## Requirement check

| # | Requirement | Solved by |
| --- | --- | --- |
| 1 | Compute-agnostic | Agent writes to DDB from anywhere |
| 2 | Fire and forget | Single orchestrator path |
| 3 | HITL notification | FanOutConsumer → Slack/email on `approval_required`; hard gates block, soft questions don't |
| 4 | Check in + steer | `bgagent ask` (instant Lambda summary or agent-directed), `bgagent nudge` (with acknowledgment), `bgagent watch` (history) |
| 5 | Update source context | FanOutConsumer → per-channel dispatch (GitHub edit-in-place, Slack, email), configurable per repo |

---

## Open design decisions

Before building, these need answers:

1. **Soft question timeout.** How long does the agent wait before proceeding with its default? Per-task config? Per-gate config? Global default?
2. **Status summary depth.** What does the Lambda-synthesized summary include? Just the latest milestone, or a full narrative? Should it call an LLM to summarize, or template from structured events?
3. **Nudge acknowledgment cost.** Appending to system context is lighter than a standalone turn, but still consumes tokens. Cap the acknowledgment length? Skip acknowledgment for simple nudges?

---

## Suggested timeline

1. **Rework PR #52** — ship only the foundational pieces (ProgressWriter, events table, FanOutConsumer skeleton, nudge, cancel fix, single IAM runtime). Exclude Runtime-JWT, SSE path, `execution_mode`, stranded-task reconciler.
2. **`bgagent ask` (Lambda summary tier)** — highest-value, lowest-effort UX addition. No agent changes needed.
3. **Nudge acknowledgment + `bgagent ask --agent`** — agent prompt changes, status_response event type.
4. **FanOutConsumer dispatch** — router + let's add one or two channels. SQS FIFO per channel.
5. **Phase 3 Cedar-HITL** — hard gates and soft questions, both using the event/response loop above. Design against DDB async delivery from day one.

What do you think ?

scoropeza:
Thanks @krokoko — this is a really well-structured proposal and I'm on board with the overall direction. Agreeing fully on the big moves and flagging a handful of smaller adjustments where I think the architecture or the Claude Agent SDK pushes us one way rather than another. Organizing this as **what we agree on**, **where I think the proposal needs a small amendment**, and **a revised phase plan** at the end.

## What I fully agree on

- **Single orchestrator path, delete Runtime-JWT.** No requirement actually demands a live SSE connection — HITL is inherently async, fire-and-forget is what users want, and the second runtime was paying for a benefit the product doesn't need.
- **Delete `execution_mode: 'interactive'`, `_SSEAdapter`, `sse_wire.py`, `sse-client.ts`, `ag-ui-translator.ts`, `bgagent run`.** The compute-agnostic durable event plane through `TaskEventsTable` covers all five requirements.
- **FanOutConsumer as router**, per-channel dispatcher Lambdas, edit-in-place GitHub comment, configurable per-repo/per-task notification routing.
- **Keep Phase 3 Cedar HITL design as-is** — it's already DDB-poll based, so approval events ride the same fan-out plane without changes.
- **Hard gates block, soft gates don't** is the right conceptual split.
- **The overall interaction model** (submit → notify → ask/nudge/approve asynchronously → finish) matches what Cursor, GitHub Copilot coding agent, Codex, and Devin all ship — we're catching up, not inventing.

One correction on the LOC claim: real deletion is closer to **~4.5k–6k LOC** once you include tests (`test_sse_wire.py`, `test_sse_adapter.py`, `cli/test/sse-client.test.ts`, `cli/test/ag-ui-translator.test.ts`, plus SSE paths in `agent/src/server.py` which is ~50% of that file). Your ~2.5k number was conservative — we get even more cleanup than advertised.

## Where I'd amend the proposal

### 1. Keep the stranded-task reconciler (just collapse the two timeouts)

`cdk/src/handlers/reconcile-stranded-tasks.ts:59-70` actually has two thresholds — `STRANDED_INTERACTIVE_TIMEOUT_SECONDS=300` and `STRANDED_ORCHESTRATOR_TIMEOUT_SECONDS=1200`. The orchestrator branch covers **"OrchestratorFn async-invoke crashed between TaskTable write and the Runtime-IAM sync invocation."** That failure mode persists after we remove the interactive path — we still need a watchdog that catches tasks stuck in `SUBMITTED` or `HYDRATING`. What we can delete is the interactive-specific timeout branch; the reconciler itself stays. ~30 LOC cleanup, not a removal.

### 2. `bgagent ask` — honest positioning: deterministic `status` vs. LLM-synthesized summary

I love the "manager asks the teammate" UX, but I want us to be explicit about what the Lambda summary actually produces, because the events in `TaskEventsTable` are **structured machine telemetry, not narrative**. Look at `agent/src/progress_writer.py:111-175`:

- `agent_tool_call`: `{tool_name, tool_input_preview (200 chars), turn}`
- `agent_turn`: `{turn_number, model, thinking_preview (200 chars), tool_calls_count}`
- `agent_milestone`: `{label, details_preview (200 chars)}`
- `agent_cost_update`, `agent_error`, etc.

A deterministic Lambda reading these can produce "Step 7 of ~12: ran Bash 14 times, wrote utils.ts + auth.ts, last milestone `nudge_applied` 45s ago." That's useful, but it's not a synthesized narrative — it's a templated status.

We should pick one of two explicit options and commit, because the shape of what ships differs:

- **Option A — `bgagent status` (deterministic, no LLM).** Rename to make the semantics clear. Last N milestones + elapsed time + current turn counter + cost so far + last activity timestamp. Fast, free, never hallucinates. This is what Cursor's task view and Copilot's PR view already do.
- **Option B — `bgagent ask` (LLM-synthesized).** Genuine narrative summary. Costs per-ask on the platform's Bedrock budget (not the user's `max_budget_usd`, which is scoped to the task pipeline), adds a new failure mode ("summary Lambda hallucinated"), and would be a real industry differentiator — **no shipped product does this today**, per my research.

The middle path — "Lambda pretends to synthesize" — is the worst of both worlds, so let's not ship that. Happy to discuss which one lands in v1.

### 3. `bgagent ask --agent` — realistic latency + a foreground-wait UX

One thing I want to nail down before we commit to copy: the "agent responds on next between-turns hook" has meaningful latency. Looking at `agent/src/hooks.py:324-380`, the Stop hook only fires between turns — and a single turn with a long Bash or test run (e.g., `npm test`, `pytest`) is one turn. Realistic **P95 is in the range of minutes, not seconds**. We need to be honest about that in the UX.

**How the CLI should behave (proposal):** default to a **foreground block-and-poll** so it feels like an API call from the user's perspective, with durability as a safety net:

```
$ bgagent ask --agent abc123 "why did you change the retry logic?"
⠋ queued as ask_01JHXY... — waiting for agent
⠙ agent is running tool: Bash (turn 7/~12) — 42s elapsed
✓ agent responded (1m 14s)

The existing retry used exponential backoff with no jitter, causing thundering
herd under load. Added jitter to spread retries across the window.
```

Mechanism (same shape as `submit --wait`):

1. CLI `POST /tasks/{id}/asks` → returns `{ask_id, cursor: <last_event_id_before_write>}`
2. CLI polls `GET /events?after=<cursor>&type=status_response&correlation_id=<ask_id>` with **adaptive interval** (500ms while last poll had events, backoff to 5s when idle)
3. Spinner shows the agent's **current activity** (last `agent_turn` / `agent_tool_call` milestone) so the user can see it's alive during long turns
4. CLI exits on: `status_response` arrival → print response; task terminal state → print "task finished without answering"; `--timeout` expiry (default 5 min, hard cap 10 min) → print "still pending — retrieve later with `bgagent asks show <ask_id>`"

**Durability is preserved** — the ask row lives in DDB regardless of whether the CLI is still running. If the terminal is closed mid-wait (or the user Ctrl-Cs), the ask still executes; the response is retrievable via `bgagent asks show <ask_id>`, `bgagent watch`, or Slack if configured.

**Flags:**
- Default → foreground block with spinner (above)
- `--no-wait` → returns immediately with `{ask_id}`, response delivered via Slack/watch
- `--timeout <N>` → overrides default 5-minute wait

**New surface:**
- `POST /tasks/{id}/asks` (creates ask row)
- `GET /tasks/{id}/asks/{ask_id}` (retrieves an answered ask, for `bgagent asks show`)
- New event type `status_response{ask_id, content, turn}` emitted by the agent hook alongside the turn where it answers
- Rate limit: cap 1 open ask per task (return 429 otherwise). This also solves the concatenation issue in `nudge_reader.read_pending` + `format_as_user_message` (`hooks.py:245`), which today merges all pending items into one `<user_nudge>` blob — without a cap, rapid-fire asks drown each other.

### 4. Nudge acknowledgment — "appended to system context" is not something the Claude Agent SDK supports

I want to push back on this one with sources, because I dug in and it changes the cost model. The proposal says acknowledgment is "appended to the agent's system context for the current turn — not a standalone turn, so it doesn't burn a full turn from the budget." This is not achievable with the Claude Agent SDK (Python) today.

The SDK's mid-run extension surface is:

- **`HookMatcher`** for `PreToolUse`, `PostToolUse`, `Stop` / between-turns events. The full `HookEvent` enum is fixed: `PreToolUse | PostToolUse | PostToolUseFailure | UserPromptSubmit | Stop | SubagentStart | SubagentStop | PreCompact | PermissionRequest | Notification`. There is no `SystemPromptUpdate`, `BetweenTurns`, `ContextAppend`, or `MessageInject` event.
- **`decision: "block"` + `reason: <text>`** from a Stop hook — this is how today's `_nudge_between_turns_hook` injects text (see `agent/src/hooks.py:380`). That decision causes the SDK to **emit a new user turn** with `reason` as the content. It is unambiguously a full turn.
- **`updatedMCPToolOutput`** from PostToolUse for redacting tool output — not relevant here.
- **`ClaudeAgentOptions.system_prompt`** is set once, at construction time (`agent/src/runner.py:262`). The SDK has no public method to mutate it mid-conversation, and there is no `ClaudeSDKClient.update_system_prompt()` or `inject_message()` API.
- The `systemMessage` field on hook return values looks promising but is **user-visible only** — it does NOT enter Claude's conversation context. Same for `hookSpecificOutput.additionalContext` on PostToolUse. Both are confirmed by Anthropic as "not planned" — see [claude-code#18427](https://github.com/anthropics/claude-code/issues/18427) and [claude-code#19643](https://github.com/anthropics/claude-code/issues/19643), both closed with `not-planned`.

Primary sources:
- SDK source: https://github.com/anthropics/claude-agent-sdk-python/blob/main/src/claude_agent_sdk/types.py
- SDK hooks docs: https://docs.claude.com/en/api/agent-sdk/hooks
- Our own code as corroborating evidence: the fact that `_nudge_between_turns_hook` uses `decision: "block"` at `hooks.py:380` is itself proof that no lighter primitive exists — if a "system-context append" existed, the current code would already be using it (it's cheaper for regular nudges too).

What this means in practice: **acknowledging a nudge burns a turn** just like applying the nudge does today. The "savings" claim doesn't hold.

Two honest alternatives that actually work with the SDK:

- **Combined-turn ack (cost-neutral):** the hook emits a `nudge_acknowledged` milestone to `TaskEventsTable` **before** returning `decision: "block"`, and the `reason` string carries both the nudge AND "I'll acknowledge this in my next response." Still one turn, same cost as today, but the user sees the acknowledgment event land on their watch/Slack feed immediately.
- **Zero-turn milestone-only ack:** the hook writes `nudge_acknowledged` milestone, returns `{}` (no block), and the nudge text queues for the next turn that fires naturally (a tool-call Stop, etc.). Genuinely cheaper — zero turns burned by the ack itself — but the nudge text only arrives when some other Stop happens. Fine for nudges, probably too slow for asks.

I'd lean toward **combined-turn ack** for v1 — same cost as today, better event emission — and document the cost honestly. If the SDK grows a `SystemPromptUpdate` or `BetweenTurnsAppend` hook later, we revisit.

### 5. Soft questions — "proceed with default, incorporate late response" doesn't work with tool calls

Want to argue for a tighter definition, because as written the feature isn't implementable. Tool calls are committed, one-shot, fire-and-once. Consider the example:

> Agent: "Found dead code in utils.ts. Remove it? Going ahead unless you object."

The agent's decision point is synchronous: it emits an `Edit(utils.ts, delete 40 lines)` tool call **within the same turn** as the question. The file is deleted before the user's "wait, don't delete that" reaches the next Stop hook. "Incorporates late response if still relevant" requires the agent to:

1. Remember what it did in turn T
2. On turn T+k, receive the late objection
3. Inspect state, decide whether the objection is still actionable
4. Synthesize a reverse operation (re-insert deleted code, un-commit, un-push)

That's a transactional undo stack over arbitrary filesystem + Bash + git operations. The Claude Agent SDK doesn't have one. Tools don't have compensating actions defined. And primary research across Cursor, Codex, GitHub Copilot coding agent, Replit, and Devin confirms **no shipped product implements "proceed with default if no response."** GitHub Actions `environment` protection rules have auto-approve-after-timeout, but that's at the infra deploy layer, not inside an agent's tool-call loop.

**Concrete redefinition I'd propose:**

> **Soft question = bounded wait inside the hook.** Agent emits `approval_required{blocking: true, soft: true, timeout_s: 60}`. The hook polls DDB for the user's decision with a 60s timeout (configurable). If the user responds in time, the agent proceeds with the user's choice. If not, it proceeds with the default. **No post-hoc revert.**

This is implementable today — it's just a shorter version of the hard-gate poll loop. The agent does wait, just not for a long time. The "feels non-blocking" UX comes from the short timeout, not from a separate execution model. If someone genuinely needs "fire-and-later-adjust," we scope it to PR-review tasks where "adjust" = "make another commit on the branch" and the agent is explicitly built to iterate that way.

Document loudly: *soft question = bounded wait, not fire-and-forget with retroactive correction.*

### 6. Skip SQS FIFO for v1 — DDB Stream ordering + GitHub ETag `If-Match` is simpler

Two reasons:

- **DDB Streams with `ParallelizationFactor: 1` (the default) already give per-shard ordering.** Events for the same `task_id` hash to the same shard, so the Lambda processes them one-at-a-time per shard. That solves ~95% of the race.
- For the remaining race (Lambda retries after partial failure), **GitHub's issue-comment PATCH API supports `If-Match` with the comment's ETag** — optimistic concurrency at the call site beats queue serialization at the source. If we get a 412 Precondition Failed, we re-GET, re-render, re-PATCH.

Ref: https://docs.github.com/en/rest/issues/comments

Adding SQS FIFO on top = extra latency per event, + another DLQ to reason about, + the 300 msg/s per-group throughput ceiling. Defer FIFO until a concrete race is observed in production.

### 7. Notification defaults must ship with v1 (not "later")

Biggest UX risk in the whole proposal IMO. `fanout-task-events.ts:52-61` today filters to `task_*` + `agent_milestone` + `pr_created` + `agent_error`. A 20-min task produces 8–15 milestones + a couple of errors + terminal events. **If Slack pings for every milestone, users mute the bot and never unmute.**

Per-channel defaults I'd propose:

| Channel | Default events | Opt-in verbose |
|---|---|---|
| Slack | terminal (`task_completed`/`task_failed`/`task_cancelled`) + `pr_created` + `agent_error` + `approval_required` | `--verbose` adds `agent_milestone` |
| Email | terminal + `approval_required` only | — |
| GitHub issue comment | `pr_created` + terminal (single edit-in-place comment) | — already minimal |

`bgagent notifications configure` command ships in v1. Per-repo default config file ships in v1.

### 8. Debug escape hatch — `--trace` flag

Without SSE, if an agent misbehaves the only debug tool is `GET /events?after=` polling + CloudWatch logs. Event previews are truncated to 200 chars (`progress_writer.py` `_PREVIEW_MAX_LEN`). That's fine for normal runs but painful when you're debugging "why did my agent fail on step 5."

Proposed: `bgagent submit --trace` flag that (a) raises preview truncation to 4KB for that task only, (b) emits a `trajectory_uploaded` milestone with an S3 URI where the full agent trajectory is dumped post-task. Cheap (opt-in DDB storage), concrete replacement for "watch the SSE stream."

### 9. Do we need SSE for `watch` at all? I think no.

You floated this in chat and I agree: **polling is fine.** Three reasons:

1. Cost is trivial — eventually-consistent DDB query with cursor ≈ 0.5 RCU per page, concurrent watchers × 2s poll over active minutes = pennies
2. Cursor, Copilot coding agent, and Codex's external view all use polling; primary research (Cursor docs, Copilot docs) shows no user complaints about lag — expectations are calibrated to async
3. The only perceptible wart is bursty delivery during active work (4–10 events in 2s hiccups). Fix with **adaptive polling**: 500ms while last poll returned ≥1 event, back off to 5s when idle. Same endpoint, no new Lambda, no Cognito Identity Pool, no 100KB-buffer workarounds for Function URL streaming.

This also shares the exact polling mechanism with `bgagent ask --agent`'s foreground block — one implementation, two UX surfaces.

## Revised phase plan

| Phase | Scope |
|---|---|
| **1 (PR #52 rework)** | Delete JWT + all SSE machinery; single orchestrator path; `bgagent status` (deterministic) OR `bgagent ask` (LLM — pick one); polling `watch` with adaptive interval; `--trace` mode; per-channel notification defaults in FanOutConsumer; ETag-based GitHub comment edits; collapse stranded-reconciler timeouts |
| **2** | `bgagent ask --agent` with foreground block-and-poll + combined-turn ack semantics; real Slack dispatcher |
| **3** | Cedar HITL hard gates + bounded-wait soft gates |
| **4** | GitHub + email dispatchers; per-repo/task notification config; `--verbose` opt-in |
| **Deferred** | Outbound WebSocket from agent (only if a concrete sub-200ms latency requirement emerges) |

## Summary

We're aligned on every big move. The amendments are mostly making implicit assumptions explicit so we don't ship a design that promises something the SDK can't deliver:

- **Soft questions** → bounded-wait, not post-hoc undo (tool calls are committed; no undo stack exists)
- **Nudge ack** → combined-turn (same cost as today), not "system-context append" (SDK has no such hook; Anthropic has closed the related issues as `not-planned`)
- **`bgagent ask`** → honest about deterministic vs. LLM-synthesized, pick one and commit
- **`ask --agent`** → foreground block-and-poll by default (spinner shows live activity, durability preserved if terminal closes); `--no-wait` for scripts; cap to 1 open ask per task
- **Stranded reconciler** → stays (covers orchestrator crashes), just collapses timeouts
- **Notification defaults** → ship v1 or users mute the bot
- **SQS FIFO** → skip, DDB + ETag is simpler
- **`watch`** → polling-only, adaptive interval (shares mechanism with `ask --agent`)
- **Debug** → `--trace` flag as escape hatch

Happy to draft the §9.13 rewrite for `INTERACTIVE_AGENTS.md` next if this lands, or the exact file-level delete/add/modify list for the rework PR. Whichever is more useful to you.

krokoko:
Awesome @scoropeza, thank you! Let me close the loop on each of your amendments and then we can lock the plan.

  Full agreements (no further discussion needed)

  - Keep the stranded-task reconciler, collapse timeouts. You're right — orchestrator crashes can strand tasks regardless of the interactive path. My original "remove it" was sloppy. Keep, simplify, move on.
  - Skip SQS FIFO, use DDB Stream ordering + ETag. Simpler, fewer resources, solves the real problem. If we observe races in production we add FIFO then. Agreed.
  - Notification defaults ship v1. If we don't ship sensible defaults, users mute the bot day one and never come back. Your proposed per-channel defaults (terminal + approval + error for Slack, terminal + approval for email, single edit-in-place for GitHub) are the right starting point.
  - watch is polling-only with adaptive interval. No SSE, no WebSocket, no new auth surface. Shares mechanism with ask --agent. Clean.
  - --trace flag. Good escape hatch. Raises truncation to 4KB + dumps full trajectory to S3. Low cost, opt-in, concrete replacement for "stare at the SSE stream."
  - Real LOC cleanup is ~4.5-6k, not 2.5k. Even better. More confidence in the removal.

The two decisions that need to be made

  Decision 1: bgagent status (deterministic) vs. bgagent ask (LLM-synthesized)

  Your framing is exactly right: templated status from structured events vs. LLM narrative summary. These are different products with different failure modes.

  My recommendation: ship both, name them differently, v1 is deterministic only.

  - bgagent status <task-id> — deterministic Lambda, structured output. Fast, free, never hallucinates. This is v1.
  - bgagent ask <task-id> "why did you..." — always goes to the agent (your "Tier 2"). No confusing --agent flag. If you're asking a question in natural language, it goes to the agent. If you want machine-readable status, use status.

  The reason: bgagent ask "status?" where sometimes it hits a Lambda and sometimes the agent is confusing UX. Two commands with two clear contracts is better than one command with a flag that changes the execution model.

  bgagent ask always blocks foreground (your spinner UX), always burns a turn, always has latency. Users understand this because they typed a natural-language question — it's obviously going to a thinking entity. bgagent status is
  obviously a dashboard read. No ambiguity.

  Decision 2: Soft questions — bounded wait vs. don't ship

  Your pushback on "proceed with default, incorporate late response" is devastating. You're right:

  - Tool calls are committed and one-shot
  - No undo stack exists in the SDK or in the filesystem/git abstraction
  - No shipped product does this
  - "Incorporates late response if still relevant" is hand-waving over transactional rollback

  My recommendation: don't ship soft questions in v1. Hard gates only.

  Here's why: bounded-wait (your "60s timeout, then proceed with default") is implementable but creates a worse product than no soft questions at all. Consider:

  - User sees "Agent wants to delete dead code. Proceeding in 60s unless you object."
  - User is in a meeting. Comes back in 5 minutes.
  - Code is already deleted. User is annoyed.
  - Or: user scrambles to respond in time and feels pressured.

  Both are bad UX. The "ticking clock" feeling is anxiety-inducing for what should be a calm async workflow.

  The right v1 behavior: the agent makes its own judgment call silently (it already does this — it decides things every turn without asking). If we want the agent to be more conservative about certain actions, that's what Cedar policies
   are for — make them hard gates. The spectrum isn't "hard gate vs. soft question" — it's "gate vs. no gate." If an action is risky enough to ask about, it's risky enough to wait for an answer.

  Soft questions can be revisited post-v1 as a Cedar policy type: effect: "advise" that emits an event but doesn't block. The user sees it in their notification feed, the agent has already moved on. That's not "proceed unless you
  object" — it's "FYI, I did this." Different framing, no timeout, no anxiety.

  Nudge acknowledgment

  Your combined-turn ack is the right mechanism. Same cost as today, better observability (milestone lands immediately before the turn starts). The proposal was wrong about system-context append — thanks for the SDK deep-dive with sources. Let's go with:

  1. Hook fires, reads nudge
  2. Emits nudge_acknowledged milestone to ProgressWriter (instant, visible to watch/Slack)
  3. Returns decision: "block", reason: <nudge text> — starts the next turn
  4. Agent incorporates the nudge in its response (which is the turn)

  The user sees the acknowledgment event land within their polling interval. The "burned turn" is the same turn that would have happened anyway — the nudge just redirects its content. Document it honestly: "nudges are consumed as turns."

  bgagent ask --agent foreground UX

  Your proposed spinner + adaptive polling + timeout + durability-on-disconnect is the right UX. One amendment:

 Rate limit: cap 1 open ask per task

  Agreed, but make it 1 open ask per task per user if we ever support multi-user teams watching the same task. For v1 (single user), 1 per task is fine.

  The POST /tasks/{id}/asks → GET /events?correlation_id=<ask_id> pattern is clean. Ship it.

image

Does that work for you ? Thank you !

scoropeza:
Fully aligned @krokoko — let's lock it. Your three refinements are all improvements on what I had:

1. **`bgagent status` + `bgagent ask` as separate commands (no `--agent` flag).** You're right — command-level split is cleaner than a flag that silently changes the execution model. `status` = dashboard read (deterministic Lambda), `ask` = natural-language question to the agent (always blocks, always burns a turn, always has latency). This also resolves my open Decision 1: v1 is deterministic-only for `status`; LLM-synthesized summary is punted to a future version if it earns its keep.

2. **No soft questions in v1 — hard gates only.** Your "gate vs. no gate" framing is sharper than my "hard vs. soft" compromise. The ticking-clock UX is actively hostile in an async workflow, and the bounded-wait version I proposed was a worse product than just not having the feature. Post-v1 we can explore a Cedar `effect: "advise"` that emits an event without blocking — "FYI, I did this" framing, no timeout, no pressure. Agreed, dropping from v1 scope.

3. **Rate limit: 1 open ask per task per user.** Forward-compat for future multi-user team-view scenarios. No extra work in v1 (single user = same as today), just the right schema.

Locked plan:

- **Phase 1 (PR #52 rework)**: delete JWT + all SSE machinery; single orchestrator path; `bgagent status` (deterministic); polling `watch` with adaptive interval; `--trace` mode; per-channel notification defaults in FanOutConsumer; ETag-based GitHub comment edits; collapse stranded-reconciler timeouts.
- **Phase 2**: `bgagent ask` (foreground block-and-poll, spinner, durability-on-disconnect, 1 open per task per user); nudge combined-turn ack; real Slack dispatcher.
- **Phase 3**: Cedar HITL hard gates only.
- **Phase 4**: GitHub + email dispatchers; per-repo/task notification config; `--verbose` opt-in.
- **Deferred**: LLM-synthesized summary; Cedar `effect: "advise"` soft gates; outbound WebSocket.

Starting on the `docs/design/INTERACTIVE_AGENTS.md` rewrite next (retire §9.13, revise to async-only), then the `PHASE3_CEDAR_HITL.md` revision (drop soft-gate scope for v1). Will push those as a docs-only commit for your review before touching any code. Thanks for the great back-and-forth on this one.

scoropeza:
Updated both design docs to reflect the locked architecture. Pushed as `7b3b5b1`.

**Changes:**

- **`docs/design/INTERACTIVE_AGENTS.md`** — full rewrite to rev 6. States only the async-only architecture (single runtime, polling CLI, notification plane, `bgagent status`/`watch`/`nudge`/`ask` UX, FanOutConsumer router, `--trace` escape hatch, unified stranded-reconciler timeout). Architectural decisions captured as a concise §11 summary. 1723 → 809 lines.

- **`docs/design/PHASE3_CEDAR_HITL.md`** — revised for hard-gates-only v1. `@tier` vocabulary shifted to `hard-deny` + `hard-gate` throughout. §8.3 became "Notification UX" (fan-out plane with Slack action buttons / email / GitHub; no streaming prompts). Starter set reclassified — `force_push_main` and `write_credentials` promoted to hard-deny; `push_to_protected_branch` and `write_env_files` remain hard-gate; `force_push_any` deferred. `@tier("advise")` documented in §17 as a post-v1 non-blocking advisory tier. Scenarios rewritten to match the new rule set.

- Starlight mirrors regenerated via `docs/scripts/sync-starlight.mjs`; docs build verified green (`mise //docs:build` — 38 pages, 4.5s).

**Not touched in this commit (follow-up):** the 12-page `docs/diagrams/phase3-cedar-hitl.drawio` still reflects the old three-tier model — planning to regenerate that alongside the code rework so the diagram + code + docs land consistent.

**Starting the code rework next.** Scope per the locked plan:

1. Delete Runtime-JWT + all SSE machinery (`sse_wire.py`, `sse_adapter.py`, `sse-client.ts`, `ag-ui-translator.ts`, `bgagent run`, `execution_mode`)
2. `bgagent status` (deterministic Lambda)
3. `bgagent watch` adaptive polling
4. Combined-turn nudge ack
5. FanOutConsumer router + per-channel default filters (Slack/email/GitHub stubs, dispatched to real integrations in Phase 2)
6. GitHub edit-in-place via `If-Match` ETag
7. Unified stranded-task reconciler timeout
8. `--trace` flag

Will keep this PR as the v1 rework and open follow-ups for Phase 2 (`ask` + real Slack) and Phase 3 (Cedar hard gates) after v1 lands. Thanks again for the thorough review.

scoropeza:
Hi @krokoko — this PR is ready for another look.

Since we locked the rev-6 async-only architecture on 2026-04-30, the branch has been fully implemented, deployed to a dev stack, and E2E-validated. A few notable things happened along the way:

## Deploy validation caught three real runtime bugs (all fixed on-branch)

All three passed 100% of the unit tests but failed on live AWS. Documented in the updated PR description:

1. **`costUsd.toFixed is not a function`** — DDB Document-client returns `Number` as strings; fanout dispatcher was not coercing. Every terminal event on `pr_iteration` / `pr_review` was silently rejected in production.
2. **`agent_milestone` wrapper events never routed** — fanout filter matched `event_type` literally; missed `pr_created` wrapped in an `agent_milestone` event. Zero comments ever posted on any PR.
3. **GitHub rejects `If-Match` on PATCH `/issues/comments/{id}`** with HTTP 400. The ETag optimistic-concurrency scheme from the original §6.4 design never worked in production. AD-9 and §6.4 have been rewritten to rely on DDB-Stream `ParallelizationFactor: 1` + single-writer invariant instead. The "rejected alternative" is documented in the design doc with the verbatim GitHub error string so the next person to reach for ETag hits the explanation immediately.

## Hands-on testing — 15 scenarios against `scoropeza/agent-plugins`

All CLI surfaces validated end-to-end: `bgagent submit / list / status / cancel / nudge / events / watch / trace download / webhook`, plus `--trace` / `--wait` / `--force` / `--idempotency-key` / `--verbose` / `--max-turns` / `--max-budget` / `--pr` / `--review-pr`. Each hands-on session also surfaced UX gaps that are now fixed on-branch (snapshot missing Type / Channel / Description / Reason; generic error classifications; `channel_source` silently dropped from API responses; `--wait` formatter bifurcation; stale watch terminal message).

## Test baselines updated

- CDK: 1001 passing (was 879)
- CLI: 190 passing (was 215 — count dropped because we deleted SSE tests during rev-6, then rose again with the hardening)
- Agent: 468 passing

## Phase 3 Cedar HITL — design + plumbing only, no engine

`docs/design/PHASE3_CEDAR_HITL.md` (1,883 lines) and the 12-page companion diagram are fully locked. Orchestrator + agent payload shape threads `cedar_policies` through end-to-end so the next PR that implements the engine doesn't need to touch the transport layer. No Cedar evaluation; no approval flow.

## Follow-ups

6 carry-forward issues drafted to `.l4-issue-drafts/` on the branch. Notable ones for post-merge:

- `failTask` double-emits `task_failed` + over-decrements concurrency on Durable Execution step retry (pre-existing bug this PR surfaced; documented with root cause + fix plan).
- Bedrock Guardrail `PROMPT_ATTACK` over-triggers on imperative-mood prompts even at MEDIUM (several example prompts captured).
- Idempotent submit returns `DUPLICATE_TASK` error instead of the original `task_id` — non-standard semantics.
- `webhook_url` missing from `CreateWebhookResponse`.

Ready for review whenever you have time. Happy to walk through any section or specific commit in a sync if that's faster.

krokoko:
Awesome, just a few findings from automated review, after that let's merge

1. Fanout handler never returns DynamoDBBatchResponse despite reportBatchItemFailures: true

  [CDK Code Review] cdk/src/constructs/fanout-consumer.ts:146 + cdk/src/handlers/fanout-task-events.ts:697

  The construct sets reportBatchItemFailures: true but the handler returns void. When any record fails (e.g., resolveTokenSecretArn throws
  AccessDeniedException), Lambda retries the entire batch instead of isolating the poisonous record. This defeats the per-task ordering guarantee and can cause
   cascading retries.

  Fix: Change handler return type to Promise<DynamoDBBatchResponse>, wrap per-record processing in try/catch, and return { batchItemFailures: [{itemIdentifier:
   record.eventID}] } for failed records.

  2. turns_attempted / turns_completed bypass coerceNumericOrNull — same bug class as the fixed costUsd.toFixed crash

  [Type Design + Error Handling] cdk/src/handlers/shared/types.ts:304-305

  The toTaskDetail mapper passes these fields through with only ?? null:
  turns_attempted: record.turns_attempted ?? null,
  turns_completed: record.turns_completed ?? null,
  These are written via the same DDB Document-client path that caused Bug #1 (costUsd.toFixed is not a function). Any downstream caller doing arithmetic on
  these will crash at runtime.

  Fix: Apply coerceNumericOrNull consistently:
  turns_attempted: coerceNumericOrNull(record.turns_attempted, 'turns_attempted', logger),
  turns_completed: coerceNumericOrNull(record.turns_completed, 'turns_completed', logger),

  3. Cancel hook doesn't short-circuit nudge processing — nudges consumed but never delivered

  [Agent Runtime] agent/src/hooks.py:374-385

  When the cancel hook fires and sets ctx["_cancel_requested"] = True, the loop continues to _nudge_between_turns_hook, which reads DDB, marks nudges as
  consumed, and adds them to _INJECTED_NUDGES. Since cancel takes precedence, these nudges are consumed (DDB state changed) but never actually injected into
  the agent's context. The user sees 202 Accepted for their nudge but it silently disappears.

  Fix: Check if ctx.get("_cancel_requested"): break after the cancel hook, or have _nudge_between_turns_hook early-return when cancel is set.

  4. CLI/CDK type drift: turns_attempted and turns_completed optionality mismatch

  [Type Design + CLI Review] cli/src/types.ts:66-69

  CDK TaskDetail declares these as required (number | null), but CLI marks them optional (?). This violates the documented contract that these files must stay
  in sync.

  Fix: Remove the ? from both fields in cli/src/types.ts.

  ---
  Important Issues (8 found — should fix)

  5. Unhandled exception in routeEvent crashes entire fanout batch

  [CDK Code Review] cdk/src/handlers/fanout-task-events.ts:734

  routeEvent is called with no try/catch inside the record loop. If resolveTokenSecretArn throws (intentional for AccessDeniedException), the entire handler
  crashes — combined with Issue #1, all records in the batch retry.

  6. progress_writer.py circuit breaker trips on ValidationException — kills all progress writes permanently

  [Error Handling] agent/src/progress_writer.py:138-152

  The bare except Exception feeds all errors into the same circuit breaker. A persistent schema error (e.g., item > 400KB) trips the breaker for all event
  types, silencing the entire progress stream. Transient vs. permanent errors should be distinguished.

  7. Reconciler silently succeeds when ALL transitions fail (e.g., IAM/DDB outage)

  [Error Handling] cdk/src/handlers/reconcile-stranded-tasks.ts:261-270

  Per-task failures are caught and logged at WARN, but the Lambda returns success. A systemic DDB failure strands all tasks indefinitely with only WARN-level
  logs as signal.

  8. Dual _ProgressWriter instances with independent circuit breakers

  [Agent Runtime] agent/src/runner.py:240 + agent/src/pipeline.py:303

  Two independent writers (runner-level for turn events, pipeline-level for milestones) each have their own failure counter. One can silently disable while the
   other continues, creating invisible event gaps.

  9. renderCommentBody not self-defending against uncoerced DDB strings

  [Error Handling] cdk/src/handlers/shared/github-comment.ts:329

  Defense is entirely at the single fanout call site. A future caller passing raw DDB values will hit the same toFixed is not a function crash.

  10. channel_source typed as string instead of 'api' | 'webhook' literal union

  [Type Design] cdk/src/handlers/shared/types.ts:54,158

  Loses exhaustiveness checking in downstream switches. The internal CreateTaskContext correctly narrows it but the external types don't.

  11. Python TaskConfig missing @model_validator for trace=True + user_id=""

  [Type Design] agent/src/models.py:111-124

  This combination causes a runtime failure at S3 upload time rather than at construction. A validator would surface it immediately.

  12. Markdown injection possible via prUrl in GitHub comment body

  [CDK Code Review] cdk/src/handlers/shared/github-comment.ts:315-325

  While sanitizeEventType strips dangerous characters, prUrl goes unsanitized into a markdown link. A compromised agent writing a crafted URL could break the
  table layout or inject content.

scoropeza:
Thanks @krokoko for the thorough 12-finding review — this was exactly the kind of scrutiny the change needed. All 12 are now addressed across 5 thematic commits plus one hotfix caught by integration re-testing.

## Finding → commit map

| # | Finding | Commit |
|---|---|---|
| 1, 5, 9, 12 | Fanout partial-batch retry + GitHub-comment defense-in-depth | [`1b8ffa6`](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/52/commits/1b8ffa6) |
| 2, 4, 10 | `TaskDetail` numeric coercion + `ChannelSource` literal union | [`d6ad9a5`](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/52/commits/d6ad9a5) |
| 3 | Cancel hook short-circuits nudge consumption | [`9e6c23f`](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/52/commits/9e6c23f) |
| 6, 8 | Progress writer error classification + shared circuit breaker | [`db55bfa`](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/52/commits/db55bfa) |
| 7, 11 | Reconciler alarming + `TaskConfig` trace-without-user_id validator | [`331e283`](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/52/commits/331e283) |
| — | Hotfix: drop unused `_context`/`_callback` args (nodejs24 rejects callback-style signatures) | [`d8a98d6`](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/52/commits/d8a98d6) |

### Notes on the hotfix

The hotfix landed after integration re-testing of the partial-batch fix. Under `nodejs24.x`, Lambda rejects the legacy 3-arg `(event, context, callback)` handler signature and throws `Runtime.CallbackHandlerDeprecated` on every invocation. The fanout consumer was crashlooping silently — visible only in CloudWatch. Switching to the 2-arg `async (event)` form restored it. Post-hotfix CloudWatch shows zero `CallbackHandlerDeprecated` errors and `fanout.batch.complete` reports `failed: 0`.

## Updated test counts

- CDK: 1001 → **1036**
- CLI: **190** (unchanged)
- Agent: 468 → **500**

All suites green at tip `7e170da`.

## Live integration validation

Post-hotfix re-run against a deployed dev stack confirmed:

- Fanout partial-batch: `fanout.batch.complete` with `failed: 0`; POST + PATCH dispatched against the same comment_id across `pr_created` → `task_completed`; zero `fanout.dispatcher.rejected`.
- Cancel-vs-nudge: `task_cancelled` observed without a corresponding `nudge_acknowledged` when cancel won the race.
- Progress writer: no regressions in error classification path; circuit breaker holds.
- Reconciler alarming: INFO with `errors: 0` on cold-start sweep.

Ready for another pass when you have time.

krokoko (approved):

PR #79, opened 2026-05-12
feat(fanout): migrate SlackNotifyFn to FanOutConsumer subscriber (#64)

### LINKED ISSUE #64: Migrate SlackNotifyFn to FanOutConsumer subscriber
## Context

PR #52 introduced the `FanOutConsumer` dispatcher pattern — a single `TaskEventsTable` stream consumer that routes events to channel-specific subscribers. The intent is to keep the stream-consumer count at 1 regardless of how many channels the platform supports, and to centralise the channel-filter logic so each subscriber doesn't reimplement it.

PR #42 shipped `SlackNotifyFn` as a **direct** `TaskEventsTable` stream consumer — written before #52 landed. It was not migrated onto the dispatcher when #52 merged.

The platform currently sits at **2 concurrent readers per shard** on `TaskEventsTable` (`SlackNotifyFn` + `FanOutConsumer`). Per the [DynamoDB Streams docs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.html), "no more than two processes at most should be reading from the same stream's shard at the same time" — exceeding this causes throttling. It's an architectural constraint, not a raisable service quota.

## The problem

- **Shard-reader cap blocker.** Any future channel that wants stream-based outbound (e.g. email notifier, additional GitHub fanout subscribers, a future Teams/Discord adapter) cannot be added without first migrating `SlackNotifyFn` off the stream — there is no "raise a limit" path here.
- **Duplicated filter logic.** The concern Alain raised on PR #42 — "channel-filter logic duplicated inside each consumer" — is still present: `SlackNotifyFn` filters events inline rather than subscribing to the dispatcher's `channelSource === 'slack'` fanout.
- **Inconsistent with #52's stated pattern.** #52 establishes the fan-out pattern as the project's idiomatic channel-outbound shape. The Slack notifier is currently the exception.

## Proposed shape

1. Add a `SlackNotify` subscriber under `FanOutConsumer` using the same subscription interface as the GitHub edit-in-place subscriber.
2. Remove the direct DynamoDB Streams event-source mapping from `SlackNotifyFn`.
3. Fold the Slack-specific channel filter into the dispatcher's routing so the subscriber receives pre-filtered events.
4. Preserve existing behaviour — threaded notifications, emoji reactions, cancel button — bit-for-bit; regression-test via the existing `slack-notify.test.ts`.

After this lands:

- `TaskEventsTable` concurrent shard readers: **1** (FanOutConsumer only).
- New channels can subscribe to the dispatcher without touching the stream config.
- The Slack notifier stops reimplementing the channel-source filter.

## References

- PR #42 (Slack integration, merged): <https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/42>
- PR #52 (FanOutConsumer, merged): <https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/52>
- DynamoDB Streams shard-reader guidance: <https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.html>

isadeks:
## Summary

Closes #64. Moves Slack outbound delivery off its own `TaskEventsTable` DynamoDB Streams consumer onto `FanOutConsumer` as a per-channel dispatcher. After this PR, `TaskEventsTable` has exactly one stream reader (FanOutConsumer), restoring headroom for future channels (Email, Teams, etc.) without exceeding DynamoDB's documented 2-reader-per-shard practical limit.

The single commit message walks through the migration (a) and four review-fix layers (b–f) — please read it before the diff for context. Quick map:

- (a) Migration — slack-notify rewritten as `dispatchSlackEvent()`, FanOut wires it in, IAM grant migrated, SlackNotifyFn + its DynamoEventSource deleted from SlackIntegration.
- (b) BLOCKER — restored partial-batch retry semantics for infra errors (originally regressed by the migration).
- (c) Split SlackApiError into terminal (swallow) vs retryable (escalate to retry).
- (d) Fixed NOTIFIABLE_EVENTS / CHANNEL_DEFAULTS drift; added render cases for `task_stranded`, `agent_error`.
- (e) Conditional UpdateItem on `task_created` / `session_started` to prevent duplicate root messages on retry.
- (f) Dropped `pr_created` from Slack defaults (verified live; was visually duplicating with `task_completed`'s View PR button).

## Test plan

- [x] `mise //cdk:compile` clean
- [x] `mise //cdk:test` — 1183 / 1183 pass (8 net-new tests added for the review fixes)
- [x] `mise //cdk:eslint` clean
- [x] `mise //cdk:synth` shows exactly one `AWS::Lambda::EventSourceMapping` on `TaskEventsTable`, pointing at `FanOutFn`
- [x] Dev-stack: Slack @mention happy path (👀 → ⏳ → ✅ + View PR + intermediate cleanup)
- [x] Dev-stack: Cancel button (🚫 + task_cancelled in thread)
- [x] Dev-stack: CLI submit (channel_source=api) — zero Slack dispatches, GitHub edit-in-place still fires
- [x] Dev-stack: agent_error / pr_created paths exercised end-to-end via real PR creation

## Notes

After deploying:
```
aws lambda list-event-source-mappings \
  --query 'EventSourceMappings[?contains(EventSourceArn,\`TaskEventsTable\`)].FunctionArn'
```
Should return exactly one ARN: the FanOutFn.


krokoko:
Automated review:

This is well-considered work. The partial-batch retry semantics, the terminal/retryable Slack error classification, and the conditional UpdateItem idempotency guards are all the kind
  of defensive engineering I'd expect at this maturity level. The commit message structure (a–f) is exemplary — it tells the story of how adversarial review improved the code, which is
  itself a teaching artifact.

  That said, the migration introduces coupling trade-offs and error-path gaps that warrant attention.

  ---
  CRITICAL (must fix before merge)

  1. GitHub 403/429 rate-limit responses are swallowed as terminal

  File: cdk/src/handlers/fanout-task-events.ts (4xx swallow branch)

  The new else if (httpStatus >= 400 && httpStatus < 500) swallow path treats ALL GitHub 4xx as channel-terminal. But HTTP 403 from GitHub frequently means rate limit exceeded (a transient condition), and HTTP 429 (Too Many Requests) is also 4xx. Under a reconciliation wave touching many tasks, an entire window of GitHub comment updates would be permanently lost with only a warn log.

  Fix: Carve out 403 and 429:
  && httpStatus !== 403  // Rate limit — retryable
  && httpStatus !== 429  // Too Many Requests — retryable

  2. Unconditional Secrets Manager grant violates the construct's guard pattern

  File: cdk/src/constructs/fanout-consumer.ts

  Every other external-service grant in FanOutConsumer (taskTable, repoTable, githubTokenSecret) is guarded by if (props.X). The new bgagent/slack/* Secrets Manager grant is unconditional — meaning dev deployments without Slack onboarding get a dangling IAM permission. This breaks the construct's documented contract and will trigger a cdk-nag AwsSolutions-IAM5 finding with a misleading suppression reason.

  Fix: Add slackSecretPrefix?: string prop, guard the policy statement, and add a dedicated NagSuppression with the correct reason.

  ---
  HIGH (strongly recommended)

  3. Missing TASK_TABLE_NAME silently returns as "successful dispatch"

  File: cdk/src/handlers/slack-notify.ts — early guard in dispatchSlackEvent

  When the env var is missing, the function returns (no throw), so the router counts Slack as dispatched. A broken deployment silently drops ALL Slack notifications indefinitely. This should throw an Error (infra-class) so the rejected-rate alarm fires and operators notice.

  4. Slack ratelimited retry can starve sibling channels

  When Slack returns ratelimited, the record lands in batchItemFailures. On retry, all three dispatchers re-execute — including GitHub and Email that already succeeded. For non-terminal events without conditional-persist guards, this means duplicate GitHub PATCHes and double log entries. Under sustained Slack rate-limiting, every record retries 3× before DLQ, each time re-invoking GitHub unnecessarily.

  Mitigation options: Per-channel "already-dispatched" bitmap, or document as accepted trade-off at current scale and surface Slack's Retry-After header in the warn log.

  5. Reaction/delete helpers eat ALL errors including network failures

  File: cdk/src/handlers/slack-notify.ts — addReaction, removeReaction, deleteMessage

  These functions catch everything (DNS failures, SyntaxError, timeouts) and log at warn. Under the new architecture where infra errors should propagate for retry, these swallows are inconsistent. The user-visible symptom: stale emoji reactions (hourglass remains after completion), orphaned intermediate messages.

  Recommendation: At minimum, log network/infra failures at error with a dedicated event key so operators can alarm.

  6. task_created / session_started duplicate-delete has no observability

  When ConditionalCheckFailedException fires (sibling retry won the race), the code calls deleteMessage which is best-effort. If the delete fails, a permanent duplicate Slack message
  remains in the thread with no dedicated metric/event for operators to detect accumulating ghosts.

  Recommendation: Add a fanout.slack.dup_delete_failed event on the delete failure path.

  ---
  MEDIUM (address in follow-up)

  7. instanceof SlackApiError is brittle across module boundaries

  The fanout handler catches err instanceof SlackApiError. If the bundler ever duplicates the module (rare but possible), instanceof fails silently — terminal errors would trigger infinite retries. Add a property-based fallback:
  if (err instanceof SlackApiError || (err instanceof Error && err.name === 'SlackApiError')) {

  8. classifySlackError defaults unknowns to "retryable" — missing terminal codes

  The set omits ekm_access_denied, team_access_not_granted, missing_scope, invalid_arguments, posting_to_general_channel_denied — all permanent failures that will burn retries until DLQ.

  9. Type design: RouteOutcome arrays should be ReadonlyArray

  readonly dispatched: NotificationChannel[] prevents reassignment but allows outcome.dispatched.push(...). Inconsistent with ReadonlySet used for CHANNEL_DEFAULTS in the same file.

  10. SlackDispatchEvent is structurally identical to FanOutEvent

  The decoupling is purely nominal — both have identical fields. A type alias (type SlackDispatchEvent = FanOutEvent) or Pick<FanOutEvent, ...> would prevent silent drift and make the
  relationship explicit.

  ---
  Test Coverage Gaps

  ┌────────────────────────────────────────────────────────────────────────┬─────────────┐
  │                                  Gap                                   │ Criticality │
  ├────────────────────────────────────────────────────────────────────────┼─────────────┤
  │ Conditional UpdateItem race (task_created retry → deleteMessage)       │ 9/10        │
  ├────────────────────────────────────────────────────────────────────────┼─────────────┤
  │ taskStrandedMessage and agentErrorMessage renderers in slack-blocks.ts │ 7/10        │
  ├────────────────────────────────────────────────────────────────────────┼─────────────┤
  │ effectiveEventType rewrite reaching the Slack dispatcher mock          │ 7/10        │
  ├────────────────────────────────────────────────────────────────────────┼─────────────┤
  │ FanOutConsumer IAM policy (no construct-level test exists)             │ 6/10        │
  ├────────────────────────────────────────────────────────────────────────┼─────────────┤
  │ task_stranded through the terminal dedup path                          │ 8/10        │
  └────────────────────────────────────────────────────────────────────────┴─────────────┘

  ---
  Architecture Trade-offs (Accepted)

  These are design decisions I'd want documented but not necessarily changed:

  1. Shared timeout/concurrency — FanOutConsumer processes batches of 100 with a 60s timeout. The old SlackNotifyFn had batch=10, timeout=30s. Acceptable because channel_source gate short-circuits 90%+ of records, but worth a comment explaining why 100 is safe.
  2. Deployment ordering — During CDK deploy, there's a brief window where both consumers exist (old ESM not yet deleted). The conditional-persist guards make this safe (worst case: a duplicate that gets chat.deleted).
  3. chat.delete as best-effort cleanup — Correct design. Slack has no transactional guarantee. Just needs the observability improvement in point 6.

  ---
  What's Done Well

  - The partial-batch retry restoration (fix b) is the single most important change — without it, the migration would have silently regressed retry semantics
  - The TERMINAL_SLACK_API_ERRORS set with explicit documentation per code is excellent engineering
  - The cross-file consistency test (NOTIFIABLE_EVENTS ⊇ CHANNEL_DEFAULTS.slack) using jest.requireActual is an elegant drift guard
  - The conditional UpdateItem pattern for idempotent lifecycle persists is the correct defense against the retry hazard
  - The commit structure (a–f) makes the code reviewable in layers rather than as a monolithic diff

isadeks:
Addressed all 10 findings + 5 test gaps. Each in its own commit citing the review item — 16 commits total on top of the original migration.

**Critical / High fixes (numbers refer to the review-comment ordering, not GitHub issues):**

- **Finding 1** GitHub 403/429 carved out from terminal-swallow — retry path restored. The blanket 4xx swallow would have permanently dropped entire reconciliation waves under sustained rate-limiting.
- **Finding 2** Slack secret grant now guarded by an `slackSecretArnPattern` prop (matches the construct's other guarded grants — `taskTable`, `repoTable`, `githubTokenSecret`).
- **Finding 3** Missing `TASK_TABLE_NAME` now throws (was silently returning + counting as dispatched, so a broken stack would drop every Slack notification).
- **Finding 4** Generalized the dedup mechanism into a `SLACK_DEDUP_ATTRIBUTE` map covering `agent_error` (was unguarded — sibling-channel-failure retries would double-page operators). Also surfaces Slack's `Retry-After` header in the warn log.
- **Finding 5** Reaction/delete network errors promoted to `error` with dedicated event keys + `error_id`s. API-level rejections stay at `warn`.
- **Finding 6** New `fanout.slack.dup_delete_failed` event (with `FANOUT_SLACK_DUP_DELETE_FAILED` error_id) when the conditional-persist dup-delete path fails, so operators can alarm on accumulating ghost messages.
- **Finding 7** `SlackApiError` now matched by class OR `err.name` (defense against bundler module duplication).
- **Finding 8** Extended `TERMINAL_SLACK_API_ERRORS` with `ekm_access_denied`, `missing_scope`, `team_access_not_granted`, `posting_to_general_channel_denied`, `invalid_arguments`. Reorganized into commented blocks.
- **Finding 9** `RouteOutcome` arrays tightened to `ReadonlyArray<NotificationChannel>`.
- **Finding 10** `SlackDispatchEvent` is now a type alias of `FanOutEvent` so the contract can't drift silently.

**Test gaps:**

- **Gap 1** Conditional UpdateItem race + dup-delete: 4 tests covering the ghost-message paths and the `dup_delete_failed` alarm signal.
- **Gap 2** Renderers for `task_stranded` and `agent_error`: 5 tests pinning metadata propagation.
- **Gap 3** Agent_error dedup + slot isolation: 4 tests pinning the dedup map's separation between terminal and agent_error slots.
- **Gap 4** New `cdk/test/constructs/fanout-consumer.test.ts`: pins the single-ESM invariant, the conditional Slack secret grant, and the env-var wiring.
- **Gap 5** `task_stranded` through dedup: 2 tests for the reconciler-twin double-page scenario.

**Plus one extra:** commit `9ff9b45` fixes an orphan-message race I caught during live re-verification — the 🚀 *Task submitted* message sometimes lingered after fast tasks because the terminal-cleanup branch read `slack_created_msg_ts` from a stale dispatch-entry snapshot. Re-reading the TaskRecord before cleanup (after the dedup write linearizes the view) closes the window. Two tests for happy + fallback paths.

**Verification:**

- 1240/1240 unit tests
- ESLint clean, synth shows exactly one `AWS::Lambda::EventSourceMapping` on `TaskEventsTable`
- Dev-stack re-deployed and all 5 review-test scenarios re-verified live, including the orphan-cleanup fix

Ready for re-review.

krokoko (approved):

Issue #84, opened 2026-05-13
Add github:* resource tagging strategy to CDK stacks


scottschreckengaust:
## Parent

Sub-issue of #73 (RFC: GitHub Actions CI/CD with protected deployment environment and approval gates)

## Summary

Implement `github:*` resource tagging on all CDK stack resources via `Tags.of(stack).add()` in `main.ts`. Tags are injected through CDK context (`-c`) and default to sentinel `none` when not provided.

See tagging strategy addendum on #73: https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/73#issuecomment-4444794226

## Scope

- Add `github:*` tag resolution in `cdk/src/main.ts` (9 tags, all defaulting to `none`)
- Pass tags through CDK context values
- Add/update tests to verify tags are applied
- **No CI/CD workflow changes** — this is infrastructure-only

## Tags

| Tag key | Default | `$GITHUB_*` variable | `gh` CLI | `git` CLI |
|---------|---------|---------------------|----------|-----------|
| `github:sha` | `none` | `$GITHUB_SHA` | `gh api /repos/{owner}/{repo}/commits/{ref} --jq .sha` | `git rev-parse HEAD` |
| `github:ref` | `none` | `$GITHUB_REF_NAME` | `gh api /repos/{owner}/{repo}/git/ref/heads/{branch} --jq .ref` | `git branch --show-current` |
| `github:ref-type` | `none` | `$GITHUB_REF_TYPE` | `gh api /repos/{owner}/{repo}/git/ref/heads/{branch}` → `branch`; `/tags/{tag}` → `tag` | inferred from `git describe --exact-match` success/failure |
| `github:actor` | `none` | `$GITHUB_ACTOR` | `gh api /user --jq .login` | `git config user.name` |
| `github:head-ref` | `none` | `$GITHUB_HEAD_REF` | `gh pr view --json headRefName --jq .headRefName` | N/A (PR context only) |
| `github:run-id` | `none` | `$GITHUB_RUN_ID` | `gh run list --limit 1 --json databaseId --jq '.[0].databaseId'` | N/A (CI only) |
| `github:event` | `none` | `$GITHUB_EVENT_NAME` | `gh run view {run-id} --json event --jq .event` | N/A (CI only) |
[...]

Issue #73, opened 2026-05-10
RFC: GitHub Actions CI/CD with protected deployment environment and approval gates


scottschreckengaust:
### Related: #70 (context stack names), #72 (ephemeral cleanup)

---

## Current state (reconciled 2026-06-05)

The **core deploy pipeline is delivered and on `main`**:

- ✅ **Phase 1 — Foundation** (OIDC provider, `deploy` environment, GH→bootstrap-role assumption, account bootstrap)
- ✅ **Phase 2 — Build pipeline** (per-`compute_type` synth, immutable `cdk-<type>.out` artifacts, 13 `github:*` tags) — PRs #91, #93, #97
- ✅ **Phase 3 — Deploy pipeline** (artifact-based deploy, label-driven matrix, OIDC, baseline diff) — PR #98
  - ⬜ One item left: release flow (draft → deploy → tag → publish)
- ⬜ **Phase 4 — Cleanup (ephemeral stacks)** — **NOT done. Tracked by #72.** Partial work in draft **PR #109** (manual `scripts/cleanup-ephemeral-stacks.sh` only; no EventBridge schedule, no CloudWatch audit, no CDK construct, no tests). PR #109 is stale (untouched since 2026-05-18).
- ⬜ **Phase 5 — Observability** (CloudWatch alarms for stack count/ENI leaks/cost; approval checklist in `CONTRIBUTING.md`)

**Next to pick up:** ephemeral cleanup automation (Phase 4) — see acceptance criteria in **#72**, building on the script in **PR #109**.

---

## Summary

Establish a GitHub Actions deployment pipeline that:
1. Builds and synthesizes CDK **once per compute_type** in `build.yml` (always all registered types)
2. Stores `cdk-<compute_type>.out` as immutable deployment artifacts (synth once, deploy exact artifact)
3. Gates all deployments behind a **protected GitHub environment** (`deploy`) requiring manual approval — triggered by `deploy` label (with optional type qualifiers)
4. Deploys to AWS using OIDC federation assuming CDK bootstrap roles (no long-lived credentials)
5. Stack naming: `main-<compute_type>-prd` for production, ephemeral for PRs/branches
6. On successful deployment: creates a **GitHub Release** (drafted → published) with tagged `main` and `cdk-*.out` artifacts
7. Cleanup targets stacks **tagged with `github:*` context keys** (presence of any `github:sha` != `none`), gated behind approval with cancel-in-progress concurrency

---

[...]

PR #91, opened 2026-05-14
feat(ci): synth-per-variant build with github:* context in artifact

scottschreckengaust:
## Summary

- Adds `strategy.matrix` with `variant: [agentcore]` to the build job
- Generates `cdk/cdk.context.json` before build with all 13 `github:*` tags, `computeVariant`, and `stackName` — CDK reads this automatically during synth, baking tags into CloudFormation templates
- Resolves tag values per event type (`pull_request`, `merge_group`, `push`, `workflow_dispatch`) using intermediate env vars to prevent script injection (CWE-78)
- Uploads `cdk/cdk.out/` + `cdk/cdk.context.json` as immutable artifact (`cdk-<variant>-out`) for downstream `deploy.yml`
- Adds `/cdk/cdk-*.out/` to `.gitignore`

## How it works

```
build.yml:
  Resolve github:* → Generate cdk.context.json → mise run build (synth reads context) → Upload artifact
                                                                                              ↓
deploy.yml (future):                                                     Download artifact → cdk deploy --app cdk.out
```

CDK's `Tags.of(stack).add()` in `main.ts` calls `app.node.tryGetContext('github:sha')` etc. during synthesis. The values come from `cdk.context.json`, so they are baked directly into the template's resource `Tags` arrays. No context flags needed at deploy time.

## Context values generated

| Key | Source (pull_request) | Source (merge_group) | Source (push) |
|-----|----------------------|---------------------|---------------|
| `github:sha` | PR head SHA | merge group head SHA | commit SHA |
| `github:ref` | head branch | base ref (target) | ref name |
| `github:ref-type` | `branch` | `branch` | branch or tag |
| `github:actor` | `github.actor` | `github.actor` | `github.actor` |
| `github:head-ref` | source branch | queue branch | empty |
| `github:base-ref` | target branch | target branch | empty |
| `github:pr-number` | PR number | extracted from queue ref | empty |
| `github:run-id` | `github.run_id` | `github.run_id` | `github.run_id` |
| `github:run-attempt` | `github.run_attempt` | `github.run_attempt` | `github.run_attempt` |
| `github:event` | `github.event_name` | `github.event_name` | `github.event_name` |
| `github:workflow` | `github.workflow` | `github.workflow` | `github.workflow` |
| `github:repository` | `github.repository` | `github.repository` | `github.repository` |
| `github:clean` | `true` | `true` | `true` |

## Local verification

```
$ jq -n '{...}' > cdk/cdk.context.json && cd cdk && npx cdk synth -q
$ jq '[ .. | objects | .Tags? // empty | arrays | .[] | select(.Key? | startswith("github:")) ] | unique_by(.Key)' cdk.out/*.template.json
→ All 13 tags present, 132/299 resources tagged (all taggable resources)
```

## Follow-ups
- Additional variants (`ecs`, `eks`) added to matrix when compute paths are implemented
- `deploy.yml` downloads the artifact and runs `cdk deploy --app cdk.out` — no re-synthesis needed

## Test plan
- [x] CI build passes with matrix strategy (`build (agentcore)`)
- [x] `cdk-agentcore-out` artifact appears in workflow run (contains `cdk.out/` + `cdk.context.json`)
- [x] Mutation check passes (cdk.context.json is gitignored)
- [x] Local synth confirms all 13 `github:*` tags baked into CloudFormation template
- [x] Semgrep passes — no script injection (all GitHub context via `env:` + `jq --arg`)

Refs: #73, #84, #93

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,14 +56,11 @@ jobs:
         run: mise run install
       - name: build
         run: mise run build
-      - name: CDK Synth (${{ matrix.variant }})
-        run: npx cdk synth -c computeVariant=${{ matrix.variant }} --output cdk-${{ matrix.variant }}.out
-        working-directory: cdk
       - name: Upload CDK artifact (${{ matrix.variant }})
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: cdk-${{ matrix.variant }}-out
-          path: cdk/cdk-${{ matrix.variant }}.out/
+          path: cdk/cdk.out/
       - name: Find mutations
         id: self_mutation
         run: |-
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,9 @@ jobs:
       actions: write # upload-artifact when self-mutation is detected
       contents: read
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        variant: [agentcore]
     outputs:
       self_mutation_happened: ${{ steps.self_mutation.outputs.self_mutation_happened }}
     env:
@@ -53,11 +56,20 @@ jobs:
         run: mise run install
       - name: build
         run: mise run build
+      - name: CDK Synth (${{ matrix.variant }})
+        run: npx cdk synth -c computeVariant=${{ matrix.variant }} --output cdk-${{ matrix.variant }}.out
+        working-directory: cdk
+      - name: Upload CDK artifact (${{ matrix.variant }})
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: cdk-${{ matrix.variant }}-out
+          path: cdk/cdk-${{ matrix.variant }}.out/
+          retention-days: 5
       - name: Find mutations
         id: self_mutation
         run: |-
           git add .
-          git diff --staged --patch --exit-code > repo.patch || echo "self_mutation_happened=true" >> $GITHUB_OUTPUT
+          git diff --staged --patch --exit-code -- . ':!cdk/cdk-*.out' > repo.patch || echo "self_mutation_happened=true" >> $GITHUB_OUTPUT
         shell: bash
         working-directory: ./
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -53,11 +56,20 @@ jobs:
...
+        with:
+          name: cdk-${{ matrix.variant }}-out
+          path: cdk/cdk-${{ matrix.variant }}.out/
+          retention-days: 5
       - name: Find mutations
         id: self_mutation
         run: |-
           git add .
-          git diff --staged --patch --exit-code > repo.patch || echo "self_mutation_happened=true" >> $GITHUB_OUTPUT
+          git diff --staged --patch --exit-code -- . ':!cdk/cdk-*.out' > repo.patch || echo "self_mutation_happened=true" >> $GITHUB_OUTPUT
```
scottschreckengaust: If `/cdk/cdk-*.out/` is ignored, why is `-- . ':!cdk/cdk-*.out'` added?

scottschreckengaust: Good catch — you're right, it's redundant. Since `/cdk/cdk-*.out/` is in `.gitignore`, `git add .` never stages the synth output in the first place, so `git diff --staged` will never include it regardless of the pathspec exclusion.

Removed in e3875ee.

.github/workflows/build.yml
```diff
@@ -53,6 +56,14 @@ jobs:
         run: mise run install
       - name: build
         run: mise run build
+      - name: CDK Synth (${{ matrix.variant }})
```
krokoko: as part of the mise run build task, it runs the mise build cdk which itself runs the cdk synth command
can't we update these to pass the parameters so we don't need the custom npx cdk synth step ?

mayakost (dismissed):

.github/workflows/build.yml
scottschreckengaust: Good call...I'll add that

scottschreckengaust: Good call — removed the redundant `npx cdk synth` step entirely. The artifact upload now grabs `cdk/cdk.out/` that `mise run build` already produces (via `//cdk:build` → `synth:quiet`).

When variants actually diverge (once `computeVariant` context is consumed in the CDK app), we'll need to pass the variant context into the mise build chain — but that's a future concern since all variants produce the same output today.

Fixed in 6e04f52.

krokoko (approved):

Issue #72, opened 2026-05-09
feat: automated ephemeral stack cleanup (scheduled)


scottschreckengaust:
## Problem

Deploying ephemeral stacks (e.g. `commit-fa647ca...` for CI/testing) leaves orphaned CloudFormation stacks in the account. These accumulate over time and:

1. **ENIs from AgentCore/Lambda runtimes linger for 20-45+ minutes** after the runtime is deleted, blocking VPC/Security Group teardown and causing `DELETE_FAILED` stacks
2. Orphaned resources (DynamoDB tables, Cognito pools, log groups, secrets) accumulate costs
3. Manual cleanup is tedious and error-prone

## Proposal

An automated cleanup system that runs on a schedule and deletes ephemeral stacks that have exceeded their max age.

### Design

**Scheduled trigger**: EventBridge rule firing every N hours (configurable, default: 4h).

**Compute**: Lambda function (Python or Node) that:

1. Lists all CloudFormation stacks matching description `"ABCA Development Stack"`
2. Skips stacks with **termination protection** enabled (permanent environments: `-dev`, `-tst`, `-stg`, `-prd`, `-prod`)
3. Skips stacks in `*_IN_PROGRESS` states
4. For stacks exceeding max age:
   - Force-detaches Hyperplane ENIs (Lambda/AgentCore) attached to the stack's security groups
   - Waits briefly for detachment
   - Deletes now-available ENIs
   - Initiates `delete-stack`
5. Logs actions to CloudWatch for audit

### Max age behavior

| `MAX_AGE_HOURS` env var | Behavior |
|---|---|
| Set to `N` | Delete ephemeral stacks older than N hours |
| Missing or `0` | **Delete all non-protected ephemeral stacks** (default) |

The default of zero means: if you deployed it without termination protection, you intended it to be temporary. The scheduler cleans up everything that isn't explicitly protected. This is the safest default for cost — permanent stacks opt in to protection via the `-dev`/`-prd` suffix convention.

### Termination protection convention (from PR #70)

`cdk/src/main.ts` enables `terminationProtection` for stacks whose name ends in a known environment suffix:

```typescript
const protectedEnvs = ['dev', 'tst', 'stg', 'prd', 'prod'];
const isProtected = protectedEnvs.some(env => stackName.endsWith(`-${env}`));
```

All other stacks (e.g. `commit-abc123`, `feature-xyz`, `scott-test`) are ephemeral and eligible for cleanup.

[...]

PR #109, opened 2026-05-18
feat(ops): automated ephemeral stack cleanup script

scottschreckengaust:
## Summary

- Adds `scripts/cleanup-ephemeral-stacks.sh` — a shell script that identifies and deletes orphaned ABCA ephemeral CloudFormation stacks
- Handles stuck ENI cleanup (Lambda/AgentCore Hyperplane ENIs) before stack deletion to prevent `DELETE_FAILED` states
- Respects termination protection and skips stacks in active transitions
- Supports `--dry-run`, `--max-age-hours`, `--prefix` filter, and `--force-eni` options

## Status

**Script only** — this is the operational cleanup tool from Issue #72. Still needed for full issue completion:

- [ ] EventBridge scheduled rule (Lambda or Step Functions wrapper) for automated recurring execution
- [ ] CloudWatch logging/metrics for audit trail
- [ ] CDK construct or standalone stack to deploy the scheduler
- [ ] Integration tests / dry-run CI validation

The script itself is complete and production-ready for manual use today.

## Test plan

- [x] Run with `--dry-run` against an account with ephemeral stacks
- [x] Verify termination-protected stacks are never touched
- [x] Verify ENI cleanup works for stuck VPC security groups
- [x] Verify `--max-age-hours` filtering is correct

Part of #72 — this PR delivers the manual cleanup script foundation only. Full issue completion (EventBridge schedule, CloudWatch audit, CDK construct, tests) lands in a future PR.

### DIFF
--- a/scripts/cleanup-ephemeral-stacks.sh
+++ b/scripts/cleanup-ephemeral-stacks.sh
@@ -11,7 +11,6 @@
 #   --dry-run           Show what would be deleted without acting
 #   --max-age-hours N   Delete stacks older than N hours (default: 4)
 #   --prefix PREFIX     Only target stacks matching this prefix (default: all ABCA stacks)
-#   --force-eni         Force-detach ENIs even if stack deletion hasn't started yet
 #
 # Safety:
 #   - Never touches stacks with termination protection enabled
@@ -23,7 +22,6 @@ set -euo pipefail
 MAX_AGE_HOURS=${MAX_AGE_HOURS:-4}
 DRY_RUN=false
 PREFIX=""
-FORCE_ENI=false
 REGION="${AWS_DEFAULT_REGION:-us-east-1}"
 
 while [[ $# -gt 0 ]]; do
@@ -31,15 +29,33 @@ while [[ $# -gt 0 ]]; do
     --dry-run) DRY_RUN=true; shift ;;
     --max-age-hours) MAX_AGE_HOURS="$2"; shift 2 ;;
     --prefix) PREFIX="$2"; shift 2 ;;
-    --force-eni) FORCE_ENI=true; shift ;;
     *) echo "Unknown option: $1" >&2; exit 1 ;;
   esac
 done
 
+# Validate numeric input — guards the age arithmetic against injection/garbage.
+if ! [[ "$MAX_AGE_HOURS" =~ ^[0-9]+$ ]]; then
+  echo "Error: --max-age-hours must be a non-negative integer (got: '$MAX_AGE_HOURS')" >&2
+  exit 1
+fi
+
 MAX_AGE_SECONDS=$((MAX_AGE_HOURS * 3600))
 NOW=$(date +%s)
 
+# Surface the blast radius before touching anything. Confirms the operator is
+# pointed at the account/identity they think they are (defense in depth).
+CALLER_IDENTITY=$(aws sts get-caller-identity \
+  --region "$REGION" \
+  --query '[Account,Arn]' --output text 2>/dev/null) || {
+  echo "Error: unable to resolve AWS identity (sts:GetCallerIdentity failed). Check credentials." >&2
+  exit 1
+}
+ACCOUNT_ID=$(echo "$CALLER_IDENTITY" | cut -f1)
+CALLER_ARN=$(echo "$CALLER_IDENTITY" | cut -f2)
+
 echo "=== Ephemeral Stack Cleanup ==="
+echo "  Account:       $ACCOUNT_ID"
+echo "  Identity:      $CALLER_ARN"
 echo "  Region:        $REGION"
 echo "  Max age:       ${MAX_AGE_HOURS}h"
 echo "  Dry run:       $DRY_RUN"
@@ -62,6 +78,7 @@ fi
 
 DELETED=0
 SKIPPED=0
+FAILED=0
 
 while IFS=$'\t' read -r STACK_NAME CREATION_TIME; do
   # Apply prefix filter
@@ -99,8 +116,15 @@ while IFS=$'\t' read -r STACK_NAME CREATION_TIME; do
     continue
   fi
 
-  # Check age
-  CREATED_EPOCH=$(date -d "$CREATION_TIME" +%s 2>/dev/null || date -j -f "%Y-%m-%dT%H:%M:%S" "${CREATION_TIME%%.*}" +%s 2>/dev/null || echo 0)
+  # Check age. Parse the CreationTime to epoch seconds (GNU date, then BSD date).
+  # FAIL CLOSED: if both parsers fail we cannot trust the age, so SKIP rather than
+  # risk deleting a stack we can't prove is old enough.
+  CREATED_EPOCH=$(date -d "$CREATION_TIME" +%s 2>/dev/null || date -j -f "%Y-%m-%dT%H:%M:%S" "${CREATION_TIME%%.*}" +%s 2>/dev/null || echo "")
+  if ! [[ "$CREATED_EPOCH" =~ ^[0-9]+$ ]]; then
+    echo "  SKIP (unparseable creation time '$CREATION_TIME'): $STACK_NAME"
+    ((SKIPPED++)) || true
+    continue
+  fi
   AGE_SECONDS=$((NOW - CREATED_EPOCH))
 
   if [[ $AGE_SECONDS -lt $MAX_AGE_SECONDS ]]; then
@@ -129,7 +153,8 @@ while IFS=$'\t' read -r STACK_NAME CREATION_TIME; do
 
   if [[ -n "$SG_IDS" && "$SG_IDS" != "None" ]]; then
     for SG_ID in $SG_IDS; do
-      # Find ENIs attached to this security group
+      # Find ENIs attached to this security group.
+      # shellcheck disable=SC2016  # backticks are JMESPath literal syntax for --query, must NOT expand
       ENIS=$(aws ec2 describe-network-interfaces \
         --region "$REGION" \
         --filters "Name=group-id,Values=$SG_ID" \
@@ -172,12 +197,18 @@ while IFS=$'\t' read -r STACK_NAME CREATION_TIME; do
   fi
 
   # --- Delete the stack ---
+  # Only count a deletion we actually initiated. Tolerate a single failure
+  # (e.g. AccessDenied, transient throttling) without aborting the whole run —
+  # set -e would otherwise kill the loop mid-pass and orphan later stacks.
   echo "    Deleting stack $STACK_NAME..."
-  aws cloudformation delete-stack \
+  if aws cloudformation delete-stack \
     --region "$REGION" \
-    --stack-name "$STACK_NAME" 2>/dev/null
-
-  ((DELETED++)) || true
+    --stack-name "$STACK_NAME" 2>/dev/null; then
+    ((DELETED++)) || true
+  else
+    echo "    ERROR: delete-stack failed for $STACK_NAME (continuing)" >&2
+    ((FAILED++)) || true
+  fi
 
 done <<< "$STACKS"
 
@@ -185,6 +216,7 @@ echo ""
 echo "=== Summary ==="
 echo "  Deleted: $DELETED"
 echo "  Skipped: $SKIPPED"
+echo "  Failed:  $FAILED"
 
 if [[ "$DELETED" -gt 0 && "$DRY_RUN" == "false" ]]; then
   echo ""
--- a/scripts/cleanup-ephemeral-stacks.sh
+++ b/scripts/cleanup-ephemeral-stacks.sh
@@ -5,16 +5,17 @@
 # Handles stuck ENI cleanup (AgentCore/Lambda Hyperplane ENIs) before deletion.
 #
 # Usage:
-#   AWS_PROFILE=abca ./scripts/cleanup-ephemeral-stacks.sh [--dry-run] [--max-age-hours N] [--prefix PREFIX]
+#   AWS_PROFILE=abca ./scripts/cleanup-ephemeral-stacks.sh [--dry-run] [--max-age-hours N] [--prefix PREFIX] [--region REGION]
 #
 # Options:
 #   --dry-run           Show what would be deleted without acting
-#   --max-age-hours N   Delete stacks older than N hours (default: 4)
+#   --max-age-hours N   Delete stacks older than N hours (default: 48)
 #   --prefix PREFIX     Only target stacks matching this prefix (default: all ABCA stacks)
+#   --region REGION     AWS region to operate in (default: $AWS_DEFAULT_REGION or us-east-1)
 #
 # Safety:
 #   - Never touches stacks with termination protection enabled
-#   - Only targets stacks with description matching "ABCA Development Stack"
+#   - Only targets stacks whose description starts with "ABCA Development Stack"
 #   - Skips stacks in UPDATE_IN_PROGRESS or CREATE_IN_PROGRESS states
 
 set -euo pipefail
@@ -27,8 +28,15 @@ REGION="${AWS_DEFAULT_REGION:-us-east-1}"
 while [[ $# -gt 0 ]]; do
   case $1 in
     --dry-run) DRY_RUN=true; shift ;;
-    --max-age-hours) MAX_AGE_HOURS="$2"; shift 2 ;;
-    --prefix) PREFIX="$2"; shift 2 ;;
+    --max-age-hours)
+      [[ $# -ge 2 ]] || { echo "Error: --max-age-hours requires a value" >&2; exit 1; }
+      MAX_AGE_HOURS="$2"; shift 2 ;;
+    --prefix)
+      [[ $# -ge 2 ]] || { echo "Error: --prefix requires a value" >&2; exit 1; }
+      PREFIX="$2"; shift 2 ;;
+    --region)
+      [[ $# -ge 2 ]] || { echo "Error: --region requires a value" >&2; exit 1; }
+      REGION="$2"; shift 2 ;;
     *) echo "Unknown option: $1" >&2; exit 1 ;;
   esac
 done
@@ -62,14 +70,22 @@ echo "  Dry run:       $DRY_RUN"
 echo "  Prefix filter: ${PREFIX:-<none>}"
 echo ""
 
-# List all stacks (excluding deleted ones)
-STACKS=$(aws cloudformation list-stacks \
+# List all stacks (excluding deleted ones).
+# DELETE_FAILED is included so a stack that previously failed to delete is
+# re-targeted on the next run (delete-stack is idempotent and retries it).
+# Capture the exit code separately from emptiness: an API failure (auth,
+# throttle, IAM) must NOT look like "nothing to clean" — that would exit 0
+# and silently skip the whole run.
+if ! STACKS=$(aws cloudformation list-stacks \
   --region "$REGION" \
   --stack-status-filter \
     CREATE_COMPLETE UPDATE_COMPLETE ROLLBACK_COMPLETE \
     UPDATE_ROLLBACK_COMPLETE DELETE_FAILED \
   --query 'StackSummaries[*].[StackName,CreationTime]' \
-  --output text 2>/dev/null)
+  --output text); then
+  echo "Error: cloudformation:ListStacks failed. Check credentials/permissions/region." >&2
+  exit 1
+fi
 
 if [[ -z "$STACKS" ]]; then
   echo "No stacks found."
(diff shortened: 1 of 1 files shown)


scottschreckengaust:
Revived this branch: rebased onto latest `main` (kept your "Update branch" merge `971b1da`) and hardened the script for defense in depth (commit `baaaec0`).

**Security/robustness review** — ran `shellcheck` (exit 0, clean) and `semgrep --config=auto` (0 findings) on the result. Manual review found and fixed:

- **Fail-open age gate (most important):** a parse failure on `CreationTime` previously fell back to epoch `0`, making *every* matching stack look billions of seconds old and eligible for deletion. Now **fails closed** — unparseable time → SKIP.
- **Input validation:** `--max-age-hours` is validated as a non-negative integer before arithmetic (rejects garbage/injected values).
- **Blast-radius visibility:** prints account ID + caller ARN (`sts:GetCallerIdentity`) before any mutation; hard-fails if identity can't be resolved.
- **No mid-run abort:** a single `delete-stack` failure under `set -e` would abort the loop and orphan later stacks — now tolerated, counted, and reported (`Failed:` line).
- Removed dead `--force-eni` flag (SC2034); annotated JMESPath `--query` backticks (SC2016 false positive).

Scope unchanged: still the **manual script only** (Part of #72). EventBridge schedule, CloudWatch audit, CDK construct, and tests are deferred to a future PR.

scottschreckengaust:
Follow-up filed: **#278** — add `shellcheck` to the toolchain (mise + prek hooks + CI). This PR added the first *destructive* shell script and surfaced that shell is the only language surface with no static-analysis gate. Out of scope for this PR (toolchain/CI change, separate governance per ADR-003); tracked separately.

isadeks:
## Structural

This PR ships the script alongside a design doc (`docs/superpowers/specs/2026-06-08-ephemeral-stack-cleanup-eni-aware-design.md`) that explicitly declares the script's "force-detach → `sleep 15` → `delete-stack`" model "structurally broken" for stacks containing a Bedrock AgentCore Runtime. The doc's evidence is solid (`ela-attach` ENIs reject `--force`; `--dry-run` validates IAM only). So we'd be merging code already documented as wrong.

Two clean paths:
1. **Split.** Drop the script from this PR; land the design doc (or fold into #72), then open a follow-up implementing the gated-retry / `Pending`-exit-0 design. My preference.
2. **Implement the redesign now.** Replace lines 146–197 with the read-only ENI gate, add `Pending` branch, ship together.

Either way, ordering is wrong: `delete-stack` first, then **observe** what's stuck, then act on confirmed orphans — not act first, then delete. And the pre-clean is redundant for non-Hyperplane ENIs too, since CloudFormation's own teardown handles regular Lambda VPC ENIs correctly.

## Concrete bugs (fix in either path)

1. **Silent "no stacks" on API failure.** Lines 66-77 — `list-stacks` is wrapped in `2>/dev/null`. If it fails (auth, throttle, IAM), `STACKS` is empty and the script prints "No stacks found." and exits 0. Operator gets the same output for "nothing to clean" and "we don't have permission to ask." Check the exit code separately from emptiness.

2. **BSD `date` loses timezone.** Line 122 — `${CREATION_TIME%%.*}` strips both the fractional seconds and the trailing `Z`. BSD `date -j` then parses as local time. On a Mac in PST, `AGE_SECONDS` is off by 8h, so a 48h-old stack reads as 40h and gets skipped. GNU `date -d` (Linux/CI) handles `Z` correctly, so this is local-Mac-only. Use `date -j -u -f` or keep the timezone in the input.

3. **`delete-stack` failure reason swallowed.** Line 206 — `2>/dev/null` discards the actual API error. The script then prints a generic "ERROR: delete-stack failed (continuing)" but you can't tell AccessDenied from ValidationError. This is the one place you definitely want stderr — drop `2>/dev/null` here.

4. **N+1 API calls per stack.** `list-stacks` returns enough info to pre-filter, but the script then issues `describe-stacks` per stack (line 90) and `list-stack-resources` per stack (line 148). For an account with hundreds of stacks (the case that motivates the script), that's hundreds of round-trips. Tag-based filtering on `list-stacks` plus a batched `describe-stacks` would be O(1) at API level.

5. **`MAX_AGE_HOURS` default mismatch.** Header comment says `default: 4`, code is `:-48`. Pick one.

6. **Argument parsing edge case.** `--max-age-hours`/`--prefix` shift 2 without verifying `$2` exists. With `set -u`, calling `./script --max-age-hours` errors as "`$2`: unbound variable" instead of a useful message.

7. **No `--region` / `--profile` flag** though the header docs `AWS_PROFILE=abca`. Drift between docs and code.

8. **Account ID leak.** Spec doc names `46**********` and `us-east-1` as the example stuck-stack environment. Replace with "the shared dev account."

9. **Description-equality filter is fragile.** `[[ "$DESCRIPTION" != "ABCA Development Stack" ]]` (line 101) silently disqualifies stacks if the description is human-edited or renamed. Tag-based filtering would survive both. Not a blocker but a TODO.

10. **No tests.** The spec doc itself recommends extracting `classify_stack(status, eni_count) → action` as a pure function so the load-bearing branch (`DELETE_FAILED` + 0 ENIs → retry; `DELETE_FAILED` + N ENIs → `Pending`) is unit-testable without AWS. Worth doing before this becomes an EventBridge schedule.

11. **#72 governance.** Spec doc flags #72 as "not yet `approved`"; per AGENTS.md the rule is approve-first. Worth pinging an admin to label.

## What the script gets right (don't lose these in a rewrite)

- `set -euo pipefail` — proper strict mode.
- `sts:GetCallerIdentity` blast-radius print before any mutation.
- Numeric validation on `--max-age-hours` (line 37).
- BSD/GNU `date` fallback chain with fail-closed parse (line 123 — modulo the timezone bug above).
- `*IN_PROGRESS*` skip is narrow enough that `DELETE_FAILED` is *not* matched — the spec doc identifies this distinction as load-bearing for retry. Pin with a regression test in the next pass.
- Counter increments inside `<<<` (current shell), not a pipe, so counters actually persist.
- `((counter++)) || true` survives starting-from-zero (worth a one-line comment so a future "simplification" doesn't break it).
- Termination-protection check is unconditional and early.

laithalsaadoon:
**🔴 Secret scan: a real AWS account ID is committed on this branch** (flagged by `gitleaks` in the `main` Security suite — [run 27431593754](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/actions/runs/27431593754)).

@scottschreckengaust — the weekly/dispatch `mise run security` job scans **all branch history** (`gitleaks detect --source .`), so this branch's commit is keeping it red even though `main` itself is clean.

**Finding**
- **Rule:** `aws-account-id`
- **File:** `docs/superpowers/specs/2026-06-08-ephemeral-stack-cleanup-eni-aware-design.md`, **line 14** — `A real ephemeral stack (\`scoschre\`, account \`<12-digit-id>\`, us-east-1) entered DELETE_FAILED…`
- **Introduced by:** commit `aac6e12` (authored by `bgagent`)
- This is a genuine 12-digit account ID, not a placeholder — same incident class as #313.

**Remediation** (it's in committed history, so a working-tree edit alone won't clear the scan):
1. Replace the ID with a documentation placeholder — `111122223333` (AWS's reserved example account) — or redact it (`<redacted>`).
2. Because it's already in this branch's history, rewrite the commit that introduced it so the value is gone from the branch entirely — e.g. `git rebase -i aac6e12~1` and amend that commit, or `git filter-repo --replace-text`. A new commit that only edits the current file leaves the ID reachable in history and the scan stays red.
3. Force-push the cleaned branch; the Security suite should then go green once this and `feat/247-reconciler` (a separate false-positive) are clear.

Happy to prep the exact rewrite commands or a placeholder patch if useful — but the history rewrite is yours to run since it's your branch. Flagging here so it's scrubbed **before** #109 merges, not after it reaches `main`.

_— filed by Bonk during the ABCA validation-loop review; a follow-up issue to make this a per-PR required check (so secrets are caught at PR time, not weekly) is being filed separately._

docs/superpowers/specs/2026-06-08-ephemeral-stack-cleanup-eni-aware-design.md
```diff
@@ -0,0 +1,132 @@
+# Ephemeral Stack Cleanup — AgentCore ENI-Aware Redesign
```
krokoko: this file shouldn't be there

scripts/cleanup-ephemeral-stacks.sh
```diff
@@ -0,0 +1,225 @@
...
+#
+# Options:
+#   --dry-run           Show what would be deleted without acting
+#   --max-age-hours N   Delete stacks older than N hours (default: 4)
+#   --prefix PREFIX     Only target stacks matching this prefix (default: all ABCA stacks)
+#
+# Safety:
+#   - Never touches stacks with termination protection enabled
+#   - Only targets stacks with description matching "ABCA Development Stack"
+#   - Skips stacks in UPDATE_IN_PROGRESS or CREATE_IN_PROGRESS states
```
scottschreckengaust, line 19: What about stacks with `DELETE_IN_PROGRESS`?

scripts/cleanup-ephemeral-stacks.sh
```diff
@@ -0,0 +1,225 @@
...
+echo "  Dry run:       $DRY_RUN"
+echo "  Prefix filter: ${PREFIX:-<none>}"
+echo ""
+
+# List all stacks (excluding deleted ones)
+STACKS=$(aws cloudformation list-stacks \
+  --region "$REGION" \
+  --stack-status-filter \
+    CREATE_COMPLETE UPDATE_COMPLETE ROLLBACK_COMPLETE \
+    UPDATE_ROLLBACK_COMPLETE DELETE_FAILED \
```
scottschreckengaust, line 83: Then how will stacks that are `DELETE_FAILED` get re-run?

scripts/cleanup-ephemeral-stacks.sh
```diff
@@ -0,0 +1,225 @@
...
+    --stack-name "$STACK_NAME" \
+    --query 'Stacks[0].[Description,EnableTerminationProtection,StackStatus]' \
+    --output text 2>/dev/null) || continue
+
+  DESCRIPTION=$(echo "$STACK_INFO" | cut -f1)
+  TERMINATION_PROTECTED=$(echo "$STACK_INFO" | cut -f2)
+  STATUS=$(echo "$STACK_INFO" | cut -f3)
+
+  # Only target stacks from this CDK app
+  if [[ "$DESCRIPTION" != "ABCA Development Stack" ]]; then
```
scottschreckengaust: Shouldn't this be what the description is started with?

docs/superpowers/specs/2026-06-08-ephemeral-stack-cleanup-eni-aware-design.md
scottschreckengaust: deleted

scottschreckengaust:
Pushed `a845fbb` (force-pushed — history rewritten, see below). Addressed the **secret-scan blocker** and the **concrete bugs**, while keeping the PR's intent (manual script only). Structural ENI redesign and the larger items stay deferred to a follow-up per the original scope.

## 🔴 Secret-scan blocker (resolved) — @theagenticguy

The leaked account ID lived in the spec doc `docs/superpowers/specs/2026-06-08-...md`, introduced by commit `aac6e12`. Both reviewers also flagged that file shouldn't be in the PR at all (@krokoko: "this file shouldn't be there"). So rather than a placeholder swap, I **dropped both spec-doc commits** (`aac6e12` add + `519a9b0` delete — they net to zero) via `git rebase --onto origin/main`, which also flattened the stale `Merge branch 'main'` commits. The branch is now 4 clean commits on latest `main`.

Verified: `git log -p origin/main..HEAD | grep -cE 'account \`[0-9]{12}\`'` → **0**, and the spec file is absent from the entire branch history. Force-pushed with `--force-with-lease`. The weekly `gitleaks detect --source .` should go green for this branch.

## Concrete bugs fixed (review by @isadeks)

| # | Fix |
|---|-----|
| 1 | **Silent "no stacks" on API failure** — `list-stacks` no longer swallows stderr; exit code is checked separately from emptiness, so auth/throttle/IAM failure exits 1 with a clear error instead of "No stacks found" / exit 0. |
| 2 | **BSD `date` timezone** — added `-u` to the BSD branch so UTC `CreationTime` isn't parsed as local (was up to 8h off near the boundary on a non-UTC host). |
| 3 | **`delete-stack` error swallowed** — dropped `2>/dev/null` so AccessDenied vs ValidationError vs throttling is distinguishable; this is the one call where the API error matters most. |
| 5 | **Default mismatch** — header comment `default: 4` → `48` to match code. |
| 6 | **Arg-parsing edge case** — `--max-age-hours`/`--prefix`/`--region` now verify a value follows before `shift 2` (clean error instead of `set -u` "unbound variable"). |
| 7 | **`--region` flag** — added (the `REGION` var already existed; closes doc/code drift). |
| 9 | **Description filter fragility** — exact-match → **prefix-match**. This is now load-bearing: **#293 merged** and changed the description to `'ABCA Development Stack (uksb-wt64nei4u6)'`, so the old `!= "ABCA Development Stack"` matched **zero** stacks. Prefix-match survives both the suffix present and stripped. (Tag-based filtering would be even more robust — see deferred items.) |

`shellcheck` clean, `bash -n` clean.

## Deferred (out of scope for "manual script only" — replies below)

- **Structural ENI redesign** (force-detach→sleep→delete is wrong for AgentCore `ela-attach` ENIs): the design doc that documented this has been removed from the PR; the redesign (read-only ENI gate, `DELETE_FAILED`+0 ENIs → retry, +N ENIs → `Pending`-exit-0) belongs in the follow-up PR under #72/#111. Agreed with @isadeks's preference to **split**.
- **#4 N+1 API calls / tag-based filtering**, **#9 tag filter**, **#10 `classify_stack` pure-function + tests** — all the right call *before this becomes an EventBridge schedule*; deferred to the same follow-up. Filing a tracking note.
- **#11 #72 approval** — pinging admin per ADR-003.

See inline replies for the CloudFormation-status questions (grounded in the CFN + Resource Groups Tagging API docs).

docs/superpowers/specs/2026-06-08-ephemeral-stack-cleanup-eni-aware-design.md
scottschreckengaust: Agreed and resolved. The spec doc has been removed from the branch entirely — not just deleted at HEAD, but its introducing commit (`aac6e12`) was dropped via history rewrite, because it also carried the leaked AWS account ID flagged by gitleaks. The file is now absent from the full branch history (`git log origin/main..HEAD -- docs/superpowers/specs/` → empty). The design content it held will live in the follow-up redesign PR under #72/#111.

scripts/cleanup-ephemeral-stacks.sh
scottschreckengaust: `DELETE_IN_PROGRESS` is intentionally **not** in the `--stack-status-filter` list (line ~74), so such stacks are never enumerated and the script can't act on them — correct, since CloudFormation is already tearing them down and a second `delete-stack` would be a no-op at best.

Grounding (CloudFormation stack-status semantics): `DELETE_IN_PROGRESS` is a transient state CFN owns; the terminal outcomes are `DELETE_COMPLETE` (stack disappears from the default `list-stacks` view) or `DELETE_FAILED` (stays listable and re-deletable). So filtering to the "settled" statuses + `DELETE_FAILED` is the right set: we skip what's mid-flight and re-target what got stuck. I added a comment on the filter block making the `DELETE_FAILED` inclusion explicit.

scripts/cleanup-ephemeral-stacks.sh
scottschreckengaust: Via the `--stack-status-filter` including `DELETE_FAILED` (line ~78). A stack that failed to delete stays in `DELETE_FAILED` (a terminal-but-listable state), so the next run re-enumerates it and re-issues `delete-stack` — which is idempotent for a stuck stack and retries the teardown. I made this explicit with a comment on the list-stacks block:

> `# DELETE_FAILED is included so a stack that previously failed to delete is re-targeted on the next run (delete-stack is idempotent and retries it).`

Note the `*IN_PROGRESS*` skip (line ~129) is deliberately narrow so it does **not** match `DELETE_FAILED` — that distinction is load-bearing for retry, and the deferred redesign will add a regression test pinning it (review #10).

scripts/cleanup-ephemeral-stacks.sh
scottschreckengaust: Yes — fixed in `a845fbb`. Changed exact-equality to a prefix match:

```bash
if [[ "$DESCRIPTION" != "ABCA Development Stack"* ]]; then
```

This is now load-bearing: **#293 merged** and set the description to `'ABCA Development Stack (uksb-wt64nei4u6)'` (`cdk/src/main.ts:41`), so the old `!= "ABCA Development Stack"` exact check matched **zero** stacks. Prefix-match works whether the `(uksb-…)` solution-id suffix is present or stripped. (Tag-based filtering would be more robust still — noted as a follow-up; deferred with the other API-shape items.)

docs/superpowers/specs/2026-06-08-ephemeral-stack-cleanup-eni-aware-design.md
scottschreckengaust: Confirmed — the file is gone from the branch history (not just the working tree). Its introducing commit was dropped in the `a845fbb` rewrite to also clear the committed account ID. Safe to resolve.

theagenticguy (approved):
## ✅ Approved — secret-scan blocker resolved, in-scope bugs fixed

Re-reviewed `a845fbb` end to end. The 🔴 secret-scan blocker I filed is cleared, every in-scope concrete bug from @isadeks's review is fixed, and the script is clean for its stated scope (**manual script only, Part of #72**).

### Secret-scan blocker — verified resolved
Cloned the branch fresh and checked full history, not just the working tree:
- Branch is **4 clean commits** on latest `main`, touching exactly one file (`scripts/cleanup-ephemeral-stacks.sh`, +254).
- `git log -p origin/main..HEAD | grep -E '\b[0-9]{12}\b'` → **0 matches**. No 12-digit account ID in any blob reachable from the branch.
- The spec doc that carried the leak (`docs/superpowers/specs/2026-06-08-...md`) is **absent from the entire branch history**, not just deleted in a later commit. Dropping both spec-doc commits via `rebase --onto` was the right call — it also addressed @krokoko's "this file shouldn't be there."

`security-pr` and the PR-time secret scan are green on this branch.

### Concrete bugs — verified fixed in code
| # | Finding | Status |
|---|---------|--------|
| 1 | Silent "no stacks" on API failure | ✅ `if ! STACKS=$(...)` checks exit code separately from emptiness; auth/throttle/IAM failure exits 1 |
| 2 | BSD `date` drops timezone | ✅ `-u` added to the BSD branch (`date -j -u -f`); UTC `CreationTime` no longer parsed as local |
| 3 | `delete-stack` error swallowed | ✅ `2>/dev/null` dropped on the delete call; AccessDenied vs ValidationError vs throttle now distinguishable |
| 5 | Default mismatch (`4` vs `48`) | ✅ Header comment now reads `48` |
| 6 | Arg-parse `shift 2` under `set -u` | ✅ `[[ $# -ge 2 ]]` guard before each `shift 2` |
| 7 | No `--region` flag | ✅ Added |
| 9 | Exact-match description filter | ✅ Prefix-match. Load-bearing: #293 changed the description to add the `(uksb-...)` suffix, so the old exact-match would have matched **zero** stacks |

Static checks reproduced locally: `shellcheck` clean (exit 0), `bash -n` clean. All CI green (CodeQL, Analyze ×3, build, lint).

### Deferred — agreed out of scope for "manual script only"
These are correctly tracked for the follow-up under #72/#111, not blockers here:
- **Structural ENI redesign** (force-detach→`sleep 15`→delete is ineffective for AgentCore `ela-attach` ENIs). The current logic is `|| true`-tolerated and idempotent: a stack that hits `DELETE_FAILED` from a stuck Hyperplane ENI is simply re-targeted next run, so it is safe, just not always effective for AgentCore stacks. Matches @isadeks's preference to split.
- **#4 N+1 API calls / tag-based filtering**, **#10 `classify_stack` pure function + unit tests** — both belong before this becomes an EventBridge schedule. Agreed.

### One non-code gate before merge
Per `AGENTS.md` approve-first (ADR-003), **#72 still carries no `approved` label** (state: OPEN, labels: enhancement/infra-cdk/ci-cd/cost-finops/P1). That governance ping is the remaining merge gate — the code itself is ready. `mergeStateStatus` is `BLOCKED` on `REVIEW_REQUIRED`, which this approval addresses on the code side.

Nice work on the history rewrite and the fail-closed age gate. Approving.

— reviewed by Bonk

Issue #126, opened 2026-05-19
feat(bootstrap): live-account preflight validator


scottschreckengaust:
## Parent

Sub-issue 6 of #120 (RFC: Least-privilege CDK bootstrap policies as code)

## Branch strategy

`feat/bootstrap-preflight` → targets `feat/bootstrap-aspect`

## Summary

Implement a validator that checks the deployed CDKToolkit stack's CloudFormation outputs against the required bootstrap version/hash before deploying. This prevents the "pull latest, deploy, fail 15 minutes later" failure mode.

## Deliverables

- [ ] Create `cdk/src/bootstrap/preflight/validator.ts`:
  - Reads CDKToolkit stack CF outputs (`BootstrapPolicyVersion`, `BootstrapPolicyHash`) via `CloudFormation.describeStacks`
  - Compares against required version/hash from the app's policy code
  - Returns structured result:
    - **PASS**: version matches, hash matches → safe to deploy
    - **WARN**: version compatible but hash mismatch → possible console drift, recommend re-bootstrap
    - **FAIL**: version too old → lists missing permissions, provides exact remediation command
    - **FAIL**: no outputs found → default bootstrap detected (AdministratorAccess), provide bootstrap command
  - Outputs both structured JSON (for CI parsing) and human-readable summary (for terminal)
- [ ] Create `cdk/test/bootstrap/validator.test.ts`:
  - Mock `CloudFormation.describeStacks` responses for each scenario
  - Version match → PASS
  - Hash mismatch with compatible version → WARN with advisory message
  - Version too old → FAIL with specific missing actions listed
  - No CDKToolkit stack → FAIL with "not bootstrapped" message
  - No version/hash outputs → FAIL with "default bootstrap detected" message
  - AWS SDK error (permissions, network) → clear error message
- [ ] Add `mise //cdk:preflight` task in `cdk/mise.toml`:
  - Runs the validator
  - Exits 0 on PASS or WARN
  - Exits 1 on FAIL
  - Prints remediation: exact `mise //cdk:bootstrap` command to fix

## Key design decisions

- Requires only `cloudformation:DescribeStacks` permission (read-only, safe for CI with minimal credentials)
- WARN does **not** block deploy — hash drift from console edits is advisory, not fatal
- FAIL prints the exact `mise //cdk:bootstrap` command needed (copy-paste friendly)
- Validator is importable as a library (for integration tests) and runnable as a CLI entry point (for mise task)
[...]

Issue #125, opened 2026-05-19
feat(bootstrap): CDK Aspect for policy envelope checking


scottschreckengaust:
## Parent

Sub-issue 5 of #120 (RFC: Least-privilege CDK bootstrap policies as code)

## Branch strategy

`feat/bootstrap-aspect` → targets `feat/bootstrap-action-map`

## Summary

Implement a CDK Aspect that runs during `cdk synth` and validates every CfnResource's required IAM actions are covered by the declared bootstrap policy envelope. This catches "your new construct needs permissions the bootstrap role doesn't have" at dev time — before a failed deploy.

## Deliverables

- [ ] Create `cdk/src/bootstrap/preflight/aspect.ts`:
  - `BootstrapPolicyAspect` implementing `IAspect`
  - Visits all `CfnResource` nodes in the construct tree
  - Looks up each resource type in the resource-action-map
  - Compares required actions against the combined action set from all three policies
  - Emits `Annotations.of(node).addError()` for actions absent from all policy versions
  - Emits `Annotations.of(node).addWarning()` for:
    - Actions in a newer policy version than currently declared
    - Resource types not found in the map (unknown/unmapped)
- [ ] Register Aspect in `cdk/src/main.ts` alongside existing `AwsSolutionsChecks`:
  ```typescript
  Aspects.of(app).add(new BootstrapPolicyAspect());
  ```
- [ ] Create `cdk/test/bootstrap/aspect.test.ts`:
  - Covered resource type → no annotations (pass)
  - Uncovered resource type (requires action not in any policy) → error annotation
  - Unknown resource type (not in map) → warning annotation
  - Policy version mismatch (action exists in v2 but current is v1) → warning with upgrade message
  - Validates Aspect does not interfere with cdk-nag Aspect
- [ ] Verify: `mise //cdk:synth` passes cleanly with current policies and current resources

## Key design decisions

- Aspect runs **after** cdk-nag in the Aspect registration order (cdk-nag validates runtime IAM; this validates deployment IAM — different concerns)
- **Error** = deploy will fail. **Warning** = deploy may succeed but policy is stale.
- Aspect is **enabled by default** but can be disabled via CDK context flag (`bootstrapPolicyCheck: false`) for escape-hatch scenarios
- Does not modify the template — read-only inspection only

## Acceptance criteria

- `mise //cdk:synth` passes with zero errors on current codebase (all resources are covered by current policies)
[...]

Issue #123, opened 2026-05-19
feat(bootstrap): custom bootstrap template generation


scottschreckengaust:
## Parent

Sub-issue 3 of #120 (RFC: Least-privilege CDK bootstrap policies as code)

## Branch strategy

`feat/bootstrap-template` → targets `feat/bootstrap-policies`

## Summary

Generate a custom CDK bootstrap CloudFormation template that replaces `AdministratorAccess` with the three IaCRole-ABCA policies, and adds version/hash outputs for preflight validation.

## Deliverables

- [ ] Create `cdk/src/bootstrap/bootstrap-template.ts` — programmatically generates the custom CF template:
  - Based on CDK's default bootstrap template structure
  - Substitutes `AdministratorAccess` on the CloudFormation execution role with the three managed policies
  - Adds CF outputs: `BootstrapPolicyVersion`, `BootstrapPolicyHash`
  - Retains all other default resources (S3 staging bucket, ECR repo, deploy/lookup/publishing roles)
- [ ] Create `cdk/test/bootstrap/bootstrap-template.test.ts`:
  - Generated template is valid YAML/JSON
  - Execution role references exactly three managed policies (not AdministratorAccess)
  - `BootstrapPolicyVersion` and `BootstrapPolicyHash` outputs are present
  - All default bootstrap resources still exist
  - Template parameters match CDK bootstrap expectations
- [ ] Generate `cdk/bootstrap/bootstrap-template.yaml` — the operator-consumable artifact
- [ ] Add `mise //cdk:bootstrap:generate` task in `cdk/mise.toml` — regenerates all `cdk/bootstrap/` artifacts
- [ ] Add `mise //cdk:bootstrap` task — wraps `cdk bootstrap --template cdk/bootstrap/bootstrap-template.yaml`

## Key design decisions

- Template is **generated** (not hand-edited) so it stays in sync with policy source code
- Uses CF pseudo-parameters (`AWS::AccountId`, `AWS::Region`) for account/region — no hardcoding
- Managed policies are created inline in the template (not referencing external ARNs) — eliminates the "create policies first, then bootstrap" two-step
- Qualifier remains `hnb659fds` (CDK default) for compatibility with existing deployments

## Acceptance criteria

- `mise //cdk:bootstrap:generate` produces a valid CF template
- Template passes `cfn-lint` validation (if available)
- Existing `mise //cdk:synth` and `mise //cdk:test` still pass (no regression)
[...]

Issue #120, opened 2026-05-19
RFC: Least-privilege CDK bootstrap policies as code (with preflight validation)


scottschreckengaust:
## Summary

Convert the three hand-documented IaCRole-ABCA policies (`DEPLOYMENT_ROLES.md`) into versioned, testable TypeScript code under `cdk/src/bootstrap/`. Provide a custom CDK bootstrap template that embeds these policies with per-compute-variant selection. Add a two-layer preflight validation system: a CDK Aspect that warns at synth time when the template exceeds the declared policy envelope, and a `mise //cdk:preflight` task that validates the live account's bootstrap state before deploy. Use stacked PRs for incremental review.

## Use case and motivation

**Who it's for:** Operators deploying ABCA, CI/CD pipelines (`deploy.yml`), and agents modifying CDK stacks.

**Problem today:**
1. `cdk bootstrap` grants `AdministratorAccess` to the CloudFormation execution role — violates least-privilege and may breach organizational compliance gates (SCPs, permission boundaries).
2. The three IaCRole-ABCA policies exist only as JSON blobs in a design doc. An operator must manually create them in IAM, then re-bootstrap — error-prone and unversioned.
3. When a new release adds resources (e.g., SQS queue, Step Functions), operators who pull-and-deploy hit a mid-rollback failure because their bootstrap policy predates the new permissions. No warning until CloudFormation fails 15 minutes into the update.

**After this RFC:**
- Policies are code: versioned, tested, diffable in PRs.
- `mise //cdk:synth` warns immediately when a CDK construct requires permissions not covered by the declared policy.
- `mise //cdk:preflight` (and CI) verifies the deployed bootstrap version is compatible before `cdk deploy` runs — failing fast with an actionable message ("re-bootstrap required: v1.2 → v2.0, adds SQS and StepFunctions permissions").
- Operators run a single `mise //cdk:bootstrap` to provision least-privilege roles.

### Location

- **Policy source code** — `cdk/src/bootstrap/policies/` — co-located with the CDK app for testability, agent routing, Aspect import at synth-time, and self-containment.
- **Template generator** — `cdk/scripts/generate-bootstrap-template.ts` — build-time script, not part of CDK synthesis (bootstrap is an operational prerequisite, not an app construct).
- **Generated artifacts** — `cdk/bootstrap/` — committed, operator-consumable.

### Policy architecture: core + compute variants

| Policy | Always required? | Scope |
[...]

PR #165, opened 2026-05-21
feat(bootstrap): resource-action-map for synth-time validation

### LINKED ISSUE #124: feat(bootstrap): resource-action-map for synth-time validation
## Parent

Sub-issue 4 of #120 (RFC: Least-privilege CDK bootstrap policies as code)

## Branch strategy

`feat/bootstrap-action-map` → targets `feat/bootstrap-template`

## Summary

Create a mapping from CloudFormation resource types to the IAM actions required to manage them (CRUD lifecycle), scoped to the ~30 resource types used by this app. This map powers the synth-time Aspect (Sub-issue 5) and informs the preflight validator (Sub-issue 6).

## Deliverables

- [ ] Create `cdk/src/bootstrap/preflight/resource-action-map.ts`:
  - Export a typed map: `Record<string, { create: string[], read: string[], update: string[], delete: string[] }>`
  - Cover all CF resource types present in `cdk.out/` after synth (e.g., `AWS::DynamoDB::Table`, `AWS::Lambda::Function`, `AWS::ApiGateway::RestApi`, `AWS::EC2::VPC`, `AWS::Cognito::UserPool`, `AWS::WAFv2::WebACL`, `AWS::Bedrock::Guardrail`, etc.)
  - Each entry's action list is the minimum CloudFormation needs to manage the resource lifecycle
- [ ] Create `cdk/src/bootstrap/preflight/index.ts` — barrel export
- [ ] Create `cdk/test/bootstrap/resource-action-map.test.ts`:
  - All resource types in the synthesized template have a mapping entry (synth → extract types → assert coverage)
  - Each action list is non-empty
  - All actions use valid IAM action format (`service:ActionName`)
  - Combined action set is a subset of the three policy documents' action sets (the map doesn't require more than what policies allow)

## Key design decisions

- **Scoped to this app's resources** — not all 800+ AWS CF resource types. Keeps maintenance tractable.
- Actions derived from [CloudFormation resource type documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-template-resource-type-ref.html) and validated against CloudTrail from prior deployments.
- Unknown resource types (not in map) produce a **warning**, not an error — allows new constructs to be added without immediately updating the map.
- Map is a plain TypeScript object (not loaded from JSON) — benefits from IDE autocomplete and type checking.

## Acceptance criteria

- `mise //cdk:synth` followed by resource-type extraction shows 100% map coverage of current template
- All tests pass
- No circular imports between `preflight/` and `policies/`

### LINKED ISSUE #164: refactor(compute): gate ECS construct on compute_type context instead of comment toggle
## Summary

Replace the commented-out ECS construct in `cdk/src/stacks/agent.ts` with a proper `compute_type` context gate.

## Problem

Currently, enabling ECS requires manually uncommenting code in `agent.ts` (lines ~46, ~524-533). This:
- Requires source code changes to change deployment topology
- Is invisible to the bootstrap/preflight system (commented code doesn't synthesize)
- Defeats the compute-variant architecture established in #123

## Solution

```typescript
const computeType = app.node.tryGetContext('compute_type') ?? 'agentcore';

// In the stack:
if (computeType === 'ecs') {
  new EcsAgentCluster(this, 'EcsAgentCluster', { vpc, ... });
}
```

The construct is always in source, always compilable, always tested. The context variable governs synthesis:
- `mise //cdk:synth` → agentcore only (default)
- `mise //cdk:synth -- -c compute_type=ecs` → agentcore + ECS resources

## Acceptance criteria

- [ ] ECS construct uncommented and gated behind `if (computeType === 'ecs')`
- [ ] `mise //cdk:synth` (default) produces same template as today (no ECS resources)
- [ ] `mise //cdk:synth -- -c compute_type=ecs` produces template with ECS resources
- [ ] `mise //cdk:test` passes for both configurations
- [ ] No "uncomment to enable" prose remains in the file

## Relationship to #120 stack

Prerequisite for #124 (resource-action-map) — the synth-coverage test needs to run with `compute_type=ecs` to verify ECS resource types are mapped. Without this issue, ECS resources never appear in the synthesized template.

Blocked by: #123 (compute-ecs policy must exist)
Blocks: #124 (ECS coverage in resource-action-map)

scottschreckengaust:
## Summary

Closes #124
Closes #164

Creates a mapping from CloudFormation resource types to required IAM actions (CRUD lifecycle), scoped to all resource types in this app's synthesized template. Introduces `getRequiredBootstrapPolicies()` for downstream consumption by the Aspect (#125) and preflight validator (#126). Gates ECS construct on `compute_type` context variable (replaces comment toggle).

## Stack position

PR 5 for #120 — least-privilege CDK bootstrap policies as code

## Key decisions
- **ECS context gate (#164):** Construct is always in source, compute_type governs synthesis — no commenting/uncommenting
- **getRequiredBootstrapPolicies(computeType):** Single function declaring what the app needs, consumed by Aspect and preflight
- **Dual-config synth-coverage test:** Validates map completeness for both agentcore and ecs configurations
- **Map scoped to this app resources (~60 types):** Unknown types produce warnings, not errors
- **All map actions within configured policy set:** Test enforces the map never requires more than policies allow

## Deliverables
- [ ] Gate ECS construct on compute_type context (#164)
- [ ] cdk/src/bootstrap/required-policies.ts
- [ ] cdk/src/bootstrap/preflight/resource-action-map.ts
- [ ] cdk/src/bootstrap/preflight/index.ts
- [ ] Tests: required-policies, resource-action-map, dual-config synth-coverage
- [ ] Investigate SQS gap (queue/policy in template, no SQS actions in policies)

## Test plan
- [ ] All existing CDK tests pass
- [ ] Map covers all resource types in both synth configurations
- [ ] All mapped actions exist in the combined policy set (wildcard-aware)
- [ ] getRequiredBootstrapPolicies returns correct sets for each compute type
- [ ] tsc --noEmit compiles cleanly
- [ ] No circular imports between preflight/ and policies/

## Open questions
- SQS: AWS::SQS::Queue is in the template but no policy has SQS actions — needs investigation (may require policy update + version bump)

## Implementation plan
See: docs/superpowers/plans/2026-05-21-resource-action-map.md

Blocked by: #123 (PR #162)
References: RFC #120, ADR-002


scottschreckengaust:
```table
┌─────────┬──────┬───────────────────────────────────────────┐
  │ Commit  │ Task │                   What                    │
  ├─────────┼──────┼───────────────────────────────────────────┤
  │ d3a9804 │ 0    │ ECS context gate (closes #164)            │
  ├─────────┼──────┼───────────────────────────────────────────┤
  │ 5ed8db3 │ 1    │ getRequiredBootstrapPolicies(computeType) │
  ├─────────┼──────┼───────────────────────────────────────────┤
  │ 83099e1 │ 2    │ Resource-action-map (57 CF types)         │
  ├─────────┼──────┼───────────────────────────────────────────┤
  │ ed0cf6b │ 3    │ Dual-config synth-coverage test           │
  └─────────┴──────┴───────────────────────────────────────────┘
```

*Note:* this branch currently sits on top of feat/bootstrap-template (#162). When #162 merges to main, I'll retarget and rebase per ADR-001 §8 — the scaffold commit
  (f46cfb7) will be skippable and the #123 commits will drop out, leaving just the 4 clean #124 commits on main.

cdk/src/bootstrap/required-policies.ts
```diff
@@ -0,0 +1,36 @@
...
+ *  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ *  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ *  SOFTWARE.
+ */
+
+const CORE_POLICIES = [
+  'infrastructure',
+  'application',
+  'observability',
+  'compute-agentcore',
```
scottschreckengaust: `compute-agentcore` should not be a core policy

cdk/src/bootstrap/required-policies.ts
```diff
@@ -0,0 +1,36 @@
...
+] as const;
+
+const COMPUTE_VARIANT_POLICIES: Record<string, string[]> = {
+  ecs: ['compute-ecs'],
```
scottschreckengaust, line 28: these are the actual compute variants:

```suggestion
  agentcore: ['compute-agentcore'],
  ecs: ['compute-ecs'],
```

cdk/src/bootstrap/required-policies.ts
scottschreckengaust: Fixed in 3128560 — moved compute-agentcore out of CORE_POLICIES. Core is now just [infrastructure, application, observability]. Each compute variant is an independent choice.

cdk/src/bootstrap/required-policies.ts
scottschreckengaust: Applied in 3128560 — exactly this structure. agentcore and ecs are peer variants, not base+optional.

cdk/src/stacks/agent.ts
```diff
@@ -519,26 +518,25 @@ export class AgentStack extends Stack {
...
-    //   vpc: agentVpc.vpc,
-    //   agentImageAsset,
-    //   taskTable: taskTable.table,
-    //   taskEventsTable: taskEventsTable.table,
-    //   userConcurrencyTable: userConcurrencyTable.table,
-    //   githubTokenSecret,
-    //   memoryId: agentMemory.memory.memoryId,
-    // });
+    // --- ECS Fargate compute backend (enabled when compute_type=ecs) ---
+    const computeType = this.node.tryGetContext('compute_type') ?? 'agentcore';
```
krokoko, line 643: why do we need to have this uncommented ?

scottschreckengaust: Later on (lines 523-539) `computeType` to use it (rather than a comment/uncomment)

krokoko: ah yes I missed that we were reading the context, I thought we were always going to provision the ecs version, all good

krokoko:
on it

krokoko:
## Review — `feat(bootstrap): resource-action-map for synth-time validation`

Thanks for this — the direction is great. Converting the bootstrap permissions into a versioned, testable map advances the "bounded blast radius" tenet from RFC #120, and replacing the comment-toggle with a real `compute_type` context gate (#164) is a clear improvement that keeps the ECS path always-compilable. The data model is clean and the pure-function unit tests are genuinely strong.

I do think a few things should be addressed before merge. Verdict: **request changes** — all correctable in a focused follow-up.

### Blocking

**B1 — The map is missing the ECS resource types this PR enables, and the included test is red.**
`EcsAgentCluster` synthesizes `AWS::ECS::Cluster` and `AWS::ECS::TaskDefinition` (`cdk/src/constructs/ecs-agent-cluster.ts:79,108`), but `RESOURCE_ACTION_MAP` has no `AWS::ECS::*` entry and neither type is in the test's `SKIP_TYPES`. Running `cdk synth -c compute_type=ecs` and then the suite fails:
```
Test "all ecs resource types have map entries"
  Received + ["AWS::ECS::Cluster", "AWS::ECS::TaskDefinition"]
```
The sibling policy `compute-ecs.ts` already grants the matching `ecs:*` actions, so the map is the only thing out of sync. Suggested fix — add both entries, deriving actions from `compute-ecs.ts`:
```ts
'AWS::ECS::Cluster': {
  create: ['ecs:CreateCluster', 'ecs:TagResource', 'ecs:PutClusterCapacityProviders'],
  read:   ['ecs:DescribeClusters', 'ecs:ListTagsForResource'],
  update: ['ecs:UpdateCluster', 'ecs:UpdateClusterSettings', 'ecs:PutClusterCapacityProviders', 'ecs:TagResource', 'ecs:UntagResource'],
  delete: ['ecs:DeleteCluster'],
},
'AWS::ECS::TaskDefinition': {
  create: ['ecs:RegisterTaskDefinition', 'ecs:TagResource'],
  read:   ['ecs:DescribeTaskDefinition', 'ecs:ListTaskDefinitions', 'ecs:ListTagsForResource'],
  update: ['ecs:RegisterTaskDefinition', 'ecs:TagResource', 'ecs:UntagResource'],
  delete: ['ecs:DeregisterTaskDefinition'],
},
```

**B2 — Both "Synth coverage" tests pass/skip silently, so the map's only drift-defense isn't load-bearing.** (`cdk/test/bootstrap/resource-action-map.test.ts`)
- The agentcore test reads `cdk.out/backgroundagent-dev.template.json`; if it's missing → `if (types.length === 0) return;` runs **zero assertions**. `cdk.out/` is gitignored and `cdk:test` depends only on `:compile` (not `:synth`), running parallel to `:synth:quiet` — so on a clean CI runner the template may not exist yet and the test passes vacuously.
- The ECS test wraps the whole synth in a bare `catch { …; return; } // skip gracefully`, which swallows *every* failure mode (a real synth/cdk-nag regression, a timeout, an `npx` blip) and reports green. This is most likely why CI is green while the test is actually red locally — CI hits the skip path.

Suggested fix: synthesize in-process so the tests are self-contained, and add a non-vacuous guard:
```ts
expect(types.length).toBeGreaterThan(0); // fail if synth produced nothing
expect(types.filter(t => !SKIP_TYPES.has(t) && !RESOURCE_ACTION_MAP[t])).toEqual([]);
```
For the ECS shell-out, capture the error and `throw` with stderr rather than `return`; if a no-tooling skip is genuinely needed, gate it on an explicit env flag and use `it.skip` so the skip is *reported*, not hidden.

**B3 — The `compute_type` gate (the actual behavioral change) has no direct test.** (`cdk/src/stacks/agent.ts:566-619`)
Nothing asserts that default context produces *no* ECS resources, that `compute_type=ecs` produces them, or that `ecsConfig` is wired into `TaskOrchestrator` only in the ECS case. The touched assertion in `github-tags.test.ts:142` was changed `'ecs'` → `'custom-compute'`, which (understandably, to avoid the Docker asset build) removes the sole incidental exercise of the ECS path. Suggested:
```ts
test('default → no ECS', () => { const t = synth({}); t.resourceCountIs('AWS::ECS::Cluster', 0); });
test('compute_type=ecs → ECS cluster + task def', () => {
  const t = synth({ compute_type: 'ecs' });
  t.resourceCountIs('AWS::ECS::Cluster', 1);
  t.resourceCountIs('AWS::ECS::TaskDefinition', 1);
});
```
plus an assertion that the orchestrator receives `ecsConfig` only when `ecs`.

### Non-blocking suggestions

- **Unused for now:** `getRequiredBootstrapPolicies`, `RESOURCE_ACTION_MAP`, `getActionsForResource`, `getAllMappedActions` have no production caller yet (only barrels + tests). Totally fine as staged scaffolding for the Aspect (#125) / preflight (#126) — just worth stating explicitly in the description so reviewers don't expect synth-time enforcement yet.
- **Untracked gaps:** `KNOWN_GAP_SERVICES`/`KNOWN_GAP_ACTIONS` (`sqs:*`, `s3:CreateBucket`+lifecycle, Lambda `EventSourceMapping`/`LayerVersion`) are genuine — the stack creates those resources but no policy grants the actions, so the "all mapped actions exist in policies" test passes only by excluding the cases it most needs to catch. Could each be tied to a tracking issue (`// gap tracked in #NNN`), noted in `DEPLOYMENT_ROLES.md`, and guarded with a "gap set only shrinks" assertion?
- **Test polish:** temp-dir `rmSync` is duplicated rather than in a `finally` (`:233,241`); a couple of `throw`s after an `expect().toBe(true)` that already aborts are dead (`:101-104,:116-119`); `getActionsForResource` only asserts `create`/`delete`, never `read`/`update`.
- The `...(ecsCluster && { ecsConfig: {...} })` spread in `agent.ts:596` is correct (`ecsConfig` is optional and spreading a falsy is a no-op) — flagging only because it's an unusual pattern.

### Docs / security
No docs changes needed for this internal scaffolding, and the Starlight mirror sync isn't triggered (no `docs/`/`CONTRIBUTING.md` edits). The map is inert until #125/#126 consume it, so there's no new IAM/network surface at deploy time — but when the Aspect lands, please make sure the `KNOWN_GAP_*` exclusions don't translate into under- or over-scoped roles.

Really nice foundation overall — just want the coverage tests to actually fail when they should, and the ECS entries added, before this goes in. 🙏

*Review assisted by Claude Code (code-reviewer, silent-failure-hunter, pr-test-analyzer agents).*

scottschreckengaust:
Question for reviewers:

Option 1:
Keep the stacked PRs and continue with incremental changes to the main.

Option 2:
Start from the end of the stack with a large PR to be merged to main at the end.

scottschreckengaust:
@krokoko Picking this back up — rebased onto `main` (`d3974f7a`) and worked your blocking review. Answering your Option 1/Option 2 question from 2026-06-08 with **Option 1 (incremental to main)**: the six reactive permission fixes that landed while this sat (#351, #403, #405, #408, #410, #494, #595) are the evidence that deferring costs more than it saves.

**One structural decision that changes where the fixes land.** While this PR was idle, `main` grew its *own* `cdk/src/bootstrap/resource-action-map.ts` reactively via #351 (issue #350) — outside the RFC — and that is the map wired into the live gate (`synth-coverage.test.ts`). This branch's `bootstrap/preflight/` copy has no production consumer. Fixing the unreachable one would have left the real gate blind, so **B1/B2 are fixed on main's live map**. `preflight/` stays as the verification/testing layer per @scottschreckengaust.

### B1 — ECS blind spot: fixed
Confirmed with a real gated synth: `--context compute_type=ecs` emits exactly `AWS::ECS::Cluster` + `AWS::ECS::TaskDefinition` as unmapped, so `compute-ecs.ts`'s 14 `ecs:*` grants were unverified. Both added to the live map, actions derived from `compute-ecs.ts` as you suggested.

### B2 — vacuous tests: fixed, and I confirmed your diagnosis
You were right that CI was hitting the skip path. The `catch { … return }` swallowed a **nonzero synth exit** — locally `npx cdk synth -c compute_type=ecs` exits 1 (an `ec2:DescribeAvailabilityZones` denial in my account), so the test burned ~86s, reported green, and asserted nothing.

Rather than repair the shell-out, I took your "synthesize in-process" suggestion into `synth-coverage.test.ts` — no child process, so there is no exit code to swallow — plus the non-vacuous guard you asked for:

```ts
expect([...typesInTemplate]).toContain('AWS::ECS::Cluster');
expect([...typesInTemplate]).toContain('AWS::ECS::TaskDefinition');
```

**Mutation-tested both directions**, since a coverage test that cannot fail is the whole defect:
| mutation | result |
|---|---|
| remove the two ECS map entries | ✅ fails — reports both as unmapped |
| hard-code `computeType = 'agentcore'` (silently broken gate) | ✅ fails the `toContain` guard |

Both passed before the fix.

### B3 — already satisfied by #596
`agent.test.ts:674` now has the ECS-gate describe block you specified: cluster + both task-defs, the `ComputeSubstrate` output, and the default no-gate case. No new work needed; #164 is delivered.

### Non-blocking items
- **`getRequiredBootstrapPolicies` now has a production caller.** Your note that it was inert was fair. `collectBootstrapAllowActions()` called `allPolicies()` unconditionally — validating against the **union** — so RFC #120's `deployed ⊇ required` model was unimplemented and an agentcore-only operator's app was checked against `ecs:*` grants their IaCRole cannot perform. Added `policiesForComputeType()`, routed through the salvaged function (fails loud on an unregistered name). Measured: union 357 actions / 14 `ecs:*`; agentcore scope 343 / **0** `ecs:*`; ecs scope 356 / 14 `ecs:*` / 0 `bedrock-agentcore:*`. The argument is optional, so the union behaviour is preserved for callers that want it.
- **`KNOWN_GAP_*` — closed by granting, not excluding.** You flagged these as "genuine gaps … the test passes only by excluding the cases it most needs to catch." Granted all four: `s3:GetBucketPolicy` + `s3:GetEncryptionConfiguration` (CFN reads both back on stack update; every other `Put*` in that statement already had its `Get*` pair) and `sqs:AddPermission` + `sqs:RemovePermission` (`AWS::SQS::QueuePolicy` is managed via Add/RemovePermission, not `SetQueueAttributes`, and the stack creates one for DLQ redrive). `BOOTSTRAP_VERSION` → 1.3.0, artifacts regenerated, `DEPLOYMENT_ROLES.md` updated for golden parity, Starlight mirror re-synced.

### One thing found on the way — filed, not fixed here
**`BOOTSTRAP_HASH` is blind to every policy action** (#732). I added four IAM actions and the committed hash was **byte-identical**. `computeBootstrapHash` passes `Object.keys(json).sort()` as `JSON.stringify`'s second argument, which is a *replacer allowlist*, not a sort — so `Statement` array elements serialize to `{}` and only statement *counts* are protected. Swapping an action, or widening a resource ARN to `*`, leaves the digest untouched. Pre-existing on `main` from #122; left out of this PR to keep it reviewable, but #125/#126 will depend on that digest meaning something.

### Still to come on this PR
Deepening the live map to full CRUD and retiring the duplicate data (keeping `preflight/` as a facade). That is where the real value of the 428-line map is: it carries **48 create-phase actions the flat map lacks** on shared types, and `Update*`/`Tag*`/`Delete*` depth is precisely what the six reactive fixes kept rediscovering.

178 suites / 3533 tests pass.

scottschreckengaust:
Follow-up to the previous comment — the remaining work is done. The duplicate map is retired and the CRUD depth is now enforced rather than decorative.

## One map, on the live path

Merged **programmatically**, not by hand, with the safety invariant asserted mechanically: every action from the create-only map must survive in the merged entry's `create` phase. Verified **0 lost** across 52 types.

| | before | after |
|---|---|---|
| live map | 52 types, create-only | **64 types, 430 actions, CRUD** |
| `preflight/resource-action-map.ts` | 428 lines of duplicate data | **62-line facade, no data** |

The CRUD map contributed **48 create-phase actions the live map lacked** on shared types, plus `AWS::IAM::ManagedPolicy`; main's 6 extra types (CloudFront, `Custom::*`, `CDK::Metadata`) are preserved. `preflight/` stays as the verification layer per @scottschreckengaust — it now re-exports the single map and keeps the query helpers #125/#126 will read it through, so there is no second copy to drift.

`findMissingBootstrapActions` defaults to `['create']`, preserving the pre-CRUD contract for existing callers; pass `phases` to widen.

## `KNOWN_GAP_*` removed entirely — the gate now has zero exemptions

This is the part I want to flag, because the exclusions turned out to be broader than the problem. I checked all 11 excluded actions against the policies: **every one was already covered.** The `sqs` and `s3` entries excluded *whole services*, and all 7 lambda actions were stale. They were hiding nothing.

With the 4 genuine gaps closed by granting (previous commit), the coverage assertion now runs over **all 430 actions in all four phases with no exemptions** — which is what you asked for: *"the test passes only by excluding the cases it most needs to catch."*

Mutation-tested: revoking `sqs:AddPermission` fails with `1 actions not covered by bootstrap policies: sqs:AddPermission`.

Also deleted the two vacuous `Synth coverage` tests from this file — both bailed silently and the ECS one burned ~86s asserting nothing. `synth-coverage.test.ts` now covers both configs in-process and fails loudly.

## Structural pins so the depth can't erode

Two new tests: every entry must declare all four phases as arrays (a regression to `readonly string[]` would make `actionsForResource` silently skip phases), and ≥45 entries must carry real `update`/`delete` actions.

## One bug found and fixed along the way

The new in-process ECS synth builds the agent `DockerImageAsset`, which fingerprints the repo root — and Jest workers evict `.jest-cache/jest-transform-cache-*` entries mid-run, so the walk can hit `ENOENT` on a path another worker just deleted. It surfaced once in a full `mise run build` and would **not** reproduce across three cold-cache runs, so I fixed it structurally rather than retrying.

`.dockerignore` already documents this exact vanishing-file class for pytest-cov's `.coverage.*` temp files. `.jest-cache` was in `.gitignore` but not `.dockerignore`, and `.dockerignore` is what CDK's fingerprint honours. Verified by asserting no staged asset directory contains `cdk/.jest-cache`.

## Final state

**cdk 178 suites / 3533 tests · cli 55 / 695 · agent 1460 · `//cdk:eslint` clean, no mutations.**

`//cdk:synth:quiet` fails locally on `ec2:DescribeAvailabilityZones` — an IAM gap in my sandbox account, reproduced identically on a near-main branch, unrelated to this PR.

All of krokoko's blockers are addressed: **B1** (ECS entries on the live map, mutation-tested), **B2** (loud in-process coverage, mutation-tested both ways), **B3** (already delivered by #596). Non-blocking items: `getRequiredBootstrapPolicies` has a production caller, `KNOWN_GAP_*` closed by granting, CRUD depth merged and enforced. Ready for re-review.

Still open and deliberately out of scope: **#732** (`BOOTSTRAP_HASH` is blind to every policy action — pre-existing from #122, and #125/#126 will need it to mean something).

isadeks:
Heads up before we go further on this one: **the branch is conflicting with `main` and hasn't been updated since 2026-08-06**, so it can't merge as-is. Given the size (18 files, +1147/-70) and how much has moved on `main` since, worth a rebase pass before a fresh review — the current diff may not reflect what actually lands.

Happy to re-review deeply once it's rebased. Ping me.

cdk/src/bootstrap/required-policies.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const CORE_POLICIES = [
+  'infrastructure',
+  'application',
+  'observability',
+] as const;
+
+const COMPUTE_VARIANT_POLICIES: Record<string, string[]> = {
+  agentcore: ['compute-agentcore'],
+  ecs: ['compute-ecs'],
```
isadeks, line 28: **B1 (blocking).** The ECS gate on `main` is additive, not exclusive — `AgentMemory` is unconditional (`stacks/agent.ts:446`) and `agent.test.ts:982` says "alongside the always-present AgentCore runtime". So an `ecs` deploy still creates `AWS::BedrockAgentCore::Runtime`/`Memory`, and an operator who bootstraps exactly this list rolls back on `bedrock-agentcore:CreateMemory` AccessDenied. Suggest `ecs: ['compute-agentcore', 'compute-ecs']`, with `required-policies.test.ts:33` inverted to assert the additive relationship. Same question for `lambda-microvm`, which `main` added while this sat.

cdk/src/bootstrap/version.ts
```diff
@@ -22,7 +22,7 @@ import { createHash } from 'node:crypto';
 import { allPolicies } from './policies';
 
 /** Semantic version of the bootstrap policy bundle. */
-export const BOOTSTRAP_VERSION = '1.2.0';
+export const BOOTSTRAP_VERSION = '1.3.0';
```
isadeks, line 25: **B2a (blocking).** 1.3.0 is already published on `main` — it went to the `compute-lambda-microvm` policy (#645 / ADR-021) and the bundle is now at 1.6.0 with a per-bump changelog in this docstring. `BootstrapPolicyVersion` is an operator-visible `>=` contract, so two different bundles cannot share a number. This needs to be 1.7.0 with a matching history entry after the rebase.

cdk/test/bootstrap/synth-coverage.test.ts
```diff
@@ -72,6 +72,48 @@ describe('Bootstrap policy synth coverage', () => {
...
+    const missingByType: Record<string, string[]> = {};
+    for (const cfnType of typesInTemplate) {
+      if (CFN_TYPES_WITHOUT_EXEC_ROLE_IAM.has(cfnType)) {
+        continue;
+      }
+      if (!(cfnType in RESOURCE_ACTION_MAP)) {
+        unmapped.push(cfnType);
+        continue;
+      }
+      const missing = findMissingBootstrapActions(cfnType, allowedActions);
```
isadeks, line 107: **B5 (blocking).** `allowedActions` is the union (`collectBootstrapAllowActions()` with no argument, `:47`), so the ECS-gated check validates ECS-path resources against grants an ECS-only operator never deploys — the over-permissive direction this PR says it closes. `policiesForComputeType` has no non-test caller as a result. Computing the set per test (`('agentcore')` above, `('ecs')` here) makes the new function load-bearing and would have failed on the `bedrock-agentcore:*` actions B1 leaves out.

cdk/src/bootstrap/policies/application.ts
```diff
@@ -210,6 +210,13 @@ export function applicationPolicy(): iam.PolicyDocument {
           'sqs:UntagQueue',
           'sqs:GetQueueUrl',
           'sqs:ListQueueTags',
+          // AWS::SQS::QueuePolicy is a distinct CFN resource from the queue, and
+          // CloudFormation manages it with Add/RemovePermission — not
+          // SetQueueAttributes. The stack creates one (the DLQ redrive policy),
+          // so without these a queue-policy create/update/delete fails
+          // (#124 review — previously excluded via KNOWN_GAP rather than granted).
+          'sqs:AddPermission',
```
isadeks, line 218: Nit, but please re-derive before granting: the `AWS::SQS::QueuePolicy` resources come from `enforceSSL: true` on the queues, not the DLQ redrive policy (redrive is a queue attribute, not a separate resource). `enforceSSL` emits a `Deny` statement, which `sqs:AddPermission` cannot express — so `SetQueueAttributes`, already granted, is presumably what CFN fires, which is why `main` has `AWS::SQS::QueuePolicy` in `CFN_TYPES_WITHOUT_EXEC_ROLE_IAM`. Can you confirm from CloudTrail on a real deploy? If it is `SetQueueAttributes`, the least-privilege answer is to drop both grants and the map entry.

isadeks (dismissed):
## Review — `feat(bootstrap): resource-action-map for synth-time validation`

**Verdict: request changes** (6 blocking). The direction is right and the work done since krokoko's review is substantial and honest — I re-verified all three of her blockers against the code rather than the write-up and they are genuinely closed, which is not something I say often about a re-review. Two things stop me signing off. First, the branch is 40 commits behind `main` and `mergeable: CONFLICTING`, and the merge is not mechanical: `main` has moved through this exact code four times since 2026-08-06 (#733, #739, #831, #867), and three of those collisions *reduce* coverage if the conflict is resolved by taking this branch's side. Second, the "compute variants are independent choices" model that this PR builds its new selection logic on does not match how the ECS gate actually behaves on `main` — the gate is additive, not exclusive, so the list an ECS operator is told to bootstrap is missing the AgentCore policy their own deploy needs.

A note on CI before anything else: with no computable merge commit, `pull_request` workflows cannot dispatch, so the absent checks here are **not** passing checks — nothing has run against this head. I could not run the suites either (no `node_modules` / `agent/.venv` in my review worktree), so every claim below comes from reading the code and from diffing this branch against current `upstream/main` (`5e10038c`). The only thing I take on trust is your reported `178 suites / 3533 tests`, and that run predates 40 commits of `main`.

### Blocking

**B1 — `getRequiredBootstrapPolicies('ecs')` omits `compute-agentcore`, but the ECS gate is additive, so an operator who follows this list cannot deploy.** `required-policies.ts:26-29` treats `agentcore` and `ecs` as peer variants, and `required-policies.test.ts:33` pins `expect(result).not.toContain('compute-agentcore')`. On `main` the gate does not swap substrates — it adds one. `AgentMemory` is constructed unconditionally (`stacks/agent.ts:446`), the stack's own output text says the gate "adds the Fargate substrate **alongside** AgentCore" (`stacks/agent.ts:1139`), and `cdk/test/stacks/agent.test.ts:982-983` states it outright: "provisions the Fargate substrate alongside the always-present AgentCore runtime." So a `compute_type=ecs` deploy still creates `AWS::BedrockAgentCore::Runtime` and `AWS::BedrockAgentCore::Memory`. Scenario: an operator bootstraps the four policies this function names for `ecs`, runs `cdk deploy -c compute_type=ecs`, and the stack rolls back on `bedrock-agentcore:CreateMemory` AccessDenied — the exact reactive-permission-failure class RFC #120 exists to eliminate, now emitted by the tool meant to prevent it. The `0 bedrock-agentcore:*` figure quoted as verification of the ecs scope is the defect rather than the evidence. The pre-`71d06a56` shape was right about this one; the reviewer suggestion it responded to holds only if the substrates are mutually exclusive, and on `main` they are not. Suggest: `ecs: ['compute-agentcore', 'compute-ecs']` (and the same question for `lambda-microvm` — check whether that gate is additive too), with the test inverted to assert the additive relationship and a comment recording *why*, since this is exactly the kind of thing that gets "simplified" back.

**B2 — Rebase first; the merge is not mechanical.** Merge base is `d3974f7a`; `main` is `5e10038c`. Four collisions, in the order they bite:

*(a) `BOOTSTRAP_VERSION 1.3.0` is already published, for a different bundle.* `version.ts:25` takes 1.3.0; on `main` that number went to the `compute-lambda-microvm` policy (#645 / ADR-021) and the bundle is now at **1.6.0** with a per-bump changelog in the docstring. The version is an operator-visible contract (the `CDKToolkit` stack's `BootstrapPolicyVersion` output, which operators are told to `>=` check), so a second, different 1.3.0 makes two bundles indistinguishable. This needs to become 1.7.0 with a matching entry in that history.

*(b) The rewritten map drops 14 things `main` added.* This PR replaces `RESOURCE_ACTION_MAP` wholesale (`Record<string, readonly string[]>` → `Record<string, ResourceActions>`), so resolving in favour of the branch silently deletes every type `main` mapped in the interim — 13 entries plus one exemption:

```
AWS::BedrockAgentCore::Gateway        AWS::SNS::Topic
AWS::BedrockAgentCore::GatewayTarget  AWS::SNS::Subscription
AWS::CloudFormation::Stack            AWS::StepFunctions::StateMachine
AWS::Cognito::UserPoolGroup           Custom::AgentRegistry
AWS::KMS::Key                         Custom::CDKBucketDeployment
AWS::Lambda::MicrovmImage             Custom::LinearWorkloadIdentity
AWS::Lambda::NetworkConnector
exempt set: AWS::SNS::TopicPolicy
```

They cannot be pasted back as create-only arrays either — this PR's own pin at `cdk/test/bootstrap/resource-action-map.test.ts:77-87` requires all four phases on every entry. Each of those 13 needs CRUD authored, which is real work and is the argument for rebasing *before* re-review.

*(c) There is a third compute substrate now, and the selection logic does not know it.* `COMPUTE_VARIANT_POLICIES` enumerates only `agentcore` and `ecs`. `main` ships `lambda-microvm` with its own policy document (`policies/compute-lambda-microvm.ts`, in `allPolicies()`), its own gate (`stacks/agent.ts:348`), its own resource types, and its own test block (`agent.test.ts:1058`). Post-rebase, `policiesForComputeType('lambda-microvm')` returns core-only — B4 is why that is silent.

*(d) The regenerated `bootstrap-template.yaml` is in the pre-#867 shape.* #867 restructured that artifact to fit CloudFormation's inline body limit and added a hard generator budget (`cdk/src/bootstrap/template-size.ts:62`, `TEMPLATE_SIZE_BUDGET = 49_152`). The committed artifact here is ~50KB against `main`'s ~46KB. Four actions is a couple hundred bytes and the headroom looks sufficient, so I expect this to be fine — but it must be regenerated and re-measured against the new `bootstrap-template.test.ts` budget assertion, not carried over.

**B3 — This PR makes 11 of its 64 map entries unreachable by the synth-coverage gate, and `main` does not have this problem.** `CFN_TYPES_WITHOUT_EXEC_ROLE_IAM` and `RESOURCE_ACTION_MAP` are disjoint on `upstream/main` by design — the first set means "needs no exec-role IAM". After the preflight merge in `58ad265` they overlap on eleven types: `AWS::ApiGateway::Account`, `Deployment`, `Stage`, `AWS::EC2::SubnetRouteTableAssociation`, `VPCGatewayAttachment`, `AWS::Lambda::Alias`, `Permission`, `Version`, `AWS::Logs::ResourcePolicy`, `AWS::S3::BucketPolicy`, `AWS::SQS::QueuePolicy`. Both loops `continue` on the skip set *before* the map lookup (`cdk/test/bootstrap/synth-coverage.test.ts:58`, `:100`), so those entries are dead on the live path. The sharpest instance is self-referential: the two SQS grants this PR adds exist specifically to make `AWS::SQS::QueuePolicy` deployable, the map now describes that type at `resource-action-map.ts:588-593` — and the gate skips it because `resource-action-map.ts:37` says it needs no IAM at all. Drop `sqs:AddPermission` from `application.ts` tomorrow and the synth gate stays green; your mutation test caught it via the union-coverage test at `resource-action-map.test.ts:124`, not via the gate. Fix: decide per type — either it needs actions and comes out of the exempt set, or it does not and the entry goes — then pin it, `expect([...CFN_TYPES_WITHOUT_EXEC_ROLE_IAM].filter((t) => t in RESOURCE_ACTION_MAP)).toEqual([])`, in the same spirit as the structural pins you already added.

**B4 — An unrecognised compute type silently under-scopes, and a test pins that as intended.** `required-policies.ts:34` is `if (variants) base.push(...variants);` — an unknown or misspelled substrate yields the three core policies and no error, and `required-policies.test.ts:45-50` asserts exactly that. This defeats the fail-loud guarantee documented at `policies/index.ts:63-71`: that `throw` can only fire for a name that *is* in the list but has no document, which is the drift that will not realistically happen; the drift that will — a substrate nobody registered — takes the silent path. Concretely, once B2(c) lands, `collectBootstrapAllowActions('lambda-microvm')` returns an allow-set with no compute grants at all and no signal the name was unrecognised. `compute_type` is a closed set on `main`, so I would make it one here: type the parameter as a union, or throw on an unknown name, rather than degrading to core.

**B5 — The live gate still validates against the union, so the compute-type-aware selection is not load-bearing — and it would have caught B1.** `cdk/test/bootstrap/synth-coverage.test.ts:47` calls `collectBootstrapAllowActions()` with no argument in `beforeAll`, and the new ECS-gated test reuses that union set at `:107`. `policiesForComputeType` and the `computeType` parameter have zero non-test callers, so RFC #120's `deployed ⊇ required` model is exercised only by `required-policies.test.ts`'s own unit tests and the drift gate behaves exactly as before. This is not just an unused capability: passing `'ecs'` into the ECS-gated check would have failed on the `bedrock-agentcore:*` actions that B1 leaves out of the ECS bootstrap list. Fix: compute the allow-set per test — `collectBootstrapAllowActions('agentcore')` for the default synth, `('ecs')` for the gated one — which makes the new function load-bearing, turns B4 into a caught error, and gives B1 a regression test in one move.

**B6 — The two ECS entries this PR was blocked on are create-only, so 12 of the 14 `ecs:*` grants stay unverified.** `resource-action-map.ts:324-335` gives `AWS::ECS::Cluster` and `AWS::ECS::TaskDefinition` `read: []`, `update: []`, `delete: []`. `compute-ecs.ts:35-48` grants the full lifecycle — `ecs:DeleteCluster`, `ecs:DeregisterTaskDefinition`, `ecs:DescribeClusters`, `ecs:UpdateCluster`, `ecs:UpdateClusterSettings`, `ecs:PutClusterCapacityProviders`, `ecs:UntagResource`, `ecs:ListTagsForResource` — and krokoko's suggested fix spelled those phases out. This is a validation hole rather than a deploy failure: trim `ecs:DeleteCluster` and nothing fails, on a substrate whose stack-delete path is precisely the class the CRUD shape is argued to catch. It also reads oddly against the thesis — the pin at `resource-action-map.test.ts:89-99` demands update/delete depth on 45 entries, and the two newest entries opt out. Five of the seven create-only entries are `Custom::*`/CloudFront inherited from `main`; ECS is the pair this PR authored. Fill the phases from `compute-ecs.ts`.

### Prior review — verified, and closed

| krokoko blocker | state |
|---|---|
| B1 ECS types missing | fixed, and still needed |
| B2 vacuous coverage tests | fixed, convincingly |
| B3 no ECS-gate test | already on `main` |

B1: `upstream/main`'s map still has **no** `AWS::ECS::*` entry, so this is not a stale complaint — `compute-ecs.ts`'s 14 `ecs:*` grants are unverified on `main` today, and `resource-action-map.ts:324-335` closes that (partially — see B6). B2: the shell-out and its `catch { return }` are gone; the in-process synth at `synth-coverage.test.ts:82-95` has no exit code to swallow, and the `toContain` guards mean it cannot pass vacuously if the gate regresses. Right call, and the mutation evidence in both directions is exactly what I want to see on a test whose whole job is to fail. B3: genuinely satisfied on `main` — `agent.test.ts:978-1000` (cluster count, both task defs, `ComputeSubstrate=ecs`) plus the default no-gate case at `:148-151`; note `main` now provisions **two** task definitions, so any assertion you carry forward should expect 2. The non-blocking `KNOWN_GAP_*` point is properly closed too: the exclusion sets are gone entirely, the `.ts` sources and generated `.json` artifacts agree, and closing a gap by granting rather than excluding is the right resolution — with one caveat in the nits below about *which* gap the SQS pair actually closes.

Your #732 disclosure also still holds: `computeBootstrapHash` on current `main` is unchanged and still passes `Object.keys(json).sort()` as `JSON.stringify`'s *replacer*, so `Statement` elements digest to `{}` and the hash is blind to every action. Filing rather than fixing was right. Worth noting this PR is the first bundle change where that blindness has an operator-visible consequence — an operator whose drift check keys on `BootstrapPolicyHash` sees no change and does not re-bootstrap, so their IaCRole never gets the four new actions. The `BootstrapPolicyVersion` bump still carries the signal, so it degrades rather than breaks; but #732 should land before #125/#126 build on that digest.

### Non-blocking

- `policies/application.ts:213-217` — the justification is wrong on both halves, and the grant may be unnecessary. The `AWS::SQS::QueuePolicy` resources come from `enforceSSL: true` on the queues (`fanout-consumer.ts`, `orchestration-reconciler.ts`, `approval-metrics-publisher-consumer.ts`, `github-screenshot-integration.ts`), not "the DLQ redrive policy" — redrive is an attribute of the queue itself, not a separate resource. And `enforceSSL` emits a `Deny` statement, which `sqs:AddPermission` cannot express at all, so `SetQueueAttributes` (already granted) is what CloudFormation has to use — which is presumably why `main` put `AWS::SQS::QueuePolicy` in the no-exec-role-IAM set in the first place. I can't observe CFN's internal calls from here, so I'll put it as a question: can you re-derive this pair from CloudTrail on a real deploy? If `SetQueueAttributes` is what fires, the least-privilege answer is to drop both grants and the map entry and keep the exemption, rather than widening the IaCRole for a resource that never needed it. (`s3:GetBucketPolicy` / `s3:GetEncryptionConfiguration` I have no such doubt about — the Put/Get-pair reasoning there is sound.)
- `resource-action-map.ts:547-574` — by that same Put/Get-pair rationale, `AWS::S3::Bucket` omits lifecycle from all four phases while four constructs set `lifecycleRules`. `s3:PutLifecycleConfiguration` is granted (`observability.ts:137`) but unmapped, so a trim would pass the gate; `s3:GetLifecycleConfiguration` is granted nowhere, which is the same missing-`Get*` defect one bucket property over from the two you just fixed.
- `resource-action-map.test.ts:70-73` — the `throw` after `expect(hasCreateOrDelete).toBe(true)` is still dead (`expect` aborts first). krokoko flagged this; the sibling at `:113-118` got fixed by reordering, this one did not.
- `resource-action-map.ts:690` — `phases.flatMap((phase) => entry[phase])` on an entry missing a phase puts `undefined` *inside* the action list, which poisons `getAllMappedActions()` and then throws in `action.split(':')` somewhere unrelated. The comment at `resource-action-map.test.ts:78-80` says such an entry would be "silently skipped"; it would actually crash. `?? []` plus a comment correction — the pin is good, the described failure mode is wrong.
- The `>= 55` / `>= 45` thresholds (`:63`, `:98`) go slack once B2(b) adds 13 entries. Consider `count - 1` so they keep ratcheting.
- `.dockerignore` — the `.jest-cache` exclusion is correct and the comment explains the vanishing-file class well. `cdk/cdk.out.*/` looks vestigial now that the ECS synth is in-process; harmless, but it no longer has a producer.
- `preflight/` as a directory holding a 62-line re-export plus two three-line helpers is thin for its own package boundary. It reads fine as the declared seam for #125/#126 — worth one sentence in the module docstring saying that is the only reason it is not folded into `bootstrap/`.
- PR body is stale in three ways: the `#164` ECS gate is already on `main`, so `Closes #164` no longer describes this diff (#164 can be closed on its own); the "Open questions — SQS gap" is resolved; and every deliverable checkbox is still unticked. Please rewrite it after the rebase so a cold reviewer is not reconstructing scope from six comments.

### Docs

Correct and complete for what the diff does. `docs/design/DEPLOYMENT_ROLES.md` carries all four new actions for golden-baseline parity, and the Starlight mirror (`docs/src/content/docs/architecture/Deployment-roles.md`) is regenerated in the same PR with a matching diff — no stale-mirror mutation failure. Both need re-running after the rebase, since `main` added ~140 lines to that baseline in the interim. Governance is fine: #120 is the approved RFC and #124 is the backing issue; branch naming is a waived nit here.

### Tests & CI

Test *design* is strong, and I want to be clear that the B3–B6 findings are about placement rather than effort: the two-directional mutation testing, the zero-exemption union assertion, and the structural pins are the right instincts for a file whose only job is to fail when someone forgets something. I could not execute anything (no `node_modules`), so this is a read-only assessment plus your reported local run — and no CI has run against this head, per the conflict. ADR-002's checklist is satisfied on this branch's terms: policy sources, generated `policies/*.json`, `bootstrap-template.yaml`, `BOOTSTRAP_VERSION`, and the golden baseline all move together; `BOOTSTRAP_HASH` is the one artifact that did not, and #732 explains why. Bootstrap synth-coverage: cannot be executed here, and post-rebase it will fail until B2(b) is resolved, since `main` synthesizes 13 types this map no longer knows.

### Review agents run

`code-reviewer`, `silent-failure-hunter` and `pr-test-analyzer` scope, run over the `d3974f7a..HEAD` range; B1, B3 and the two comment-accuracy items came out of that pass and I re-verified each against `upstream/main` before writing it here. Omitted: `type-design-analyzer` — the one new type, `ResourceActions`, is a four-field record already pinned structurally by test; `/security-review` as a separate pass — the IAM delta is four read/permission-management actions on existing ARN-scoped statements, reviewed directly. On that: `sqs:AddPermission` does let the IaC role grant cross-account access to those queues, but it is scoped to `arn:aws:sqs:*:*:backgroundagent-dev-*`; my concern there is the necessity question in the nits, not the scope.

### Human heuristics

- **Proportionality** — pass. Retiring the duplicate map rather than maintaining two is the proportionate move, and the CRUD shape is justified by the seven reactive permission fixes it would have caught.
- **Coherence** — concern. B3 is a coherence failure: two sets `main` keeps disjoint now disagree about eleven types and no test notices. B1 is the deeper one — the policy layer models the substrates as exclusive while the stack layer treats them as additive.
- **Clarity** — concern at `required-policies.ts:34` (B4), where the code degrades exactly where its own docstring promises to fail loud, and at `application.ts:213-217`, where a comment written to explain a new grant misdescribes the resource that motivated it.
- **Appropriateness** — concern, and it is B2. This map is a statement about what `main` synthesizes; a version of it that is 40 commits stale is wrong in the one way this file cannot afford. Rebase, re-derive against a current synth, fix B1, and I will re-review deeply — the substance is good and I want it to land.

cdk/src/bootstrap/version.ts
isadeks, line 25: **B1(a) — 1.3.0 is already taken on `main`, by a different bundle.** `main` consumed 1.3.0 for the `compute-lambda-microvm` policy (#645 / ADR-021) and is now at **1.6.0**, with a per-bump changelog in this docstring. `BootstrapPolicyVersion` is an operator-visible `CDKToolkit` output that operators are told to `>=` check, so two different bundles publishing 1.3.0 makes them indistinguishable. On rebase this becomes 1.7.0 plus a bump-history entry.

cdk/src/bootstrap/required-policies.ts
```diff
@@ -0,0 +1,36 @@
...
+};
+
+export function getRequiredBootstrapPolicies(computeType: string): string[] {
+  const base: string[] = [...CORE_POLICIES];
+  const variants = COMPUTE_VARIANT_POLICIES[computeType];
+  if (variants) base.push(...variants);
```
isadeks, line 34: **B3 — silent under-scoping.** An unknown or misspelled compute type falls through to core-only with no error, and `required-policies.test.ts:45-50` pins that as intended. That defeats the fail-loud guarantee documented at `policies/index.ts:63-71`: the `throw` there fires only for a name that *is* in this list but has no document — the drift that can actually happen (a substrate nobody registered) takes the quiet path. Concretely, `main` already ships `lambda-microvm`, so post-rebase `collectBootstrapAllowActions('lambda-microvm')` returns an allow-set with none of that substrate's grants. `compute_type` is a closed set on `main` — type this parameter as a union, or throw on an unrecognised name.

cdk/src/bootstrap/resource-action-map.ts
```diff
@@ -41,63 +41,591 @@ export const CFN_TYPES_WITHOUT_EXEC_ROLE_IAM = new Set([
...
+    delete: ['s3:DeleteBucketPolicy'],
+  },
+  // ─── SQS ───────────────────────────────────────────────────────────
+  'AWS::SQS::Queue': {
+    create: ['sqs:CreateQueue', 'sqs:TagQueue', 'sqs:GetQueueUrl', 'sqs:GetQueueAttributes'],
+    read: ['sqs:GetQueueAttributes', 'sqs:GetQueueUrl'],
+    update: ['sqs:SetQueueAttributes', 'sqs:TagQueue', 'sqs:UntagQueue'],
+    delete: ['sqs:DeleteQueue', 'sqs:GetQueueUrl'],
+  },
+  'AWS::SQS::QueuePolicy': {
```
isadeks, line 588: **B2 — this entry is unreachable by the live gate.** `AWS::SQS::QueuePolicy` is also in `CFN_TYPES_WITHOUT_EXEC_ROLE_IAM` (line 37), and the exempt check short-circuits first in `synth-coverage.test.ts:58` / `:100`. So the two SQS grants this PR adds *specifically* to make `QueuePolicy` deployable are never validated by the synth-time gate — drop `sqs:AddPermission` from `application.ts` and the gate stays green. Eleven of the 64 entries overlap the exempt set this way; the two sets are disjoint on `upstream/main`, so this is introduced by the preflight-map merge in `58ad265`. Decide per type, then pin the invariant: `expect([...CFN_TYPES_WITHOUT_EXEC_ROLE_IAM].filter((t) => t in RESOURCE_ACTION_MAP)).toEqual([])`.

cdk/src/bootstrap/resource-action-map.ts
```diff
@@ -41,63 +41,591 @@ export const CFN_TYPES_WITHOUT_EXEC_ROLE_IAM = new Set([
...
+  'AWS::EC2::VPCGatewayAttachment': {
+    create: ['ec2:AttachInternetGateway'],
+    read: ['ec2:DescribeInternetGateways'],
+    update: ['ec2:AttachInternetGateway', 'ec2:DetachInternetGateway'],
+    delete: ['ec2:DetachInternetGateway'],
+  },
+  // ─── ECS ───────────────────────────────────────────────────────────
+  'AWS::ECS::Cluster': {
+    create: ['ecs:CreateCluster', 'ecs:TagResource'],
+    read: [],
```
isadeks, line 326: **B5 — create-only on the two entries this PR was blocked on.** `compute-ecs.ts:35-48` grants the full lifecycle (`ecs:DeleteCluster`, `ecs:DeregisterTaskDefinition`, `ecs:DescribeClusters`, `ecs:UpdateCluster`, `ecs:UpdateClusterSettings`, `ecs:PutClusterCapacityProviders`, `ecs:UntagResource`, `ecs:ListTagsForResource`) and krokoko's suggested fix spelled those phases out. Not a deploy failure today, but 12 of the 14 `ecs:*` grants are unverified: trim `ecs:DeleteCluster` and nothing fails, on the substrate whose stack-delete path is exactly the reactive-fix class the CRUD shape exists to catch. It also reads oddly against the pin at `resource-action-map.test.ts:89-99`, which demands update/delete depth on 45 entries while the two newest entries opt out.

cdk/test/bootstrap/synth-coverage.test.ts
isadeks, line 107: **B4 — the gate still validates against the union.** `allowedActions` comes from `collectBootstrapAllowActions()` with no argument (line 47), so this ECS-substrate check validates ECS-path resources against a set that includes `compute-agentcore`'s grants — which an ECS-only operator never deploys. A required action granted only there passes here and `AccessDenied`s on their real IaCRole, which is the exact failure the compute-type-aware selection was added to close. Two arguments fix it: `collectBootstrapAllowActions('agentcore')` for the default synth and `('ecs')` here, computed per-test rather than once in `beforeAll`.

cdk/src/bootstrap/resource-action-map.ts
```diff
@@ -137,19 +675,37 @@ export function collectBootstrapAllowActions(): Set<string> {
...
+/** Flattens an entry's phases into one action list (deduplicated, order kept). */
+export function actionsForResource(
+  cfnType: string,
+  phases: readonly LifecyclePhase[] = ['create', 'read', 'update', 'delete'],
+): string[] {
+  const entry = RESOURCE_ACTION_MAP[cfnType];
+  if (!entry) {
+    return [];
+  }
+  return [...new Set(phases.flatMap((phase) => entry[phase]))];
```
isadeks, line 690: Nit — an entry missing a phase makes this yield `undefined` *inside* the action list, which poisons `getAllMappedActions()` and then throws in `action.split(':')` somewhere unrelated. The comment at `resource-action-map.test.ts:78-80` describes it as "silently skipped"; it crashes instead. `entry[phase] ?? []` plus a comment correction — the pin itself is good.

cdk/test/bootstrap/resource-action-map.test.ts
```diff
@@ -0,0 +1,181 @@
...
+      expect(entryCount).toBeGreaterThanOrEqual(55);
+    });
+
+    it('every entry has at least one action in create or delete', () => {
+      for (const [type, entry] of Object.entries(RESOURCE_ACTION_MAP)) {
+        const hasCreateOrDelete = entry.create.length > 0 || entry.delete.length > 0;
+        expect(hasCreateOrDelete).toBe(true);
+        if (!hasCreateOrDelete) {
+          // Extra info for debugging (won't reach here if assertion passes)
+          throw new Error(`${type} has no create or delete actions`);
```
isadeks, line 72: Nit — dead code: `expect(hasCreateOrDelete).toBe(true)` on the line above already aborts, so this `throw` never runs. krokoko flagged this pair; the sibling at `:113-118` got fixed by putting the `throw` *before* the `expect`. Same treatment here, or drop it.

isadeks (changes requested):
## Review — `feat(bootstrap): resource-action-map for synth-time validation`

**Verdict: request changes** (6 blocking). The direction is right and the work done since krokoko's review is substantial and honest — I re-verified all three of her blockers against the code rather than the write-up and they are genuinely closed, which is not something I say often about a re-review. Two things stop me signing off. First, the branch is 40 commits behind `main` and `mergeable: CONFLICTING`, and the merge is not mechanical: `main` has moved through this exact code four times since 2026-08-06 (#733, #739, #831, #867), and three of those collisions *reduce* coverage if the conflict is resolved by taking this branch's side. Second, the "compute variants are independent choices" model that this PR builds its new selection logic on does not match how the ECS gate actually behaves on `main` — the gate is additive, not exclusive, so the list an ECS operator is told to bootstrap is missing the AgentCore policy their own deploy needs.

A note on CI before anything else: with no computable merge commit, `pull_request` workflows cannot dispatch, so the absent checks here are **not** passing checks — nothing has run against this head. I could not run the suites either (no `node_modules` / `agent/.venv` in my review worktree), so every claim below comes from reading the code and from diffing this branch against current `upstream/main` (`5e10038c`). The only thing I take on trust is your reported `178 suites / 3533 tests`, and that run predates 40 commits of `main`.

### Blocking

**B1 — `getRequiredBootstrapPolicies('ecs')` omits `compute-agentcore`, but the ECS gate is additive, so an operator who follows this list cannot deploy.** `required-policies.ts:26-29` treats `agentcore` and `ecs` as peer variants, and `required-policies.test.ts:33` pins `expect(result).not.toContain('compute-agentcore')`. On `main` the gate does not swap substrates — it adds one. `AgentMemory` is constructed unconditionally (`stacks/agent.ts:446`), the stack's own output text says the gate "adds the Fargate substrate **alongside** AgentCore" (`stacks/agent.ts:1139`), and `cdk/test/stacks/agent.test.ts:982-983` states it outright: "provisions the Fargate substrate alongside the always-present AgentCore runtime." So a `compute_type=ecs` deploy still creates `AWS::BedrockAgentCore::Runtime` and `AWS::BedrockAgentCore::Memory`. Scenario: an operator bootstraps the four policies this function names for `ecs`, runs `cdk deploy -c compute_type=ecs`, and the stack rolls back on `bedrock-agentcore:CreateMemory` AccessDenied — the exact reactive-permission-failure class RFC #120 exists to eliminate, now emitted by the tool meant to prevent it. The `0 bedrock-agentcore:*` figure quoted as verification of the ecs scope is the defect rather than the evidence. The pre-`71d06a56` shape was right about this one; the reviewer suggestion it responded to holds only if the substrates are mutually exclusive, and on `main` they are not. Suggest: `ecs: ['compute-agentcore', 'compute-ecs']` (and the same question for `lambda-microvm` — check whether that gate is additive too), with the test inverted to assert the additive relationship and a comment recording *why*, since this is exactly the kind of thing that gets "simplified" back.

**B2 — Rebase first; the merge is not mechanical.** Merge base is `d3974f7a`; `main` is `5e10038c`. Four collisions, in the order they bite:

*(a) `BOOTSTRAP_VERSION 1.3.0` is already published, for a different bundle.* `version.ts:25` takes 1.3.0; on `main` that number went to the `compute-lambda-microvm` policy (#645 / ADR-021) and the bundle is now at **1.6.0** with a per-bump changelog in the docstring. The version is an operator-visible contract (the `CDKToolkit` stack's `BootstrapPolicyVersion` output, which operators are told to `>=` check), so a second, different 1.3.0 makes two bundles indistinguishable. This needs to become 1.7.0 with a matching entry in that history.

*(b) The rewritten map drops 14 things `main` added.* This PR replaces `RESOURCE_ACTION_MAP` wholesale (`Record<string, readonly string[]>` → `Record<string, ResourceActions>`), so resolving in favour of the branch silently deletes every type `main` mapped in the interim — 13 entries plus one exemption:

```
AWS::BedrockAgentCore::Gateway        AWS::SNS::Topic
AWS::BedrockAgentCore::GatewayTarget  AWS::SNS::Subscription
AWS::CloudFormation::Stack            AWS::StepFunctions::StateMachine
AWS::Cognito::UserPoolGroup           Custom::AgentRegistry
AWS::KMS::Key                         Custom::CDKBucketDeployment
AWS::Lambda::MicrovmImage             Custom::LinearWorkloadIdentity
AWS::Lambda::NetworkConnector
exempt set: AWS::SNS::TopicPolicy
```

They cannot be pasted back as create-only arrays either — this PR's own pin at `cdk/test/bootstrap/resource-action-map.test.ts:77-87` requires all four phases on every entry. Each of those 13 needs CRUD authored, which is real work and is the argument for rebasing *before* re-review.

*(c) There is a third compute substrate now, and the selection logic does not know it.* `COMPUTE_VARIANT_POLICIES` enumerates only `agentcore` and `ecs`. `main` ships `lambda-microvm` with its own policy document (`policies/compute-lambda-microvm.ts`, in `allPolicies()`), its own gate (`stacks/agent.ts:348`), its own resource types, and its own test block (`agent.test.ts:1058`). Post-rebase, `policiesForComputeType('lambda-microvm')` returns core-only — B4 is why that is silent.

*(d) The regenerated `bootstrap-template.yaml` is in the pre-#867 shape.* #867 restructured that artifact to fit CloudFormation's inline body limit and added a hard generator budget (`cdk/src/bootstrap/template-size.ts:62`, `TEMPLATE_SIZE_BUDGET = 49_152`). The committed artifact here is ~50KB against `main`'s ~46KB. Four actions is a couple hundred bytes and the headroom looks sufficient, so I expect this to be fine — but it must be regenerated and re-measured against the new `bootstrap-template.test.ts` budget assertion, not carried over.

**B3 — This PR makes 11 of its 64 map entries unreachable by the synth-coverage gate, and `main` does not have this problem.** `CFN_TYPES_WITHOUT_EXEC_ROLE_IAM` and `RESOURCE_ACTION_MAP` are disjoint on `upstream/main` by design — the first set means "needs no exec-role IAM". After the preflight merge in `58ad265` they overlap on eleven types: `AWS::ApiGateway::Account`, `Deployment`, `Stage`, `AWS::EC2::SubnetRouteTableAssociation`, `VPCGatewayAttachment`, `AWS::Lambda::Alias`, `Permission`, `Version`, `AWS::Logs::ResourcePolicy`, `AWS::S3::BucketPolicy`, `AWS::SQS::QueuePolicy`. Both loops `continue` on the skip set *before* the map lookup (`cdk/test/bootstrap/synth-coverage.test.ts:58`, `:100`), so those entries are dead on the live path. The sharpest instance is self-referential: the two SQS grants this PR adds exist specifically to make `AWS::SQS::QueuePolicy` deployable, the map now describes that type at `resource-action-map.ts:588-593` — and the gate skips it because `resource-action-map.ts:37` says it needs no IAM at all. Drop `sqs:AddPermission` from `application.ts` tomorrow and the synth gate stays green; your mutation test caught it via the union-coverage test at `resource-action-map.test.ts:124`, not via the gate. Fix: decide per type — either it needs actions and comes out of the exempt set, or it does not and the entry goes — then pin it, `expect([...CFN_TYPES_WITHOUT_EXEC_ROLE_IAM].filter((t) => t in RESOURCE_ACTION_MAP)).toEqual([])`, in the same spirit as the structural pins you already added.

**B4 — An unrecognised compute type silently under-scopes, and a test pins that as intended.** `required-policies.ts:34` is `if (variants) base.push(...variants);` — an unknown or misspelled substrate yields the three core policies and no error, and `required-policies.test.ts:45-50` asserts exactly that. This defeats the fail-loud guarantee documented at `policies/index.ts:63-71`: that `throw` can only fire for a name that *is* in the list but has no document, which is the drift that will not realistically happen; the drift that will — a substrate nobody registered — takes the silent path. Concretely, once B2(c) lands, `collectBootstrapAllowActions('lambda-microvm')` returns an allow-set with no compute grants at all and no signal the name was unrecognised. `compute_type` is a closed set on `main`, so I would make it one here: type the parameter as a union, or throw on an unknown name, rather than degrading to core.

**B5 — The live gate still validates against the union, so the compute-type-aware selection is not load-bearing — and it would have caught B1.** `cdk/test/bootstrap/synth-coverage.test.ts:47` calls `collectBootstrapAllowActions()` with no argument in `beforeAll`, and the new ECS-gated test reuses that union set at `:107`. `policiesForComputeType` and the `computeType` parameter have zero non-test callers, so RFC #120's `deployed ⊇ required` model is exercised only by `required-policies.test.ts`'s own unit tests and the drift gate behaves exactly as before. This is not just an unused capability: passing `'ecs'` into the ECS-gated check would have failed on the `bedrock-agentcore:*` actions that B1 leaves out of the ECS bootstrap list. Fix: compute the allow-set per test — `collectBootstrapAllowActions('agentcore')` for the default synth, `('ecs')` for the gated one — which makes the new function load-bearing, turns B4 into a caught error, and gives B1 a regression test in one move.

**B6 — The two ECS entries this PR was blocked on are create-only, so 12 of the 14 `ecs:*` grants stay unverified.** `resource-action-map.ts:324-335` gives `AWS::ECS::Cluster` and `AWS::ECS::TaskDefinition` `read: []`, `update: []`, `delete: []`. `compute-ecs.ts:35-48` grants the full lifecycle — `ecs:DeleteCluster`, `ecs:DeregisterTaskDefinition`, `ecs:DescribeClusters`, `ecs:UpdateCluster`, `ecs:UpdateClusterSettings`, `ecs:PutClusterCapacityProviders`, `ecs:UntagResource`, `ecs:ListTagsForResource` — and krokoko's suggested fix spelled those phases out. This is a validation hole rather than a deploy failure: trim `ecs:DeleteCluster` and nothing fails, on a substrate whose stack-delete path is precisely the class the CRUD shape is argued to catch. It also reads oddly against the thesis — the pin at `resource-action-map.test.ts:89-99` demands update/delete depth on 45 entries, and the two newest entries opt out. Five of the seven create-only entries are `Custom::*`/CloudFront inherited from `main`; ECS is the pair this PR authored. Fill the phases from `compute-ecs.ts`.

### Prior review — verified, and closed

| krokoko blocker | state |
|---|---|
| B1 ECS types missing | fixed, and still needed |
| B2 vacuous coverage tests | fixed, convincingly |
| B3 no ECS-gate test | already on `main` |

B1: `upstream/main`'s map still has **no** `AWS::ECS::*` entry, so this is not a stale complaint — `compute-ecs.ts`'s 14 `ecs:*` grants are unverified on `main` today, and `resource-action-map.ts:324-335` closes that (partially — see B6). B2: the shell-out and its `catch { return }` are gone; the in-process synth at `synth-coverage.test.ts:82-95` has no exit code to swallow, and the `toContain` guards mean it cannot pass vacuously if the gate regresses. Right call, and the mutation evidence in both directions is exactly what I want to see on a test whose whole job is to fail. B3: genuinely satisfied on `main` — `agent.test.ts:978-1000` (cluster count, both task defs, `ComputeSubstrate=ecs`) plus the default no-gate case at `:148-151`; note `main` now provisions **two** task definitions, so any assertion you carry forward should expect 2. The non-blocking `KNOWN_GAP_*` point is properly closed too: the exclusion sets are gone entirely, the `.ts` sources and generated `.json` artifacts agree, and closing a gap by granting rather than excluding is the right resolution — with one caveat in the nits below about *which* gap the SQS pair actually closes.

Your #732 disclosure also still holds: `computeBootstrapHash` on current `main` is unchanged and still passes `Object.keys(json).sort()` as `JSON.stringify`'s *replacer*, so `Statement` elements digest to `{}` and the hash is blind to every action. Filing rather than fixing was right. Worth noting this PR is the first bundle change where that blindness has an operator-visible consequence — an operator whose drift check keys on `BootstrapPolicyHash` sees no change and does not re-bootstrap, so their IaCRole never gets the four new actions. The `BootstrapPolicyVersion` bump still carries the signal, so it degrades rather than breaks; but #732 should land before #125/#126 build on that digest.

### Non-blocking

- `policies/application.ts:213-217` — the justification is wrong on both halves, and the grant may be unnecessary. The `AWS::SQS::QueuePolicy` resources come from `enforceSSL: true` on the queues (`fanout-consumer.ts`, `orchestration-reconciler.ts`, `approval-metrics-publisher-consumer.ts`, `github-screenshot-integration.ts`), not "the DLQ redrive policy" — redrive is an attribute of the queue itself, not a separate resource. And `enforceSSL` emits a `Deny` statement, which `sqs:AddPermission` cannot express at all, so `SetQueueAttributes` (already granted) is what CloudFormation has to use — which is presumably why `main` put `AWS::SQS::QueuePolicy` in the no-exec-role-IAM set in the first place. I can't observe CFN's internal calls from here, so I'll put it as a question: can you re-derive this pair from CloudTrail on a real deploy? If `SetQueueAttributes` is what fires, the least-privilege answer is to drop both grants and the map entry and keep the exemption, rather than widening the IaCRole for a resource that never needed it. (`s3:GetBucketPolicy` / `s3:GetEncryptionConfiguration` I have no such doubt about — the Put/Get-pair reasoning there is sound.)
- `resource-action-map.ts:547-574` — by that same Put/Get-pair rationale, `AWS::S3::Bucket` omits lifecycle from all four phases while four constructs set `lifecycleRules`. `s3:PutLifecycleConfiguration` is granted (`observability.ts:137`) but unmapped, so a trim would pass the gate; `s3:GetLifecycleConfiguration` is granted nowhere, which is the same missing-`Get*` defect one bucket property over from the two you just fixed.
- `resource-action-map.test.ts:70-73` — the `throw` after `expect(hasCreateOrDelete).toBe(true)` is still dead (`expect` aborts first). krokoko flagged this; the sibling at `:113-118` got fixed by reordering, this one did not.
- `resource-action-map.ts:690` — `phases.flatMap((phase) => entry[phase])` on an entry missing a phase puts `undefined` *inside* the action list, which poisons `getAllMappedActions()` and then throws in `action.split(':')` somewhere unrelated. The comment at `resource-action-map.test.ts:78-80` says such an entry would be "silently skipped"; it would actually crash. `?? []` plus a comment correction — the pin is good, the described failure mode is wrong.
- The `>= 55` / `>= 45` thresholds (`:63`, `:98`) go slack once B2(b) adds 13 entries. Consider `count - 1` so they keep ratcheting.
- `.dockerignore` — the `.jest-cache` exclusion is correct and the comment explains the vanishing-file class well. `cdk/cdk.out.*/` looks vestigial now that the ECS synth is in-process; harmless, but it no longer has a producer.
- `preflight/` as a directory holding a 62-line re-export plus two three-line helpers is thin for its own package boundary. It reads fine as the declared seam for #125/#126 — worth one sentence in the module docstring saying that is the only reason it is not folded into `bootstrap/`.
- PR body is stale in three ways: the `#164` ECS gate is already on `main`, so `Closes #164` no longer describes this diff (#164 can be closed on its own); the "Open questions — SQS gap" is resolved; and every deliverable checkbox is still unticked. Please rewrite it after the rebase so a cold reviewer is not reconstructing scope from six comments.

### Docs

Correct and complete for what the diff does. `docs/design/DEPLOYMENT_ROLES.md` carries all four new actions for golden-baseline parity, and the Starlight mirror (`docs/src/content/docs/architecture/Deployment-roles.md`) is regenerated in the same PR with a matching diff — no stale-mirror mutation failure. Both need re-running after the rebase, since `main` added ~140 lines to that baseline in the interim. Governance is fine: #120 is the approved RFC and #124 is the backing issue; branch naming is a waived nit here.

### Tests & CI

Test *design* is strong, and I want to be clear that the B3–B6 findings are about placement rather than effort: the two-directional mutation testing, the zero-exemption union assertion, and the structural pins are the right instincts for a file whose only job is to fail when someone forgets something. I could not execute anything (no `node_modules`), so this is a read-only assessment plus your reported local run — and no CI has run against this head, per the conflict. ADR-002's checklist is satisfied on this branch's terms: policy sources, generated `policies/*.json`, `bootstrap-template.yaml`, `BOOTSTRAP_VERSION`, and the golden baseline all move together; `BOOTSTRAP_HASH` is the one artifact that did not, and #732 explains why. Bootstrap synth-coverage: cannot be executed here, and post-rebase it will fail until B2(b) is resolved, since `main` synthesizes 13 types this map no longer knows.

### Review agents run

Ran the repo's `code-review` agent over the `d3974f7a..HEAD` range (`code-reviewer` / `pr-test-analyzer` / `silent-failure-hunter` scope); its pass had not returned by the time I finished, so to be straight with you: **every finding above is one I verified myself**, mechanically where the claim is set-shaped — the exempt-set/map overlap in B2, the 14 dropped types in B1(b), and the create-only ECS entries in B5 were each computed by diffing this branch's declarations against `upstream/main`'s, not eyeballed. Omitted: `type-design-analyzer` (the one new type, `ResourceActions`, is a four-field record already pinned structurally by test), `comment-analyzer` (folded into the read — one inaccuracy found, noted in the nits), and `/security-review` as a separate pass: the IAM delta is four read/permission-management actions on existing ARN-scoped statements, which I reviewed directly. On that — `sqs:AddPermission` does let the IaC role grant cross-account access to those queues, but it is scoped to `arn:aws:sqs:*:*:backgroundagent-dev-*` and is the only way CloudFormation can manage `AWS::SQS::QueuePolicy`, so it is inherent to the resource rather than a widening.

### Human heuristics

- **Proportionality** — pass. Retiring the duplicate map rather than maintaining two is the proportionate move, and the CRUD shape is justified by the seven reactive permission fixes it would have caught.
- **Coherence** — concern. B3 is a coherence failure: two sets `main` keeps disjoint now disagree about eleven types and no test notices. B1 is the deeper one — the policy layer models the substrates as exclusive while the stack layer treats them as additive.
- **Clarity** — concern at `required-policies.ts:34` (B4), where the code degrades exactly where its own docstring promises to fail loud, and at `application.ts:213-217`, where a comment written to explain a new grant misdescribes the resource that motivated it.
- **Appropriateness** — concern, and it is B2. This map is a statement about what `main` synthesizes; a version of it that is 40 commits stale is wrong in the one way this file cannot afford. Rebase, re-derive against a current synth, fix B1, and I will re-review deeply — the substance is good and I want it to land.

PR #200, opened 2026-05-27
feat(linear): multi-workspace support — add-workspace, per-workspace webhook signing, update-webhook-secret

isadeks:
## Summary

Three CLI commands + a webhook architecture fix that together make ABCA's Linear integration usable against multiple Linear workspaces from one stack. 

- `bgagent linear add-workspace <slug>` — install OAuth in additional workspaces without re-running the full setup wizard
- `bgagent linear update-webhook-secret <slug>` — rotate or fix a workspace's webhook signing secret without re-running OAuth
- Webhook receiver now picks the signing secret per `organizationId` (workspace-scoped Linear webhooks were silently dropping signatures from any workspace beyond the first)

## Why

Single-workspace deploys work great after #160. The moment you try to onboard a second workspace, two things break:

1. **OAuth Client ID/Secret entry** — `setup` is heavy (full OAuth dance, registry write, user mapping, webhook prompt). `add-workspace` skips the parts you've already done.
2. **Webhook signature verification** — Linear generates a fresh signing secret per webhook subscription, and webhook subscriptions are workspace-scoped. Storing a single stack-wide signing secret means events from any workspace beyond the one whose secret was pasted silently fail HMAC verification.

Both surfaced concretely while onboarding a second Linear workspace this week — `add-workspace` got an `access_denied` from Linear when the OAuth app was private to the origin workspace; webhook events 401'd silently when the new workspace was finally authorized.

### 1. `bgagent linear add-workspace <slug>`

Lighter-weight install for additional workspaces. Reuses an existing workspace's OAuth Client ID as a default (Enter to keep it, or paste a new one for a per-workspace OAuth app). Always interactive — secrets never on the command line. Refuses to silently overwrite an already-onboarded workspace's registry row.

### 2. Per-workspace webhook signing secret

`StoredOauthToken` / `StoredLinearOauthToken` get an optional `webhook_signing_secret` field. Webhook receiver:
1. Parses the body to extract \`organizationId\` (untrusted — used only to select WHICH secret to verify against)
2. Looks up that workspace's per-workspace secret. If found and signature matches → trusted. **If found and mismatches → 401, no fallback** (cross-workspace impersonation defense).
3. If no per-workspace secret → falls back to the stack-wide \`LinearWebhookSecret\` (back-compat for installs predating per-workspace signing).

Cross-language schema parity test extended to track required vs optional fields so the validator const can't drift from the interface.

### 3. \`bgagent linear update-webhook-secret <slug>\`

Rotation / recovery primitive. The OAuth dance can't re-run when the app is already installed in a workspace — Linear returns \`access_denied\`. This command sidesteps it: read the existing OAuth bundle, prompt for the new signing secret, re-upsert the bundle. No DDB writes, no Linear API calls, no stack-wide writes.

This made \`setup --rotate-webhook-secret\` redundant — the flag is removed in this PR.

### 4. \`list-projects\` fixes + setup-guide rewrite

- \`list-projects\` now reads from the OAuth secret model (was still pointing at the parked PAK secret)
- Empty-workspace message clearer: \"Workspace 'X' has no projects yet\" instead of \"No projects visible\"
- Setup guide rewritten end-to-end:
  - Pruned parked Phase 2.0a flow (oauth-register-workspace, AgentCore Identity callbacks)
  - New step-by-step walkthrough for adding a second workspace, with the OAuth-app-public-vs-private decision up front
  - New section: \"How webhook signature verification works\" with the trust model explained
  - Troubleshooting points at \`update-webhook-secret\` for rotation cases

## Backward compatibility

Designed for zero impact on existing single-workspace deploys:

- Webhook receiver tries per-workspace path first, falls back to stack-wide. Existing installs keep working without re-onboarding.
- Re-running \`setup\` on an existing single-workspace install auto-mirrors the stack-wide signing secret onto the workspace's OAuth bundle (no prompt). After that the workspace works on the per-workspace path.
- Multi-workspace was structurally impossible before this PR; nobody could be relying on the broken behavior.

## Trust model

The \`organizationId\` in the webhook body is attacker-controlled. But it only **selects** which secret to verify against — they still need the matching signing secret to forge a valid signature. The no-fallback-on-per-workspace-mismatch rule prevents an attacker from bypassing a per-workspace secret by tricking the receiver into re-checking against the stack-wide one.

The \`linear-webhook-multi-workspace.test.ts\` suite covers this:
- workspace A signed with workspace B's secret → 401, no dispatch
- registry status \`revoked\` + signature matches the stored per-workspace secret → 401 (revoked workspaces don't trigger tasks)

## Tests

- 6 new tests in \`linear-webhook-multi-workspace.test.ts\` covering happy multi-workspace, cross-workspace impersonation rejection, registry-miss fallback, missing-field fallback, revoked status
- 5 new tests in \`linear.test.ts\` covering \`findReusableOauthAppCredentials\` (empty registry, happy path, missing fields, corrupted JSON, missing SecretString)
- Schema parity test extended to allow optional fields and verify required-set matches the validator const
- Existing 12 webhook tests + 27 resolver tests all pass unchanged (back-compat path verified)

CDK: 1789 / 1789. CLI: 286 / 286 (after rebuild).

## Smoke test

Deployed to a personal AWS account against two Linear workspaces with separate OAuth apps. Verified:
- Setup against workspace A: writes per-workspace secret, mirrors to stack-wide.
- \`add-workspace\` against workspace B with a different OAuth app: prompts for new client_id/secret, completes OAuth dance, refuses to silently overwrite registry row.
- Webhook from workspace A → verified via per-workspace path → dispatched.
- Webhook from workspace B → verified via per-workspace path → dispatched.
- (Negative) workspace A signed with workspace B's secret → 401 (manual curl).
- \`update-webhook-secret\` after rotating workspace B's signing secret in Linear: bundle updated, next event verified successfully.

## Test plan

- [ ] CI green
- [ ] Maintainer review of trust-model section in the setup guide + the no-fallback-on-mismatch invariant in webhook handler
- [ ] Spot-check that re-running \`setup\` on a single-workspace install auto-migrates without prompting


krokoko:
## Review summary — multi-workspace Linear support

Thanks for this — it's a well-engineered PR that closes a real multi-workspace gap, and the docs rewrite is genuinely excellent. The core trust model (per-workspace HMAC, mismatch-is-fatal, no fallback) is sound and the new "How webhook signature verification works" section documents it clearly. A few items I'd like to resolve before merge.

**Verdict: Request changes** — mostly one security item and a shipped broken string; the rest are nits or follow-ups.

### Blocking

1. **Fail-open on transient infra errors** — `verifyLinearRequestForWorkspace` (`cdk/src/handlers/shared/linear-verify.ts:200-207`) can't tell "no per-workspace secret" from "DynamoDB/Secrets Manager threw": both collapse to `null` in `getRegistryRow`/`getOauthSecret` (`linear-oauth-resolver.ts:232-238`, `:313-319`) and fall through to the **stack-wide** secret (`linear-webhook.ts:127`). A DDB throttle during a webhook burst silently downgrades a per-workspace-secured workspace to stack-wide verification. Suggest giving the auth path *strict* lookups that **throw** on infra error (keeping the lenient `null`-returning helpers for `resolveLinearOauthToken`, whose graceful-no-op contract is intentional) — the handler's outer try/catch already returns 500, so Linear just retries.

2. **Revoked first-install not enforced at the gate** — a revoked workspace returns `'no-per-workspace-secret'` → stack-wide fallback, and since `setup` mirrors the first workspace's secret into the stack-wide one and revocation never clears it, that workspace's events still verify. Downstream, the processor logs a warning but still calls `createTaskCore` (`linear-webhook-processor.ts:254-270`), so revocation is only incidentally blocked by the project/user mapping tables. Suggest a distinct `'revoked'` outcome treated like `'mismatch'` (401, no fallback), and/or `return` in the processor when the workspace can't be resolved. (Note: the revoked test at `linear-webhook-multi-workspace.test.ts:285` only passes because the stack-wide secret differs there — worth a case where it matches.)

3. **Broken command in user-facing output** — `cli/src/commands/linear.ts:1665` tells the operator to run `bgagent linear link-user-self <slug>`, which isn't a registered command. Self-link is inline in `setup`/`add-workspace` now; suggest pointing there or at `invite-user`. Same stale reference in comments at `:628`, `:976`, `:1605` (cite `link-user`) and `linear-link.ts:63-67` / `linear-webhook-multi-workspace.test.ts:278` (cites the removed `--rotate-webhook-secret` flag).

### Governance

The branch is `feat/linear-add-workspace` — no issue number (expected `feat/<issue>-...` per ADR-003), and I don't see a linked `approved` issue or labels. Flagging for a maintainer to confirm this is authorized.

### Non-blocking nits

- `findReusableOauthAppCredentials` swallows a JSON-parse failure and surfaces "no active workspace, run setup" (`linear.ts:285-293` → `:816`), nudging toward a duplicate install when the secret is just corrupt.
- No log/metric on the *successful* stack-wide fallback (`linear-webhook.ts:131`) — it's the breadcrumb that would make items 1–2 debuggable.
- `link` makes a wasted dry-run call in `--output json` mode (`linear.ts:391`); gate behind `opts.output !== 'json'`.
- `list-projects` doesn't paginate `ListSecretsCommand` (`linear.ts:903`) — silently caps at 100 workspaces.
- Minor: invite-code modulo bias (`b % 31`, immaterial here); cdk-nag IAM5 suppression text doesn't describe the new Secrets Manager grant.

### Tests

The multi-workspace HMAC tests are the right kind (real signing, documents the attack). Gaps worth closing: the `linear-link.ts` `dry_run` branch (`:86-95`) is untested — that's the whole preview-without-write safety rail; `generateInviteCode`'s format is only asserted consumer-side; and the new commands / `runSelfLinkPicker` / `linearLink({dryRun})` have no behavioral coverage (extracting pure functions, as you did for `findReusableOauthAppCredentials`, would make the high-value bits testable). Also: `build (agentcore)` is still pending — needs to be green.

I verified the IAM prefix grant (read-only `GetSecretValue`, consistent with the processor), the HMAC `timingSafeEqual` comparison, the docs mirror sync, and types.ts parity — all clean. Happy to dig into any of the above.

🤖 Review assisted by code-reviewer, silent-failure-hunter, security-review, pr-test-analyzer, and comment-analyzer agents.

isadeks:
Thanks @krokoko — addressed in three commits on the branch tip. Mapping each review item to where it landed:

### Blockers
- **#1 fail-open on transient infra errors** → [`4570b9a`](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/200/commits/4570b9a). Added `getRegistryRowStrict` / `getOauthSecretStrict` that throw on infra error; `verifyLinearRequestForWorkspace` now uses them. Lenient `null`-returning helpers retained for `resolveLinearOauthToken`'s graceful-no-op contract.
- **#2 revoked-not-enforced** → [`4570b9a`](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/200/commits/4570b9a). Distinct `'revoked'` outcome, treated like `'mismatch'` (401, no fallback). Processor now early-returns when `resolveLinearOauthToken` returns null instead of falling through to `createTaskCore`. New test pins the exact bypass scenario you flagged (stack-wide secret matches AND workspace is revoked → 401). Second new test pins infra-error → 500.
- **#3 broken `link-user-self` + stale references** → [`4da242f`](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/200/commits/4da242f). `linear.ts:1665` now points at re-running setup/add-workspace's inline picker; comments at :628/:976/:1605 + `linear-link.ts:63` + multi-workspace test :278 cleaned up.

### Nits
All in [`4da242f`](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/200/commits/4da242f):
- `findReusableOauthAppCredentials` throws `CliError` on corrupt/missing-string/missing-fields (was collapsing to "no active workspace, run setup").
- Stack-wide-fallback success now logs an info breadcrumb for diagnosability.
- `bgagent linear link --output json` skips the dry-run (no interactive prompt to feed it).
- `list-projects` paginates `ListSecretsCommand`.
- IAM5 suppression text now mentions the SM `bgagent-linear-oauth-*` prefix grant the wildcard covers.

### Tests
[`28599dd`](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/200/commits/28599dd):
- `linear-link` dry_run preview branch (returns identity, doesn't write the confirmed mapping or delete the pending row).
- dry_run on a non-existent code still 404s — preview must not leak existence.
- `generateInviteCode` format / alphabet / distinctness.

`runSelfLinkPicker` behavioural coverage **deferred** — needs the helper extracted from the giant setup/add-workspace action first; that refactor didn't fit this PR.

### Governance
Branch is `feat/linear-add-workspace` without an issue number — that's a maintainer call to confirm authorization (per ADR-003).

### Verification
Full CDK suite green (1799 tests). Smoke-tested deployed to dev: Linear-triggered task created, agent ran end-to-end, comments posted, no regressions on the happy path.

krokoko (approved):

PR #208, opened 2026-05-28
feat(cdk): add CloudWatch alarms for FanOut + ApprovalMetricsPublisher DLQs

### LINKED ISSUE #117: Add CloudWatch alarms for FanOutConsumer + ApprovalMetricsPublisher DLQs
> **Follow-up from [PR #88](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/88)** — the design doc (§11.5) calls these out as planned but they didn't ship; tracking now so they don't drift, but ABCA is currently a reference application, so this is low-priority polish for whenever real production traffic shows up.

## Functional description

Two Lambdas in ABCA route through SQS Dead Letter Queues:

- **FanOutConsumer** (`cdk/src/handlers/fanout-task-events.ts`): consumes `TaskEventsTable` DDB streams and dispatches notifications to GitHub / Slack / email. Has `FanOutDlq`.
- **ApprovalMetricsPublisher** (`cdk/src/handlers/approval-metrics-publisher.ts`): consumes the same stream and publishes EMF metrics for the Cedar HITL dashboard. Has `ApprovalMetricsPublisherDlq`.

Both DLQs are wired up correctly. **Neither has a CloudWatch alarm on `ApproximateNumberOfMessagesVisible > 0`**. A poison-pill event (malformed JSON, downstream throttling, a code bug in the consumer) would silently accumulate in the DLQ until someone happens to look at the SQS console.

**Why this is lower-priority today than it would be in production:** ABCA is currently a reference application under active development. There's no on-call rotation, no production traffic, and the failure mode (silent notification dropout) hits zero customers today. This issue is a placeholder for "when a real adopter takes ABCA into production, they should ship these alarms before going live." Filing now so the design intent (already written in §11.5) doesn't decay.

**User-visible impact (when ABCA reaches production):**

- A regression in the FanOut handler ships without anyone noticing for hours-to-days, until a team member asks "did task X complete? I never got the Slack message."
- Approval-flow regressions show up as "the dashboard graph went flat" rather than as a paged alert.
- No SLO can be defined against "the FanOut path is healthy" because there's no signal.

**User-visible impact today:** none — but the design doc claims these alarms exist and they don't, which is a documentation/code drift.

## Technical context

**Where the DLQs are wired:**

- `cdk/src/constructs/fanout-consumer.ts` — defines `FanOutDlq` and attaches it as the failure target for the EventSourceMapping on the DDB stream.
- `cdk/src/constructs/approval-metrics-publisher-consumer.ts` — same pattern with `ApprovalMetricsPublisherDlq`.

**What's missing:**

- `cloudwatch.Alarm` resources on the `ApproximateNumberOfMessagesVisible` metric for both queues.
- Alarm action — for now, a CloudWatch alarm is enough; integrate to SNS / PagerDuty later.
- Optional: alarm on `ApproximateAgeOfOldestMessage > <threshold>` to detect "messages going in but DLQ consumer (none today) not draining" — but since there's no DLQ consumer planned, the count alarm covers it.

**Design reference:** `docs/design/CEDAR_HITL_GATES.md` §11.5 calls out:
> "DLQ alarms (planned, not in this milestone): both FanOutDlq and ApprovalMetricsPublisherDlq need `ApproximateNumberOfMessagesVisible > 0` CloudWatch alarms with SNS notification destinations."

That paragraph has been there since the design doc landed; the implementation kept slipping.

## Proposed fix

Two minimal additions in `cdk/src/constructs/fanout-consumer.ts` and `cdk/src/constructs/approval-metrics-publisher-consumer.ts`:

```typescript
new cloudwatch.Alarm(this, 'DlqMessageAlarm', {
  metric: this.dlq.metricApproximateNumberOfMessagesVisible({
    period: Duration.minutes(5),
    statistic: 'Maximum',
  }),
  threshold: 1,
  evaluationPeriods: 1,
  comparisonOperator: ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
  alarmDescription: 'FanOutConsumer DLQ has at least one message; investigate poison records',
  treatMissingData: TreatMissingData.NOT_BREACHING,
});
```

For SNS topic wiring, take the existing `OperatorAlarmTopic` if one exists, or add a new `OperatorAlarmTopic` construct + email subscription parameter. **Recommend keeping the SNS work scoped tight**: ship the alarms first, wire the topic in a follow-up PR if not already present.

## Acceptance criteria

- [ ] CloudWatch alarm exists for `FanOutDlq` on `ApproximateNumberOfMessagesVisible >= 1` over a 5-minute window
- [ ] CloudWatch alarm exists for `ApprovalMetricsPublisherDlq` on the same metric/threshold
- [ ] Both alarms have `alarmDescription` text that points to a runbook (can be a TODO link if the runbook doesn't exist yet)
- [ ] Construct unit tests verify alarm presence + threshold value
- [ ] (If `OperatorAlarmTopic` exists) both alarms send to it

## Out of scope

- Building a DLQ consumer Lambda that auto-retries (separate enhancement; harder design).
- PagerDuty / OpsGenie integration (separate operational concern).
- Alarms on the FanOut/Publisher Lambda functions themselves (errors / throttles) — that's a separate broader observability issue worth its own ticket if the appetite is there.

## References

- `cdk/src/constructs/fanout-consumer.ts` (FanOutDlq definition)
- `cdk/src/constructs/approval-metrics-publisher-consumer.ts` (ApprovalMetricsPublisherDlq definition)
- `docs/design/CEDAR_HITL_GATES.md` §11.5 (planned-but-not-shipped note)
- AWS DLQ alarm pattern reference: <https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html>

nizar-lahlali:
## Summary

- Adds `ApproximateNumberOfMessagesVisible >= 1` CloudWatch alarms on both `FanOutDlq` and `ApprovalMetricsPublisherDlq` (5-min window, Maximum statistic)
- Exposes alarms as `public readonly dlqAlarm` for future SNS topic wiring
- Adds construct-level unit tests verifying alarm presence and threshold configuration

Closes #117

## Test plan

- [x] `tsc --noEmit` compiles cleanly
- [x] `fanout-consumer.test.ts` passes (8 tests including new alarm test)
- [x] `approval-metrics-publisher-consumer.test.ts` passes (8 tests including new alarm test)
- [x] CI green


krokoko:
## Review summary 🧾

Thanks for picking up #117 — silent DLQ accumulation is exactly the kind of invisible failure ABCA should surface, and the alarm configuration itself is clean and idiomatic (correct metric, `Maximum`/5-min window, `>= 1` threshold, `NOT_BREACHING`). Tests pass locally (14/14). A few things to resolve before this is merge-ready.

**Verdict: Request changes** — the code is sound, but the change needs to be reconciled with the design decision it reverses and the operator copy finished.

### Blocking 🚧

1. **Governance — issue #117 isn't `approved`.** The issue currently has no labels (a member comment even flagged that `P3`/`enhancement` couldn't be applied and asked a triager to add them). Per [ADR-003](docs/decisions/003-contribution-governance.md), the `approved` label needs to land before merge.

2. **Doc drift — §11.5 says the opposite of what we're shipping.** `docs/design/CEDAR_HITL_GATES.md` §11.5 (lines 1588–1601) and the roadmap note at line 2073 still describe these DLQ alarms as *deferred / out-of-scope for v1 until a notification channel exists* — and the new code comments cite `§11.5` as justification for the reverse. Since the alarms ship with no `addAlarmAction`/SNS target, they currently reproduce the exact "fires into the void" condition §11.5 warned about. Please update §11.5 + line 2073 to record this decision (alarms now ship without a notification action, SNS wiring as follow-up), then regenerate the Starlight mirror via `cd docs && node scripts/sync-starlight.mjs` and commit it.

3. **`Runbook: TODO` in operator-facing `alarmDescription`** (`fanout-consumer.ts:200`, `approval-metrics-publisher-consumer.ts:166`). This string shows up in the CloudWatch console / notifications, so a `TODO` reads as unfinished during an incident. The diagnostic sentence after it is genuinely useful — suggest dropping the `TODO` token or pointing at #117.

### Non-blocking / nits 💡

- **Misleading `// §11.5:` comments and `(§11.5)` test titles** — same root as #2; reword to reference the implementing issue once the doc is updated.
- **`dlqAlarm` is unreachable today** — both constructs are instantiated in `agent.ts` (lines 672, 696) without capturing the reference, so the "for future SNS wiring" property has no consumer yet. Mirrors the existing (also-unconsumed) `errorAlarm`, so not a regression — but consider wiring the SNS action here, or keeping the alarm private until a consumer exists.
- **Narrow the type to `cloudwatch.IAlarm`** — `addAlarmAction` lives on `IAlarm`; exposing the concrete `Alarm` leaks the full mutable surface.
- **Add a JSDoc to `dlqAlarm`** to match the `errorAlarm` precedent in `task-orchestrator.ts:162`.
- **Removed deferral rationale** — the deleted comment explaining *why* alarms were deferred is still relevant (the SNS precondition is unmet); consider a one-liner noting the alarm intentionally has no action yet.

### Tests 🧪

Coverage of the scalar alarm config is solid. Two cheap additions worth making while you're here — they guard the most plausible failure given the two constructs are near-identical copies:

1. **Assert the `QueueName` dimension** so the alarm is pinned to *its own* DLQ — nothing currently catches a metric wired to the wrong queue.
2. **`template.resourceCountIs('AWS::CloudWatch::Alarm', 1)`** — matches the count-assertion convention the rest of these test files already use.

### Review agents run

`code-reviewer`, `type-design-analyzer`, `comment-analyzer`, `pr-test-analyzer` ✅ — `silent-failure-hunter` and `/security-review` omitted (no error-handling and no IAM/Cedar/network/secrets surface in the diff).

Really nice, focused change overall — once the doc reconciliation and runbook text are sorted it'll be in good shape. 🙏

krokoko (dismissed):
See comment

isadeks:
LGTM: the alarm config (threshold, statistic, window, treatMissingData) is right and the doc reframing of §11.5 is honest about the action-less limitation.
  
One non-blocking design question: an action-less alarm is roughly invisible — the only surface is the CloudWatch Alarms console. A 5-line AlarmStatusWidget row in TaskDashboard would deliver the "operator visibility" the docs describe today without waiting for the SNS notification channel. Worth doing in this PR or the immediate follow-up — your call.

cdk/src/constructs/fanout-consumer.ts
```diff
@@ -243,8 +245,11 @@ export class FanOutConsumer extends Construct {
       }),
       threshold: 1,
       evaluationPeriods: 1,
+      comparisonOperator: cloudwatch.ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD,
       alarmDescription:
-        'Fan-out DLQ has undelivered task-event records — Slack/GitHub/Linear notifications are failing',
+        'Fan-out DLQ has undelivered task-event records — Slack/GitHub/Linear notifications are failing. ' +
```
scottschreckengaust: Nit (non-blocking, operator copy): this dropped `Jira` from the channel list — `main` currently ships `…Slack/GitHub/Linear/Jira notifications are failing`, and the merge landed `…Slack/GitHub/Linear…`. It's cosmetic. Note the fanout handler's `NotificationChannel` union is `'slack' | 'email' | 'github' | 'linear'` (no Jira), so this text is arguably *more* accurate — but it silently diverges from what main ships. Decide deliberately: either restore `Jira` to match main, or leave it and know you've corrected an over-broad description.

cdk/test/constructs/fanout-consumer.test.ts
```diff
@@ -204,4 +204,24 @@ describe('FanOutConsumer', () => {
...
+
+    template.hasResourceProperties('AWS::CloudWatch::Alarm', {
+      MetricName: 'ApproximateNumberOfMessagesVisible',
+      Namespace: 'AWS/SQS',
+      Threshold: 1,
+      EvaluationPeriods: 1,
+      ComparisonOperator: 'GreaterThanOrEqualToThreshold',
+      TreatMissingData: 'notBreaching',
+      Statistic: 'Maximum',
+      Period: 300,
```
scottschreckengaust: Nit (coherence, non-blocking): the twin `approval-metrics-publisher-consumer.test.ts` alarm test adds `template.resourceCountIs('AWS::CloudWatch::Alarm', 1)` as a duplicate-guard (the exact failure mode this branch's merge had to fix). Adding the same one-line count assertion here would give the fanout alarm test symmetric protection against an accidental second alarm.

scottschreckengaust (dismissed):
## Verdict: Approve (one non-blocking nit)

Clean, tightly-scoped observability change that closes #117 and — importantly — this revision has resolved every prior blocking concern from the earlier `Request changes` review. I re-verified each prior claim against the current head (`9740b34`) rather than repeating it:

1. **Governance / `approved` label** — RESOLVED. Issue #117 now carries `approved` (plus `P2`, `infra-cdk`, `observability`, `enhancement`). ADR-003 gate satisfied.
2. **Doc drift, §11.5 said the opposite** — RESOLVED. `docs/design/CEDAR_HITL_GATES.md` §11.5 and the roadmap line (~2108) are rewritten to state the DLQ-depth alarms ship *without* an `addAlarmAction`/SNS target as an intentional intermediate step, with SNS wiring as a bounded follow-up. This is honest about the action-less limitation. The Starlight mirror (`docs/src/content/docs/architecture/Cedar-hitl-gates.md`) is in sync — I ran `node docs/scripts/sync-starlight.mjs` and it produced zero drift, so CI's "Fail build on mutation" step will pass.
3. **`Runbook: TODO` in operator-facing `alarmDescription`** — RESOLVED. Both descriptions now carry actionable diagnostic text ("Check CloudWatch Logs for the FanOutFn / ApprovalMetricsPublisherFn error…") plus a link to #117; no `TODO` token remains.
4. **Type narrowed to `cloudwatch.IAlarm`**, **JSDoc added** to match the `errorAlarm` precedent (`task-orchestrator.ts:187`), **`QueueName` dimension assertion**, and **`resourceCountIs('AWS::CloudWatch::Alarm', 1)`** guard on the approval-metrics test — all present.

### Vision alignment

Advances the **bounded blast radius & observability** tenet (VISION.md): surfaces silent poison-pill accumulation in two best-effort DLQs that would otherwise sit unnoticed for the 14-day retention window. The action-less-alarm trade is now explicitly documented in §11.5, so it is a *documented* tenet trade, not a silent one. Outcome remains reviewable (durable ALARM state, queryable in console/dashboard).

### Merge-conflict resolution (the interesting part)

Main independently shipped `FanOutDlqDepthAlarm` while this branch was open. The head merge commit correctly **drops this branch's duplicate `DlqMessageAlarm` on the fanout construct** and folds its richer diagnostic + #117 link into the surviving `FanOutDlqDepthAlarm`, keeping only the net-new `ApprovalMetricsPublisher` `dlqAlarm`. I confirmed the fanout construct creates exactly one alarm (no double-create) and that `agent.ts` instantiates both consumers without capturing the alarm reference — so `dlqAlarm`/`dlqDepthAlarm` have no consumer yet, but this mirrors the existing (also-unconsumed) `errorAlarm`, so it is not a regression.

### Blocking issues

None.

### Non-blocking / nits

1. **`fanout-consumer.ts:250` — dropped `Jira` from the channel list.** Main currently ships `…Slack/GitHub/Linear/Jira notifications are failing`; the merged text here is `…Slack/GitHub/Linear…` (no Jira). Purely cosmetic operator copy. Note the fanout handler's `NotificationChannel` union is `'slack' | 'email' | 'github' | 'linear'` (no Jira), so this version is arguably *more* accurate — but it does silently diverge from what main ships today. Pick one deliberately. (Inline.)
2. **Test asymmetry (coherence).** `approval-metrics-publisher-consumer.test.ts:199` adds a `resourceCountIs('AWS::CloudWatch::Alarm', 1)` duplicate-guard; the twin `fanout-consumer.test.ts` alarm test does not. Since main's fanout construct has one alarm, a symmetric count guard there would catch the same accidental-duplicate failure mode the approval test now guards against. Cheap to add. Non-blocking.
3. **`task-orchestrator.ts:187`** still exposes `errorAlarm` as the concrete `cloudwatch.Alarm`. Out of scope for this PR, but the same `IAlarm`-narrowing this PR applies would tighten that surface — worth a follow-up.

### Documentation

Correctly updated and in scope: §11.5 rewrite + roadmap line + regenerated Starlight mirror (verified in sync). New behavior (two shipped alarms) is documented. Backing issue #117 filed with `P2`. No AGENTS.md/README flow change needed.

### Tests & CI

Coverage is solid: both tests pin metric name, namespace, threshold=1, single evaluation period, `GreaterThanOrEqualToThreshold`, `notBreaching`, `Maximum`, 300s period; approval test additionally pins the `QueueName` dimension to its own DLQ and asserts exactly one alarm. **#366 (synth perf): no regression** — bundling stays globally disabled via `test/setup/disable-bundling.ts`; the new tests reuse each file's existing `new App()` / `createStack()` convention and add no bundled-asset synth. **Bootstrap coverage (ADR-002): N/A** — `AWS::CloudWatch::Alarm` is already in `cdk/src/bootstrap/resource-action-map.ts:62` (`cloudwatch:PutMetricAlarm`) and main already ships alarms of this type, so no net-new CFN resource type is introduced; no bootstrap bump required.

### Review agents run

- **code-reviewer** — ran (CDK L2 idiom, removal policy, cdk-nag, routing all clean; alarm config idiomatic).
- **type-design-analyzer** — ran (new `dlqAlarm: IAlarm` and the `Alarm`→`IAlarm` narrowing are the correct minimal-surface choice: `addAlarmAction` lives on `IAlarm`).
- **comment-analyzer** — ran (comments now cite the implementing issue #117, not the stale §11.5 justification; the `ApprovalMetricsPublisherDlq` comment was correctly updated from "deferred" to "alarmed below").
- **pr-test-analyzer** — ran (findings: nit #2 above; happy-path + wrong-queue guard covered).
- **silent-failure-hunter** — OMITTED: no error-handling / catch-block code in the diff (pure CDK resource declaration + docs + tests).
- **/security-review** — OMITTED: no IAM policy, Cedar policy, network, secrets, or input-gateway surface changed; a `cloudwatch.Alarm` on an existing metric adds no principal-facing permission.

### Human heuristics

- **Proportionality** — Pass. Two ~20-line construct additions + doc reconciliation; no new abstraction for a one-off.
- **Coherence** — Concern (minor): the two near-identical constructs use different alarm logical IDs (`FanOutDlqDepthAlarm` vs `DlqMessageAlarm`) and different public property names (`dlqDepthAlarm` vs `dlqAlarm`), and the tests are asymmetric (nit #2). Same concept, slightly different vocabulary. Non-blocking but worth a unifying pass someday.
- **Clarity** — Pass. Names communicate intent; `DLQ_ALARM_PERIOD_MINUTES` is a named constant, not a magic value; alarm descriptions surface the failure and where to look.
- **Appropriateness** — Pass. Idiomatic CDK maintainable by this team; tests assert what the alarm *should* be (threshold/metric/dimension), not merely a synth snapshot.

ayushtr-aws (approved):

scottschreckengaust (approved):
## Verdict: **Approve** (three non-blocking nits)

Tightly-scoped observability change. Net-new behavior is one `AWS::CloudWatch::Alarm` on `ApprovalMetricsPublisherDlq` plus explicitness/copy hardening on the FanOut alarm `main` already ships. I verified every prior blocking claim against the current head (`20c71ea5`) rather than re-asserting it, and I re-verified the type decision the last revision reversed — the reversal is **correct**.

---

## 1. Prior review threads — RESOLVED / STILL OPEN

### krokoko `CHANGES_REQUESTED` (2026-05-28)

| # | Claim | Status | Evidence |
|---|-------|--------|----------|
| 1 | Issue #117 not `approved` | **RESOLVED** | `gh issue view 117` → labels `enhancement`, **`approved`**, `infra-cdk`, `observability`, `P2`. ADR-003 gate satisfied. |
| 2 | §11.5 says the opposite of what ships; mirror stale | **RESOLVED** | `docs/design/CEDAR_HITL_GATES.md:1587-1600` rewritten to "DLQ-depth alarms (shipped)" and states explicitly they ship **without** an `addAlarmAction`. Roadmap line 2111 updated. Mirror verified: I ran `node docs/scripts/sync-starlight.mjs` in the worktree → exit 0, `git status --porcelain` **empty**. CI's "Fail build on mutation" will pass. |
| 3 | `Runbook: TODO` in operator-facing `alarmDescription` | **RESOLVED** | `fanout-consumer.ts:300-303` and `approval-metrics-publisher-consumer.ts:183-186` both carry actionable diagnostic text + the #117 link. `grep -n TODO` over both constructs: no hits. |
| nit | Misleading `// §11.5:` comments / test titles | **RESOLVED** | Comments and test names now cite `#117`. |
| nit | `QueueName` dimension assertion | **RESOLVED** | `approval-metrics-publisher-consumer.test.ts:207-217`, `fanout-consumer.test.ts:157-165`. |
| nit | `resourceCountIs('AWS::CloudWatch::Alarm', 1)` | **RESOLVED both sides** | `approval-...test.ts:198` and `fanout-consumer.test.ts:147`. This also closes the symmetry nit from the dismissed review. |
| nit | **Narrow the type to `cloudwatch.IAlarm`** | **CORRECTLY REJECTED — do not re-apply** | This was the one piece of prior guidance that was *wrong*, and head `20c71ea5` reverts it. `addAlarmAction` is declared on the abstract class `AlarmBase`, **not** on `interface IAlarm`. Verified against `aws-cdk-lib@2.261.0` `aws-cloudwatch/lib/alarm-base.d.ts`: `IAlarm` (line 18) exposes only `alarmArn` + `alarmName`; `addAlarmAction` is at line 63 inside `declare abstract class AlarmBase`. Typing the property as `IAlarm` would have **broken the exact SNS wiring (#228/#629) it claimed to enable**. Keeping `cloudwatch.Alarm` matches `task-orchestrator.ts:212` (`errorAlarm`) and `github-screenshot-integration.ts:139` (`processorDlqDepthAlarm`). Good catch by the author; the JSDoc at `fanout-consumer.ts:163-166` documents the reasoning so nobody re-narrows it. |

### scottschreckengaust `DISMISSED` (2026-07-21)

Both inline nits are now moot: the `Jira` channel copy is restored (`fanout-consumer.ts:301` reads `Slack/GitHub/Linear/Jira`, matching `main`), and the fanout `resourceCountIs` symmetry nit is fixed (`fanout-consumer.test.ts:147`).

### isadeks comment (2026-06-05) — `AlarmStatusWidget` in `TaskDashboard`

**STILL OPEN, and it is the one substantive gap.** `grep -i alarm cdk/src/constructs/task-dashboard.ts` → **zero hits**; the dashboard is `LogQueryWidget` + `GraphWidget` only. See nit 1 — this makes one sentence in the new doc prose inaccurate.

### ayushtr-aws `APPROVED` (2026-08-07)

Noted, but I judged the code independently rather than deferring to it.

---

## 2. The scope call: alarms with no action — **acceptable, ship it**

Stating this explicitly rather than leaving it implicit, since "an alarm with no action notifies nobody" is a real objection.

**Why it is acceptable here:**

1. **The unit of value is state, not the notification.** An `AWS::CloudWatch::Alarm` maintains durable, queryable `ALARM` state and 15-month state-transition history. `DescribeAlarms --state-value ALARM` becomes a valid health probe; `aws sqs get-queue-attributes` polling is not. That is a genuine step up from today's "nothing."
2. **The alternative is worse.** Deferring the alarm until SNS exists means the queue metric is unwatched *and* has no retained history — so when the channel does land, there is no backfill of "was this ever breaching?"
3. **The wiring is genuinely one line, and the surface is now correct for it.** Because the reverted type keeps `dlqAlarm: cloudwatch.Alarm`, #228/#629 is literally `alarm.addAlarmAction(new SnsAction(topic))` with zero construct restructuring. No metric, threshold, or dimension rework.
4. **It is documented as a trade, not hidden.** `CEDAR_HITL_GATES.md:1587-1600` names the limitation in bold and points at the follow-up. Per the review rubric an undocumented tenet trade is blocking; a documented one is a judgment call, and this one is defensible.
5. **Blast radius is zero.** Two alarms at $0.10/alarm-month, no new IAM, no new network path.

What would flip this to blocking: if the PR *claimed* operators get notified. It does not — it says the opposite. So this is (a) acceptable incremental scope, not (b) a dead-observability half-feature.

**Tracking debt — #228 and #629 are duplicates.** Both are OPEN with identical labels (`enhancement`, `infra-cdk`, `observability`) and describe the same work: create an SNS topic and `addAlarmAction` it to both DLQ alarms. #228 was filed off this PR; #629 restates it with more detail and correctly notes #208 is unmerged. **Recommend closing #228 as a duplicate of #629** (#629 has the better write-up and the accurate not-yet-merged framing) — or the reverse, but pick one. Carrying two is how a bounded follow-up quietly becomes never-done. Note also that #629's text says the properties are exposed as `cloudwatch.IAlarm`, which head `20c71ea5` deliberately reversed — that sentence should be corrected on the surviving issue so the follow-on implementer isn't misled.

---

## 3. Vision alignment

Advances **bounded blast radius & observability** (VISION.md). Both DLQs sit on best-effort, batch-non-failing paths (`reportBatchItemFailures: true`, `retryAttempts: 3`), so the DLQ is the *only* persistent signal that fan-out or approval-metrics is down. Before this PR a poison record could sit unnoticed for the full 14-day retention. Outcomes stay inspectable. No ADR needed — no tenet is traded away.

---

## 4. Domain review: alarm configuration — correct

This is the part worth getting right, and it is right.

- **`treatMissingData: NOT_BREACHING`** — correct, and the classic footgun is avoided. SQS publishes `ApproximateNumberOfMessagesVisible` on a 5-minute cadence but **stops publishing entirely when a queue has been idle for ~6 hours**. Under `MISSING` (the CloudWatch default) an idle DLQ drifts to `INSUFFICIENT_DATA` and the alarm looks broken during a review; under `BREACHING` it false-fires on every idle window. `NOT_BREACHING` correctly reads "no data on an empty DLQ = healthy." Consistent with all three sibling alarms (`fanout:304`, `github-screenshot-integration:238`, `task-orchestrator:430`).
- **`period: 5min` vs SQS publication interval** — correctly matched. A 1-minute period against a 5-minute publication cadence would produce 4-of-5 empty datapoints and, combined with `NOT_BREACHING`, would suppress real breaches. 5/5 is right.
- **`Maximum` statistic** — right for a gauge metric. `Sum` over a gauge is meaningless; `Average` would dilute a single-message spike below the threshold-1 boundary.
- **`threshold: 1` + `GREATER_THAN_OR_EQUAL_TO_THRESHOLD`** — `>= 1` is the correct encoding of "any message." Note `>0` and `>=1` are equivalent here but `>=1` is the idiomatic integer form.
- **`evaluationPeriods: 1` / no `datapointsToAlarm`** — correct for this signal, and deliberately *unlike* #284's webhook Errors alarm, which uses 1-of-2 `datapointsToAlarm` to damp a spiky rate metric. DLQ depth is monotonic-until-drained, not spiky: a message that arrived is still there next period, so M-of-N adds only detection latency with no flap protection. Right call; the asymmetry with #284 is justified, not inconsistent.
- **L2 usage** — `dlq.metricApproximateNumberOfMessagesVisible()` rather than a hand-rolled `new Metric({ namespace: 'AWS/SQS', dimensions: ... })`. Correct: the L2 helper wires the `QueueName` dimension from the queue's own token, so the alarm cannot drift onto the wrong queue on rename.
- **`comparisonOperator` addition on the FanOut alarm (`fanout-consumer.ts:299`)** — behaviorally a **no-op**. I confirmed in `aws-cdk-lib@2.261.0` `alarm.js` that the constructor does `props.comparisonOperator || ComparisonOperator.GREATER_THAN_OR_EQUAL_TO_THRESHOLD`. So no synth diff on `main`'s existing alarm. Still worth keeping — relying on an implicit default for the operator that decides alarm polarity is exactly the thing that breaks silently if CDK ever changes it, and the paired test assertion now pins it.
- **No hardcoded ARNs / account IDs; cdk-nag clean** (existing `AwsSolutions-SQS3` suppression on the DLQ is unrelated and correctly reasoned).

---

## 5. Bootstrap policy coverage — **PASS, no updates required**

Worked the Stage 3 checklist. No new CFN resource *type* is introduced — `AWS::CloudWatch::Alarm` already ships on `main`:

1. `cdk/src/bootstrap/policies/observability.ts:76-79` — `cloudwatch:PutMetricAlarm`, `DeleteAlarms`, `DescribeAlarms` present, `resources: ['*']` (line ~102). No ARN-pattern gap: alarms are named by CDK logical id, and the statement is already unscoped, so a new alarm needs no new pattern. This is the "action present but resource too narrow" failure mode from #350 — checked, does not apply.
2. `cdk/src/bootstrap/resource-action-map.ts:62` — `'AWS::CloudWatch::Alarm': ['cloudwatch:PutMetricAlarm']` already mapped.
3. **`BOOTSTRAP_VERSION` bump correctly NOT taken** — `version.ts` stays `1.2.0`. No new permission is granted, so a bump would churn `BOOTSTRAP_HASH` and force every adopter to re-bootstrap for zero policy delta. Right restraint.
4. `docs/design/DEPLOYMENT_ROLES.md` golden baseline — unchanged, consistent with (3).
5. **`cdk/test/bootstrap/synth-coverage.test.ts` — I ran it.** `npx jest test/bootstrap` → **6 suites, 91 tests, all pass**, including `synth-coverage`, `golden-baseline`, `artifact-sync`, `version`. This is the check that would have caught a missing map entry; it is green.

---

## 6. Tests & CI — good

- `npx jest test/constructs/approval-metrics-publisher-consumer test/constructs/fanout-consumer` → **18/18 pass**. `approval-metrics-publisher-consumer.ts` at **100% statements / 100% branches**.
- `npx tsc --noEmit` → exit 0. `npx eslint` on all four touched files → exit 0, no fixes pending (so CI's "Fail build on mutation" is safe).
- Assertions are strong, not tautological: `MetricName`, `Namespace`, `Threshold`, `EvaluationPeriods`, `ComparisonOperator`, `TreatMissingData`, `Statistic`, `Period` **and** the `QueueName` `Fn::GetAtt` dimension. The dimension assertion is the valuable one — the two constructs are near-identical copies, and nothing else would catch an alarm accidentally bound to the sibling DLQ. Plus `resourceCountIs(..., 1)` on both sides against the accidental-duplicate failure mode this branch's own merge history actually hit.
- **Test performance (#366) — clean.** No `aws:cdk:bundling-stacks` manipulation; global `test/setup/disable-bundling.ts` untouched. The new approval test reuses the existing module-level `createStack()` helper and adds exactly one synth — and the last commit *deleted* a redundant duplicate FanOut alarm test rather than adding a second synth there. That is the right direction on this rubric.

---

## 7. Base freshness / rebase — **not warranted, and I verified it**

Branch is 5 commits behind `origin/main` (`c927a209`). Rather than guess, I test-merged:

- `git diff HEAD...origin/main -- <the 4 touched cdk files>` → **empty**. None of the 5 drift commits touch either construct or either test.
- The drift *does* touch `cdk/src/bootstrap/{version,resource-action-map,policies/index}.ts` and adds `policies/compute-lambda-microvm.ts` (#689 Lambda MicroVMs) — the plausible collision surface, since bootstrap has hash/version/golden-baseline coupling. But because this PR deliberately makes **no** bootstrap change (§5), there is nothing to conflict.
- `git merge --no-commit --no-ff origin/main` → *"Automatic merge went well"*, no conflicts. On the merged tree: `npx jest test/constructs/... test/bootstrap` → **8 suites, 124 tests, all pass**, including `golden-baseline` and `artifact-sync` against `main`'s new `BOOTSTRAP_HASH`.

So the drift is inert. GitHub reports `MERGEABLE`. A rebase is optional hygiene, not a precondition. One caveat worth naming: per repo convention, if you *do* merge `main`, re-run `mise //cdk:eslint` and commit any auto-fixes before pushing, or CI's mutation check rejects it.

---

## 8. Documentation — complete and honest

- `docs/design/CEDAR_HITL_GATES.md` §11.5 rewritten from "deferred" to "shipped, without an action target," with the residual alarms (timeout rate, stuck-in-`AWAITING_APPROVAL`, write-failure rate, gate cap, `MetricEmitSkipped`, heartbeat) correctly left deferred and the now-shipped DLQ bullet removed from that list. Roadmap line 2111 updated to match.
- Starlight mirror regenerated and **verified in sync** (sync script → zero drift). Generated file was not hand-edited.
- Issue tracking: #117 exists, `approved`, `P2`, and its acceptance criteria are all met (alarm on each DLQ ✓, runbook-pointing `alarmDescription` ✓, construct unit tests for presence + threshold ✓, `OperatorAlarmTopic` — does not exist, correctly out of scope). No AGENTS.md/README change needed — no developer flow or command changed.

---

## Non-blocking nits

**1. `CEDAR_HITL_GATES.md:1594` — "appear in the console/dashboard" is half-inaccurate.** The alarms appear in the CloudWatch **Alarms console**, but not in any *dashboard*: `TaskDashboard` (`cdk/src/constructs/task-dashboard.ts`) contains zero `AlarmStatusWidget` / `AlarmWidget` — grep for `alarm` returns nothing. The sentence's own next clause ("operators must check the CloudWatch Alarms console") contradicts the "dashboard" claim. Either drop "/dashboard", or take isadeks' suggestion and add a ~5-line `AlarmStatusWidget` row to `TaskDashboard` — which would make the sentence true and deliver most of the operator visibility the doc promises without waiting on SNS. My preference is the widget (it is genuinely small and it is the cheapest thing that makes these alarms *seen*), but fixing the prose is sufficient to merge. Comment-accuracy issue, not a correctness one.

**2. `DLQ_ALARM_PERIOD_MINUTES = 5` is now triplicated** — `fanout-consumer.ts:41`, `approval-metrics-publisher-consumer.ts:36`, `github-screenshot-integration.ts:42`. Three copies is the threshold where these drift. Since the 5-minute value is not arbitrary (it tracks the SQS metric publication cadence — see §4), it is a real shared contract and belongs in `contracts/constants.json` or a small shared `observability-defaults.ts`, with a comment recording *why* 5. Not worth blocking a 2-file change over; worth folding into the #629 SNS PR, which will touch all three call sites anyway.

**3. `fanout-consumer.ts:174-175` — the "three times" comment fix is right but the reasoning is subtler than stated.** `retryAttempts: 3` on `DynamoEventSource` (line ~318) does make "five times" wrong, so the correction is correct and worth having. Precision note: CDK's `retryAttempts` defaults to `-1` (infinite), so the explicit `3` is what bounds it — the comment reads as if 3 were inherent. Also note `retryAttempts` bounds *batch* retries, and with `reportBatchItemFailures: true` the batch is split on partial failure, so "the record was attempted 3 times" is approximately-but-not-exactly true. Fine as operator-facing prose; flagging only so it is not treated as a precise contract. Same wording now appears in `approval-metrics-publisher-consumer.ts:113`.

---

## Human heuristics

- **Proportionality — pass.** Two `cloudwatch.Alarm` instantiations and two test blocks for two alarms. No factory, no `AlarmFactory`, no props plumbing for a value nothing configures. The last commit *removed* a redundant test rather than adding surface. Correct altitude.
- **Coherence — pass.** Fourth instance of an established shape (`task-orchestrator.ts:423`, `github-screenshot-integration.ts:229`, `fanout-consumer.ts:292`, now `approval-metrics-publisher-consumer.ts:175`) with identical statistic/period/`treatMissingData`, and the two tests are now genuinely symmetric (both have `resourceCountIs` + `QueueName` dimension) rather than parallel-looking boilerplate. Constant triplication (nit 2) is the one coherence smudge.
- **Clarity — pass, with nit 1.** Names say what they are (`dlqAlarm`, `dlqDepthAlarm` — the fanout one is marginally better; the `dlqAlarm`/`dlqDepthAlarm` split across two constructs for the identical concept is a small same-concept-different-term wobble, not worth churning a public property over). `alarmDescription` strings are real runbooks: what broke, which log group to read, which issue. The `fanout-consumer.ts:163-166` JSDoc explaining *why* the type is `Alarm` and not `IAlarm` is exactly the comment that prevents someone "tidying" it back and silently breaking the SNS follow-up. Only inaccuracy is the "/dashboard" claim.
- **Appropriateness — pass.** Alarm semantics verified against real CloudWatch/SQS behavior (idle-queue metric suppression driving the `NOT_BREACHING` choice), not against self-written mocks. Tests assert what the alarm *should* be (including the wrong-queue failure mode), not merely what synth currently emits. The `IAlarm` reversal is the strongest signal here — the author verified the actual `aws-cdk-lib` type surface instead of accepting plausible-sounding review guidance, and documented the finding.

---

## Review agents run

**Process disclosure, per Stage 3.** The `pr-review-toolkit` agents (`code-reviewer`, `silent-failure-hunter`, `type-design-analyzer`, `comment-analyzer`, `pr-test-analyzer`) and the `security-review` skill were **NOT invoked — no Agent/Task dispatch tool exists in this execution context** (this review runs as a subagent of a `/review_prs` fan-out, and subagent nesting is capped at one level, so agent dispatch is unavailable). This is a tooling limitation, not a skipped step. I substituted direct verification for each agent's scope and am naming that explicitly so it is not mistaken for coverage:

| Agent | Substituted with |
|-------|------------------|
| `code-reviewer` | Manual read of all 6 files + `npx eslint` (clean) + `npx tsc --noEmit` (clean) + L2/threshold/statistic review vs. the 3 sibling alarms. |
| `type-design-analyzer` | Read `aws-cdk-lib@2.261.0` `alarm-base.d.ts` + `alarm.js` directly to adjudicate the `Alarm` vs `IAlarm` reversal (§1) and confirm the `comparisonOperator` default. |
| `comment-analyzer` | Line-by-line audit of every changed comment: found nit 1 (`:1594` dashboard claim, contradicted in-sentence) and nit 3 (`retryAttempts` framing); confirmed the "five times"→"three times" fix against `retryAttempts: 3`. |
| `pr-test-analyzer` | Ran both suites (18/18) + full `test/bootstrap` (91/91) + post-merge run (124/124); audited assertion strength, synth count, and #366 bundling rules. |
| `silent-failure-hunter` | **Genuinely out of scope** — the diff adds no error handling, no catch blocks, no fallbacks. Zero runtime code paths changed; this is declarative CDK plus doc prose. |
| `security-review` | **Genuinely out of scope** — no IAM statement, Cedar policy, network config, secret, or input-gateway surface in the diff. I still checked the bootstrap IAM implication independently (§5) since that is the adjacent security concern, and confirmed no new grant is required. |

---

## Summary

Approve. Every krokoko blocker is resolved with evidence; the one piece of prior guidance that was technically wrong (`IAlarm`) was correctly reverted with a verified rationale and a JSDoc guard. Alarm configuration is right on the details that actually matter for a DLQ depth metric. Bootstrap coverage verified green rather than assumed. Base drift verified inert by test-merge. The three nits are cheap and none blocks merge — please do fix the `:1594` "/dashboard" wording (or add the widget) since it is the only statement in the PR that is not accurate, and please close one of #228/#629 as a duplicate.

PR #242, opened 2026-06-02
chore(project): quality improvements

krokoko:
- add a failure in case of code coverage not being met (enforcing the rule)
- update the review pr command 
- add new mise commands to run drift prevention mechanism as part of build

#### Area

#### Acknowledgment

### DIFF
--- a/.abca/commands/review_pr.md
+++ b/.abca/commands/review_pr.md
@@ -78,6 +78,24 @@ Then apply principal-architect judgment over the diff:
 - **Routing** — Changes should land in the right package per the AGENTS.md routing table
   (agent runtime in `agent/`, API/Lambdas in `cdk/`, CLI in `cli/`).
 
+**Human review heuristics (non-automatable)** — After automated agents, apply these four
+dimensions. They are where agent output often looks plausible but isn't. Flag blocking issues when
+a dimension is clearly violated:
+
+- **Proportionality** — Does complexity match the problem? (new abstraction/factory/"engine"
+  for a one-off → AI002/AI003; files >800 lines — is size essential or accreted?)
+- **Coherence** — Does it belong here? Same concept = same term across the repo? Parallel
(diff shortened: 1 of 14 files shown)


isadeks (commented):
Two things to fix before this can merge — otherwise the shape looks good.
  
  1. Broken Starlight link in 
  docs/src/content/docs/developer-guide/Contributing.md:61

  The mirror points at /architecture/review-pr, which doesn't exist on the docs
  site (no Review-pr.md under docs/src/content/docs/architecture/). The root
  CONTRIBUTING.md:57 correctly uses a repo-relative path to
  .abca/commands/review_pr.md, but that path doesn't translate to the published
  site, so the mirror was hand-edited to a URL that goes nowhere.

  Suggested fix: since .abca/commands/ isn't part of the published docs, link to
  the GitHub source directly, or drop the link in the Starlight version and
  keep the prose. Whichever you pick, please confirm mise //docs:sync
  regenerates a consistent result (i.e. the source markdown's relative link is
  what produces this absolute URL, not a manual edit).

  2. agent/uv.lock revision downgraded 3 → 1

  The header flipped from revision = 3 to revision = 1, and the 1425-line diff
  is essentially stripping upload-time = "..." from every package entry — that's
  the schema an older uv produces. Adding pytest-cov==7.1.0 alone should be a
  ~10-line diff. The next maintainer running uv sync with the project's pinned
  uv will silently re-bump it to revision 3 and re-add all the metadata,
  churning the lockfile again.

  Suggested fix: regenerate the lockfile using the project-pinned uv (mise has
  0.11.x available; looks like this was generated against an older Homebrew uv
  0.9.9), then push only the actual pytest-cov addition.
  
  Once those are addressed I'm happy to re-review. The contract → enforcement →
  drift-check triad and the four-dimension heuristic in review_pr.md are well
  done.

isadeks (approved):

PR #243, opened 2026-06-02
feat(notifications): Linear dispatcher in fanout consumer (cost/turns/duration)

### LINKED ISSUE #239: feat(notifications): platform-side final-status Linear comment with cost/turns/duration
## Problem

When a Linear-triggered task finishes, the requester gets no metrics on the issue. Current behavior:

- **GitHub PR side** — the edit-in-place dispatcher (DDB Streams fanout) posts a status comment that updates with phase, milestone, **cost so far**, and a link to the task. Requester always sees the numbers.
- **Linear side** — the agent itself posts three comments via `mcp__linear-server__save_comment` (start, PR-opened, completion). The completion comment is freeform "succeeded / failed + short reason." **No cost, no turns, no duration.**

This produces real UX gaps:

1. **Requester has no usage signal.** They open Linear, see "✅ task completed", and have to dig into CloudWatch / `bgagent status` to see what it cost. They also can't tell the difference between a 5-turn $0.20 fix and a 100-turn $3.40 sprint that hit `max_turns`.
2. **Agent crashes leave Linear silent.** If the agent crashes (e.g. `error_max_turns`, `Runtime.OutOfMemory`, SDK buffer overrun) before reaching step 3 of the prompt contract, no completion comment is ever posted. The requester sees the ❌ reaction and that's it. Real example: ABCA-91 hit `error_max_turns` on turn 101 with PR opened — the PR was shipped, the agent never got to its own completion comment.

(1) is a comment-content gap. (2) is a "comment never gets posted at all" gap.

## Proposed shape

Add a Linear-side dispatcher to the existing DDB Streams fanout (`fanout-consumer.ts`), parallel to the GitHub edit-in-place dispatcher. On terminal task events (`task_completed`, `task_failed`, `error_max_turns`, `error_max_budget_usd`, etc.), post a structured final-status comment to the linked Linear issue with:

- **Status** — succeeded / failed + classifier title (e.g. "Hit max-turns cap")
- **PR** — URL if one was opened (works for the ABCA-91 case where the work shipped despite max_turns)
- **Cost** — `\$X.XX`
- **Turns** — `N / MAX`
- **Duration** — `Mm Ss`
- **Task ID** — for `bgagent status` / `bgagent watch` lookup

Lives platform-side (Lambda, fanout consumer), so it fires deterministically regardless of whether the agent reached its own step-3 completion comment. Consequence: there will be two completion comments on Linear when the agent runs cleanly (agent's freeform "✅ done" + platform's structured metrics). That's the same dual-comment pattern as GitHub today (PR comment + agent's own commit message), and the agent-side prompt can be simplified to drop step 3 once the platform side is reliable.

## Acceptance criteria

- New Linear dispatcher in the fanout consumer fires on terminal task events for `channel_source: linear` tasks.
- Comment uses Linear OAuth token resolved via `resolveLinearOauthToken` (same path the agent uses).
- All terminal subtypes are covered, not just success: `task_completed`, `task_failed`, `error_max_turns`, `error_max_budget_usd`, `error_during_execution`.
- If `pr_url` is set on the metrics, the comment treats the task as \"shipped a PR\" even on `error_max_turns` — flips the framing from \"❌ failed\" to \"⚠️ shipped a PR but ran out of turns; review and decide if more work is needed.\"
- Best-effort: Linear API failure logs but doesn't break the fanout (existing reportIssueFailure pattern).
- Tests: success case, error_max_turns + pr_url case, error_max_turns without pr_url, error_during_execution, missing OAuth token (graceful no-op).

## Out of scope

- Linear reaction-flip logic (the ❌ vs ⚠️ on the original issue) — separate concern, tracked separately if needed.
- Migrating the agent's three-comment prompt contract to the platform side. Step 3 (completion) becomes redundant once this lands; the agent prompt can drop it in a follow-up.
- Real-time progress comments (phase / milestone updates). The GitHub dispatcher does in-place editing; Linear's `save_comment` doesn't support edit so live updates would create N comments. Defer.

## Files

- `cdk/src/handlers/fanout-consumer.ts` (or wherever the GitHub dispatcher lives) — add Linear dispatcher branch
- `cdk/src/handlers/shared/linear-feedback.ts` — extract `postIssueComment` for reuse, add structured-metrics renderer
- `cdk/test/handlers/fanout-consumer.test.ts` — terminal-event coverage
- IAM grant: fanout consumer Lambda needs `secretsmanager:GetSecretValue` on `bgagent-linear-oauth-*` and read on `LinearWorkspaceRegistryTable` (likely already grant'd)

## Verification

Submit a Linear-triggered task that:
- Completes cleanly → Linear gets the structured "✅ … cost \$X.XX • turns N / 100 • duration Mm Ss" comment.
- Hits `max_turns` after opening a PR (like ABCA-91) → Linear gets "⚠️ shipped a PR but hit max-turns cap … cost \$X.XX • turns 100 / 100" with the PR URL.
- Fails before opening a PR → Linear gets "❌ failed: <error classifier title> … cost \$X.XX • turns N / 100".

Related: ABCA-91 in our dev workspace (turn 101 / 100, PR shipped, ❌ reaction posted).

isadeks:
Closes #239. Stacked on #241 (the screenshot pipeline) — review #241 first; this PR is the dispatcher that composes with it for the screenshot-from-Linear-issue feedback loop.

## Summary

- New Linear dispatcher in `fanout-task-events.ts`, mirroring the slack/github/email pattern. Single-entry registration via the existing `NotificationChannel + CHANNEL_DEFAULTS + DISPATCHERS` extension surface.
- Posts a deterministic final-status comment to the linked Linear issue on terminal task events. Three framing modes: ✅ task_completed, ⚠️ error_max_turns-with-PR, ❌ all other terminal states.
- Catches the case where the agent crashes (OOM, SDK buffer overflow, max_turns) before reaching its own step-3 completion comment — see ABCA-91 in #239's description.
- Reuses the existing `postIssueComment` helper; no new outbound Linear API surface.

## Architecture

- Channel registration: 1 line in `NotificationChannel`, 1 entry in `CHANNEL_DEFAULTS` (terminal events only), 1 entry in `DISPATCHERS` map.
- `dispatchToLinear` gates on `channel_source === 'linear'` so non-Linear tasks short-circuit after one `DDB Get` — same shape as the Slack dispatcher's gate.
- `renderLinearFinalStatusComment` is exported for testability and cleanly separates the (event_type, pr_url) → header logic from the body assembly.
- Construct gets two new guarded props (`linearWorkspaceRegistryTable`, `linearOauthSecretArnPattern`) matching the `slackSecretArnPattern` pattern. A deployment without LinearIntegration gets no dangling IAM grants.
- `FanOutConsumer` moved below `LinearIntegration` in `agent.ts` so it can receive the registry table reference (LinearIntegration depends on `runtime` and `orchestrator`, both already constructed earlier).

## Test plan

- [x] 6 new Linear-dispatcher tests: happy path, failed without PR, **ABCA-91 max_turns-with-PR**, channel_source short-circuit, missing channel_metadata, postIssueComment graceful no-op
- [x] Existing routing tests updated for the new dispatcher count (4 channels: slack/github/linear/email)
- [x] Full CDK suite: **1816 passing, 101 suites** (15 min on M1)
- [ ] CI: build + tests pass on PR
- [ ] Smoke test on dev stack (deferred until #241 lands so this doesn't fight against the in-flight stack)

## Out of scope (deferred per #239)

- Linear reaction-flip logic (the ❌ vs ⚠️ on the original issue)
- Migrating the agent's three-comment prompt contract to platform-side (step 3 in the prompt becomes redundant once this lands)
- Real-time progress comments — Linear's `save_comment` doesn't support edit, so this is post-once on terminal events only

### DIFF
--- a/agent/src/prompt_builder.py
+++ b/agent/src/prompt_builder.py
@@ -159,12 +159,15 @@ def _channel_prompt_addendum(config: TaskConfig) -> str:
         "transition the issue state to `In Review` (fall back to `In Progress` "
         "if that state doesn't exist). If neither exists, skip the state "
         "transition — the PR comment alone is enough. Do not invent state "
-        "names or loop on `list_issue_statuses`.\n"
-        "3. **On completion or failure** — call `mcp__linear-server__save_comment` "
-        "with the final status (succeeded / failed + short reason).\n\n"
-        "Keep comments concise. Do not mirror the full agent transcript back to "
-        "Linear. Even small tasks must post all three updates — users rely on "
-        "them to track progress."
+        "names or loop on `list_issue_statuses`.\n\n"
+        "**Do NOT post a final 'task completed' or 'task failed' comment.** "
+        "The platform fan-out plane (issue #239) posts a structured "
+        "✅/⚠️/❌ summary on terminal events with cost / turns / duration / "
+        "PR-link metrics that you don't have visibility into. A redundant "
+        "agent-side completion comment would just stack two near-identical "
+        "comments on the issue.\n\n"
+        "Keep the start + PR-opened comments concise. Do not mirror the full "
+        "agent transcript back to Linear."
     )
 
 
--- a/cdk/src/constructs/fanout-consumer.ts
+++ b/cdk/src/constructs/fanout-consumer.ts
@@ -77,6 +77,23 @@ export interface FanOutConsumerProps {
   readonly slackSecretArnPattern?: string;
 
   /**
+   * LinearWorkspaceRegistryTable — the Linear dispatcher reads this to
+   * resolve per-workspace OAuth tokens at comment-post time. Optional:
+   * when omitted, the dispatcher logs and skips so a deployment without
+   * Linear onboarding doesn't accumulate dangling IAM grants.
+   */
+  readonly linearWorkspaceRegistryTable?: dynamodb.ITable;
+
+  /**
+   * Secrets Manager ARN-prefix pattern for per-workspace Linear OAuth
+   * bundles. Mirrors ``slackSecretArnPattern`` shape — typically
+   * ``bgagent-linear-oauth-*``. Required when ``linearWorkspaceRegistryTable``
+   * is set; without it the dispatcher would resolve the registry row but
+   * fail at the SM GetSecretValue call.
+   */
+  readonly linearOauthSecretArnPattern?: string;
+
+  /**
    * Maximum batch size delivered to the Lambda per invocation.
    *
    * @default 100 (DynamoDB Stream default)
@@ -170,6 +187,30 @@ export class FanOutConsumer extends Construct {
       this.fn.addToRolePolicy(new iam.PolicyStatement({
         actions: ['secretsmanager:GetSecretValue'],
         resources: [props.slackSecretArnPattern],
+      }));
+    }
+
+    // Linear dispatcher plumbing. Same guarded shape as Slack/GitHub:
+    // a deployment without Linear onboarding gets no IAM grants and
+    // the dispatcher logs-and-skips on missing env. The registry table
+    // tells us per-workspace OAuth-secret ARN; the secret holds the
+    // access token that ``postIssueComment`` uses to drive
+    // ``commentCreate`` GraphQL.
+    if (props.linearWorkspaceRegistryTable) {
+      props.linearWorkspaceRegistryTable.grantReadData(this.fn);
+      this.fn.addEnvironment(
+        'LINEAR_WORKSPACE_REGISTRY_TABLE_NAME',
+        props.linearWorkspaceRegistryTable.tableName,
+      );
+    }
+    if (props.linearOauthSecretArnPattern) {
+      this.fn.addToRolePolicy(new iam.PolicyStatement({
+        // GetSecretValue + PutSecretValue: the resolver may rotate the
+        // OAuth token (writes the refreshed bundle back to SM) — same
+        // grants the LinearIntegration's webhook-processor Lambda holds
+        // for the same reason.
+        actions: ['secretsmanager:GetSecretValue', 'secretsmanager:PutSecretValue'],
+        resources: [props.linearOauthSecretArnPattern],
       }));
     }
 
--- a/cdk/src/handlers/fanout-task-events.ts
+++ b/cdk/src/handlers/fanout-task-events.ts
@@ -776,10 +776,16 @@ async function dispatchToEmail(event: FanOutEvent): Promise<void> {
  *
  * The framing flips between three outcomes based on `(eventType, prUrl)`:
  *
- *   1. ``task_completed``        → ✅ "Task completed"
- *   2. ``error_max_turns``  + PR → ⚠️ "Shipped a PR but hit max-turns cap" — the
- *                                  motivating GH #239 case (ABCA-91 turn 101)
- *   3. all other terminal subtypes → ❌ "Task failed" + classifier title if known
+ *   1. ``task_completed``                        → ✅ "Task completed"
+ *   2. any non-completed terminal event WITH PR  → ⚠️ "Shipped a PR but stopped early"
+ *      (the motivating ABCA-91 case is max-turns-with-PR, but the same
+ *      framing applies to any terminal failure — budget cap, agent
+ *      crash, etc. — that managed to ship a PR before stopping)
+ *   3. any non-completed terminal event NO PR    → ❌ "Task <subtype>" + classifier title
+ *
+ * The ⚠️ frame appends the classifier title when one is available so the
+ * requester sees both outcomes (the PR shipped, AND the reason it
+ * stopped — "Hit max-turns cap" for ABCA-91).
  *
  * Cost / turns / duration appear as a subtitle line. Missing values
  * (e.g. failure before the agent emitted any tokens) render as `—`.
@@ -801,7 +807,11 @@ export function renderLinearFinalStatusComment(args: {
   if (isCompleted) {
     header = '✅ **Task completed**';
   } else if (shippedDespiteFailure) {
-    header = '⚠️ **Shipped a PR but stopped early** — review and decide if more work is needed';
+    // Append the classifier title (when known) so the requester sees
+    // *why* the agent stopped, not just that it shipped a PR. For
+    // ABCA-91 this renders "...stopped early — Hit max-turns cap".
+    const reason = args.errorTitle ? ` — ${args.errorTitle}` : '';
+    header = `⚠️ **Shipped a PR but stopped early${reason}** — review and decide if more work is needed`;
   } else {
     const reason = args.errorTitle ? `: ${args.errorTitle}` : '';
     header = `❌ **Task ${args.eventType.replace(/^task_/, '')}${reason}**`;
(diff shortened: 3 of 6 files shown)


krokoko:
Looking 👁️

krokoko:
## Review — Linear dispatcher in fanout consumer (#239)

Thanks for this — it's a clean, well-tested addition and it closes a real UX gap. The ABCA-91 case (agent crashes at `max_turns` after shipping a PR, leaving the Linear requester with only a ❌ reaction and no metrics) is exactly the kind of "outcomes must stay inspectable" gap worth closing, and doing it platform-side so it fires even when the agent never reaches its own step-3 comment is the right call.

A few things I especially liked:
- The dispatcher reuses the existing `NotificationChannel + CHANNEL_DEFAULTS + DISPATCHERS` surface and mirrors Slack's `channel_source` gate and best-effort failure contract — same idiom, real per-channel substance, no copy-paste.
- Numerics are coerced at the DDB boundary with `coerceNumericOrNull`, matching the GitHub dispatcher.
- IAM grants are guarded behind optional props, so a deployment without Linear onboarding gets no dangling permissions.
- The `mockDdbSend` refactor from index-based assertions to command-type filtering is the correct defensive fix now that GitHub + Linear both hit the shared Get mock.
- Docs are in good shape — I ran `node scripts/sync-starlight.mjs` on the branch and the Starlight mirror is in sync, so CI's mutation guard should pass.

**Verdict: Approve with nits.** One item I'd like scoped down before merge, plus some polish.

### Please address before merge

**1. `bedrock-agentcore:*` action wildcard — least privilege** (`cdk/src/constructs/github-screenshot-integration.ts:195`)
```ts
actions: ['bedrock-agentcore:*'],
resources: ['*'],
```
The handler only calls `StartBrowserSession`, `StopBrowserSession`, and the SigV4-presigned `ConnectBrowserAutomationStream` WSS endpoint. `resources: ['*']` is defensible (ephemeral sessions have no stable ARN, and the IAM5 suppression documents it), but the **action** wildcard grants every AgentCore action (memory, runtime, gateway, identity, code-interpreter) we don't use. I see the commit history notes this was widened to unblock e2e with a "scope back down" follow-up — this is a good moment to do it:
```ts
actions: [
  'bedrock-agentcore:StartBrowserSession',
  'bedrock-agentcore:StopBrowserSession',
  'bedrock-agentcore:ConnectBrowserAutomationStream',
],
```
If `ConnectBrowserAutomationStream` can't be confirmed as the exact IAM action name (it wasn't in the public CLI list), CloudTrail's decoded deny message will have it. If it genuinely can't be resolved, keeping the wildcard is acceptable as long as the blocker is documented inline with a tracking issue rather than a bare "followup."

### Non-blocking nits

2. **Doc-vs-code mismatch on the ⚠️ frame** (`renderLinearFinalStatusComment` JSDoc, `fanout-task-events.ts`). The comment keys case 2 on `error_max_turns` as an `eventType`, but `error_max_turns` is an `agent_status`, never an `event_type` — the real trigger is `!isCompleted && prUrl != null` (any terminal failure that still produced a PR). The test correctly passes `event_type: 'task_failed'`. Worth aligning the comment to the real condition.

3. **`classifyError` returns `UNKNOWN_CLASSIFICATION`, not `null`, for unmatched non-empty messages** (`error-classifier.ts:359`). The dispatcher's inline comment says it "returns null … the renderer falls back to a generic frame," but a failed task with any non-empty unmatched `error_message` will actually render `❌ Task failed: Unexpected error`. Not a bug — just pick the intended behavior and align the comment with it.

4. **The classifier title is dropped on the ⚠️ path.** For the ABCA-91 case (max-turns + PR), `classification.title` ("Exceeded max turns") is computed but not rendered — the requester sees the PR but not *why* it stopped, which is the most useful context for the motivating issue. Consider appending it to the ⚠️ frame.

5. **Log severity** (`dispatchToLinear`): the `missing_env` and `post_failed` paths log at INFO, whereas the GitHub dispatcher's equivalent missing-env path is WARN. Since this comment is the *only* completion signal for the crash case, a misconfigured env var or revoked OAuth token failing every post is worth a WARN + `error_id` so it's alarmable. (The underlying `linear-feedback.ts` failures already WARN, so this is a backstop.)

6. **Stale comment** in `routeEvent` (`fanout-task-events.ts`): "at most 3 channels" — now 4 (slack/github/linear/email).

7. **Output description leftover** (`agent.ts:883`): one `ScreenshotBucketName` description still reads "Vercel-preview screenshots" after the de-Vercel-ize sweep. Cosmetic.

### Tests

Coverage is solid and the suite passes locally (86 tests in `fanout-task-events.test.ts`). A few gaps the issue's acceptance criteria named that would be cheap to close:
- **Missing `LINEAR_WORKSPACE_REGISTRY_TABLE_NAME` env path** — the env is set at module load and never unset, so the deploy-misconfig safety valve is unexercised.
- **`renderLinearFinalStatusComment` null-metric rendering** — every fixture supplies non-null metrics, so the `—` fallbacks and `formatDuration` boundaries (`<60s`, exact-minute `Nm`) aren't covered. This is exactly the crash-before-metrics case the feature is built for; a small table-driven test calling the exported renderer directly would do it.
- **`error_max_turns` without `pr_url`** — the ⚠️↔❌ boundary (the `prUrl != null` flip) isn't asserted in that direction.

(The missing-OAuth-token graceful no-op, AC item 5, *is* covered at the helper layer in `linear-feedback.test.ts` — no need to duplicate it here.)

One small heads-up: since the PR is stacked on #241, the diff against `main` carries the full screenshot pipeline too — happy to review just the #239 delta if you retarget the base, otherwise no worries.

Nice work overall — scope the IAM action and this is good to go.

---
*Reviewed with the help of automated agents (code-reviewer, silent-failure-hunter, pr-test-analyzer). 🤖*

isadeks:
@krokoko thanks for the review. All items addressed across two review-response commits.

## Shipped on this branch

**Blocking item (1) — `bedrock-agentcore:*` action wildcard** (`d8d9479`)
Narrowed `github-screenshot-integration.ts:195` to the three actions the handler actually calls: `StartBrowserSession`, `StopBrowserSession`, `ConnectBrowserAutomationStream`. Resource wildcard stays (sessions are ephemeral, no per-resource ARN); the cdk-nag IAM5 suppression already documents that.

**Non-blocking nits** (`3280d2c`, `b84ce1e`, `d8d9479`)

| # | Item | Where |
|---|---|---|
| 2 | JSDoc on `renderLinearFinalStatusComment` keyed on the real `(eventType, prUrl)` discriminator instead of `error_max_turns` | `3280d2c` |
| 3 | `classifyError` inline comment now correctly says `UNKNOWN_CLASSIFICATION` (not `null`) for non-empty unmatched messages | `3280d2c` |
| 4 | ⚠️ frame appends classifier title — for ABCA-91 the requester sees `Shipped a PR but stopped early — Exceeded max turns` | `3280d2c` |
| 5 | `missing_env` and `post_failed` paths bumped from INFO to WARN with `error_id` tags so they're alarmable | `3280d2c` |
| 6 | "at most 3 channels" → "4 channels" in `routeEvent` | `3280d2c` |
| 7 | `ScreenshotBucketName` description: "Vercel-preview screenshots" → "deploy-preview screenshots" | `d8d9479` |

**Test gaps** (`3280d2c`)

- New test: `LINEAR_WORKSPACE_REGISTRY_TABLE_NAME` unset → WARN + skip (deploy-misconfig safety valve)
- New test: `error_max_turns` without `pr_url` renders ❌ not ⚠️ (the boundary in the other direction)
- New `describe` block: 8 direct tests of `renderLinearFinalStatusComment` covering null-metric fallbacks, `formatDuration` boundaries (`<60s`, exact-minute `Nm`, mixed `Nm Ss`), classifier-title rendering on ⚠️ and ❌ frames, and the no-trailing-colon when `errorTitle` is null

102 tests in `fanout-task-events.test.ts` (was 92), full suite still green.

## Bonus from first dev smoke-test (`b84ce1e`)

After deploying to dev I noticed two near-duplicate signals on the Linear thread:
- ✅ comment carried `PR: <url>` while the agent's step-2 had already posted the same link → render PR URL only on ⚠️ now
- Agent step-3 "task completed" stacked next to the platform's structured ✅ → removed step 3 from the Linear-channel prompt contract; agent now posts start (1) + PR-opened (2), platform posts terminal ✅/⚠️/❌ (3)

You'd predicted this exact migration in your review ("the agent prompt can drop step 3 once the platform side is reliable").

## Heads-up

PR is stacked on PR #241 — the diff against `main` carries the screenshot pipeline too. Happy to retarget the base for review purposes if it'd help; otherwise the #239 delta is the three commits `3ba880d`, `3280d2c`, `b84ce1e`, `d8d9479`.

Re-requesting review.

krokoko (approved):

PR #295, opened 2026-06-08
feat(cdk): Phase-0 deploy-then-verify integ tests (#236)

### LINKED ISSUE #236: feat(cdk): CDK integ-tests for deployed runtime E2E verification
### Component

CDK / infrastructure, API or orchestration, Tooling / CI

### Describe the feature

Introduce **automated, deploy-then-verify integration tests** for the ABCA platform using the experimental [`@aws-cdk/integ-tests-alpha`](https://docs.aws.amazon.com/cdk/api/v2/docs/integ-tests-alpha-readme.html) library and the **integ-runner** workflow.

Today we have strong unit and handler-level coverage, but full-stack **runtime behavior** is validated manually (for example Cedar HITL scenarios A–E against a live `backgroundagent-dev` stack, as described in `docs/design/CEDAR_HITL_GATES.md`). ADR-008 Level 3 calls for integration or E2E coverage on critical paths; this issue tracks making that repeatable in CI/CD: **synthesize → deploy → assert → destroy**, with no long-lived drift in shared accounts.

Each integ test should be a **1:1 CDK application** (per AWS guidance): an `IntegTest` construct registering one or more stack test cases, plus optional `DeployAssert` assertions (`awsApiCall`, `httpApiCall`, `invokeFunction`, `waitForAssertions`) to poll until terminal conditions are met.

### Use case

- **Confidence before merge:** Catch regressions in orchestration, admission, hydration, agent session lifecycle, and channel hand-offs that mocks cannot surface.
- **Safer iteration on roadmap priorities:** Identity propagation, cost attribution, registry resolution, and Linear workflows all need a trustworthy “does the deployed stack actually work?” gate.
- **Replace ad-hoc smoke tests:** Operators should not be the only path to validate a post-deploy stack; failures should surface in CI with clear assertion output.
- **Document the contract:** Assertions encode expected task states, event shapes, and API responses—the living spec for platform behavior.

### Proposed solution

**Phase 0 — Foundation**
- Add `@aws-cdk/integ-tests-alpha` to `cdk/` (experimental; pin version; document upgrade risk in CONTRIBUTING or a short design note).
- Add `cdk/test/integ/` (or `cdk/integ/`) with at least one **smoke integ** that deploys the main stack (or a trimmed integ variant if full stack is too heavy) and asserts a minimal happy path, e.g.:
  - `POST /v1/tasks` (or internal test hook) creates a task
  - Poll DynamoDB / `GET /v1/tasks/{id}` until a terminal state
  - Verify required fields on the task record (`task_id`, `user_id`, status, timestamps)
- Wire **mise** / CI entry point (e.g. `mise //cdk:integ`) invoking `integ-runner` with documented prerequisites (AWS credentials, target account/region, optional dedicated integ stack name).

**Phase 1 — Core lifecycle**
- Scenarios aligned with manual Cedar HITL E2E matrix where feasible (submit → run → complete / fail / await approval → approve or deny → terminal).
- Use `waitForAssertions()` for long-running agent/orchestrator paths with conservative timeouts.
- Ensure **teardown** (`destroy` with force) runs on success and failure; document cost and account isolation expectations.

**Phase 2 — Channels and guardrails** (follow-up issues OK)
- Webhook create-task, Slack/Linear adapters (may require test fixtures or mocked externals).
- Stranded-task reconciler behavior under controlled clock/fixtures if possible.

**Design constraints**
- Integ tests must **not** run on every PR by default if cost/latency is prohibitive—options: nightly workflow, `workflow_dispatch`, or label-gated job (document the chosen policy in the issue implementation PR).
- Prefer a **dedicated integ stack name** or ephemeral account to avoid colliding with developer `backgroundagent-dev` stacks.
- Keep integ apps **separate** from production synth (`cdk.out` isolation); assertion stacks use `DeployAssert` per library docs.

**Out of scope (for this issue)**
- Replacing unit tests or moto-backed handler tests (see #61 for trace round-trip).
- Full agent “open PR on real GitHub repo” unless we add a gated fixture repo and PAT—can be a child issue.

### Other information

- **Roadmap gap:** “Deployed runtime E2E verification” is a prioritized platform item but not yet listed explicitly in `docs/guides/ROADMAP.md`—add a row when implementation starts (source + Starlight sync).
- **Related:** ADR-008 Level 3; `docs/design/CEDAR_HITL_GATES.md` §15.3 (manual E2E on `backgroundagent-dev`); ADR-013 Tier 3 (remote CI as authoritative, integ as pre-merge signal).
- **Reference:** [@aws-cdk/integ-tests-alpha README](https://docs.aws.amazon.com/cdk/api/v2/docs/integ-tests-alpha-readme.html)

### Acknowledgements

ayushtr-aws:
## Summary

Closes #236 (Phase 0). Adds the first automated **deploy-then-verify** coverage for ABCA — a real `synth → deploy → assert → destroy` against AWS, fulfilling ADR-013 Tier 3 / Phase 0. Everything to date validated locally without touching AWS; this exercises the deployed runtime.

- **`cdk/test/integ/integ.task-api-smoke.ts`** — a trimmed `TaskApiSmokeStack` (TaskApi + task/events tables) and an `IntegTest` that asserts the create-and-persist happy path: Cognito `adminCreateUser → adminSetUserPassword → initiateAuth` → `POST /tasks` → `GET /tasks/{id}` (200, `SUBMITTED`) → DynamoDB `getItem` (`user_id` persisted). Teardown forced on success **and** failure.
- **`.github/workflows/integ.yml`** — PR-driven via `workflow_run` (mirrors `deploy.yml`): runs only when the diff touches `cdk/**` or `agent/**`, behind the `integ` environment's admin-approval gate, and posts a required `integ-smoke` commit status that blocks merge. `workflow_dispatch` retained for manual runs (main only). OIDC creds, concurrency guard, pinned action SHAs, and an `if: always()` force-destroy safety net.
- **`cdk/mise.toml`** — `mise //cdk:integ`.
- **`cdk/package.json`** — pin `@aws-cdk/integ-runner@2.199.0` + `@aws-cdk/integ-tests-alpha@2.257.0-alpha.0` (devDeps); exclude `test/integ/` from jest test + coverage paths.
- **`.gitignore`** — ignore regenerated `*.snapshot/` (~100 MB of bundled Lambda assets) and `cdk-integ.out.*/`.
- **docs** — ROADMAP + ADR-013 note Phase 0 landed (Starlight mirrors regenerated).

## Trigger model (reviewer feedback)

The reviewer asked for the same experience the repo already has for `deploy.yml`: a check visible on the PR that **blocks merge**, with an **admin approving** the run that executes in the shared account — no manual dispatch per PR. Implemented by mirroring the proven `deploy.yml` pattern:

1. **`build` completes → `workflow_run` picks it up** in the trusted base-repo context (secrets/OIDC available even for fork PRs — a fork `pull_request` job gets none).
2. **`resolve` job** decides applicability by path-filtering the PR's changed files (`cdk/**` or `agent/**`). Docs/cli-only PRs get an **immediate green `integ-smoke` skip** so the required check never deadlocks; applicable PRs get a `pending` status.
3. **`integ` job** runs behind **`environment: integ`** — the required-reviewer gate is the admin approval. While it waits, `integ-smoke` stays `pending` and merge stays blocked.
4. **`report` job** posts the final `integ-smoke` status (success/failure) back to the PR head.

To make the check actually block merge, an admin marks **`integ-smoke` a required status check** in branch protection (account-side; see test plan).

⚠️ **Fork PRs run behind the same approval gate** (no hard fork-guard, unlike `deploy.yml`). The `integ` environment approver is therefore **authorizing fork-authored test code to run with the privileged AWS role** — the approver MUST review `cdk/test/integ/**` changes in fork PRs before approving. This was reconciled with the reviewer; it can be tightened to exclude forks later if desired.

## Design notes (non-obvious decisions)

- **Trimmed stack, not full AgentStack.** Omits the AgentCore Runtime (`jean_cloude`) and Memory (`bgagent_memory`) — their account-unique immutable names would collide with the live `backgroundagent-dev` in the shared account. With no orchestrator wired in, the task correctly persists at `SUBMITTED`, which is the deterministic thing to assert in Phase 0.
- **No `.expect()` on the POST `/tasks` call.** Reading `body.data.task_id` from it (for the GET + DynamoDB steps) flips that call's `flattenResponse` to true, which collapses the response into dotted leaf keys and breaks a nested `objectLike` expectation ("Expected type object but received undefined"). The create is verified **transitively** instead — the GET returns `200`/`SUBMITTED` for the returned id and DynamoDB confirms the persisted row, a stronger check than a bare `201`.
- **Cognito service name is the SDK-v2 form `CognitoIdentityServiceProvider`.** The assertion provider's v2→v3 map only resolves the v2 key; the v3-style name falls through to a nonexistent package.
- **Path-filter lives in the `resolve` job, not `on.pull_request.paths`.** The trigger is `workflow_run`, so GitHub's native path filter doesn't apply — applicability is resolved via the Files API at runtime.

## Test plan

- [x] `mise //cdk:compile` + `mise //cdk:eslint` clean (license header / lint).
- [x] Full `mise //cdk:build` green — snapshot passing, coverage thresholds held (integ dir excluded).
- [x] `npx integ-runner --dry-run` → `Tests: 1 passed` (test synthesizes/validates).
- [x] `zizmor` clean on `integ.yml` (the `workflow_run` dangerous-triggers finding is annotated with the mitigation rationale, matching `deploy.yml`).
- [x] **Real deploy-then-verify on one account** (using the OIDC role): `SUCCESS`, `Tests: 1 passed`. Stack torn down; no stranded stacks.
- [x] Also ran on a second account to confirm the stack is environment-agnostic and collision-free.
- [x] Confirmed the live `backgroundagent-dev` stack is untouched.
- [ ] **Account-side setup required before CI can run** (does not block merging this PR; blocks only the live CI check):
  1. **`integ` GitHub Environment with a required reviewer** — the admin-approval gate (same shape as the `deploy` environment).
  2. **Dedicated least-privilege role** (e.g. `GitHubIntegE2E`) scoped to the integ stack, whose **OIDC trust `sub`** includes `repo:aws-samples/sample-autonomous-cloud-coding-agents:environment:integ`, exposed to the `integ` environment as `AWS_ROLE_TO_ASSUME` + `AWS_REGION`.
  3. **Branch protection:** mark **`integ-smoke`** a **required status check** on `main` — this is what makes the check block the merge button.
- [ ] Once set up: open a PR touching `cdk/**` → build passes → `integ-smoke` pending → admin approves `integ` → deploy/assert/destroy → `integ-smoke` flips green → merge unblocks. A docs-only PR → `integ-smoke` immediate green skip.

## Deferred to follow-ups

Tracked separately so Phase 0 stays focused; none block this PR functionally:

- **CDK CLI version alignment** — `yarn.lock` resolves both `aws-cdk@2.1115.1` and `2.1126.0` (the latter pulled in by `integ-runner`). Worked on the test accounts, but the synth-vs-CLI skew is worth pinning `aws-cdk` to align with `integ-runner`'s expected CLI. Needs its own lockfile investigation.
- **Local-run stack-name collision note** — `backgroundagent-integ` is a fixed stack name (good for the GH-Actions concurrency lock), but a developer running `mise //cdk:integ` locally while a CI run is in flight would collide. Worth a README note (or `process.env.USER` interpolation for local runs).
- **Per-PR stack-name isolation** — `backgroundagent-integ-pr<N>` would allow parallel CI runs instead of serializing via the `cdk-integ` concurrency group. Deferred; current single-name + concurrency is correct for now.

## Coming in later PRs of this stack

- **Phase 1 — Core lifecycle:** submit → run → complete / fail / await-approval → approve/deny → terminal, aligned with the manual Cedar HITL E2E matrix; `waitForAssertions()` for long-running agent/orchestrator paths. Requires a gated fixture repo + PAT + the orchestrator/AgentCore stack.
- **Phase 2 — Channels & guardrails:** webhook create-task, Slack/Linear adapters (test fixtures or mocked externals), and stranded-task reconciler behavior under a controlled clock.


isadeks:
Nice work on Phase 0 — trimmed-stack design and the transitive-verify-via-GET+DDB choice are well-justified. A few items before merge:

  ### Must-fix

  1. **Restrict `workflow_dispatch` to `main`.** Right now any branch can be dispatched, so a feature-branch edit to `test/integ/*.ts` runs against the integ role. Either `if: github.ref == 'refs/heads/main'` on the job, or `environment: integ` with required reviewers configured in repo Settings → Environments.

  2. **`if: always()` cleanup step is a no-op.** `.github/workflows/integ.yml:67` runs `npx cdk destroy backgroundagent-integ --force || true` against `cdk.json`'s app (`src/main.ts`), which doesn't define that stack — exits "no stacks match" and `|| true` swallows it. If integ-runner crashes mid-deploy you keep public infra live. Either point `-a` at the integ entrypoint or fall back to `aws cloudformation delete-stack --stack-name backgroundagent-integ`.

  ### Nice-to-have

  3. `cdk/mise.toml` — `mkdir -p $TMPDIR` errors when `TMPDIR` is unset (CI default). Use `mkdir -p "${TMPDIR:-/tmp}"`.
  4. `--update-on-failed` is dead code given `.snapshot/` is `.gitignore`d (nothing to compare against). Drop the flag, or commit a snapshot and drop `--force`.
  5. CDK CLI version drift in `yarn.lock` — both `aws-cdk@2.1115.1` and `2.1126.0` resolved (the latter pulled in by integ-runner). Worked on the test accounts but synth-vs-CLI skew bites only sometimes; consider pinning`aws-cdk` to align with `integ-runner`'s expected CLI.
  6. Stack name `backgroundagent-integ` is fixed — fine for the GH-Actions concurrency lock, but a developer running `mise //cdk:integ` locally during the nightly will collide. Worth a README note or `process.env.USER` interpolation for local runs.
  7. `aws-region: ${{ vars.AWS_REGION }}` with no fallback — if the variable is unset the action runs region-less. Pin or default it.

  The detailed Cognito-service-name comment at lines 109-116 of the integ file is a great future-self gift — leave that in.

isadeks:
For AWS deployment, right now your PR is structured  "Not per-PR. The run deploys billable resources into a shared account, so it's nightly + manual only — out of the merge gate." It would be better to have it after every PR to catch early breaking changes rather than a routine job. Also deploy per-PR when the diff touches `cdk/**` or `agent/**` (skip docs or cli PRs). Worth a quick research pass on this. Later on in the future, would be nice to deploy only the necessary resources (I am not sure if this is in scope right now).

ayushtr-aws:
@isadeks Addressed you and Alain's comments. Please re-review the PR :)

krokoko:
Thanks @ayushtr-aws — this is a genuinely well-built PR and a meaningful milestone: the first deployed-runtime E2E coverage for ABCA (ADR-013 Tier 3 / Phase 0). The trigger model is carefully reasoned, the teardown is belt-and-suspenders, and the in-code comments explaining the alpha-library sharp edges (`flattenResponse`, the v2 Cognito service-name map, the env-agnostic stack rationale) are exactly the kind of "why, not what" that saves a future maintainer hours. Reviewed as a principal-architect pass with the repo's review agents (code-review, silent-failure, test-coverage, security).

## Verdict: Approve with nits — conditional on the documented go-live prerequisites

It merges safely **today** because the privileged path is inert until account-side setup exists (I confirmed the `integ` environment doesn't exist yet, `main` isn't branch-protected, and the role isn't wired — matching the 3 unchecked items in your test plan). The items below should be resolved **before the `integ` gate is activated**, not necessarily before merge.

### Most important — fork code + privileged role (go-live blocker, not merge blocker)

The `integ` job checks out fork PR-head code (`integ.yml` `repository: head_repo` / `ref: head_sha`) and runs `mise //cdk:integ` with `secrets.AWS_ROLE_TO_ASSUME`. Versus `deploy.yml`, which has a hard fork guard (`head_repository.full_name == github.repository`, line 14) **and** runs from a pre-built artifact, this deliberately drops both protections — so a fork author's test code (or a transitive import / `yarn install` build hook) can execute against a shared-account role behind only the environment approval prompt. You've documented and reconciled this trade-off already; my ask is just to make the deferred least-privilege **`GitHubIntegE2E` role a hard prerequisite of go-live** (isolated/scoped, trust pinned to `…:environment:integ`, not the shared deploy role), ideally with a maintainer-applied label gate on top of approval for fork PRs, and to record the residual-risk acceptance in ADR-013.

### Error-handling hardening (fold in before activating the gate)

- **`resolve` crash → possible merge deadlock.** `gh api …/files` (and `resolve_pr_number`) run under `set -euo pipefail` with no failure path; a transient API error aborts the job *before* any status is posted, so `integ`/`report` skip and the required `integ-smoke` check can sit pending with no surfaced failure. Suggest an `EXIT` trap that posts an `error` status, and having `report` key off `needs.resolve.result`.
- **Teardown can silently leak a billable stack.** The `always()` safety net runs region-less if the job is cancelled before `Configure AWS credentials` (the exact crash case it exists for), and `wait … || true` swallows a terminal `DELETE_FAILED`. Suggest pinning `AWS_REGION` on that step and failing loud if the stack still exists after `wait`.
- **Path-filter can false-green.** A truncated/failed `/files` response is indistinguishable from "no `cdk/**` changes," so an infra PR could get a green skip — treat only a *successful empty* result as skip.

### Test quality (transitive verification is sound — verified no false-pass hole)

The dropped inline `.expect()` on POST is well-compensated: a failed POST has no `data.task_id`, so the GET (404→fails) and the DynamoDB `getItem` both fail — genuinely stronger than a bare `201`. Two small adds:
- **One negative-auth assertion** (unauthenticated POST → 401/403) would be the highest value-per-line for Phase 0 — the Cognito authorizer wiring is deployed-only and currently never asserted.
- `getItem`'s `stringLikeRegexp('.+')` accepts any non-empty `user_id`; the GET-200 already pins `user_id === caller` transitively, so just tightening that comment (optionally the regex) would reflect what's actually proven.

### Nits

- `@aws-cdk/integ-runner@2.199.0` is ~58 minors behind `integ-tests-alpha@2.257`/`aws-cdk-lib`; you've flagged the alignment as a follow-up — reasonable, just noting a runner/library skew is a known source of assertion-provider mismatches.
- `resolve` job grants `actions: read` but never reads the Actions API — can drop it (least privilege).
- The new `aws-account-id` gitleaks rule is a nice, well-scoped hardening; `keywords` gates it though, so a *context-free* bare 12-digit ID won't match — the comment slightly overstates coverage. Adding `secretGroup = 1` would report just the ID.

### Well done (no action)

- All four `uses:` are SHA-pinned ✅; untrusted `workflow_run.*` values are correctly routed through env vars (no script-injection) ✅; `permissions: {}` deny-all default with per-job grants ✅; per-synth random Cognito password (no credential literal) ✅; docs + both Starlight mirrors in sync ✅; governance clean (issue #236 `approved`, branch name conventional) ✅.

Net: I'd merge once `build` is green, and treat the fork-role design + the two error-handling items as the checklist for the same change that wires up the `integ` environment. Nice work pushing the validation pyramid into real deploy-then-verify territory. 🚀

krokoko (approved):

Issue #246, opened 2026-06-03
feat(registry): central agent asset registry for capabilities, skills, plugins, and MCP servers


krokoko:
### Component

CDK / infrastructure, API or orchestration, Agent (Python runtime), Scripts / CLI

### Describe the feature

Introduce a **central agent asset registry** — a versioned, platform-managed catalog from which agents and blueprints resolve runtime artifacts instead of vendoring them in source or redeploying CDK for every tool/skill change.

The registry holds **typed records** for multiple artifact kinds, including (non-exhaustive):

| Asset kind | Examples |
|------------|----------|
| **Capabilities** | Verification strategies, knowledge-tool configs, task workflow definitions |
| **Skills** | Agent skill packages (prompt + tool guidance) |
| **Plugins** | Claude Code / agent plugin bundles |
| **MCP servers** | Server definitions, tool descriptors, connection config |
| **Prompt fragments** | Reusable system-prompt sections |
| **Cedar policy modules** | Soft-deny packs, repo-specific policy extensions |
| **Event rule packs** | Lifecycle governance rules (pairs with #230) |
| **Notification profiles** | Channel rendering templates |

Each record is **immutable at a given version**; blueprints **pin** semver constraints (no floating `latest` in production). At task start the orchestrator resolves pins, fetches artifacts, and provisions the agent environment — decoupling asset authoring from infrastructure deployment.

This enables **more flexible, capability-driven tasks** beyond the three hardcoded task types (`new_task`, `pr_iteration`, `pr_review`) without forking core orchestrator logic for every new workflow.

### Use case

- **Generic / composable tasks:** Submit work defined by a capability bundle (tools + prompts + validation + policy) rather than a fixed enum task type.
- **Org-wide rollout:** Publish a new MCP server or skill version; blueprints re-pin when ready — running tasks keep their original pin.
- **Reproducibility:** Every execution records which asset versions were resolved (audit + eval).
- **Extensibility without forking:** Teams add skills/plugins/MCP defs through the registry instead of patching `agent/` or CDK per tenant.
- **Governance integration:** Registry becomes the source of truth for tool-surface validation (#230 Change Manifest L1/L5 concepts).

### Proposed solution

**Phase 0 — Schema and API contract**
- Define asset record schema: `{ id, kind, version, semver, descriptor, artifact_ref, publisher, created_at, status }`.
[...]

PR #296, opened 2026-06-09
feat(tasks): workflow-driven tasks — declarative steps + repo-less workflows (#248)

### LINKED ISSUE #248: feat(tasks): capability-driven tasks — declarative steps, non-coding workflows, migrate existing task types
### Component

Agent (Python runtime), API or orchestration, Scripts / CLI

### Describe the feature

Introduce **capability-driven tasks**: instead of hardcoding workflows in Python (`pipeline.py` branches on `task_type`) and a fixed enum (`new_task` | `pr_iteration` | `pr_review`), the platform loads a **capability file** that declares the ordered **steps** to run inside the agent container — prompts, tool profiles, pre/post hooks, hydration requirements, and terminal outcomes.

**Existing coding workflows become capability files too** — `new_task`, `pr_iteration`, and `pr_review` are the first shipped capabilities, not special cases in core code. New domains (research, document drafting, data analysis, email triage, etc.) are new capability files, not new orchestrator branches.

Inspired by the AKW vision in #99 (blueprint-driven, task-mode-agnostic agent loop) but scoped to **current ABCA architecture**: reuse durable orchestration, AgentCore isolation, Cedar HITL, Guardrails, and the agent asset registry (#246) as the home for published capability artifacts.

### Use case

- **Non-coding tasks:** Submit work with no GitHub repo — e.g. "summarise these papers", "draft a design doc from attachments" — without forcing clone/PR scaffolding.
- **Composable workflows:** Operators and blueprint authors define steps declaratively (hydrate → agent run → verify → deliver) instead of patching `pipeline.py` for every new workflow.
- **One platform, many domains:** Coding (PR open/review/iterate) and knowledge work share admission, memory, policy, cost limits, and observability — only the capability file differs.
- **Gradual migration:** Shipped task types keep working via capability files that encode today's behavior; callers can migrate from `task_type` enum to `capability_ref` over time.

### Proposed solution

**Capability file (conceptual schema)**

A versioned document (YAML or JSON — format TBD in design PR) describing at minimum:

| Field | Purpose |
|-------|---------|
| `id`, `version` | Stable identity; pinned via registry (#246) or repo-local path |
| `domain` | e.g. `coding` \| `knowledge` \| `hybrid` — drives admission defaults |
| `requires_repo` | Whether GitHub clone / PR finalization is mandatory |
| `steps` | Ordered pipeline phases executed in the container |
| `tool_profile` | Allowed tools / MCP servers / Cedar policy module refs |
| `prompt` | System prompt fragment or reference to registry prompt asset |
| `hydration` | Which context sources to assemble (issue, PR, memory, attachments, URLs) |
| `post_hooks` | e.g. `verify_build`, `verify_lint`, `ensure_pr`, custom deliverables |
| `terminal_outcomes` | What "done" means (PR URL, review JSON, artifact upload, comment) |

**Example step kinds (extensible):**

- `clone_repo` / `skip_repo` — coding vs knowledge path
- `hydrate_context` — shared hydration contract
- `run_agent` — Claude Agent SDK loop with capability prompt + tools
- `verify_build` / `verify_lint` — optional quality gates
- `ensure_pr` / `post_review` / `deliver_artifact` — domain-specific completion

**Platform changes (high level)**

1. **Create-task API:** Accept `capability_ref` (registry ID + semver constraint) or legacy `task_type` (maps to built-in capability for backward compatibility).
2. **Orchestrator:** Resolve capability at admission; skip repo pre-flight when `requires_repo: false`; pass resolved capability bundle to compute session.
3. **Agent pipeline:** Replace `if task_type == "pr_review"` branches with a **step runner** that interprets the capability file's `steps` list (deterministic phases + agent invocation).
4. **Shipped capabilities:** Publish `new_task`, `pr_iteration`, `pr_review` as first-party capability files — behavior parity with today is an acceptance criterion.
5. **Example non-coding capability:** One reference capability (e.g. `document_draft` or `web_research`) proving repo-optional execution end-to-end.

**Phasing**

| Phase | Deliverable |
|-------|-------------|
| 0 | Capability schema + step runner skeleton; design doc |
| 1 | Migrate `new_task` to capability file; legacy `task_type` alias |
| 2 | Migrate `pr_iteration`, `pr_review`; parity tests |
| 3 | Repo-optional knowledge capability + CLI/API `capability_ref` submit path |
| 4 | Registry-native capabilities (#246); inline/repo-local capabilities for dev |

### Acceptance criteria

- [ ] Capability file schema documented with step types and validation rules.
- [ ] Step runner executes a capability's `steps` in order inside the container; failures surface as today (terminal FAILED with structured error).
- [ ] `new_task`, `pr_iteration`, `pr_review` each implemented as capability files with **no behavioral regression** vs current shipped paths (covered by existing tests + #236 integ when available).
- [ ] At least one **non-coding** reference capability runs end-to-end without `repo` (attachments + description sufficient).
- [ ] `POST /v1/tasks` accepts `capability_ref`; legacy `task_type` continues to work (deprecated alias mapping documented).
- [ ] CLI: `bgagent submit --capability <id>@<constraint> ...` (or equivalent); help text explains migration from `--task-type`.
- [ ] Resolved capability id/version recorded on task metadata for audit and eval.
- [ ] Types synced: `cdk/src/handlers/shared/types.ts` ↔ `cli/src/types.ts`.

### Other information

- **Inspired by:** [#99](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/99) (AKW integration — `task_mode`, blueprint registry, knowledge tasks). This issue is the **focused, incremental** track aligned with current codebase — not the full AKW merge (Mem0, ToolBuilderAgent, Trust graduation, etc. remain out of scope or separate issues).
- **Depends on / pairs with:** [#246](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/246) (registry stores published capabilities); [#245](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/245) (attribution on resolved capability); [#236](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/236) (E2E verification).
- **Roadmap alignment:** Tool capability tiers (extended tool surface per repo); Agent asset registry capability descriptors.
- **Out of scope for this issue:**
  - Meta-agents that generate capabilities at runtime (#99 ToolBuilderAgent / BlueprintBuilderAgent)
  - Mem0 / alternate memory backends
  - Full replacement of Cedar tool-level HITL with event rules (#230)
  - Auto-spawn child tasks for unknown capabilities (DEFER → builder agent)

### Acknowledgements

krokoko:
Workflow-driven tasks (#248): replace the hardcoded `task_type` enum with declarative, versioned **workflow** files that describe how the agent runs one kind of task — and unlock **repo-less** (knowledge) workflows that run end-to-end with no GitHub repo. Implements ADR-014.

#### Area

#### Related

- Closes #248
- ADR: [ADR-014 — Workflow-driven tasks](docs/decisions/ADR-014-workflow-driven-tasks.md) (with 2026-06-08 addenda: Phase-2a sequencing; repo-optional open questions / schema freeze)
- Design: [docs/design/WORKFLOWS.md](docs/design/WORKFLOWS.md)
- Pairs with #246 (registry — workflows are its first `capability`-kind consumer; registry resolution is Phase 4, out of scope here)

#### Changes

**Workflow model + runner (agent).** A versioned YAML **workflow** declares ordered `steps`, `domain`, `requires_repo`, `read_only`, prompt, `agent_config`, hydration, and terminal outcomes. A step runner (`agent/src/workflow/`) interprets `workflow.steps` in-container; the scattered `task_type` branches become step handlers. The three shipped task types are now first-party workflow files (`coding/new-task-v1`, `coding/pr-iteration-v1`, `coding/pr-review-v1`) plus the platform default `default/agent-v1`. A cross-field validator + `contracts/workflow-validation/` golden corpus gate authoring.

**`task_type` → `workflow_ref` (breaking, pre-1.0).** The enum is removed end-to-end (API/CLI/agent); the request field is `workflow_ref`, recorded metadata is `resolved_workflow {id, version}`. Legacy callers migrate via the published one-to-one map (`new_task → coding/new-task-v1`, etc.). CDK↔CLI types kept in sync.

**Cedar read-only migration (security, ADR-014 §Phase 2a).** Read-only enforcement moved from a literal `Agent::TaskAgent::"pr_review"` principal match to a `context.read_only == true` rule, applied uniformly to *every* read-only workflow. Kept byte-for-byte identical across both Cedar bindings (`agent/policies/hard_deny.cedar` + `cdk/.../builtin-policies.ts`, drift-guard enforced) and locked by new `contracts/cedar-parity/` fixtures verified on **both** the `cedarpy` (Python) and `cedar-wasm` (TypeScript) engines. Fails closed (missing attribute → DENY).

**Repo-optional / repo-less tasks (Phase 3, acceptance criterion #4).**
- Admission: `repo` is optional across CDK + CLI types; required only when the resolved workflow's `requires_repo` is true (`requires_repo:false` = repo-*optional*, not forbidden). Onboarding/blueprint lookup runs only when a repo is present.
- Pipeline: a repo-less task branches to `_run_repoless_task`, driving `hydrate_context → run_agent → deliver_artifact` through the workflow runner with no clone/build/PR. The coding path is unchanged.
- `deliver_artifact`: a registered-deliverer dispatcher (`agent/src/workflow/deliverers.py`); the `s3` deliverer uploads the agent's result text to `artifacts/{task_id}/result.md` (new SessionRole `s3:PutObject` grant scoped to `artifacts/${task_id}/*`, 5 MiB cap, `ARTIFACTS_BUCKET_NAME` env), surfaced on `TaskDetail.artifact_uri`; the `comment` deliverer records a `delivered_comment` milestone (event-stream visible; external-channel routing is a tracked follow-up).
- Memory: repo-less tasks key long-term memory on `user:{user_id}` instead of `repo`.
- A runnable reference knowledge workflow ships: `agent/workflows/knowledge/web-research-v1.yaml`.

**Verification.** agent 954 / cdk 1833 / cli 300 green; `ty` + `ruff` + ESLint + CDK↔CLI type-sync clean; `mise run build` PASS; `cdk synth` + cdk-nag clean. `mise run security` clean for this branch's changes (the 2 findings it surfaces — semgrep `scripts/check-types-sync.ts`, zizmor `build.yml` — are pre-existing on `main`, not introduced here; the semgrep one is tracked separately). Reviewed across four passes (correctness, silent-failure/fail-open, test coverage, type design, comments).

**Reviewer notes / call-outs**
- **Breaking change:** removing `task_type` contradicts the issue's "legacy `task_type` alias" acceptance line — this is an intentional, ADR-documented hard cutover (no dual-field grace period). Please confirm this deviation is acceptable, or I'll restore an alias.
- **Out of scope / follow-ups:** Phase 4 registry resolution (#246); rendering the `delivered_comment` milestone to an external channel (Slack/email/GitHub); `post_review` handler (no shipped workflow uses it).

Draft pending the breaking-change sign-off and a green CI run.

Some tests:

A1 — Coding (new task → PR) — the most common path; verifies #1 (bare submit still opens a PR)
example: 
```
node lib/bin/bgagent.js submit --repo krokoko/agent-plugins \                                                                               
    --task "Update the rfc template to add a new codeowner file" \
    --workflow coding/new-task-v1"
```
```                                                                 
Task 01KTQ49FYJYXB3SHGNY4GN21W0 — COMPLETED (4m 49s total)
  Repo:          krokoko/agent-plugins
  Channel:       api
  Description:   Update the rfc template to add a new codeowner file
  Turn:          25 / ~—
  Last milestone: pr_created (5s ago)
  Current:       task completed
  Cost:          $0.23 / budget —
  PR:            https://github.com/krokoko/agent-plugins/pull/101
  Last event:    2026-06-09T21:28:16.832506+00:00
```
A2 — Repo-less knowledge task — verifies #2 (CLI reachable) + #8 (research prompt) + repo-less lane
```
node lib/bin/bgagent.js submit --workflow knowledge/web-research-v1 --task "Compare gRPC vs REST for internal microservices; cite sources"
```
```
Task 01KTQBHRXGF7K6PNQ2T3411XYY — COMPLETED (2m 37s total)
  Repo:          — (repo-less)
  Channel:       api
  Workflow:      knowledge/web-research-v1
  Description:   Compare gRPC vs REST for internal microservices; cite
                 sources
  Turn:          49 / ~—
  Last milestone: agent_execution_complete (14s ago)
  Current:       task completed
  Cost:          $0.32 / budget —
  Artifact:      s3://XXXXXXX/artifacts/01KTQBHRXGF7K6PNQ2T3411XYY/result.md
  Last event:    2026-06-09T23:32:53.977286+00:00
```
The markdown file is generated and contains the result of the research
image

A3 — PR iteration & review (PR flags still work)
```
node lib/bin/bgagent.js submit --repo krokoko/agent-plugins --review-pr 102 # read-only review
```
Result: comments correctly added to https://github.com/krokoko/agent-plugins/pull/102
```
node lib/bin/bgagent.js submit --repo krokoko/agent-plugins --pr 102 # address feedback
```
Result:
```
Task 01KTQCDSXD4ENRKSHCB8G6DFRK — COMPLETED (5m 37s total)
  Repo:          krokoko/agent-plugins
  Channel:       api
  Workflow:      coding/pr-iteration-v1 (PR #102)
  Turn:          16 / ~—
  Last milestone: pr_created (2m 06s ago)
  Current:       task completed
  Cost:          $0.18 / budget —
  PR:            https://github.com/krokoko/agent-plugins/pull/102
  Last event:    2026-06-09T23:51:37.508Z
```

A4 - Resolution-ladder / repo-optionality
```
node lib/bin/bgagent.js submit --workflow default/agent-v1 --task "Draft a 200-word summary of CQRS"
```
```
Task 01KTQCHW04E5GZ189SMDAWCRVC — COMPLETED (11s total)
  Repo:          — (repo-less)
  Channel:       api
  Workflow:      default/agent-v1
  Description:   Draft a 200-word summary of CQRS
  Turn:          2 / ~—
  Last milestone: agent_execution_complete (1m 37s ago)
  Current:       task completed
  Cost:          $0.05 / budget —
  Artifact:      s3://XXXXXXXX/artifacts/01KTQCHW04E5GZ189SMDAWCRVC/result.md
  Last event:    2026-06-09T23:48:17.547Z
```

Document created:
image

A5 - guardrails and validation (expected failures)

```
node lib/bin/bgagent.js submit --task "do something"                                                                                              ✔ 
Error: --repo is required (omit it only with --workflow for a repo-less workflow).
```
```
node lib/bin/bgagent.js submit --workflow default/agent-v123 --task "Draft a 200-word summary of CQRS"                                         
Error: Unknown workflow_ref "default/agent-v123". (VALIDATION_ERROR)
```

```
node lib/bin/bgagent.js submit --repo krokoko/agent-plugins --task 'xxx' --workflow coding/does-not-exist-v1                                  
Error: Unknown workflow_ref "coding/does-not-exist-v1". (VALIDATION_ERROR)
```
#### Acknowledgment


isadeks:
**Scope:** 113 files, +9029/−596. Replaces the hardcoded `task_type` enum with declarative YAML workflows, adds repo-less tasks, an S3 deliverer, and migrates Cedar read-only enforcement to a `context.read_only` rule.

This is a well-structured PR with strong test coverage and thoughtful docs. The findings below are concentrated in the **resolution ladder / repo-optionality seam** (where CDK admission, agent pipeline, and CLI disagree about what a default or repo-less task means) and the **half-migrated workflow runner** (declarative `steps`/`gate`/`allowed_tools` partly ignored on the coding path).

---

## Findings (ranked, most severe first)

### 1. ⚠️ Bare `bgagent submit --repo --task` silently stops producing PRs
**`cli/src/commands/submit.ts:183` + `cdk/src/handlers/shared/workflows.ts:131`**

The old `task_type` default was `new_task` (clone → build → open PR). The CLI now only maps `--pr`→`pr-iteration` and `--review-pr`→`pr-review`; a bare submit leaves `workflow_ref` undefined. The server's resolution ladder (`resolveWorkflowRef(undefined)`) falls back to **`default/agent-v1`** — `requires_repo:false`, repo-less, Read/Glob/Grep/WebFetch only, S3 artifact — **not** `coding/new-task-v1`. The documented `new_task → coding/new-task-v1` mapping exists only in a comment; the Blueprint-default rung is "Phase 4, not yet wired."

**Failure:** `bgagent submit --repo owner/repo --task "implement X"` — previously opened a PR, now produces an S3 markdown artifact with no clone/build/PR. The most common invocation regresses silently. Either map a repo-present default to `coding/new-task-v1`, or block bare submit.

### 2. ⚠️ Repo-less workflows are unreachable from the CLI (`--repo` is required)
**`cli/src/commands/submit.ts:62`**

`--repo` is a `.requiredOption(...)`, but this PR makes `repo` optional server-side for `requires_repo:false` workflows. `bgagent submit --workflow knowledge/web-research-v1 --task "..."` is rejected by commander before a request is built, even though `create-task-core.ts:128` would accept it. The headline repo-less feature can't be exercised through the CLI.

### 3. ⚠️ Repo-optional workflow given a repo: the agent silently drops it
**`agent/src/pipeline.py:645` vs `cdk/src/handlers/shared/context-hydration.ts`**

Repo-optionality is decided on two disagreeing axes. The orchestrator's hydration keys off **`task.repo` presence** (repo present ⇒ repo-bound prompt, issue/PR fetch), but `pipeline.run_task` keys off **`config.requires_repo`** (the workflow flag) to choose `_run_repoless_task`. `create-task-core.ts:128` explicitly allows `requires_repo:false` + a supplied repo.

**Failure:** submit `default/agent-v1 --repo owner/repo`. CDK accepts and persists the repo; the orchestrator assembles a repo-bound prompt; the agent sees `requires_repo == False`, takes `_run_repoless_task` (no clone, no build, no PR), and the prompt promises a repo the agent never clones. The two halves disagree on whether the repo is in play.

### 4. ⚠️ Per-workflow tool ceiling is never enforced — `allowed_tools` is hardcoded
**`agent/src/runner.py:323`**

`allowed_tools = ["Bash", "Read", "Write", "Edit", "Glob", "Grep", "WebFetch"]` is a fixed list; the workflow's `agent_config.allowed_tools` is ignored. The schema/YAML claim this list "Replaces the hardcoded list in runner.py" and that excluding Write/Edit/Bash is the "defense in depth" second layer.

**Failure:** `default/agent-v1` declares `[Read, Glob, Grep, WebFetch]` (no mutation, by design) and is `read_only:false`, so the Cedar hard-deny (`context.read_only == true`) does **not** fire. A default-workflow task that should be incapable of mutation can Write/Edit/Bash, gated only by soft-deny HITL. The intended hard tool-restriction is gone. (Read-only workflows are still protected by the `context.read_only` hard-deny — that invariant survives; the per-workflow ceiling does not.)

### 5. Post-hook `load_workflow` lacks the fail-soft fallback `build_config` uses → task FAILED after the tree is already mutated
**`agent/src/pipeline.py:853`**

`build_config` (`config.py:382`) wraps `load_workflow` in a `WorkflowValidationError` fallback (id-derived `read_only`/`requires_repo`) so a load failure still runs the task. The post-hook reload at `pipeline.py:853` is **not** wrapped — it's only caught by the broad terminal `except Exception`.

**Failure:** a workflow file that `build_config` couldn't load took the fallback and the agent ran the full loop (possibly committing). At finalize, `load_workflow` raises again; instead of falling back, the task is marked FAILED with the work stranded (no PR/commit resolution) — inconsistent with `build_config`'s identical-failure handling.

### 6. `resolveWorkflowRef` silently discards a `@version` constraint
**`cdk/src/handlers/shared/workflows.ts:147`**

`isValidWorkflowRef` accepts `@<constraint>` syntactically (regex `(@[^\s]+)?`), but `resolveWorkflowRef` does `ref.split('@', 1)[0]` (JS `limit=1` returns only the first element) and always pins to the single shipped version.

**Failure:** `workflow_ref: "coding/new-task-v1@2.0.0"` passes validation, `@2.0.0` is dropped, the task runs `1.0.0` with no warning or error. When a v2 ships with different behavior, callers pinned to `@2.0.0` silently keep getting v1. An unsatisfiable pin should 400, not be discarded.

### 7. `deliver_artifact` with an unset `target`: validator is lenient but runtime defaults to `s3` → declared `comment` outcome never delivered, yet task reports success
**`agent/src/workflow/runner.py:625` vs `agent/src/workflow/deliverers.py:170`**

`produced_outcomes(None)` returns the **full** `DELIVER_OUTCOMES` set (incl. `comment`), so a workflow declaring `terminal_outcomes.primary: comment` with a `deliver_artifact` step that omits `target` **passes** validation (rule 11). At runtime `_handle_deliver_artifact` does `target = step.target or "s3"` → only `_upload_to_s3` runs (`produces={'artifact'}`), never `_post_comment`.

**Failure:** the declared terminal deliverable (`comment`) is never produced, but the S3 upload succeeds so `succeeded=True` — a silent "succeeded with the declared product never delivered." The lenient validator default and the runtime `s3` default must agree.

### 8. `knowledge/web-research-v1` has no registered prompt → silent fallback to the generic default-agent prompt
**`agent/src/prompts/__init__.py:24`**

CDK accepts `knowledge/web-research-v1` as valid and the YAML ships in the image, but `_PROMPTS` has no entry for that id. `get_system_prompt` logs a WARN and falls back to `default/agent-v1`'s prompt.

**Failure:** a task submitted with the reference knowledge workflow runs with the generic default-agent system prompt instead of a research-specialized one — the workflow silently degrades to generic behavior with no error surfaced to the user. (CDK admission and agent capability diverge.)

### 9. S3 artifact 5 MiB cap is checked *after* the full result is encoded → cap never bounds memory
**`agent/src/workflow/deliverers.py:77`**

`_artifact_body` does `body = text.encode("utf-8")` (a second full copy of `ctx.agent_result.result_text`, already resident) and only then checks `len(body) > MAX_ARTIFACT_BYTES`.

**Failure:** a knowledge task that ingests a large page and emits a multi-hundred-MB result has both the `str` and its UTF-8 `bytes` resident before the cap fires — an OOM risk on the constrained agent MicroVM, which is exactly what the cap was meant to prevent. Check `len(text)` (or character budget) up front; for `s3_and_comment`, `_artifact_body` is also re-encoded twice.

### 10. `read_only` / `requires_repo` are two hand-maintained sources (CDK `DESCRIPTORS` vs agent YAML) with no parity test
**`cdk/src/handlers/shared/workflows.ts:64` + `agent/src/config.py:27`**

CDK admission/preflight derives `readOnly`/`requiresRepo` from the TS `DESCRIPTORS` table; the agent derives the same from `agent/workflows/*.yaml`. The drift test (`builtin-policies.test.ts`) covers only Cedar policy **text**, not these descriptors, and the workflows fixture only checks `id`/`version`. `_KNOWN_WRITEABLE_WORKFLOW_IDS` is a third hand-maintained copy of the writeable set.

**Failure:** edit a YAML to `read_only:true` but leave TS `readOnly:false` and preflight scopes a `contents:write` token for a task the agent runs read-only (over-grant); or add a writeable workflow and forget `_KNOWN_WRITEABLE_WORKFLOW_IDS` and a load-failure fallback mis-denies its writes. Nothing in CI catches the divergence. Generate the descriptor table from the YAML (codegen/JSON sidecar), or add a parity test over all four fields + `requiredInputs`.

---

## Lower-priority (noted, not in the top 10)

- **Dead `post_review` plumbing** (`runner.py:613`): a registered handler that raises `NotImplementedError`, plus the `post_review`/`review_posted` vocabulary threaded through the schema, models, and validator special-cases — no shipped workflow uses it. Inviting a runtime error where a validation error belongs.
- **Workflow YAML loaded/parsed 3–4× per task** (`server.py`, `config.py:382`, `pipeline.py:172`, `pipeline.py:853`): `load_workflow` isn't memoized though the file is immutable per process — add `@lru_cache` or thread the resolved `Workflow` forward.
- **Half-migrated runner**: on the repo-bound path `run_workflow(only_kinds={"run_agent"})` drives only the agent step; `verify_build`/`ensure_pr` stay inline and the inline gate logic ignores the per-step `gate` field declared in the coding YAMLs. The declarative `steps`/`gate` are effectively decorative for coding workflows — worth flagging as an altitude concern for follow-up.
- **Cedar `context.read_only == true` is structurally fail-open** on a missing/non-bool attribute. Not currently reachable (`policy.py` always threads a real bool), but a latent fragility — consider a fail-closed default in the policy or an explicit assertion.

---

## Bottom line

The architecture (declarative workflows, registry-driven deliverers, Cedar-by-property) is sound and the right direction. The blocking issues are the **resolution-ladder/repo-optionality seam** (#1–#3) — a default coding submit silently stops opening PRs, and repo-optional + repo disagree across the boundary — and the **unenforced per-workflow tool ceiling** (#4). I'd hold the breaking cutover until #1–#4 are resolved; #5–#10 are fixable in follow-up but #7/#10 risk silent divergence.

laithalsaadoon:
Reviewed at HEAD `3e3a66f`, re-fetched after `d26860c` and `3e3a66f`. Thanks @krokoko. I read this through the lens of where ABCA is heading: one engine, with coding and knowledge lanes. So the framing leans architectural. Sophia's review is the line-by-line pass and is already in flight, so I've tried not to duplicate it.

**My two findings are verified closed, at the code level rather than the commit message.**

First, `allowed_tools` now reaches the SDK. `_resolve_allowed_tools` (`agent/src/runner.py:285`) reads `config.allowed_tools`. `build_config` populates that from `workflow_obj.agent_config.allowed_tools` (`agent/src/config.py:386`). It then flows to `ClaudeAgentOptions` verbatim (`runner.py:387`). The empty-list fallback to the full surface fires only on the `WorkflowValidationError` path. On that path `read_only` is also fail-closed to `true` (`config.py:402`), so Write and Edit drop anyway. The defense-in-depth layer the design promised is real now.

Second, the Cedar missing-attribute path is pinned. `contracts/cedar-parity/read-only-missing-attribute.json` is honest in the best way. It asserts both engines fail open identically on a missing `read_only`. It also documents in-fixture that safety rests on `policy.py::_eval_tier` re-raising on `diagnostics.errors`, which maps to `Outcome.DENY`, plus the always-inject discipline. It does not rest on Cedar's native semantics. `test_eval_tier_reraises_on_missing_read_only_attribute` locks the re-raise. This is the ADR-014 Phase-2a "silently weakens" tripwire. Resolved.

Both of mine are done. The rest is architectural color on the work Sophia already surfaced.

**The architectural read: this PR ships the lane-router, and the lane-router is the bet.**

Strip away the YAML and #248 ships a routing primitive. A submission arrives, and one decision picks the lane: coding, which clones, builds, and opens a PR, versus knowledge, which hydrates, runs the agent, and delivers an artifact. That decision is the control point for everything ABCA is growing into. So the findings worth holding the cutover for are not the ones with the scariest blast radius. They are the ones where the lane decision is computed in more than one place, and the places disagree. Today it is computed in four: CLI flag inference, server `resolveWorkflowRef`, the `orchestrate-task` missing-id default, and the agent's `config.requires_repo`. Two of Sophia's findings are symptoms of those four not agreeing.

I'd gate the cutover on Finding #1, where a bare `submit --repo --task` stops producing a PR. I want to underline it, because the fix commit is titled for the workflow seam but this regression is still live. Traced at HEAD: the CLI bare submit sends no `workflow_ref` (`cli/src/commands/submit.ts`). The server then resolves `default/agent-v1` (`cdk/src/handlers/shared/workflows.ts:173`), which sets `requiresRepo:false`. The agent takes `_run_repoless_task` (`agent/src/pipeline.py:660`) and writes an S3 markdown artifact. No clone, no build, no PR.

The `coding/new-task-v1` mapping lives only in a comment. The `orchestrate-task.ts` default to `coding/new-task-v1` fires only when `resolved_workflow` is wholly absent, which means legacy records, not this case. So the most common historical invocation, "here is my repo, do this task," silently changes from "opens a PR" to "writes a markdown file." For a pre-1.0 breaking cutover, that burns trust on day one. The cheapest correct fix: when a repo is present and no `workflow_ref` was given, resolve to `coding/new-task-v1`. Repo-presence is the signal. That keeps the ladder honest and puts the lane decision on one axis.

Finding #3, a repo-optional workflow given a repo, is the same root cause. Hydration keys the lane off `task.repo` presence (`context-hydration.ts`) and builds a repo-bound prompt. The agent keys it off `config.requires_repo` (`pipeline.py:660`) and runs repo-less. Submit `default/agent-v1 --repo owner/repo` and the orchestrator promises the agent a repo it never clones. Repo-presence and `requires_repo` are two independent booleans that must be one decision. This is invisible today because nobody submits that combination. But once knowledge workflows are real and a user attaches a repo for context, the prompt and the runtime diverge. Better to collapse it now, while there are two lanes, than after there are nine.

I'd treat #1 and #3 as one fix. Make "is this task repo-bound?" a single value, resolved at the create-task boundary and threaded everywhere. Then the CLI, server, orchestrator, and agent cannot hold four opinions.

**Independent corroboration on three of Sophia's open items.** These are being worked, so this is a severity read, not new findings.

Finding #5 I'd rank High. The post-hook `load_workflow` (`pipeline.py:868`) is a bare call with no `WorkflowValidationError` fallback, unlike `build_config` (`config.py:381`). It is the only spot that can fail a task after `run_agent` already mutated the tree. The work gets stranded with no commit and no PR. This is distinct from the "no PR" findings: here a PR should exist and does not. Wrap it in the same fallback `build_config` uses.

Finding #10 is mostly closed, with one axis left. The new `workflows.test.ts` drift guard over `id`, `version`, `requires_repo`, `read_only`, and `model` genuinely closes the TS-to-YAML axis. But the third copy Sophia named, the Python `_KNOWN_WRITEABLE_WORKFLOW_IDS` and `REPO_LESS_DEFAULT_WORKFLOW_ID` (`config.py:19,27`), still has no parity test against the YAML. That set is the fail-closed fallback's source of truth. If it drifts, the load-failure path mis-classifies read-only. One more assertion closes it.

Finding #6 is Low, but make it loud. The `@version` constraint is discarded at `workflows.ts:177` via `split('@',1)[0]`, which pins to the single shipped version. A test now codifies this as intended, a reasonable Phase-1 stance. But an unsatisfiable pin like `...-v1@2.0.0` should `400`, not silently run v1. The day a v2 ships, anyone pinned to `@2.0.0` keeps getting v1 with no signal. Cheap to reject unknown versions now.

Two smaller items I'd not block on but would log so they don't get lost. `load_workflow` is parsed three times per repo-bound task with no `@lru_cache` (`loader.py`). And the coding path still runs `run_workflow(only_kinds={"run_agent"})` with `verify_build` and `ensure_pr` inline, ignoring each step's declared `gate` field. So `steps` and `gate` are decorative for coding workflows today. Both are fine as follow-ups. They are worth a tracking note so "declarative steps" doesn't read as fully wired on the coding lane.

**Bottom line.** The direction is right and the security core is solid. My two findings are genuinely closed, and the byte-equality and cedar-parity guards are real; the full agent suite was green when I ran it. I'd hold the breaking `task_type` removal only until #1 and #3 collapse onto a single repo-bound decision, and #5 gets the fail-soft wrapper. Everything else is good follow-up material. Nice work on the convergence groundwork. The repo-less lane is the part that matters most long-term, and it is structurally sound.

Reviewed via background-agent tooling. Happy to pair on the #1 and #3 single-axis refactor if useful.

cdk/src/handlers/shared/workflows.ts
```diff
@@ -0,0 +1,217 @@
...
+ *   2. (Blueprint default — Phase 4, not yet wired);
+ *   3. the platform default ``default/agent-v1``.
+ *
+ * Returns ``null`` when an explicit ref names an unknown workflow (caller
+ * returns 400). A constraint suffix is accepted syntactically and, in Phase 1,
+ * pins to the single shipped version of that id.
+ */
+export function resolveWorkflowRef(ref?: string | null): ResolvedWorkflow | null {
+  if (ref === undefined || ref === null || ref === '') {
+    const fallback = DESCRIPTORS[DEFAULT_WORKFLOW_ID];
```
theagenticguy, line 193: **[#1 — the one I'd gate the cutover on]** A bare `bgagent submit --repo owner/repo --task "…"` sends no `workflow_ref`, so this fallback resolves `default/agent-v1` (`requiresRepo:false`) → the agent takes `_run_repoless_task` (`pipeline.py:660`) → S3 markdown artifact, **no clone/build/PR**. That's the most common historical invocation silently changing from "opens a PR" to "writes a file." The `new_task → coding/new-task-v1` mapping only exists in a comment; `orchestrate-task.ts`'s `coding/new-task-v1` default fires only when `resolved_workflow` is wholly absent (legacy records), not here. Suggest: when a repo is present and no ref was given, resolve to `coding/new-task-v1` — repo-presence is the signal — so the lane decision lives on one axis (see #3, same root cause).

agent/src/pipeline.py
```diff
@@ -623,19 +860,35 @@ def _on_trace_truncated(max_bytes: int, first_dropped: int) -> None:
                     "turns_attempted": agent_result.num_turns or agent_result.turns,
                 }
 
+            # Resolve the workflow once for post-hook gating: read_only (replaces
+            # the former task_type=='pr_review' checks) and the ensure_pr
+            # strategy (create / push_resolve / resolve) the workflow declares.
+            from workflow import load_workflow
+
+            _workflow = load_workflow(
```
theagenticguy: **[#5 — High: can FAIL a task *after* the tree is mutated]** This `load_workflow(...)` is a bare call with no `try/except WorkflowValidationError`, unlike the guarded reload in `build_config` (`config.py:381-404`). If the same workflow file that `build_config` already fell back on (corrupt YAML, registry-only id) is reloaded here at finalize — *after* `run_agent` has already edited/committed the tree — it raises into the broad terminal `except` and the task is marked FAILED with the work stranded: no `ensure_pr`, no commit resolution. Distinct from the "no PR" findings — here a PR *should* exist and doesn't. Fix: wrap this in the same id-derived `read_only`/strategy fallback `build_config` uses, so an identical load failure is handled identically on both ends of the pipeline.

laithalsaadoon:
**Follow-up — re-verified at `67d71ec`** (you pushed it while I was typing the comment above 😄, so updating to keep my comments honest against the current tree).

`67d71ec` lands fixes for most of what I'd flagged. Re-checked each against the code:

- ✅ **#1 (bare `submit --repo` → no PR)** — `cli/src/commands/submit.ts:199-208` now sends `workflow_ref: 'coding/new-task-v1'` when a repo is present and no `--workflow`/`--pr`/`--review-pr` is given. That's the repo-present-is-the-signal fix; the old `new_task` default is preserved. My inline note on `workflows.ts:174` is now **resolved** — leaving it for thread continuity only.
- ✅ **#3 (repo-optional + repo disagreed)** — `pipeline.py:670` now takes the repo-less branch only when `not requires_repo AND not repo_url`, so a repo-optional workflow *given* a repo runs the repo-bound path (clone/build/PR), matching CDK admission and the hydrated prompt. The two axes are one decision now. 👍
- ✅ **#6 (`@version` silently discarded)** — `parseWorkflowRef` + `resolveWorkflowRefError` (`workflows.ts:162,186`) now honor the constraint and 400 on an unsatisfiable pin instead of quietly running the shipped version.
- ✅ **#7 (deliver_artifact unset-target)** — `DEFAULT_DELIVER_TARGET` is now the single source of truth shared by validator and runtime, so a `primary: comment` workflow that omits `target` is correctly flagged by rule 11.

**Two still open after `67d71ec`:**

1. **#5 — post-hook `load_workflow` is still a bare call (`pipeline.py:878`), no `WorkflowValidationError` fallback.** Not mentioned in the commit message and confirmed unchanged. This is the one I'd still treat as **High**: it's the only load that can raise *after* `run_agent` has mutated/committed the tree, stranding the work with no `ensure_pr`. My inline comment on it stands — it just needs the same id-derived fallback `build_config` (`config.py:381-404`) already uses.
2. **#10 — residual third copy.** The new `workflows.test.ts` drift guard over `id/version/requires_repo/read_only/model/required_inputs` is a genuine close of the **TS↔YAML** axis 👏. But the Python constants `_KNOWN_WRITEABLE_WORKFLOW_IDS` / `REPO_LESS_DEFAULT_WORKFLOW_ID` (`agent/src/config.py:19,27`) — the source of truth for the *fail-closed* load-failure path — still have no parity test against the YAML (`agent/tests/` only checks `resolved_requires_repo` per file). If that set drifts, the fallback mis-classifies read-only silently. One assertion closes it.

Net: the resolution-ladder seam I'd have gated on is now collapsed onto a single axis — nice turnaround. I'd hold only on **#5** (cheap, but it's a post-mutation failure) and treat **#10's third-copy** as a should-fix-before-merge. Everything else from my pass is closed. 🚢

krokoko:
@isadeks thanks for the thorough, accurate pass — this was exactly the right place to concentrate scrutiny (the resolution-ladder/repo-optionality seam). Status against the current tree, by finding:

**Fixed**

- **#1 — bare `submit --repo --task` stopped opening PRs** ✅ `67d71ec`. The CLI now sends `workflow_ref: 'coding/new-task-v1'` when a repo is present and no `--workflow`/`--pr`/`--review-pr` is given (`cli/src/commands/submit.ts`). Repo-presence is the signal; the server ladder stays repo-agnostic by design.
- **#2 — repo-less unreachable from CLI** ✅ `3e3a66f` (before your re-fetch). `--repo` is no longer a hard `requiredOption`; it's required unless `--workflow` selects a repo-less workflow (still required with `--pr`/`--review-pr`).
- **#3 — repo-optional + repo disagreed** ✅ `67d71ec`. `pipeline.run_task` takes the repo-less branch only when `not requires_repo AND not repo_url`; a repo-optional workflow *given* a repo runs the repo-bound path. The two axes are one decision now.
- **#4 — `allowed_tools` hardcoded** ✅ `3e3a66f`. `_resolve_allowed_tools(config)` threads the workflow's list to the SDK and drops Write/Edit when `read_only`; the empty-list fallback only fires on the `WorkflowValidationError` path, where `read_only` is already fail-closed.
- **#6 — `@version` silently discarded** ✅ `67d71ec`. `parseWorkflowRef` + `resolveWorkflowRefError` honor the constraint; an unsatisfiable pin 400s ("version not available") instead of running the shipped version.
- **#7 — `deliver_artifact` unset-target mismatch** ✅ `67d71ec`. `DEFAULT_DELIVER_TARGET` is now the single source of truth shared by validator and runtime, so a `primary: comment` workflow that omits `target` is correctly flagged by rule 11.

**Still open / partial — agree with your severity**

- **#5 — post-hook `load_workflow` has no fail-soft fallback** — confirmed unchanged (`pipeline.py:878`). I agree with the High ranking (@theagenticguy independently flagged the same): it's the only load that can raise *after* `run_agent` has mutated/committed, stranding the work with no `ensure_pr`. Fix is the id-derived fallback `build_config` already uses. **Will address next.**
- **#10 — partially closed.** The new `workflows.test.ts` drift guard now covers the TS↔YAML axis (`id/version/requires_repo/read_only/model/required_inputs`) **and** cross-checks the agent-side `_KNOWN_WRITEABLE_WORKFLOW_IDS` against the writeable repo-bound descriptors. The residual is `REPO_LESS_DEFAULT_WORKFLOW_ID` (`config.py`), which still has no parity assertion — one more check closes it.
- **#8 (web-research prompt falls back to generic)** and **#9 (5 MiB cap checked after encode)** — agree these are real; tracking as follow-ups. #9's up-front `len(text)` check is a small, worthwhile fix.

**Lower-priority noted:** `load_workflow` parsed 3–4×/task (no `@lru_cache`), and the coding path's inline `verify_build`/`ensure_pr` ignore the per-step `gate` field (declarative `steps`/`gate` are decorative on the coding lane today). Both logged as follow-ups so "declarative steps" isn't read as fully wired on coding.

Net: the resolution-ladder seam (#1–#3) is collapsed onto a single repo-bound decision, and the tool-ceiling (#4) is enforced. Remaining before I'd call the cutover clean: **#5** and **#10's `REPO_LESS_DEFAULT_WORKFLOW_ID` residual**. Thanks again.

krokoko:
Follow-up — pushed `9eacdd8` closing the remaining items from both passes. Thanks @isadeks and @theagenticguy for the thorough reviews.

**Now fixed:**

- **#5 (High) — post-hook `load_workflow` fail-soft** ✅ The reload at `pipeline.py` runs *after* `run_agent` has mutated/committed, so a failure there no longer strands the work. `read_only` now comes from `config` (already fallback-computed by `build_config`, and the verdict that drove Cedar during the run — so the post-hook can't diverge); the reload — needed only for the `ensure_pr` strategy — is wrapped in the same `WorkflowValidationError` fallback `build_config` uses, defaulting to `"create"` so the PR still opens. Test: `test_post_hook_workflow_reload_failure_still_opens_pr`.
- **#8 — `knowledge/web-research-v1` prompt** ✅ Added a research-specialized, repo-less prompt (`prompts/web_research.py`) and registered it, so it no longer silently inherits the generic default-agent prompt.
- **#9 — artifact cap before encode** ✅ `_artifact_body` now rejects on the character count up front (chars ≤ bytes in UTF-8) before materializing the UTF-8 copy, so the 5 MiB cap actually bounds memory on the MicroVM. Test proves `encode()` is never reached on the oversize path.
- **#10 residual — `REPO_LESS_DEFAULT_WORKFLOW_ID` parity** ✅ @theagenticguy — you were right that a residual remained. For the record, `_KNOWN_WRITEABLE_WORKFLOW_IDS` *was* already cross-checked in `67d71ec`'s `workflows.test.ts` (it reads `config.py`); `9eacdd8` adds the missing assertion for `REPO_LESS_DEFAULT_WORKFLOW_ID` / `DEFAULT_WORKFLOW_ID` — the repo-less default must match the CDK platform default and be `requires_repo:false` in the YAML; the coding default must be repo-bound. That closes the last hand-maintained-copy axis.

**Also addressed (perf):** `load_workflow` is now memoized (`@cache`) — files are image-baked and immutable per process and `Workflow` is `frozen=True`, so the 3–4 loads per task parse once.

**Tracked follow-up (not in this PR):** the coding-lane decorative-`gate` observation — `run_workflow(only_kinds={"run_agent"})` drives only the agent step while `verify_build`/`ensure_pr` stay inline and ignore each step's declared `gate`. That's a larger half-migrated-runner refactor both of you flagged as non-blocking; logging it so "declarative steps" isn't read as fully wired on the coding lane, to be picked up separately.

Full suite green before push: agent 975, CLI, CDK 1844 — lint/typecheck/types-sync clean. Deploying to dev now for manual verification.

theagenticguy (approved):
## Approved ✅

Re-verified at `9eacdd8` against the code, not the commit messages. Every item I gated on is closed, and I ran the suite locally rather than trust the green claim.

**My two findings:** both closed. `allowed_tools` reaches the SDK and drops Write/Edit on read-only (`runner.py:285`, `config.py:386`). The Cedar missing-attribute path is pinned by `read-only-missing-attribute.json`, with the re-raise locked by `test_eval_tier_reraises_on_missing_read_only_attribute`.

**The one I held the cutover on, #5:** fixed well in `9eacdd8`. The post-hook now reads `read_only` from `config`. That is the same verdict that drove Cedar during the run, so it cannot diverge. The strategy-only `load_workflow` reload is wrapped in the `WorkflowValidationError` fallback and defaults to `"create"`. So a load failure after the tree is mutated still opens the PR instead of stranding the work. `test_post_hook_workflow_reload_failure_still_opens_pr` covers it. Reusing the config verdict is cleaner than the fix I suggested.

**The rest of Sophia's pass:** #1, #2, #3, #4, #6, #7 confirmed closed. #8 now ships a registered research prompt. #9 checks the character count before encoding, so the 5 MiB cap actually bounds MicroVM memory. #10's residual parity assertion for `REPO_LESS_DEFAULT_WORKFLOW_ID` is in `workflows.test.ts`.

**Local verification:** full agent suite ran green for me, 975 passed in 5.4s. The workflow, policy, cedar-parity, pipeline, deliverer, and prompt paths pass; all three new named tests exist and pass.

**One tracked follow-up, not a blocker:** the coding lane still runs `run_workflow(only_kinds={"run_agent"})` with `verify_build` and `ensure_pr` inline, ignoring each step's declared `gate`. So `steps` and `gate` are declarative only on the knowledge lane today, not on coding. Both reviewers flagged this as a separate half-migrated-runner refactor, and it is logged for follow-up. Worth a short tracking issue so "declarative steps" is not read as fully wired on the coding lane.

The convergence groundwork is solid. The repo-less lane is the part that matters most for where ABCA is heading, and it is structurally sound. Nice work turning all of this around fast. 🚢

Reviewed via background-agent tooling.

agent/src/pipeline.py
krokoko: Done in subsequent commits

cdk/src/handlers/shared/workflows.ts
krokoko: done in subsequent commits

laithalsaadoon:
Opened #301 to track the coding-lane decorative-`gate` follow-up we agreed was non-blocking (per-step `gate` is honored on the repo-less lane via `_gate_status` but not on the coding lane, where `verify_build`/`ensure_pr` run inline). Scoped to making `gate` honored without the larger half-migrated-runner unification. cc @krokoko @isadeks

Issue #288, opened 2026-06-08
feat(jira): Feature request for Jira Cloud integration (parity with Linear)


mayakost:
# Jira Integration for ABCA — Scaffold Plan

## Context

ABCA (Autonomous Background Coding Agents on AWS) currently ingests coding tasks from CLI, GitHub webhooks, Slack, and Linear, then opens PRs autonomously. Linear is the only "issue tracker" channel today. Many teams use Jira instead, so we want a parity-level Jira Cloud integration: an issue gets a `bgagent` label → ABCA picks it up → an agent run produces a PR → status flows back into the Jira issue.

The existing Linear integration (`cdk/src/constructs/linear-integration.ts` + sibling handlers + `agent/src/channel_mcp.py`) is the canonical template. Confirmed scope:

- **Jira Cloud only** (REST v3, cloud webhooks)
- **Per-workspace OAuth 3LO** stored via AgentCore credential provider, mirroring `bgagent-linear-oauth-*`
- **Trigger**: label added (default `bgagent`), parity with Linear
- **Outbound**: Atlassian's Remote MCP Server, registered into `.mcp.json` when `channel_source == "jira"` — same shape as the Linear MCP path

## Reference files (read these first)

Linear is the closest analog and should be copied/adapted, not reinvented.

- `cdk/src/constructs/linear-integration.ts` — webhook + processor + link Lambdas, dedup table, secrets, API routes
- `cdk/src/constructs/linear-project-mapping-table.ts`, `linear-user-mapping-table.ts`, `linear-workspace-registry-table.ts` — DDB schemas
- `cdk/src/handlers/linear-webhook.ts` — HMAC verify, dedup, async-invoke processor
- `cdk/src/handlers/linear-webhook-processor.ts` — label-trigger logic, calls `createTaskCore`, writes channel metadata
- `cdk/src/handlers/linear-link.ts` — Cognito-authenticated user→workspace linking
- `cdk/src/stacks/agent.ts` — how `LinearIntegration` is wired (api, userPool, taskTable, taskEventsTable, repoTable, orchestratorFunctionArn, guardrails) and how the orchestrator is granted read on its workspace registry
- `agent/src/channel_mcp.py` — gates Linear MCP server entry into `.mcp.json` based on `channel_source`
- `agent/src/linear_reactions.py` — outbound progress hooks (consult for parity, may not need a Jira twin if MCP covers it)
[...]

PR #302, opened 2026-06-10
feat(jira): Jira Cloud integration -- parity with Linear (#288)

mayakost:
## Summary

Full **Jira Cloud integration**, bringing Jira to parity with the existing Linear adapter: a Jira issue gets the `bgagent` label → ABCA picks it up → an agent run produces a PR → progress flows back as comments on the originating issue. Implements aws-samples/sample-autonomous-cloud-coding-agents#288.

The Linear adapter was the file-for-file template; this PR diverges only where Jira's API forces it (see *Jira-specific divergences* below). Design rationale is captured in **ADR-014** (`docs/decisions/ADR-014-jira-integration.md`).

**Scope:** Jira **Cloud** only (REST v3 + Cloud webhooks), per-tenant OAuth 3LO, label-only trigger. Inbound-only adapter — no DynamoDB Streams consumer, no outbound-notify Lambda. Out of scope: Jira Server/Data Center, Forge/Connect distribution, status-transition/comment-command triggers, bidirectional state sync.

#### Related

This PR resolves issue #288 to add a Jira integration

## Architecture

**Inbound (Jira → ABCA):**
```
Jira Cloud webhook
  → POST /jira/webhook (API GW, no Cognito, HMAC-verified)
  → JiraWebhookFn      (verify X-Hub-Signature, dedup, async-invoke processor)
  → JiraWebhookProcessorFn (resolve cloudId→tenant, project→repo, build task, createTaskCore)
  → existing orchestrator pipeline (unchanged)
```

**Outbound (Agent → Jira):** progress comments posted on the originating issue at task start and completion.

### Contracts (Phase 1)
- `'jira'` added to the `ChannelSource` union on both sides of the wire (`cdk/src/handlers/shared/types.ts`, `cli/src/types.ts`) plus the `agent/src/models.py` doc comment. `check-types-sync.ts` allowlists `JiraLinkResponse` as CLI-only (parity with `Slack`/`Linear` link responses).

### CDK constructs & DynamoDB (Phases 2 & 4)
- `JiraIntegration` construct (`cdk/src/constructs/jira-integration.ts`) — 3 mapping tables + an 8-hour-TTL webhook-dedup table, 3 Lambdas (webhook / processor / link), `/jira/*` API routes, per-tenant `bgagent-jira-oauth-*` IAM grants, and cdk-nag suppressions.
- Three DDB table constructs, all keyed on `cloudId` as the tenant prefix so the same project key / account id stays unambiguous across tenants:
  | Table | PK |
  |---|---|
  | `JiraProjectMappingTable` | `{cloudId}#{projectKey}` → `owner/repo` |
  | `JiraUserMappingTable` | `{cloudId}#{accountId}` (+ `PlatformUserIndex` GSI) |
  | `JiraWorkspaceRegistryTable` | `jira_cloud_id` → OAuth provider name |
- Stack wiring (`cdk/src/stacks/agent.ts`): grants the orchestrator read on the workspace registry + `Get/PutSecretValue` on the per-tenant prefix (mirrors Linear's pre-container failure-feedback path).

### Lambda handlers & shared helpers (Phase 3)
- `jira-webhook.ts` — HMAC-SHA256 verify over the **raw** body (`X-Hub-Signature: sha256=<hex>`, constant-time compare), event filtering (`jira:issue_created` / `jira:issue_updated`), dedup, async invoke. Silent 200 for unsupported events.
- `jira-webhook-processor.ts` — label-add detection by **diffing `changelog.items[]`** (not `issue.fields.labels`), ADF→markdown for the task description, `cloudId`→tenant resolution, `createTaskCore` with `channelSource: 'jira'`.
- `jira-link.ts` — Cognito-authenticated Jira-account → platform-user linking with dry-run preview.
- Shared: `jira-oauth-resolver.ts` (per-tenant token resolve/refresh), `jira-verify.ts` (signature), `jira-feedback.ts` (REST-based failure comment at the orchestrator boundary).

### Agent runtime (Phase 5)
- `channel_mcp.py` refactored from a single-channel gate to a `CHANNEL_MCP_BUILDERS` dispatch dict — adding a channel is now one entry, not a rewrite.
- `resolve_jira_oauth_token()` added to `config.py`, mirroring the Linear resolver's race-handling and fail-closed semantics (differs only in endpoint `auth.atlassian.com` + JSON body + `JIRA_API_TOKEN`).

### Outbound progress comments — `agent/src/jira_reactions.py` (Phase 7)
Jira-origin tasks comment on the issue at start (`🤖 picked up…`) and completion (`✅ finished — PR: <url>` / `❌ …`), wired into `pipeline.py` at start / finish / crash paths, parallel to the Linear hooks. Comments are advisory — all network/auth errors are swallowed and never gate the pipeline (with an auth circuit-breaker mirroring `linear_reactions`).

> **Note — outbound is REST, not MCP.** ADR-014 originally specified the Atlassian Remote MCP server for outbound. In practice the hosted MCP (`mcp.atlassian.com`) requires an interactive browser-based OAuth 2.1 flow with dynamic client registration and will **not** accept the stored REST OAuth token as a Bearer header, so it fails to load from a headless agent. The Jira REST API accepts the same token (it carries `write:jira-work`), so comments go via `POST /rest/api/3/issue/{key}/comment`. This is the ADR's documented "Plan B" fallback, promoted to the implemented path.

### CLI (Phase 6)
`bgagent jira` with 4 v1 subcommands (`app-template`, `setup`, `link`, `map`) + `jira-oauth.ts` (ports `linear-oauth.ts`; Atlassian uses a JSON token endpoint and requires `offline_access` for a refresh token). Deferred: `add-workspace`, `update-webhook-secret`, `invite-user`, `list-projects`.

### Docs
`JIRA_SETUP_GUIDE.md`, ADR-014, README / USER_GUIDE / ROADMAP channel listings, and regenerated Starlight mirrors.

## Jira-specific divergences from the Linear copy
1. **Label-add on updates** — Jira reports label changes in `changelog.items[]` (`field: "labels"`, `fromString`/`toString`), not as a full label list; the processor diffs the changelog so re-saving an already-labeled issue doesn't re-trigger.
2. **Operator-chosen signing secret** — Atlassian doesn't auto-generate a per-subscription secret; the operator sets it at webhook-create time and pastes it during `bgagent jira setup`. Stored on the per-tenant OAuth bundle with a stack-wide fallback.
3. **`cloudId` is the tenant key** everywhere — not domain or site name.
4. **UI-created webhooks omit `cloudId`** — the processor falls back to the sole `active` tenant in the registry, and deliberately drops (rather than guesses) when zero or multiple active tenants exist, preserving multi-tenant safety.
5. **ADF descriptions** — flattened to markdown (text/headings/lists), not a full ADF converter.
6. **Dedup key** `{issueKey}#{webhookEventTimestamp}`, 8-hour TTL.

## Testing
- `mise //cdk:compile` — clean; `check-types-sync.ts` — passing (verified on this branch).
- Per the final integration commit: **294 CLI + 837 agent + 1896 CDK tests passing**, `mise run build` green.
- New suites: `jira-webhook.test.ts`, `jira-webhook-processor.test.ts`, `jira-link.test.ts`, `jira-oauth-resolver.test.ts` (32), `test_jira_reactions.py`, `test_channel_mcp.py` (jira cases), plus `agent.test.ts` updated for the 4 new tables (13→17).

---
*Branch rebased onto the fork's `main` so the diff is Jira-only (46 files); the 4 unrelated upstream commits it was originally cut from have been dropped.*

#### Acknowledgment


krokoko (changes requested):
Thanks for this PR — it's an impressively thorough port of the Linear template. The OAuth refresh/rotation handling in `jira-oauth-resolver.ts` (including the Atlassian rotated-refresh-token race) and the four-way `verifyJiraRequestForTenant` outcome union are genuinely excellent, and the divergences from the Linear copy (changelog diffing, `cloudId` tenancy, ADF flattening) are all well-justified. The review below is detailed because the PR is large and security-relevant, not because the foundation is weak.

**Verdict: Request changes** — CI is currently red, the docs mirror will fail the mutation gate, there's a coupled multi-tenant signature-binding gap, and the ADR/docs still describe the pre-pivot MCP outbound design.

## Blocking issues

1. **CI is red — `agent/tests/test_config.py:11` is missing a comma after `PR_WORKFLOW_IDS`.** This is a `SyntaxError` at pytest collection time, so the *entire* agent suite (including the new `test_jira_reactions.py` circuit-breaker tests and the 12-case `TestResolveJiraOauthToken` suite) is currently not running. Once fixed, those suites look well-designed.

2. **Stale Starlight mirrors — the "Fail build on mutation" gate will reject.** Running `node docs/scripts/sync-starlight.mjs` at PR HEAD dirties `docs/src/content/docs/using/Overview.md` and `getting-started/Quick-start.md` (the latter inherited via the `main` merge, but it needs a re-sync here). Please run `mise //docs:sync` and commit the regenerated mirrors.

3. **Multi-tenant signature binding (coupled pair):**
   - `cli/src/commands/jira.ts:539-547` mirrors the *stack-wide* signing secret into every per-tenant OAuth bundle, so one shared secret effectively verifies for all tenants.
   - A payload verified via the stack-wide fallback carries no binding between the verified secret and `payload.cloudId` — the processor then trusts the body-supplied `cloudId` to select the tenant, project→repo mapping, and OAuth bundle (`cdk/src/handlers/jira-webhook.ts:136-147`, `jira-webhook-processor.ts:208,234`). A holder of the stack-wide secret can steer a webhook at any tenant's mappings.
   - Related fail-open: `jira-webhook.ts:152` skips the replay-window check entirely when `timestamp` is absent (Linear rejects a missing timestamp), and timestamp-less deliveries collapse to a single `…#unknown` dedup key.

   Suggested fix: don't mirror the stack-wide secret into per-tenant bundles; after a stack-wide-fallback verification, refuse any `cloudId` other than the sole-tenant fallback result; and log (or reject) when the replay check is skipped so the bypass is observable. Given the project's fail-closed tenet, I'd treat this as blocking.

4. **ADR-014 and several docs/docstrings still describe the abandoned MCP-outbound design.** The final commit correctly pivoted outbound to REST (`jira_reactions.py`) — and `agent/src/prompt_builder.py:135-146` documents the reality perfectly — but the following still claim MCP outbound: `docs/decisions/ADR-014-jira-integration.md` (still says "No `jira_reactions.py` REST module", status `proposed`), `JIRA_SETUP_GUIDE.md:31-40,163-168`, `USER_GUIDE.md:14`, the new `ROADMAP.md` entry, `agent/src/channel_mcp.py:55-96`, `agent/src/config.py:333-344`, `jira-webhook-processor.ts:172`, and `jira-integration.ts:70-76`. Additionally, `channel_mcp.py` still writes a Jira MCP entry that the PR's own docstrings say cannot connect from a headless agent (and emits the SSE URL with `"type": "http"`) — please either drop the `"jira"` entry from `CHANNEL_MCP_BUILDERS` or gate it with an in-band log explaining it's expected to fail. The ADR should also be updated for the actual dedup key (`{issueKey}#{webhookEvent}#{timestamp}` — the ADR says timestamp-only) and, since #296 already merged an ADR-014, renumbered to **ADR-015**.

5. **User-facing instructions reference a non-existent CLI command.** The unmapped-project feedback comment (`jira-webhook-processor.ts:247`) tells admins to run `bgagent jira onboard-project …` — the real command is `map` and requires a `<cloud-id>` positional. The CLI's own "Next steps" hint (`cli/src/commands/jira.ts:555`) also omits the required `<cloud-id>`. Both printed commands fail verbatim.

6. **Dead orchestrator wiring with surplus IAM.** `cdk/src/stacks/agent.ts:878-898` grants the orchestrator registry read + `GetSecretValue`/`PutSecretValue` on `bgagent-jira-oauth-*` "so the concurrency-cap rejection path can post a Jira comment" — but `orchestrate-task.ts` only implements `notifyLinearOnConcurrencyCap`; there is no Jira equivalent. Please either implement `notifyJiraOnConcurrencyCap` (parity) or remove the grant until the feature lands — unused write-capable IAM on every tenant's credentials is worth avoiding.

## Test coverage gaps (would love to see these here, given the surfaces they guard)

- **No CLI Jira tests** (~993 lines across `cli/src/jira-oauth.ts` + `commands/jira.ts`) — Linear has both `linear-oauth.test.ts` and command tests. The token exchange/refresh-rotation and the missing-`refresh_token` branch are the highest-consequence untested code in the PR.
- **No multi-tenant signature test** — Linear has a 9-case `linear-webhook-multi-workspace.test.ts` covering exactly the `mismatch`/`revoked`/no-downgrade distinctions that matter for blocking issue 3.
- `jira-feedback.ts` is mocked in every test that touches it — the ADF body shape (which Jira REST v3 will 400 on if wrong), timeout, and non-2xx paths have zero direct coverage. Linear ships `linear-feedback.test.ts`.
- No construct test for `jira-integration.ts` — the only construct-level coverage is the table-count bump in `agent.test.ts`; Linear's construct test asserts key schemas, the dedup TTL attribute, and env wiring.
- Suggest parameterizing `cdk/test/contracts/stored-oauth-token-parity.test.ts` to also cover the Jira `StoredOauthToken` ↔ `StoredJiraOauthToken` pair — currently that cross-language contract is prose-only.

## Non-blocking suggestions

- `jira-link.ts:96-112`: the Put (active mapping) + Delete (pending row) aren't atomic — consider `TransactWriteItems`, or return 200 with a WARN if the delete fails after a successful Put (today the user is told linking failed when it actually succeeded).
- `USER_GUIDE.md:7`: "There are five ways to interact" — the list now has six.
- Consider keeping `client_secret` out of the per-tenant bundle the agent role can read — a prompt-injected agent could exfiltrate `client_secret` + `refresh_token` and mint tokens outside AWS. Splitting client credentials into a separate secret would harden this.
- `resolveSoleTenantCloudId` (`jira-webhook-processor.ts:89-109`): the `ScanCommand` is unguarded and the processor has no top-level catch/DLQ (parity with Linear, so platform-wide rather than this PR's regression) — a try/catch + a metric on fallback usage would make silent drops countable.
- Tiny key-builder helpers for the composite keys (`{cloudId}#{projectKey}`, `{cloudId}#{accountId}`, `pending#{code}`) would single-source formats currently interpolated independently at write and read sites (`jira-webhook-processor.ts:234,585`, `jira-link.ts:69,99,111`) — the existing `jiraOauthSecretName()` is the precedent.
- `jira-webhook-processor.ts:123`: the `| string` in the `webhookEvent` union collapses the literal narrowing; and the `cloudId` doc comment at `jira-webhook.ts:61` sits above `matchedWebhookIds` rather than the field it describes.
- The 24h replay freshness window is generous; ~1h matches Atlassian's actual retry behavior.

## What's notably good

- `jira-oauth-resolver.test.ts` (37 cases) is a model suite — `invalid_grant` with rotated vs. same refresh token, concurrent-refresh recovery that skips the second POST, non-fatal `PutSecretValue` persistence failure.
- The strict (throwing) registry/secret lookups on the verification path, so a transient infra error can't silently downgrade a per-tenant-secured tenant to the stack-wide fallback — exactly the right fail-closed choice.
- Dedup rollback on processor-invoke failure (`jira-webhook.ts:200-227`), including the don't-mask-the-original-error nested case.
- The advisory comment swallowing in `jira_reactions.py` is logged, circuit-broken, and faithfully mirrors `linear_reactions.py`.
- `prompt_builder.py:135-146` is the single source of truth on the REST pivot — every other doc just needs to be brought in line with it.

Happy to re-review quickly once the CI fix + mirror sync land; the security binding (issue 3) and the ADR reconciliation (issue 4) are the two I'd most like to see addressed before merge.

ayushtr-aws:
## Code Review — Jira Cloud Integration

Reviewed the security-critical paths (webhook HMAC verification, OAuth resolution/refresh, task creation, agent runtime). Overall this is solid, production-minded work that faithfully mirrors the Linear adapter with well-documented divergences. Below are the findings.

### Significant issues

**1. Agent-side token refresh can permanently break a tenant (Atlassian rotating refresh tokens)**
`agent/src/config.py::resolve_jira_oauth_token` refreshes an expiring token in-memory but, by design, cannot persist it (the agent role has `GetSecretValue` only). The TS resolver itself documents that *"Atlassian rotates refresh_tokens on every use"* (`jira-oauth-resolver.ts`). So when the agent path actually fires a refresh:
- It consumes the stored `refresh_token`; Atlassian invalidates it.
- The new `refresh_token` lives only in the agent's memory and is discarded at task end.
- The Secrets Manager copy now holds a **dead** refresh token → the next Lambda/agent resolve gets `invalid_grant` → tenant requires re-onboarding.

This differs from Linear, whose copied trust model assumes the stored refresh token survives. It only triggers if the agent catches a token within 60s of expiry before any Lambda refreshes it, but when it hits it silently breaks the tenant. Consider having the agent **never** refresh — fail closed and let the orchestrator/Lambda path (which has `PutSecretValue`) own all refreshes. Worth confirming against Atlassian's rotating-refresh-token semantics before merge.

**2. Image-attachment extraction is likely dead code for real Jira issues**
`jira-webhook-processor.ts::extractImageUrlAttachments` scans the *converted markdown* for `` syntax. But the ADF walker (`walkAdf`) has no `media`/`mediaSingle` case — Jira embeds images as media nodes with `attrs` (id/url), never as markdown image text. The default branch descends into `content`, and media nodes carry no `text`, so no image URL is ever emitted. Net effect: the regex matches almost nothing for typical Jira issues. Low severity (documented as a minimal converter), but it gives a false impression of working. Either add a `media` case or drop the extraction until it's real.

### Minor issues

**3. ADF→markdown conversion runs twice per issue.** `buildTaskDescription` calls `extractDescriptionMarkdown`, then the attachments line calls it again on the same `issue.fields?.description`. Compute once and reuse.

**4. Multi-tenant security of the stack-wide fallback secret.** When a `cloudId` isn't in the registry (or its bundle lacks `webhook_signing_secret`), verification falls back to the single stack-wide secret, which is **not bound to any cloudId**. In a multi-tenant install, anyone holding the stack-wide secret can forge events with an arbitrary `cloudId`. This is documented as single-tenant back-compat, but consider explicitly disabling the stack-wide fallback when more than one active tenant is registered (mirroring the `resolveSoleTenantCloudId` safety logic).

**5. `resolveSoleTenantCloudId` does a full table `Scan`** on every UI-created (cloudId-less) webhook. Fine at current scale and the `>1 active tenant` short-circuit helps; worth a comment noting the table is expected to stay small, or a GSI on `status` if it could grow.

**6. `isWebhookTimestampFresh` uses `Math.abs`**, so far-future timestamps (clock skew or crafted) pass the freshness check. Post-signature it's only advisory, so low risk — but a one-sided check (`Date.now() - timestamp <= MAX_AGE`) is more correct.

**7. HMAC over `event.body` and `isBase64Encoded`.** If API Gateway ever delivers the webhook base64-encoded (binary media settings), both the JSON parse and HMAC will fail. Likely fine given JSON content-type and parity with the Linear template, but worth a guard or an explicit assumption comment.

### Done well
- **HMAC verification is correct**: raw-body HMAC-SHA256, `sha256=` prefix stripped, `timingSafeEqual` with length-mismatch caught.
- **Strict vs non-strict registry/secret lookups** prevent a transient DDB/SM throttle from silently downgrading per-tenant verification to stack-wide — a thoughtful trust-boundary detail.
- **Dedup happens after verification** via conditional `PutItem`, with rollback on processor-invoke failure so retries aren't permanently swallowed.
- **Fail-closed parsing**: `parseRegistryRow` treats unknown/missing status as `revoked`; `parseOauthSecret` validates required fields.
- **Per-tenant `mismatch`/`revoked` are fatal** (no fallback); only `no-per-tenant-secret` falls through — exactly the right asymmetry.
- **Secret redaction** extended to `JIRA_API_TOKEN`; reactions/feedback are truly best-effort with an auth circuit breaker and never gate the pipeline.

### Test coverage
Strong — OAuth refresh state machine (concurrent-refresh race, invalid_grant, malformed expiry, network failure), circuit breaker, channel gating, dry-run-must-not-write, and case-sensitive link codes are all covered. Suggested additions: a test asserting the **agent-side** refresh's non-persistence behavior (issue # 1), and a processor test against a realistic ADF media node (would surface the dead-code behavior in # 2).

### Recommendation
I'd treat **# 1 (agent refresh burning the rotating refresh token)** as a blocker pending confirmation of Atlassian's token semantics, and **# 2** as should-fix-or-remove. The rest are fine as follow-ups.

<sub>🤖 Generated with Claude Code</sub>

mayakost:
Thanks both — this was an excellent, security-focused review. Pushed `2dcfb3c` addressing it. Point-by-point below.

## @krokoko (blocking)

1. **CI red (missing comma).** Fixed in `1e4a4a0`; the agent suite collects and runs again (including `TestResolveJiraOauthToken*` and `test_jira_reactions.py`). CI is green.
2. **Stale Starlight mirrors.** Re-synced in `ae7e70f` and again here; `node docs/scripts/sync-starlight.mjs` now produces no diff.
3. **Multi-tenant signature binding.** Fixed end-to-end:
   - The receiver now passes `verified_via_stack_wide` to the processor. On a stack-wide-verified delivery the processor **ignores the body `cloudId`** and binds to the sole active tenant (dropping when zero/multiple are active), so the stack-wide secret can no longer steer a webhook at an arbitrary tenant's mappings.
   - `bgagent jira setup` no longer mirrors the stack-wide secret into new per-tenant bundles — each tenant stores its own operator-chosen secret, and the stack-wide fallback is seeded **once** from the first tenant for single-tenant back-compat.
   - A missing `timestamp` no longer silently skips the replay check — it's logged (`replay-window check skipped`), and the window is tightened 24h → 1h.
4. **ADR/docs still describing MCP outbound.** Renumbered **ADR-014 → ADR-015** (the #296 collision) and rewrote it to `accepted` with the implemented REST-outbound design, the actual dedup key (`{issueKey}#{webhookEvent}#{timestamp}`), and new *Multi-tenant signature binding* + *Token refresh ownership* sections. Reconciled `JIRA_SETUP_GUIDE.md`, `USER_GUIDE.md` ("six ways"), `ROADMAP.md`, `channel_mcp.py`, `config.py`, `jira-webhook-processor.ts`, and `jira-integration.ts`. The `jira-server` MCP entry is kept as an explicitly-labelled non-functional placeholder and now emits an in-band log saying it's expected to fail and that outbound goes via `jira_reactions.py`.
5. **Non-existent CLI command.** Both the processor feedback and the CLI "Next steps" hint now print `bgagent jira map <cloud-id> <project-key> --repo …` (the real command, with the required positionals).
6. **Dead orchestrator IAM grant.** Implemented `notifyJiraOnConcurrencyCap` (Linear parity) so the grant is used and Jira users get feedback on the concurrency cap instead of a silent drop; the grant comment now explains why `PutSecretValue` is needed (the orchestrator is the trusted refresh path).

## @ayushtr-aws (significant)

1. **Agent burning the rotating refresh token — confirmed and fixed by removing agent-side refresh entirely.** You're right about Atlassian's rotation semantics. The agent has `GetSecretValue` only, so a refresh would consume the stored `refresh_token`, keep the replacement in memory for one task, and leave Secrets Manager holding a dead token. The agent now uses the Lambda-written token verbatim and **fails closed** (skips the advisory comment) when it's expiring; the trusted Lambda path owns all refreshes. The old refresh-path tests are replaced by `TestResolveJiraOauthTokenNoRefresh`, which asserts no `/oauth/token` POST is ever made.
2. **ADF image extraction was dead code — fixed.** Added `media`/`mediaSingle`/`mediaGroup` cases to the ADF walker; `external` media now render to `` (so extraction works), while `file`-type attachment media (needing a Jira API round-trip) are skipped. New processor test covers a realistic media node. Also fixed the **double ADF→markdown conversion** (#3) — it's computed once and reused.

**Minor:** timestamp freshness is now one-sided with a clock-skew allowance (#6); a base64-encoded body is rejected before verification (#7); `resolveSoleTenantCloudId`'s Scan has a "table stays small / add a GSI if it grows" note (#5).

## Deferred (happy to do in a follow-up)
The broader Linear-parity **test** suites — full CLI `jira-oauth`/command tests, direct `jira-feedback.ts` tests, a `jira-integration` construct test, and parameterizing `stored-oauth-token-parity` for the Jira pair — aren't in this push. I added the security-critical tests tied to the blockers (multi-tenant binding, agent no-refresh, base64/timestamp, ADF media, concurrency-cap parity) and would rather land the parity-coverage sweep separately so this review's fixes can merge. Also left as follow-ups: the `jira-link.ts` Put+Delete atomicity and splitting `client_secret` out of the agent-readable bundle.

Verification: agent 1007 tests green; CDK Jira handler suites + synth suites pass under the pinned Node (22); `check:types-sync` and ESLint clean; docs mirrors in sync.

isadeks:
**Verdict: One blocker, then merge-ready.** The inbound pipeline, multi-tenant routing, token model, IAM scoping, and OAuth dance are all well-engineered and correctly internalize the accumulated Linear-integration lessons. A single correctness bug breaks the Lambda-side user-feedback path; once fixed, this is a strong PR.

Reviewed via the structured `/code-review` methodology (xhigh: 7 finder angles × 6 candidates → recall-biased verify) **and** a manual line-by-line pass of every Jira source file. The structured pass independently reproduced the headline blocker — strong corroboration. 67 agent tests pass.

> **Scope note:** the `gh pr diff` merge-base predates the already-merged #248/#296 workflow refactor, so it shows ~50 files; the *actual* new surface (two-dot diff vs `upstream/main`) is the Jira files. The non-Jira diff (server.py/config.py workflow plumbing) is already in `main` and won't double-apply. I did not execute the deployed Lambdas/webhooks live — findings are from static analysis + tests.

---

## 🔴 Blocker

### 1. Lambda-side Jira feedback uses the wrong REST base — every comment 401s silently
**`cdk/src/handlers/shared/jira-feedback.ts:70`**

```ts
const base = siteUrl.replace(/\/+$/, '');                 // → https://acme.atlassian.net
const url = `${base}/rest/api/3/issue/${encodeURIComponent(issueIdOrKey)}/comment`;
```

The OAuth token is minted with `audience=api.atlassian.com` (`cli/src/jira-oauth.ts:180`), which is valid **only** against the Atlassian gateway base `https://api.atlassian.com/ex/jira/{cloudId}/rest/...`. The **agent** path gets this right and documents why:

```python
# agent/src/jira_reactions.py:42,111
JIRA_API_BASE = "https://api.atlassian.com/ex/jira"
url = f"{JIRA_API_BASE}/{cloud_id}/rest/api/3/issue/{issue_key}/comment"   # ✅
```

…but the Lambda path posts to the raw `*.atlassian.net` site host → **401**.

**Blast radius** — the entire Lambda-side outbound feedback surface, both callers:
- `jira-webhook-processor.ts:59` (`safeReportIssueFailure`) — every pre-container rejection: project not onboarded, user not linked, missing `accountId`, `createTaskCore` non-201.
- `orchestrate-task.ts` (`notifyJiraOnConcurrencyCap`) — concurrency-cap rejections.

Errors are swallowed by design (advisory comments must never gate the pipeline), so the user sees **silence** instead of the "here's why nothing happened" message. This is the valuable Linear v1.1 pre-container-feedback polish — built here but **dead on arrival**.

**Why it's an original bug, not template drift:** Linear uses a fixed `api.linear.app/graphql` URL with no per-tenant base, so there was nothing to copy wrong; the site-URL base was introduced fresh on the one path that diverges from the template.

**Fix (~3 lines):** `JiraFeedbackContext` already carries `cloudId` — build `https://api.atlassian.com/ex/jira/${cloudId}/rest/api/3/issue/${encodeURIComponent(issueIdOrKey)}/comment` and drop the `siteUrl` base. Add a test asserting the request host is `api.atlassian.com`. Also fix the misleading `jira-workspace-registry-table.ts:54` comment (`site_url … (display + REST base)` — it is **not** a valid REST base for a 3LO token). Why tests missed it: `jira-feedback` has no dedicated test, and processor tests mock the resolver/`fetch`, so the wrong host sails through (textbook "mock one layer above the wiring bug").

---

## 🟡 Non-blocking (quality / efficiency)

### 2. Dead Jira MCP entry written on every task hot path
**`agent/src/channel_mcp.py:95`** — `CHANNEL_MCP_BUILDERS` registers a `jira` entry whose own docstring states it **cannot connect** from a headless agent (interactive OAuth 2.1 only); outbound is REST via `jira_reactions.py`. Every Jira task writes it into `.mcp.json` (`pipeline.py:815`), pays an SDK connection-failure cost at session start, and logs a spurious "Failed to connect" line that mis-leads operators. **Altitude:** a known-broken special case on shared MCP infra — cleaner to not register the entry until Atlassian ships a token-compatible MCP. (Intentional & documented, so not a bug — but worth reconsidering.)

### 3. Setup writes the OAuth secret twice; partial failure strands an active tenant
**`cli/src/commands/jira.ts:530`** — `upsertOauthSecret` is called at line 479 (token) and again at 530 (re-write with `webhook_signing_secret` merged), with `status:'active'` written to the registry in between (487-496). If the operator aborts at the signing-secret prompt or the second write fails, the tenant is **active** but lacks its per-tenant signing secret → `verifyJiraRequestForTenant` returns `'no-per-tenant-secret'` → falls back to stack-wide. For a 2nd+ tenant (stack-wide is pinned to tenant #1), its webhooks fail verification until setup is re-run. **Fix:** prompt for the signing secret first, build the merged object once, write the registry row last.

### 4. Wasted hot-path I/O before the trigger filter
**`cdk/src/handlers/jira-webhook-processor.ts:232`** — `resolveSoleTenantCloudId()` (full-table registry **Scan**, lines 232/241) and the project-mapping `GetCommand` (line 269) run **before** the pure `shouldTrigger()` label filter (line 288). On a noisy Jira project where most `issue_updated` events don't touch the trigger label, every such event pays a Scan + GetCommand the filter would have short-circuited. **Fix:** run the cheap `shouldTrigger()` check as early as the payload allows. *(Note: `resolveJiraOauthToken` correctly runs AFTER the filter at line 347 — the earlier "refresh-token churn" concern was refuted on trace.)*

---

## What held up under scrutiny (the good)

- **Multi-tenant routing** — stack-wide-verified deliveries can't steer `cloudId`; per-tenant mismatch/revoked are fatal with no fallback. Stronger than the Linear template.
- **Token model** — agent `GetSecretValue`-only, never refreshes; Lambda owns refresh with documented `invalid_grant` concurrent-rotation retry.
- **IAM / API auth** — `bgagent-jira-oauth-*` prefix scoping; `/jira/webhook` = `NONE` (HMAC-verified), `/jira/link` = Cognito.
- **OAuth dance** — `crypto.randomBytes` PKCE S256, `state` validated, `offline_access` present, JSON token endpoint with `audience`.
- **Webhook layer** — HMAC over raw body, timing-safe compare, one-sided timestamp freshness (avoids the `Math.abs` future-timestamp trap), dedup with invoke-failure rollback so retries aren't swallowed.
- **Pipeline hooks** — start / success-finish / crash-finish all fire and no-op safely for non-Jira and on empty token.

**Bottom line:** fix the `jira-feedback.ts` REST base (one small change + a host-asserting test) and this PR is merge-ready.

mayakost:
@isadeks thanks — great catch on the blocker. Confirmed and fixed in 9dd66b8.

**🔴 #1 (REST base) — fixed.** `jira-feedback.ts` now builds the URL from `cloudId` against the gateway base `https://api.atlassian.com/ex/jira/{cloudId}/rest/api/3/issue/{key}/comment` (matching the agent-side `jira_reactions.py`) and drops the `siteUrl` base entirely. `JiraFeedbackContext` already carried `cloudId`, so no caller changes were needed. Also corrected the misleading `jira-workspace-registry-table.ts` comment that called `site_url` a "REST base" — it's display-only for a 3LO token.

Added the dedicated test you suggested: `jira-feedback.test.ts` asserts the request host is `api.atlassian.com` (and explicitly *not* `atlassian.net`), plus issue-key encoding, no-token→false, non-2xx→false, and never-throws-on-network-error.

**CI verification** (ran the `build.yml` gates locally on the repo's pinned Node 22):
- `//cdk:test` — **115 suites / 2091 tests pass** (full suite, coverage thresholds met)
- `//cdk:compile` — clean
- `//cdk:eslint` — 0 errors, no `--fix` mutations
- `:drift-prevention` — all 4 checks pass
- mutation gate — `git status` clean after build

The three 🟡 non-blocking items (dead MCP entry, setup double-write ordering, hot-path I/O before `shouldTrigger`) are **not** addressed in this commit — happy to take them in a follow-up. And appreciated the refuted "refresh-token churn" note; agreed the resolver correctly runs after the filter.

krokoko:
## Review — Jira Cloud Integration (parity with Linear)

Thanks for this — it's a genuinely high-quality, faithful parity port of the Linear adapter. The trust model, prefix-scoped IAM, OAuth refresh-race handling, and the fail-closed agent token path are all well-reasoned, and the docs (ADR-015, setup guide, roadmap check-off, regenerated Starlight mirrors) are in great shape. Governance is clean too: #288 is `approved`, assigned, and the branch name follows convention.

I'm landing on **Request changes** for three small-but-important items. None require redesign.

### Blocking

**B1 — Fail-open HMAC: empty/whitespace signing secret is forgeable** — `cdk/src/handlers/shared/jira-verify.ts:108-128`
Linear, Slack, and GitHub all route every HMAC secret through `isUsableHmacSecret()` in **two** places (on fetch and inside the verify fn), because `HMAC('', body)` / `HMAC(' ', body)` is computable by anyone. The Jira copy imports it nowhere. `verifyJiraRequestForTenant` only checks `!stored.webhook_signing_secret` (`:202`), so a whitespace-only per-tenant secret — operator-pasted free text during `bgagent jira setup` — is truthy and flows into `verifyJiraSignature`, making every signature forgeable for that tenant. The stack-wide path has the same gap. This is a fail-closed→fail-open regression vs. the template.
_Fix (~2 lines):_ `import { isUsableHmacSecret } from './hmac-secret';`, add `if (!isUsableHmacSecret(webhookSecret)) return false;` as the first line of `verifyJiraSignature`, and swap `getJiraSecret:74`'s `!result.SecretString` for `!isUsableHmacSecret(result.SecretString)`.

**B2 — Multi-tenant verification boundary has no test coverage** — `cdk/src/handlers/shared/jira-verify.ts:188-208`
`verifyJiraRequestForTenant`'s four outcomes (`verified` / `mismatch` / `revoked` / `no-per-tenant-secret`) are the whole multi-tenant security model, and none of the four branches is exercised — there's no `jira-verify.test.ts` (Linear has `linear-verify.test.ts`, 21 tests), and `jira-webhook.test.ts` runs with `JIRA_WORKSPACE_REGISTRY_TABLE_NAME` unset, so the per-tenant block (`jira-webhook.ts:133-154`) is dead code in tests. A regression letting `mismatch` fall through to stack-wide would pass CI today. Stale/far-future timestamp rejections are also untested.
_Fix:_ add `jira-verify.test.ts` mirroring `linear-verify.test.ts`, plus stale/future-timestamp cases in `jira-webhook.test.ts`.

**B3 — Registry-table doc describes an architecture that was never built** — `cdk/src/constructs/jira-workspace-registry-table.ts:47-70` (echoed in `jira-integration.ts:89,104-108`)
The docstring documents a `provider_name` field and an AgentCore Identity flow (`call AgentCore Identity with userId='jira-tenant-<cloudId>'`). The actual `RegistryRow` has no `provider_name` — it carries `oauth_secret_arn` and the resolver reads directly from Secrets Manager. Leftover from the abandoned "Phase 2.0a" design. On a security-critical table a maintainer will hunt for a field/path that don't exist.
_Fix:_ doc-only — replace the `provider_name` bullet with `oauth_secret_arn` (Secrets Manager) and rewrite the closing paragraph to point at `jira-oauth-resolver.ts`.

### Non-blocking nits
- Stale "MCP token" comment `jira-webhook-processor.ts:345-348` contradicts the REST-outbound decision (the handler docstring at `:190` is correct) — reword to "no resolvable Jira OAuth token."
- `nosemgrep` masking annotations dropped vs Linear in `jira-verify.ts`, `jira-oauth-resolver.ts` (3 sites), `jira_reactions.py` (config.py correctly has 5). Build is currently **green**, so this is a parity inconsistency, not empirically CI-blocking — but worth matching the convention.
- Token-refresh `PutSecretValue` failure is swallowed (`jira-oauth-resolver.ts:480-490`): the rotated token is cached + returned but never persisted, so the next cold Lambda gets `invalid_grant` → tenant needs re-onboarding. Don't cache `next` on persist failure; escalate the log.
- `cloudId` isn't URL-encoded/validated (`jira-feedback.ts:78`, `jira_reactions.py:111`) where `issueKey` is — low risk (registry-sourced), but validate UUID format at `parseRegistryRow` + encode the segment.
- `memorySize: 512` inlined (`jira-integration.ts:200`) — Linear extracts `WEBHOOK_PROCESSOR_MEMORY_MB`; the sibling timeout constant was kept (AI007).
- `map` command `<cloud-id>` arg unvalidated (`cli/src/commands/jira.ts`) — flows into a secret name + DDB key.
- Consider tightening `TaskConfig.channel_source` to a `Literal[...]` and keying `CHANNEL_MCP_BUILDERS` off it; and adding a Jira `StoredOauthToken` parity fixture (Linear has one) so CDK/CLI/agent field drift fails at test time, not runtime.
- `lookupPlatformUser` excludes only `status === 'pending'` (`jira-webhook-processor.ts:657`) — fail-open to other statuses, but **identical to the Linear baseline**, so parity not regression; an `=== 'active'` allowlist in both would be a nice shared follow-up.

### Notes
- **Bootstrap synth-coverage: N/A** — no new CFN resource types beyond what `LinearIntegration` already established; `cdk/src/bootstrap/` correctly untouched.
- **Types sync verified** — `ChannelSource` in lockstep on both sides; `JiraLinkResponse` correctly allowlisted in `check-types-sync.ts`.
- Reviewed with all six pr-review-toolkit/security agents (code-review, silent-failure, security, type-design, comment-accuracy, test-coverage) plus a principal-architect pass; B1 was independently corroborated by two of them.

Fix B1 (2 lines), B2 (port the Linear test suite), and B3 (doc edit) and this is an approve. Really nice work overall. 🙏

krokoko (approved):

ayushtr-aws (approved):

krokoko:
## Review fixes pushed — commit `2dd9743`

I went ahead and implemented the review items directly on this branch. All three blocking items plus the nits are addressed; `mise run build` is green (**2177 CDK + 355 CLI + 1090 agent** tests passing, eslint/synth/docs-sync/masking-semgrep all clean).

### Blocking
- **B1 — fail-open HMAC fixed** (`jira-verify.ts`): the signing secret now routes through `isUsableHmacSecret()` in **both** `getJiraSecret` (on fetch) and `verifyJiraSignature` (defense-in-depth), matching the Linear/Slack/GitHub invariant. A whitespace-only per-tenant secret can no longer make `HMAC(' ', body)` forgeable.
- **B2 — added `cdk/test/handlers/shared/jira-verify.test.ts`** (30 cases): covers `verifyJiraRequestForTenant`'s four outcomes (`verified`/`mismatch`/`revoked`/`no-per-tenant-secret`), the strict-lookup rethrow (so a transient DDB/SM error can't silently downgrade to stack-wide), `verifyJiraRequest` rotation-refetch, empty/whitespace-secret rejection, and the one-sided timestamp window (stale / far-future / within-skew). `jira-verify.ts` goes from no coverage on this path to 100% stmt/func.
- **B3 — registry-table docs corrected** (`jira-workspace-registry-table.ts`, `jira-integration.ts`): the docstring now describes the real `oauth_secret_arn` → Secrets Manager resolution path instead of the never-implemented `provider_name` / AgentCore Identity model.

### Nits
- Reworded the stale "MCP token" comment in `jira-webhook-processor.ts` to the REST-outbound reality.
- Added `ts-`/`py-silent-success-masking` nosemgrep justifications to the best-effort swallows in `jira-verify.ts`, `jira-oauth-resolver.ts` (×3), and `jira_reactions.py`, matching the Linear annotations.
- On a refresh `PutSecretValue` failure, no longer cache the rotated token (SM still holds the now-dead refresh_token) — invalidate the cache and escalate the log so the breakage surfaces promptly instead of after the cache TTL.
- URL-encoded the `cloudId` path segment in `jira-feedback.ts` and both `cloud_id` + `issue_key` in `jira_reactions.py` (defense-in-depth; `issueKey` was already encoded on the TS side).
- Re-extracted the inlined `512` into `WEBHOOK_PROCESSOR_MEMORY_MB` (AI007).

### Deliberately deferred
The `lookupPlatformUser` `status === 'pending'`-only check (potential fail-open to other statuses) is **identical to the Linear baseline**, so I left it rather than introduce Jira-only asymmetry — better as a shared follow-up across both adapters that flips to an `=== 'active'` allowlist.

PR #307, opened 2026-06-10
feat(agent-runtime): honor per-step gate on the coding lane (#301)

### LINKED ISSUE #301: feat(agent-runtime): honor per-step `gate` on the coding lane (declarative steps decorative for verify_build/ensure_pr)
## Summary

On the **coding lane**, a workflow's per-step `gate` field is declared in the YAML but never read at runtime. The coding path runs only the agentic step through the workflow runner (`run_workflow(wf, ctx, only_kinds={"run_agent"})`, `agent/src/pipeline.py:191`); `verify_build` / `verify_lint` / `ensure_pr` then run **inline** in the post-hook block, using hardcoded gate logic rather than each step's declared `gate`. So `steps[].gate` is **live on the knowledge (repo-less) lane** — where the full step list runs through the runner and `_gate_status` (`agent/src/workflow/runner.py:486`) honors it — but **decorative on the coding lane**.

This was flagged during PR #296 review by both reviewers (@isadeks and @theagenticguy) and explicitly agreed as a **non-blocking follow-up**, since it is a larger half-migrated-runner change. Opening it so "declarative steps" is not read as fully wired on the coding lane.

## Why it's latent today (not a live bug)

The inline `verify_build` in `pipeline.py` happens to implement the same semantics an unset/`regression_only` gate would (`_gate_status` documents that the legacy behavior is regression-only for every task), and `pr-review-v1` is `read_only:true` where `_gate_status` never gates regardless. So the three shipped coding workflows behave correctly **by coincidence of alignment**, not because the inline path consults `gate`.

## Why it matters

The risk surfaces the moment a coding workflow author writes a gate expecting it to take effect:

- A `verify_build` step with `gate: strict` on a coding workflow will **not** gate — a build regression that should fail the task would still open the PR.
- Conversely, an author cannot make a coding-lane build `informational` via the YAML; the inline path ignores it.

The declarative contract silently does not hold on the lane where it matters most.

## Current references (at PR #296 HEAD `9eacdd8`)

- `agent/src/pipeline.py:191` — `run_workflow(wf, ctx, only_kinds={"run_agent"})` restricts the runner to the agent step on the coding path.
- `agent/src/pipeline.py` post-hook block (~`855`–`905`) — inline `verify_build` / `verify_lint` / `ensure_pr`; reads the `ensure_pr` *strategy* from the workflow but not the `verify_*` steps' `gate`.
- `agent/src/workflow/runner.py:486` — `_gate_status(...)`, the single place gate semantics live; exercised only on the repo-less path.
- `agent/workflows/coding/new-task-v1.yaml:45`, `pr-iteration-v1.yaml:42` — declare `gate: regression_only`; `pr-review-v1.yaml:46` declares `gate: informational` — none currently consulted on the coding lane.

## Acceptance criteria

- [ ] The coding lane's `verify_build` / `verify_lint` honor each step's declared `gate` (`informational` | `regression_only` | `strict`) via the same `_gate_status` path the repo-less lane uses — one implementation, not two.
- [ ] A coding workflow declaring `gate: strict` fails the task on a build regression (and `informational` never gates), proven by a test.
- [ ] No behavior change for the three shipped coding workflows (their current effective gating is preserved) — locked by a test asserting the inline-vs-runner verdicts match for `regression_only` and `read_only` cases.
- [ ] Either route the coding post-hooks through the runner's step handlers, or have the inline path consult `step.gate` — whichever keeps gate semantics in one place. Document the choice.

## Out of scope

- The broader half-migrated-runner unification (driving the *entire* coding step list through `run_workflow` rather than `only_kinds={"run_agent"}` + inline post-hooks) is a larger refactor; this issue is scoped to making `gate` honored, not to fully collapsing the two paths. Note the relationship to that future work.

## References

- ADR-014 — workflow-driven tasks (`docs/decisions/ADR-014-workflow-driven-tasks.md`)
- `docs/design/WORKFLOWS.md` §"Step kinds" (the `gate` field) and §"The agent-side step runner"
- PR #296 — where this was flagged and deferred.

krokoko:
## What this does (plain English)

After a coding agent finishes its work, the platform runs checks (like "does the build still pass?") before opening a pull request. Workflow authors can declare how strict each check is in the workflow YAML:

- **strict** — any failure fails the task.
- **regression_only** — fails only if the agent *broke* something that was working before.
- **informational** — reported, but never fails the task.

**The bug:** on coding tasks, that setting was silently ignored — the pipeline used hardcoded regression-only logic instead. You could write `gate: strict` and a build the agent broke would still pass. The three built-in workflows behaved correctly only by coincidence (their declared gates happened to match the hardcoded behavior).

**The fix:** the coding path now actually reads and applies the declared gate, using the *same single piece of logic* the repo-less lane already used — one definition of gate semantics, not two copies that can drift. Even when a check fails the task, the PR is still opened so the agent's work stays visible and reviewable, just marked failed. Tests prove `strict`/`informational` now work as declared and that the three shipped workflows behave exactly as before.

---

## Summary

Fixes #301. On the coding lane, `steps[].gate` was declarative-only: the post-hook block in `pipeline.run_task` hardcoded regression-only build gating, ignoring each `verify_*` step's declared `gate`. This PR makes the coding lane honor the declared gate through the **same implementation the repo-less lane uses** — the runner's `gate_status` (formerly `_gate_status`, now public) — so gate semantics live in exactly one place.

**Decision (documented in `_apply_post_hook_gates` and WORKFLOWS.md §"Step kinds"):** the inline path *consults* `step.gate` via `gate_status` rather than routing the post-hooks through the runner's step handlers. Routing through the runner would change failure-path side effects (a gating `verify_build` with `on_failure: fail` stops the runner *before* `ensure_pr`, stranding committed work with no PR) — that is the broader half-migrated-runner unification #301 explicitly defers. The inline ordering (verify → `ensure_pr` always runs) is preserved; only the task verdict honors the gate.

### Changes

- `agent/src/workflow/runner.py` — `_gate_status` → `gate_status` (public; exported from `workflow`).
- `agent/src/pipeline.py` — new `_apply_post_hook_gates(workflow, ...)` replaces the hardcoded `build_ok` block:
  - declared `verify_build`/`verify_lint` gate per their `gate` (`strict` | `regression_only` | `informational`; unset = `regression_only`), only when `on_failure: fail` (advisory `continue`/`skip_remaining` never gate, matching the runner);
  - undeclared `verify_build` keeps legacy regression-only gating; undeclared `verify_lint` never gates (legacy);
  - `read_only` workflows never gate (legacy);
  - workflow reload failure falls back to legacy gating (work is never stranded — preserves PR #296 finding #5 behavior).
- `agent/tests/test_pipeline_post_hook_gates.py` — 29 new tests.
- `docs/design/WORKFLOWS.md` — documents the `gate` field, the one-implementation rule, and the deferred runner unification (+ synced Starlight mirror).

### Acceptance criteria

- [x] Coding lane `verify_build`/`verify_lint` honor declared `gate` via the same `gate_status` path the repo-less lane uses — one implementation, not two.
- [x] `gate: strict` fails the task on a build regression (and even on a pre-existing failure); `informational` never gates — proven end-to-end through `run_task` (`TestRunTaskHonorsGate`).
- [x] No behavior change for the three shipped coding workflows — `TestShippedWorkflowParity` asserts inline == runner == legacy verdicts across the full `(build_passed, build_before)` matrix for `regression_only` and `read_only` cases.
- [x] Choice (consult `step.gate` inline vs route through runner) documented in code and WORKFLOWS.md.

## Test plan

- [x] `mise //agent:quality` — 1004 passed (29 new), ruff clean, coverage 76.05%
- [x] `mise run build` — full monorepo build green (CDK 1844 tests, CLI, docs sync + build)

### DIFF
--- a/agent/src/pipeline.py
+++ b/agent/src/pipeline.py
@@ -8,6 +8,7 @@ import os
 import subprocess
 import sys
 import time
+from typing import TYPE_CHECKING
 
 from pydantic import ValidationError
 
@@ -37,6 +38,9 @@ from telemetry import (
     print_metrics,
     upload_trace_to_s3,
 )
+
+if TYPE_CHECKING:
+    from workflow import Workflow
 
 _SDK_NO_RESULT_MESSAGE = (
     "Agent SDK stream ended without a ResultMessage (agent_status=unknown). "
@@ -364,6 +368,84 @@ def _run_repoless_task(
     terminal_status = "COMPLETED" if overall_status == "success" else "FAILED"
     task_state.write_terminal(config.task_id, terminal_status, result_dict)
     return result_dict
+
+
+def _apply_post_hook_gates(
+    workflow: Workflow | None,
+    *,
+    read_only: bool,
+    build_passed: bool,
+    lint_passed: bool,
+    build_before: bool,
+    lint_before: bool,
+) -> bool:
+    """Resolve the coding lane's post-hook verify gates against the workflow (#301).
+
+    Decision (issue #301 acceptance criteria): the inline post-hook path
+    CONSULTS each declared ``verify_build`` / ``verify_lint`` step's ``gate``
+    through the runner's ``gate_status`` — the single place gate semantics live —
+    rather than routing the post-hooks through the runner's step handlers.
+    Routing through the runner would also change failure-path side effects (a
+    gating ``verify_build`` with ``on_failure: fail`` stops the runner *before*
+    ``ensure_pr``, stranding committed work with no PR), which is the broader
+    half-migrated-runner unification the issue defers. Here the inline ordering
+    (verify → ensure_pr always runs) is preserved; only the task verdict honors
+    the declared gate.
+
+    Per-step semantics:
+
+    - A declared step gates per its ``gate`` (``strict`` | ``regression_only`` |
+      ``informational``; unset = ``regression_only``), but only when its
+      ``on_failure`` is ``fail`` — ``continue``/``skip_remaining`` steps are
+      advisory for the task verdict, matching the runner.
+    - An undeclared ``verify_build`` keeps the legacy regression-only gating
+      (identical to ``gate_status`` with ``gate=None``).
+    - An undeclared ``verify_lint`` never gates (legacy: lint is not used for
+      terminal status unless a workflow opts in by declaring the step).
+    - ``workflow is None`` (post-hook reload failed) falls back to the legacy
+      gating for both, so a corrupt file cannot strand the agent's work.
+    """
+    from workflow import gate_status
+
+    steps = list(workflow.steps) if workflow is not None else []
+    gates_ok = True
+    for kind, passed, was_passing_before in (
+        ("verify_build", build_passed, build_before),
+        ("verify_lint", lint_passed, lint_before),
+    ):
+        step = next((s for s in steps if s.kind == kind), None)
+        if step is None:
+            if kind == "verify_lint":
+                continue
+            gate, gating, on_failure = None, True, "fail"
+        else:
+            gate, gating, on_failure = step.gate, step.on_failure == "fail", step.on_failure
+        status = gate_status(
+            passed=passed,
+            gate=gate,
+            read_only=read_only,
+            was_passing_before=was_passing_before,
+        )
+        if passed:
+            continue
+        label = gate or "regression_only"
+        if status == "succeeded":
+            if read_only:
+                log("INFO", f"read-only workflow: {kind} failed — informational only, not gating")
+            elif gate == "informational":
+                log("INFO", f"{kind} failed — gate=informational, not gating")
+            else:
+                log(
+                    "WARN",
+                    f"Post-agent {kind} failed, but it was already failing before "
+                    "agent changes — not counting as regression",
+                )
+        elif gating:
+            log("WARN", f"{kind} failed — gate={label} gates the task")
+            gates_ok = False
+        else:
+            log("INFO", f"{kind} failed — gate={label} but on_failure={on_failure}, not gating")
+    return gates_ok
 
 
 def _resolve_overall_task_status(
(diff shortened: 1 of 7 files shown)


isadeks:
# Code Review: PR #307 — honor per-step `gate` on the coding lane

**Verdict: Approve. No blockers, no nits.**

This is a clean, well-scoped PR. I reviewed the actual diff (7 files, +512/−51) against `main`, read the full enclosing functions, ran independent finder passes, and verified the key claims empirically — including running the test suite — rather than trusting the description.

## What I checked

| Risk area | Verdict |
|---|---|
| **Inline verdict vs. runner verdict parity** | The PR's central claim is that `_apply_post_hook_gates` agrees with the runner's `gate_status` + `on_failure` outcome. Traced all `(gate, on_failure, passed, was_passing_before, read_only)` combos — they agree. ✓ |
| **No behavior change for the 3 shipped workflows** | Confirmed against the real YAMLs (`new-task-v1`/`pr-iteration-v1`: `regression_only`; `pr-review-v1`: `informational` + `read_only`). New code reproduces the old `build_ok = passed or not build_before` exactly across the full `(build_passed, build_before)` matrix. ✓ |
| **Removed-behavior re-establishment** | The deleted hardcoded block's invariants (read-only never gates, regression-only default, undeclared lint never gates, "already failing before" not counted as a regression) are all re-established. ✓ |
| **`workflow=None` fallback** | Reload failure → build keeps legacy regression-only, lint never gates; work is never stranded (preserves #296 finding #5). ✓ |
| **Cross-file breakage** (`_gate_status` → `gate_status` rename) | `grep -rn "_gate_status"` repo-wide: zero leftover references; correctly exported from `workflow/__init__.py`. ✓ |
| **`Workflow \| None` annotation under `TYPE_CHECKING`** | Safe — `pipeline.py` has `from __future__ import annotations`. ✓ |
| **Tests vacuous / wrong patch targets?** | **Ran them: `63 passed`** (29 new + 34 in the runner suite). `TestShippedWorkflowParity` loads the real YAMLs; `TestRunTaskHonorsGate` exercises `run_task` end-to-end and asserts `status == "error"` requires `build_ok == False` — not vacuous. ✓ |

## Quality notes (not findings)

- The "consult `step.gate` inline rather than route post-hooks through the runner" decision is the intentional #301 scope boundary (routing would strand committed work before `ensure_pr`). It's documented in both the code and `WORKFLOWS.md`, with the deferred runner-unification rationale called out. Right altitude.
- Gate semantics genuinely live in one place now (`gate_status`), and the diff respects `on_failure` (advisory `continue`/`skip_remaining` don't gate the verdict), matching the runner. No special-casing drift introduced.
- Minor: `build_ok` keeps its name even though it now also folds in lint gating — slightly under-named, but the function name (`_apply_post_hook_gates`) and call-site comment make intent clear. Not worth changing.

**Ship it.**

isadeks (approved):

Issue #256, opened 2026-06-03
feat(ci): structured CI output — pytest JUnit XML + SARIF from eslint/ruff/semgrep (CA-06)


krokoko:
> **This is a finding from https://github.com/krokoko/cairn** (action item **CA-06**).

### Component

Tooling / CI

### Describe the feature

Emit **structured CI output** and upload it as artifacts + to GitHub code-scanning:

- **pytest** → JUnit XML (`--junitxml=test-reports/agent.xml`) — currently the agent suite is console-only with no structured artifact.
- **ESLint / ruff / semgrep** → **SARIF**, uploaded to GitHub code-scanning (`github/codeql-action/upload-sarif`).

This makes findings **agent-routable** — every failure carries file / line / rule / suggested-fix, the same property that makes the existing fail-on-mutation `repo.patch` artifact an exemplary agent-consumable signal.

### Use case

The feedback loop is at Level 2 (routable) but several lanes are stuck at Level 1: pytest has no JUnit reporter, and eslint/ruff/semgrep emit console text only. Without structured output, an agent (or the fix-and-repush loop) can't reliably parse *what* to fix and *where*. Structured, machine-routable findings are the prerequisite for closing the loop toward Level 3 (auto re-trigger on failure).

### Proposed solution

1. Add `--junitxml=test-reports/agent.xml` to the pytest invocation; upload via `actions/upload-artifact`.
2. Emit SARIF: `eslint -f @microsoft/sarif`, `ruff --output-format sarif`, semgrep `--sarif`.
3. Upload SARIF to GitHub code-scanning with `github/codeql-action/upload-sarif`.
4. Keep the existing JUnit (`jest-junit`) and lcov/cobertura coverage outputs.

### Acceptance criteria

- [ ] pytest emits JUnit XML uploaded as a CI artifact.
- [ ] ESLint, ruff, and semgrep emit SARIF.
- [ ] SARIF is uploaded to GitHub code-scanning and findings appear in the Security tab.
- [ ] No regression to the existing `jest-junit` / coverage artifacts.

### Other information

Source reports: `verification-report.md` (Feedback Loop Completeness), `verification-strategy.md` (Phase 1 #3; Feedback Loop Improvements), `ai-smells-gates-report.md` (Strengthen existing gates #3). Effort: **S**. Per ADR-003 this issue needs the `approved` label before work begins.

PR #311, opened 2026-06-10
feat(security): custom semgrep rules for silent-success masking (AI004)

### LINKED ISSUE #257: feat(ci): custom semgrep rules for silent-success masking (AI004) (CA-09)
> **This is a finding from https://github.com/krokoko/cairn** (action item **CA-09**).

### Component

Tooling / CI

### Describe the feature

Add **custom semgrep rules for silent-success masking** — the dangerous AI004 variant that bare-except rules do not catch. Target patterns where a failure is swallowed and a plausible-but-empty default is returned, e.g.:

- `catch (...) { return [] }` / `return null` / `return {}`
- default-substitution-on-failure (returning a fallback value in a catch/except instead of surfacing the error)

Emit SARIF so findings are agent-routable (file / line / rule / fix).

### Use case

The current gates catch bare-except (`ruff E722`/`B`) and empty catches — the report found 0 empty catches and 0 `except: pass`. But the **more dangerous** fail-fast violation, silently returning a plausible default on failure, has **no detector**. This is exactly the kind of error handling that hides failures from the right actor and lets plausible-but-wrong agent code through (AI004). It compounds with AI001: a swallowed failure plus a fabricated happy path produces a green run that proves nothing.

### Proposed solution

1. Author semgrep rules matching silent-success masking in TS (`catch { return []|null|{} }`) and Python (`except ...: return <empty/default>`).
2. Tune with an allowlist for legitimate fallback sites.
3. Emit SARIF; wire into the security suite output (pairs with CA-06).
4. Start advisory, then make blocking once the baseline is clean.

### Acceptance criteria

- [ ] Custom semgrep rules detect `catch { return []|null|{} }` (TS) and equivalent default-substitution-on-failure (Py).
- [ ] An allowlist mechanism exists for intentional fallbacks.
- [ ] Rules emit SARIF.
- [ ] Existing-code baseline is documented; rule is advisory first, blocking once clean.
- [ ] Remediation text is agent-targeted.

### Other information

Source report: `ai-smells-gates-report.md` (AI004 — "No silent-success-masking detection"; Quick wins #2). Effort: **S**. Per ADR-003 this issue needs the `approved` label before work begins.

krokoko:
Fixes #257

## What this does, in plain English

When code hits an error but quietly returns an empty result instead of reporting the failure (for example `catch { return []; }`), whoever called it — a person, a test, or an AI agent — sees a "successful" empty answer and has no way to know something actually went wrong. Tests stay green, agents conclude their code works, and real failures go unnoticed. This PR adds an automatic detector for that pattern across the repo's TypeScript and Python code, runs it as part of the existing security checks, and reports each occurrence with a clear explanation of how to fix it.

**The gate is now enforced (blocking).** The initial PR landed the rules in advisory mode so 40 pre-existing occurrences could be triaged; the latest commit triages every one of them (justified inline `nosemgrep` allowlists for intentional fail-open / degraded-mode fallbacks), adds `--error` to `security:sast:masking`, and removes `.semgrep/BASELINE.md`. New silent-success masking without an allowlist fails CI and local `mise run security`.

## Summary

Adds the AI004 detector requested in #257: custom semgrep rules that catch **silent-success masking** — a `catch`/`except` that swallows a failure and returns a plausible-but-empty default (`[]`, `{}`, `null`/`undefined`/`None`, `""`), which existing bare-except/empty-catch gates do not detect.

- **`.semgrep/silent-success-masking.yaml`** — `ts-silent-success-masking` and `py-silent-success-masking`, with agent-targeted remediation messages (re-throw / re-raise / encode failure in the result shape; logging alone is not enough). Findings anchor on the offending `return` line. Covers `finally` variants, multi-`except`, and conditional-rethrow fallthrough; returns in the `try` body are not flagged.
- **`.semgrep/silent-success-masking.{ts,py}`** — `semgrep test` fixtures (`ruleid:`/`ok:` annotations), run automatically as part of the task.
- **`mise run security:sast:masking`** — validates rules against fixtures, scans the repo with **`--error` (blocking)**, and emits SARIF to `test-reports/` (gitignored) so findings are agent-routable (pairs with CA-06 / #256). Wired into `mise run security`.
- **Baseline triaged** — all 40 pre-existing findings (15 Py, 25 TS) reviewed and resolved: intentional degraded-mode / fail-open paths carry justified inline `nosemgrep: <rule-id> -- <reason>` on the return line (existing repo convention; precedent in `scripts/check-types-sync.ts`, `agent/src/config.py`). Unallowlisted matches fail the task.
- **Allowlist mechanism** — inline justified `nosemgrep: <rule-id> -- <reason>` on the return line for callers that may safely treat a failure as an empty success.

## Acceptance criteria (from #257)

- [x] Rules detect `catch { return []|null|{} }` (TS) and equivalent default-substitution-on-failure (Py)
- [x] Allowlist mechanism for intentional fallbacks (justified inline `nosemgrep`)
- [x] Rules emit SARIF (`test-reports/semgrep-silent-success-masking.sarif`)
- [x] Existing-code baseline triaged; gate flipped to blocking (`--error`, baseline doc removed)
- [x] Remediation text is agent-targeted

## Test plan

- [x] `semgrep test .semgrep/` — 2/2 rule unit tests pass (fixtures cover rethrow/re-raise, typed errors, meaningful fallbacks, `finally`, multi-except, conditional rethrow, return-in-try-body)
- [x] `mise run security:sast:masking` — exits 0 with `--error`; 0 unallowlisted findings; emits valid SARIF 2.1.0 with both rules
- [x] Existing `security:sast` config set passes on the new `.semgrep/` fixtures (no cross-rule noise)
- [x] `mise run build` — full monorepo build green (CDK tests, agent quality, cli, docs)


scottschreckengaust (approved):
Will there be a follow up to resolve the suppressions?

PR #324, opened 2026-06-12
fix(project): fix multiple issues identified

krokoko:
Fixes multiple findings

1. Auto-approve / Mergify governance gap — .github/workflows/auto-approve.yml now triggers on labeled only (was: every push/synchronize), checks the event's label is literally auto-approve, and skips fork-head PRs. .mergify.yml gained
  a dismiss_reviews rule so any new push invalidates stale approvals; re-approval requires removing and re-adding the label. zizmor reports no findings on the updated workflow.

  2. UTC timezone bug in approval-gate lifetime math — agent/src/hooks.py now parses TASK_STARTED_AT with tzinfo=UTC before calling .timestamp(). Added a TestRemainingMaxlifetime suite (4 tests) in agent/tests/test_hooks.py, including
  one that sets TZ=America/New_York to prove the regression would be caught.

  3. PAT no longer persisted in .git/config — agent/src/repo.py sets the origin remote to the plain https URL and configures credential.helper = !gh auth git-credential, which resolves the token at push time from the GH_TOKEN env var
  the pipeline already exports before setup_repo(). The token never touches disk. Also widened agent/src/output_scanner.py: GitHub token regex now matches 36+ chars and the ghr_ prefix, and a new GITHUB_URL_TOKEN pattern catches
  x-access-token:…@ URLs echoed in tool output (2 new tests).

  4. Linear webhook empty-secret forgery — cdk/src/handlers/shared/linear-verify.ts: verifyLinearSignature rejects empty/whitespace secrets (critical for the per-workspace path, whose secrets come from OAuth bundles, not
  getLinearSecret), and getLinearSecret treats whitespace-only SecretString as null — mirroring the hardened GitHub verifier. Created the previously missing cdk/test/handlers/shared/linear-verify.test.ts with 21 tests covering
  signatures, caching, rotation re-fetch, replay window, and the forgery scenario end-to-end.

  5. Screenshot frame-attribution race — cdk/src/handlers/shared/agentcore-browser.ts now records Document response statuses per frameId in the single message listener (the duplicate listener and its double JSON.parse are gone), and
  resolves the main frame's status after load using the frameId from Page.navigate. Events arriving before navigate resolves are no longer lost or misattributed; an unambiguous single status is used when navigate returns no frameId, and
  ambiguous multi-frame cases fall back to optimistic capture. Added 5 race-scenario tests (sub-frame 404 after main 200, the inverse, redirect chains, missing frameId).

  6. CLI --verbose secret leak — new redactSensitive() in cli/src/debug.ts deep-redacts secret, access_token, refresh_token, client_secret, authorization, etc. before request/response bodies are debug-logged in cli/src/api-client.ts.
  New cli/test/debug.test.ts (8 tests) covers the webhook-secret case, nesting, case-insensitivity, and non-mutation.

  7. Stale task_type contract in design docs — docs/design/ORCHESTRATOR.md Tasks table now documents workflow_ref/resolved_workflow (the task_type and never-implemented error_code rows are gone), the mid-table blockquote that broke
  rendering was moved below the table, and the hydration prose is keyed by workflow. docs/design/ARCHITECTURE.md model table and the AGENTS.md types.ts description now use workflow IDs. Starlight mirrors regenerated via
  sync-starlight.mjs.

  Verification: agent — 1016 tests pass (mise run quality, 76.3% coverage); cdk — compile clean, 2011 tests pass; cli — full build (compile + 324 tests + lint) passes; zizmor clean.

#### Area

#### Acknowledgment

### DIFF
--- a/.github/workflows/auto-approve.yml
+++ b/.github/workflows/auto-approve.yml
@@ -3,16 +3,19 @@ name: auto-approve
 # upgrade-main dependency PRs). pull_request_target is intentional and safe
 # here: the job never checks out or executes PR code — it only calls the
 # review API. Branch protection still requires the build workflow to pass.
+#
+# Trigger is `labeled` ONLY: each approval requires an explicit label event
+# from someone with triage+ permission. Approving on `synchronize` would
+# re-approve arbitrary future commits pushed to an already-labeled PR with no
+# human re-review — combined with Mergify's auto-merge that is an
+# unreviewed-merge path. The Mergify `dismiss stale approvals` rule is the
+# other half of this invariant: a push invalidates the bot's prior approval,
+# and re-approval requires removing and re-adding the label (deliberate
+# friction: new content needs a fresh, explicit trigger).
 on:
   pull_request_target:
     types:
       - labeled
-      - unlabeled
-      - opened
-      - synchronize
-      - reopened
-      - ready_for_review
-      - review_requested
 permissions:
   actions: none
   attestations: none
@@ -34,7 +37,11 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
-    if: contains(github.event.pull_request.labels.*.name, 'auto-approve')
+    # Only act on the auto-approve label itself, and never on fork PRs —
+    # a fork head can be force-pushed by the fork owner after labeling.
+    if: |
+      github.event.label.name == 'auto-approve' &&
+      github.event.pull_request.head.repo.full_name == github.repository
     steps:
       - uses: hmarr/auto-approve-action@f0939ea97e9205ef24d872e76833fa908a770363 # v4.0.0
         with:
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -11,6 +11,18 @@ queue_rules:
 
       {{ body }}
 pull_request_rules:
+  # Approvals must cover the commits actually being merged. Without this,
+  # the auto-approve bot's review (or any human review) would keep
+  # satisfying `#approved-reviews-by>=1` after arbitrary new commits are
+  # pushed, letting unreviewed content auto-merge. Dismissing on every
+  # push forces a fresh approval (for label-gated PRs: remove and re-add
+  # the `auto-approve` label) before the queue rule can fire again.
+  - name: Dismiss stale approvals when new commits are pushed
+    conditions:
+      - base=main
+    actions:
+      dismiss_reviews:
+        approved: true
   - name: Automatic merge on approval and successful build
     actions:
       delete_head_branch: {}
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -82,7 +82,7 @@ The `@backgroundagent/cli` package provides the `bgagent` executable for submitt
 - `src/api-client.ts` — HTTP client wrapping `fetch` with auth header injection
 - `src/auth.ts` — Cognito login, token caching (`~/.bgagent/credentials.json`), auto-refresh
 - `src/config.ts` — Read/write `~/.bgagent/config.json`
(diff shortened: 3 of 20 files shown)


theagenticguy (approved):
## Approve — 7 fixes, all verified correct, CI green

Reviewed each fix against full source context (not just the diff), with two independent deep-dives on the highest-risk changes. `build (agentcore)` is green; CodeQL, both CodeQL Analyze jobs, and title-validate all pass.

### Per-fix verdict

1. **Auto-approve / Mergify governance** — Correct. Trigger narrowed to `labeled` only; the gate now checks `github.event.label.name == 'auto-approve'` (the label on *this* event, not mere presence in the set) and excludes fork-head PRs. Paired with Mergify `dismiss_reviews` so any push invalidates the bot's prior approval. This closes the path where a `synchronize` re-approved arbitrary future commits under auto-merge.

2. **UTC timezone bug (`hooks.py`)** — Correct. `strptime(...).replace(tzinfo=UTC).timestamp()` fixes the naive-datetime skew. The `TZ=America/New_York` test genuinely catches the regression.

3. **PAT out of `.git/config` (`repo.py`)** — Verified the runtime concern that the credential-helper approach raises: does `GH_TOKEN`/`GITHUB_TOKEN` survive to *push* time? Yes. The token is set process-global in `pipeline.py` (and `runner.py` for the LLM path) before `setup_repo()`, is never deleted, and `run_cmd`'s `_clean_env()` strips only `OTEL_*` — so both the deterministic push (`post_hooks.py`) and the CLI-subprocess push inherit it. `gh` is built in the Dockerfile and authed purely via those env vars, same as the pre-existing `gh repo clone`. No push regression. The `output_scanner.py` widening (36+ chars, `ghr_`, `x-access-token:…@` URL pattern) is a clean complement.

4. **Linear empty-secret forgery (`linear-verify.ts`)** — Correct and the in-function guard is load-bearing, not redundant: `verifyLinearRequestForWorkspace` feeds the per-workspace OAuth secret straight to `verifyLinearSignature` without passing through `getLinearSecret`'s empty-check, and a whitespace-only `webhook_signing_secret` is truthy past the `!stored.webhook_signing_secret` check — so the `.trim()` guard is the only thing stopping `HMAC('   ')` forgery on that path. The "mirrors `verifyGitHubSignature` / `getGitHubWebhookSecret`" claim is accurate byte-for-byte. 21 tests.

5. **Screenshot frame-attribution race (`agentcore-browser.ts`)** — Correct. The per-frame `Map` keyed by the `Page.navigate` frameId fixes two real bugs in the old code: the global "latest Document status wins" let a sub-frame 404 overwrite the main 200, and the pre-navigate `mainDocumentFrameId === null` window misattributed early responses. Removing the duplicate `ws.on('message')` listener (and its second `JSON.parse`) is a correctness + efficiency win. The ambiguous-multi-frame → optimistic-capture fallback is a reasonable choice. 5 race-scenario tests.

6. **CLI `--verbose` secret leak (`debug.ts`)** — Correct. `redactSensitive` deep-redacts case-insensitively, non-mutating, applied to both request and response bodies. 8 tests, including the non-redaction of `next_token`.

7. **Stale `task_type` docs** — Accurate; also relocates the mid-table blockquote that was breaking table rendering, and the Starlight mirrors match.

### One follow-up (out of scope here, not a blocker)

Fix #4's own rationale — "an empty secret must never be used for HMAC, or `HMAC('', body)` is forgeable" — applies verbatim to two webhook verifiers this PR doesn't touch:

- `cdk/src/handlers/shared/slack-verify.ts` — `getSlackSecret` checks only `!result.SecretString`, and `verifySlackSignature` has no empty/whitespace guard before `createHmac`.
- `cdk/src/handlers/webhook-create-task.ts` — `getSecret` and `verifySignature` likewise.

Both reject a truly empty `""` secret but let a whitespace-only `"   "` secret through to HMAC. Pre-existing condition in different files; worth a follow-up PR to mirror the same two-line guard across all four verifiers.

Approving. Clean, well-tested, well-commented work.

Issue #292, opened 2026-06-08
feat(infra): add solution-tracking metrics ID to CloudFormation stack description


scottschreckengaust:
## Summary

Append the AWS solution-tracking metrics identifier `(uksb-wt64nei4u6)` to the CloudFormation stack description so deployments of this sample are counted via AWS's standard usage-tracking mechanism.

## Background

AWS samples/solutions embed a `uksb-*` tracking token in the stack description to record anonymized deployment metrics. This stack currently sets a plain description (`cdk/src/main.ts:41`):

```ts
description: 'ABCA Development Stack',
```

## Acceptance criteria

- [ ] The stack description includes the tracking ID: `ABCA Development Stack (uksb-wt64nei4u6)`
- [ ] CDK compiles and synth succeeds; existing tests pass
- [ ] No other behavior changes

## Coupling note (must address before/with the cleanup-script branch)

The in-flight `feat/cleanup-ephemeral-stacks` branch (PR #109) matches ABCA stacks by an **exact** description string (`[[ "$DESCRIPTION" != "ABCA Development Stack" ]]`). Appending the tracking ID changes the description and will break that exact match once both land. When the cleanup script merges, its matcher must switch to a **prefix** match (e.g. `== "ABCA Development Stack"*`). Tracking here so it isn't missed.

scottschreckengaust:
Starting implementation on branch `feat/update-description`.

Issue #215, opened 2026-05-29
feat: Bedrock cost attribution — session tags, request metadata, and operator FinOps guidance


krokoko:
## Summary

Per-session IAM scoping (#209 / #211) isolates **tenant data** (DynamoDB, S3) with session tags `{user_id, repo, task_id}`, but **Bedrock inference still bills under the shared AgentCore execution role**. Multi-user deployments cannot split Bedrock spend by user or repo in AWS Cost Explorer / CUR 2.0 without additional work.

This issue tracks **Bedrock-specific cost attribution** in ABCA: platform changes plus documented operator steps.

## Problem

| Dimension | In-app (`cost_usd`) | AWS billing (CUR / Cost Explorer) |
|-----------|---------------------|-----------------------------------|
| Per task | ✅ | ❌ |
| Per user | ✅ (aggregate) | ❌ |
| Per repo | ✅ (aggregate) | ❌ |

Finance / FinOps teams need AWS-native chargeback; operators need per-call forensics in invocation logs. Both are distinct from in-app budget enforcement.

## Proposed scope

Implement **complementary** mechanisms (not either/or):

### 1. Bedrock IAM session-tag attribution (platform)

Route Bedrock calls through assumed credentials that carry `{user_id, repo, task_id}` session tags — extend the SessionRole / `aws_session.py` pattern from #211.

- Bedrock `InvokeModel` uses tagged assumed-role credentials (refreshable, same 1 h chaining cap as tenant data).
- SessionRole (or a dedicated Bedrock invocation role) grants `bedrock:InvokeModel` on the existing foundation-model / inference-profile ARN allowlist.
- Claude Code / agent runtime configured to use scoped creds for Bedrock only; non-tenant services unchanged.

**Outcome:** `line_item_iam_principal` and `iamPrincipal/user_id`, `iamPrincipal/repo` (etc.) in CUR 2.0 after operator tag activation.

### 2. Bedrock per-request metadata (platform)

Pass `task_id`, `user_id`, `repo` on each Bedrock inference via request metadata (`X-Amzn-Bedrock-Request-Metadata` / Converse `requestMetadata`).

- Requires validating Claude Code / SDK support for metadata on InvokeModel.
- Leverages existing model invocation logging (CDK stack already enables CloudWatch destination).

**Outcome:** Per-call attribution in `/aws/bedrock/model-invocation-logs/<stack>`; join to CUR on `requestId` or aggregate tokens from logs.

[...]

Issue #245, opened 2026-06-03
feat(observability): end-to-end task attribution and cross-plane trace correlation


krokoko:
### Component

API or orchestration, Agent (Python runtime), CDK / infrastructure

### Describe the feature

Implement **end-to-end task attribution** and **unified cross-plane trace correlation** so every action in a task's lifecycle can be traced back to `{task_id, user_id, repo}` and a single shared trace root — across the orchestrator, agent runtime, GitHub/API calls, `TaskEvents`, OpenTelemetry spans, and S3 trace artifacts.

Today attribution is **partial**: per-session IAM scoping (#209, shipped) tags tenant-data access; git commits carry `Task-Id` trailers; `TaskEvents` and task records hold operational fields. Gaps remain in orchestrator structured logs, agent OTel baggage propagation, external API call correlation, and stitching MicroVM traces to platform-side events for incident response.

This issue is the **platform identity contract** — distinct from Bedrock billing attribution (#215) and compliance export schema (#237), though all three should share the same correlation field names.

### Use case

- **Incident response:** Answer "who triggered this, on which repo, in which task?" without manually joining CloudWatch, DDB, and S3.
- **Audit and compliance:** Feed #237's `abca.audit.v1` `correlation` block from real propagated context, not ad-hoc assembly.
- **Cost and ops debugging:** Join Bedrock invocation logs (#215 metadata) to orchestrator and agent spans via shared `trace_id` / `task_id`.
- **Vision tenet 7:** Observable, attributable, replayable — the anti-goal "autonomy without attribution" must not apply to shipped paths.

### Proposed solution

**Phase 0 — Correlation contract**
- Document a stable correlation envelope: `{ task_id, user_id, repo, trace_id, session_id? }` and where each field is authoritative.
- Define naming consistency across orchestrator logs, agent OTel resource/span attributes, `TaskEvents` optional fields, and git/API metadata.

**Phase 1 — Propagation MVP**
- Orchestrator: inject correlation into structured logs and durable-step context for every task transition.
- Agent: propagate envelope via OTel baggage / span attributes from session start through pipeline phases and progress events.
- Ensure `progress_writer` and terminal events include correlation stamps where missing.

**Phase 2 — Cross-plane trace root**
- Single trace root per task linking orchestrator spans, MicroVM agent spans, and (where available) S3 trace artifact URI.
[...]

PR #338, opened 2026-06-12
feat(observability): inject solution into outbound AWS SDK User-Agent (#319)

### LINKED ISSUE #319: feat(observability): inject solution into outbound AWS SDK User-Agent
## Summary

Stamp an ABCA solution-attribution identifier into the **outbound AWS SDK `User-Agent`** on every AWS API call made by the platform, so AWS can attribute runtime API usage to this sample. This is the per-API-call complement to the deploy-time CloudFormation attribution token added in #292 (`uksb-wt64nei4u6` in the stack description, `cdk/src/main.ts:41`).

Two cooperating fields, emitted on every request:

```
app/uksb-wt64nei4u6/{STACKNAME}             ← app-id form; STACKNAME sanitized + clipped to ≤34 chars (≤50 total)
md/uksb-wt64nei4u6#{IDENTIFIER}#{TRACE}     ← solution metadata; {TRACE} OPTIONAL, appended per-request
```

- `app/uksb-wt64nei4u6/{STACKNAME}` — the app-id form. `uksb-wt64nei4u6/` is 16 chars, so `{STACKNAME}` is the **first ≤ 34 chars** of the **sanitized** stack name to keep the whole value ≤ 50. `{STACKNAME}` is ASCII-safe with `/` and `#` (and any other non-token char) replaced by `-` — see "Separators & sanitization". Set once per environment.
- `md/uksb-wt64nei4u6#{IDENTIFIER}#{TRACE}` — solution metadata, emitted verbatim (no length cap — see "Character budget"). `{IDENTIFIER}` names **which ABCA component the calling code is** (e.g. `agent`, `orchestr`, `webhook`, `cli`) — a stable per-component label baked at client construction. `{TRACE}` is an **optional, per-component correlation token** — whatever trace handle that component has (see "The `{TRACE}` field") — appended per-request **without reissuing the connection** (see "Connection sharing"). When no trace handle is in scope, the token is `md/uksb-wt64nei4u6#{IDENTIFIER}` with no trailing separator. (The stack name lives only in the `app/` field, not here.)

## Background

AWS samples/solutions are attributed two ways:

1. **Deploy-time** — a `uksb-*` token in the CloudFormation stack description. ✅ Done in #292.
2. **Runtime** — a solution identifier injected into the SDK `User-Agent` header so individual API calls are attributable. ❌ This issue.

Today **no SDK client in the repo customizes its user agent.** There are three independent SDK surfaces, each instantiating clients with empty config:

| Surface | Lang / SDK | Examples | Injection mechanism |
|---|---|---|---|
| CDK Lambda handlers | TS / AWS SDK for JS v3 | `new DynamoDBClient({})`, `new BedrockAgentCoreClient({})`, `new ECSClient({})`, `new S3Client({})` (`cdk/src/handlers/*.ts`) | `customUserAgent` (verbatim metadata), or a UA middleware |
| Agent runtime | Python / boto3 + botocore | `boto3.client(...)` / `boto3.Session(...)` via `agent/src/aws_session.py`, `agent/src/config.py` | `Config(user_agent_extra=...)` (verbatim metadata), or a `before-send` event handler |
| CLI | TS / Cognito SDK v3 | Cognito client in `cli/src/auth.ts` | same as CDK |

> Note: `cdk/src/handlers/shared/gateway.ts:57,87` references `user_agent` but that is the **inbound** API Gateway caller UA from the request context — unrelated to outbound SDK calls.

## Separators & sanitization (authoritative — see issue comments)

Research findings on this issue (with verbatim source snippets) are in the two comments below. The relevant constraints:

- **Metadata (`md/...`)** is emitted **verbatim** — botocore wraps `user_agent_extra` as a `RawStringUserAgentComponent` (no sanitization, no truncation); JS v3 `customUserAgent` is likewise passed through. So the `#` separators inside the `md/uksb-wt64nei4u6#{IDENTIFIER}#{TRACE}` string are valid and preserved. ✅
- **The literal `/` between `uksb-wt64nei4u6` and `{STACKNAME}` in the `app/` segment** is kept. If `{STACKNAME}` were set via the SDK app-id config field (`AWS_SDK_UA_APP_ID` / botocore `user_agent_appid` / JS `userAgentAppId`), that field is **sanitized** against an allowed charset that excludes `/` (→ `-`), which would mangle the `uksb-wt64nei4u6/` separator. So the `app/uksb-wt64nei4u6/{STACKNAME}` string is contributed via the **raw user-agent component path** (same verbatim mechanism as `md/`), **not** the sanitizing app-id config field. The ≤ 50 length discipline is applied by us at construction.
- **`{STACKNAME}`, `{IDENTIFIER}`, and `{TRACE}` are sanitized by us before emission.** Because we emit via the raw path (no SDK sanitization), each value must be cleaned so it cannot break the segment structure or violate the UA token charset:
  - Replace **`/`** and **`#`** (the structural separators of this scheme) — and any other non-UA-token character — with **`-`**.
  - Enforce **ASCII-safe**: drop/replace any non-ASCII byte with `-`.
  - Target charset = UA token chars (`A–Z a–z 0–9 ! $ % & ' * + - . ^ _ \` | ~`). In practice CloudFormation stack names are already `[A-Za-z0-9-]` and most trace handles (ULIDs, Lambda request IDs, PIDs) are too, so this is largely defensive — but it MUST be applied unconditionally to all three values.
  - For `{STACKNAME}`, sanitize **first, then clip** to ≤ 34 chars (so clipping can't re-expose a partial multi-byte sequence).

## Character budget

Authoritative findings (verbatim snippets + URLs) are in the issue comments. Summary:

- **`md/` metadata has NO SDK length cap.** The 50-char limit applies only to the app-id field; `user_agent_extra` / `customUserAgent` is emitted verbatim. The only other size limit in the UA code is 1024 bytes on the unrelated `m/` business-metrics component. So `md/uksb-wt64nei4u6#{IDENTIFIER}#{TRACE}` is **not budget-constrained** — `{IDENTIFIER}` and the `{TRACE}` handle (ULID, Lambda request ID, etc.) fit with no truncation. (Keep it reasonable for HTTP-header practicality, but there is no hard SDK cap.)
- **The app-id form keeps the ≤ 50 discipline** because we mirror the documented app-id constraint even though we emit via the raw path: `uksb-wt64nei4u6/` (16) + `{STACKNAME}` ≤ 50 → **`{STACKNAME}` = first ≤ 34 chars of the sanitized stack name.** The `app/` marker itself is not counted toward the 50 (confirmed: the documented 50 is on the value, not the prefix).

Because `{TRACE}` rides in the uncapped `md/` metadata, **no `{TRACE}` elision is ever required** — the full handle is always emitted.

## The `{IDENTIFIER}` field — which component is making the call

`{IDENTIFIER}` lives in the uncapped `md/` metadata and is baked once at construction. The stack name is **no longer** carried in `md/` — it lives only in the `app/` field (as the sanitized, ≤ 34-char `{STACKNAME}`).

`{IDENTIFIER}` names the **component / "thing" the calling code is** — i.e. which part of ABCA originated the AWS API call. It is a **per-component constant**, not a per-deployment value: each emitting surface uses a fixed label describing itself (the `{STACKNAME}` in the `app/` field already carries the deployment identity).

Examples of the kind of value intended (illustrative — final names up to the implementer):

- `agent` — the Python agent runtime container
- `orchestr` — orchestrator / task-lifecycle Lambdas
- `webhook` — webhook-create-task path
- `cli` — the `bgagent` CLI

Requirements:

- **Must be UA-token-safe** (sanitized: `/`, `#`, and any non-UA-token / non-ASCII char → `-`).
- **Recommended short — ≤ 10 characters — but NOT a hard requirement.** There is no SDK length cap on `md/` metadata (see comments), so longer values are allowed; ≤ 10 is a readability/conciseness guideline only.
- A given component uses a **stable** label (don't vary it per call); the only per-request variation in `md/` is the optional `#{TRACE}` suffix.

## The `{TRACE}` field — optional per-component correlation handle

`{TRACE}` is the **optional** trailing segment. It is **not task-specific** — it carries whatever correlation/trace handle the emitting component naturally has, so the same scheme works across every surface. Each component supplies its own kind of handle (or none):

| Component (`{IDENTIFIER}`) | Natural `{TRACE}` handle | Source |
|---|---|---|
| `agent` | task ULID | the task the agent is working |
| `orchestr` (Lambda) | AWS Lambda request ID | `context.awsRequestId` (JS) / `context.aws_request_id` |
| `cli` | process PID (or session id) | `process.pid` |
| any other | whatever trace handle it carries | component-specific |

`{TRACE}` is **always optional** — when a component has no meaningful handle for a given call (or hasn't wired one up), it is omitted and the metadata is simply `md/uksb-wt64nei4u6#{IDENTIFIER}` (no trailing `#`). When present, it lets the `md/` field answer "**which ABCA component made this call, correlated to which unit of work**" — e.g. `md/uksb-wt64nei4u6#agent#01KTVY…`, `md/uksb-wt64nei4u6#orchestr#<aws-request-id>`, `md/uksb-wt64nei4u6#cli#12345`.

`{TRACE}` must be UA-token-safe (same sanitization as above) and is appended **per-request** (never baked into client config — see "Connection sharing").

## Connection sharing — `{TRACE}` must NOT pin a connection

`agent/src/aws_session.py` deliberately keeps a **module-level singleton session** (`_session`) with **refreshable** credentials so a single long-running task (up to the 8 h `maxLifetime`) reuses one session and its connection pool rather than re-assuming/reconnecting per call. There can also be durable/cached-client situations across task boundaries.

Baking a per-request `{TRACE}` handle into a client's `Config(user_agent_extra=...)` at construction time would **pin that cached client (and its HTTP connection pool) to a single trace context**, forcing a client/connection reissue just to update the UA. That is the failure mode to avoid.

**Resolution — static baked, dynamic appended per-request:**

- **Static** (`app/uksb-wt64nei4u6/{STACKNAME}` and `md/uksb-wt64nei4u6#{IDENTIFIER}`) is baked once at client/session construction. It never varies, so shared/cached clients and pooled connections are reused freely.
- **Dynamic** (`#{TRACE}`) is appended **per request** by mutating only the outgoing `User-Agent` header — never the client config — reading the current trace handle from ambient context:
  - **Python / botocore:** register a `before-send` (or `before-sign`) event handler on the session that reads a `contextvar` holding the current trace handle and appends `#{TRACE}` to `request.headers['User-Agent']`. The connection pool is untouched; only the header string changes per request.
  - **JS SDK v3:** add a `build`/`finalizeRequest` **middleware** to the client stack that reads the current trace handle from `AsyncLocalStorage` (or, in a Lambda, from `context.awsRequestId`) and appends `#{TRACE}`.
- `{TRACE}` is **optional**: when no trace handle is in scope, the header is simply `md/uksb-wt64nei4u6#{IDENTIFIER}` with no trailing separator.

Net effect on the wire: `... app/uksb-wt64nei4u6/{STACKNAME} md/uksb-wt64nei4u6#{IDENTIFIER}#{TRACE} ...`, with the `#{TRACE}` segment present only when a handle is in scope — and reusing one connection across many trace contexts.

## Acceptance criteria

- [ ] **Static metadata baked once:** all clients/sessions carry `md/uksb-wt64nei4u6#{IDENTIFIER}` baked at construction (boto3 `Config(user_agent_extra=...)`, JS `customUserAgent`), and the `app/uksb-wt64nei4u6/{STACKNAME}` form contributed via the **raw user-agent path** (NOT the sanitizing app-id field — see the sanitization section).
- [ ] **`{STACKNAME}` sanitized + clipped:** the stack name is sanitized (replace `/`, `#`, and any non-UA-token / non-ASCII char with `-`) **then** clipped to the **first ≤ 34 chars** so the whole app-id value (`uksb-wt64nei4u6/` + name) stays ≤ 50. A test asserts: (a) a stack name containing `/`, `#`, and a non-ASCII char emits with those replaced by `-`; (b) the value length for the longest realistic stack name is ≤ 50.
- [ ] **`{IDENTIFIER}` is a stable per-component label** identifying which ABCA component the calling code is (e.g. `agent`, `orchestr`, `webhook`, `cli`) — fixed per surface, not varied per call; UA-token-safe; recommended ≤ 10 chars (soft guideline, not enforced — no SDK cap on `md/`). A test asserts each surface emits its expected component label.
- [ ] **`{TRACE}` appended per-request, never baked, and OPTIONAL:** a botocore `before-send` handler (agent) and a JS SDK middleware (handlers/CLI as applicable) append `#{TRACE}` from ambient context (`contextvar` / `AsyncLocalStorage` / Lambda `context.awsRequestId` / PID) **without** reconstructing the client or its connection pool. Each component supplies its own handle kind (agent → task ULID, orchestrator Lambda → AWS request ID, CLI → PID, etc.) and may omit it. Tests assert: (a) two requests from the **same** cached client under different trace contexts emit different `#{TRACE}` segments (connection reused, not reissued); (b) with no handle, the metadata is exactly `md/uksb-wt64nei4u6#{IDENTIFIER}` (no trailing `#`); (c) `{TRACE}` is sanitized to UA-token-safe ASCII.
- [ ] **Outgoing-header inspection tests (required for validation):** tests must capture the **actual outbound request's `User-Agent` header** (not just the configured client value) and assert that **both** the `app/uksb-wt64nei4u6/{STACKNAME}` segment **and** the `md/uksb-wt64nei4u6#{IDENTIFIER}[#{TRACE}]` segment are present and intact in the emitted header — including verifying the literal `/` in the `app/` segment survives (i.e. it was NOT routed through the sanitizing app-id field). Capture the header at the wire layer per surface — e.g. a botocore `before-send`-stage assertion or a stubbed transport that records `request.headers['User-Agent']` (boto3), and a terminal middleware or request spy capturing the `user-agent`/`x-amz-user-agent` header (JS SDK v3). Verify both the trace-present (`...#{TRACE}`) and trace-absent (`...#{IDENTIFIER}` only) cases.
- [ ] **Agent runtime (boto3):** centralized in `agent/src/aws_session.py` / `agent/src/config.py` so no `boto3.client(...)` call site is missed; the singleton refreshable session keeps its pool across tasks.
- [ ] **CDK Lambda handlers (JS SDK v3):** all `new XClient({})` instantiations carry the ABCA UA via a shared client factory or UA middleware.
- [ ] **CLI (JS SDK v3):** the Cognito client in `cli/src/auth.ts` carries the ABCA UA.
- [ ] **`{STACKNAME}` sourced from the deployed stack name**, threaded to each surface (Lambda env, agent container env, CLI config) — not hard-coded; sanitized to UA-token-safe ASCII.
- [ ] CDK compiles, synth succeeds, `mise //cdk:test` passes; agent tests pass.
- [ ] No functional behavior change beyond the added UA metadata.

## Related

- #292 — solution-attribution metrics ID in the **CloudFormation stack description** (deploy-time counterpart; CLOSED/implemented). This issue is the runtime API-call counterpart.
- #215 — Bedrock cost attribution (session tags, request metadata, FinOps). UA-based attribution is adjacent; coordinate so Bedrock-client UA work isn't duplicated.
- #245 — end-to-end task attribution & cross-plane trace correlation. The `{TRACE}` handle in the UA is another attribution vector; align on the identifier.

## Out of scope

- Dashboards / querying of the attributed usage (belongs with #215 / #245).
- Changing the deploy-time `uksb` token (already handled by #292).

scottschreckengaust:
> Implements #319. This PR body is the plan of record (no plan file committed); all 7 tasks are complete — each landed as its own commit. The first commit is an empty start-signal.
>
> **Verification:** agent quality gate (1033 tests, 76.5% coverage, ruff, ty) ✅ · `mise //cdk:test` (2033 tests incl. new wire-capture + template assertions) ✅ · full CLI build (335 tests) ✅ · CDK↔CLI type-sync + constants-drift checks ✅ · CI: CodeQL ×3, security scans, PR lint, `build (agentcore)` all green. Local `cdk synth` fails on this workstation for environmental reasons only (synth requires `ec2:DescribeAvailabilityZones`, not granted to the local role; `main` also fails locally) — CI's build job performs the real synth and passes. The `integ` smoke failure is pre-existing on `main` (known issue, unrelated).

Closes #319

## What this builds

Every outbound AWS API call from ABCA carries two solution-attribution User-Agent segments:

```
app/uksb-wt64nei4u6/{STACKNAME}             ← sanitized stack name, clipped to ≤34 chars (value ≤50)
md/uksb-wt64nei4u6#{IDENTIFIER}[#{TRACE}]   ← component label + optional per-request trace handle
```

Both are emitted via the **raw/verbatim user-agent path** (botocore `user_agent_extra` / JS `customUserAgent`) — *not* the sanitizing app-id config field, which would mangle the `/` separator (`/` is not in the SDK app-id charset; verified against installed botocore 1.43.9 `useragent.py` and `@aws-sdk/middleware-user-agent`). Static parts are baked at client construction; `#{TRACE}` is appended per-request by a botocore `before-send` handler / JS middleware so cached clients and pooled connections are never reissued.

### Verified SDK mechanics this design relies on

- **JS v3 `customUserAgent` pair encoding** (`middleware-user-agent/dist-cjs/index.js`): a pair `['app/uksb-wt64nei4u6/STACK']` survives intact — the name is split on `/`, each part escaped (all our chars are token-safe), and rejoined with `/`. A pair `['md/uksb-wt64nei4u6', 'api']` renders `md/uksb-wt64nei4u6#api` (`#` is legal in values, not names). `customUserAgent` lands in **both** `user-agent` and `x-amz-user-agent` (node runtime), always at the **end** of each header — which is what lets the trace middleware do a safe suffix-append.
- **The UA middleware is named `getUserAgentMiddleware`, step `build`, priority `low`** — our trace middleware is added `addRelativeTo(..., { relation: 'after', toMiddleware: 'getUserAgentMiddleware' })`.
- **`DynamoDBDocumentClient.from(inner)` shares the inner client's `middlewareStack` and `config`** — instrumenting the inner `DynamoDBClient` covers the document client.
- **botocore renders `user_agent_extra` last** (`_build_extra` is the final component in `to_string()`), so a `before-send` handler can append `#{TRACE}` to the end of `request.headers['User-Agent']` and it lands exactly on the `md/` segment. `before-send` handlers can also return a stubbed `AWSResponse` to short-circuit the HTTP send — that's the wire-capture test mechanism (no network, no moto).

### Identity & trace mapping

| Surface | `{IDENTIFIER}` | source | `{TRACE}` handle |
|---|---|---|---|
| Python agent runtime | `agent` | hardcoded | task id (ULID), set once via `configure_session()` |
| REST API Lambdas (task-api) | `api` | `ABCA_COMPONENT` env | per-invocation `ulid()` requestId handlers already mint |
| Webhook ingest + Slack/Linear/GitHub integrations | `webhook` | `ABCA_COMPONENT` env | requestId where minted, else absent |
| Orchestrator, strategies, reconcilers, stream consumers | `orchestr` | `ABCA_COMPONENT` env | `event.task_id` (orchestrate-task); else absent |
| `bgagent` CLI | `cli` | hardcoded | `process.pid`, set once at startup |

`{IDENTIFIER}` comes from `ABCA_COMPONENT` env (not per-call-site parameters) because shared modules (`shared/orchestrator.ts`, `shared/create-task-core.ts`, …) construct module-level clients and are bundled into multiple Lambdas — the env var gives each Lambda its own label with one source of truth in CDK.

`{STACKNAME}` comes from a new `ABCA_STACK_NAME` env var (Lambdas, agent container) / optional `stack_name` CLI config field. When absent (local dev, unconfigured CLI), the `app/` segment is omitted entirely — never a placeholder.

**Trace ambient context:** module-level variable (with the JS/Lambda single-concurrent-invocation guarantee, and matching the agent's existing `aws_session._tags` module-state pattern) rather than `contextvar`/`AsyncLocalStorage`. `server.py:300` already documents why `ContextVar` is a per-thread trap in the agent (task runs on a spawned thread). The issue allows any ambient mechanism.

**Solution-ID constant placement:** per-surface literal constant (`agent/src/ua.py`, `cdk/src/handlers/shared/ua.ts`, `cli/src/ua.ts`) with cross-referencing comments — NOT `contracts/constants.json`. That contract's schema and its drift checker (`scripts/check-constants-sync.ts`) are numeric-only (`min`/`max`/`default` invariants policed in `agent/src/policy.py`); the existing repo precedent for `uksb-wt64nei4u6` is a literal (`cdk/src/main.ts:41`). Each surface's tests assert the exact emitted string, which is the de-facto drift guard.

### Shared sanitization spec (implemented identically in Python and TS, with identical test vectors)

```
ALLOWED = A–Z a–z 0–9 ! $ % & ' * + - . ^ _ ` | ~          (UA token chars; note '/' and '#' are NOT allowed)
sanitize(v): replace every char not in ALLOWED (incl. any non-ASCII byte) with '-'
stackname:   sanitize FIRST, then clip to first 34 chars   (16-char 'uksb-wt64nei4u6/' + 34 = 50)
```

---

### Task 1 — Python: `agent/src/ua.py` + tests (TDD)

**Create:** `agent/src/ua.py`, `agent/tests/test_ua.py`

- [x] `SOLUTION_ID = "uksb-wt64nei4u6"`, `COMPONENT = "agent"`, `STACK_NAME_ENV = "ABCA_STACK_NAME"`.
- [x] `sanitize_ua_value(raw: str) -> str` per the shared spec.
- [x] `static_user_agent_extra() -> str` — `app/uksb-wt64nei4u6/{stack(≤34)} md/uksb-wt64nei4u6#agent`, dropping the `app/` segment when `ABCA_STACK_NAME` is unset/empty.
- [x] `set_trace(handle: str | None)` / `get_trace()` — lock-guarded module state; sanitized on read.
- [x] `register_trace_appender(events)` — registers a `before-send.*` handler that appends `#{trace}` to `request.headers['User-Agent']` when a trace is set (no-op otherwise; never reconstructs the client). Works for both `client.meta.events` and botocore-session-level registration (sessions propagate to all derived clients **and resources**).
- [x] `client_config() -> botocore.config.Config` — `Config(user_agent_extra=static_user_agent_extra())` for direct-client call sites.
- [x] Tests (sanitize vectors incl. `/`, `#`, non-ASCII → `-`; 50-char budget; clip-after-sanitize ordering; trace-absent ⇒ no trailing `#`).
- [x] **Wire-capture test:** real botocore client (fake creds, region pinned), `register_trace_appender` + a `register_last('before-send.*', …)` stub returning a canned `AWSResponse`. Two calls on the **same client** under different `set_trace()` values assert different `#{TRACE}` suffixes, intact `app/.../` slash, and intact `md/...#agent` — straight from the captured `User-Agent` header.

### Task 2 — Python: wire into `aws_session.py` and the 8 bypass sites

**Modify:** `agent/src/aws_session.py`, `agent/src/config.py:46,110`, `agent/src/shell.py:67`, `agent/src/telemetry.py:62,170`, `agent/src/server.py:157,184`, `agent/src/memory.py:43`; tests in `agent/tests/test_aws_session.py`

- [x] `_build_scoped_session()` (aws_session.py:160): set `botocore_session.user_agent_extra = ua.static_user_agent_extra()` and `ua.register_trace_appender(...)` on the botocore session we already construct. Same for the plain-session path (aws_session.py:213) by constructing the botocore session explicitly. Covers `tenant_client`/`tenant_resource` (progress_writer, task_state, nudge_reader, attachments, deliverers, telemetry S3) and the singleton's pool is untouched across trace changes.
- [x] **Unscoped `tenant_client`/`tenant_resource` gap** (aws_session.py:240/250): these bypass the session and call `boto3.client(...)` directly (kept that way deliberately for test transparency) — apply `config=ua.client_config()` + `ua.register_trace_appender(client.meta.events)` there too, merging with any caller-supplied `config` kwarg.
- [x] New `aws_session.platform_client(service, **kwargs)` — `boto3.client(service, config=ua.client_config(), **kwargs)` + `ua.register_trace_appender(client.meta.events)`. Migrate the 8 direct `boto3.client(...)` sites (logs ×5, secretsmanager ×2, bedrock-agentcore ×1) to it. The ambient STS client at aws_session.py:142 gets `config=ua.client_config()` inline (its events come from the default session). Existing tests patch `boto3.client` globally — MagicMocks tolerate the extra `config` kwarg and the events registration.
- [x] `configure_session()` (aws_session.py:81) additionally calls `ua.set_trace(task_id)` — the agent's trace is live for the whole task without touching `pipeline.py` call sites beyond what already exists (pipeline.py:634).
- [x] `reset_session_cache()` also clears the trace (test hygiene).
- [x] Run `mise //agent:quality` (lint, typecheck, tests). Commit.

### Task 3 — JS handlers: `cdk/src/handlers/shared/ua.ts` + wire tests (TDD)

**Create:** `cdk/src/handlers/shared/ua.ts`, `cdk/test/handlers/shared/ua.test.ts`

- [x] Constants + `sanitizeUaValue()` mirroring Task 1 (same test vectors).
- [x] `abcaUserAgent(): { customUserAgent: UserAgentPair[] }` — `[['app/uksb-wt64nei4u6/' + stack]]` (only when `ABCA_STACK_NAME` set) + `[['md/uksb-wt64nei4u6', component]]` with component from `ABCA_COMPONENT` (default `api`). Spread into client config: `new DynamoDBClient({ ...abcaUserAgent() })`.
- [x] `setAbcaTrace(handle?: string)` module state; `withAbcaTrace<T>(client: T): T` — adds the suffix-append middleware after `getUserAgentMiddleware`, mutating both `user-agent` and `x-amz-user-agent` only when they end with the static `md/` segment. Defensive no-op when `middlewareStack` is absent — required because ~40 existing test files mock client constructors as `jest.fn(() => ({}))` and module-level instrumentation must not crash under those mocks (real clients always have a stack; this guard is test-environment-only and documented as such).
- [x] **Wire-capture test:** real `DynamoDBClient` with a stub `requestHandler` recording `request.headers` and returning a canned `HttpResponse`. Asserts: both segments present and intact in the emitted header (literal `/` survived — i.e. NOT routed through the app-id field); two sends on the **same client** under different `setAbcaTrace()` values emit different suffixes; trace-absent ⇒ exactly `md/uksb-wt64nei4u6#api` with no trailing `#`; sanitization of a hostile trace value. Commit.

### Task 4 — JS handlers: migrate all client sites + trace at handler entries

**Modify:** all `new XClient(` sites under `cdk/src/handlers/` (≈60, enumerated during exploration — every one currently passes `{}` or `{ region }`), plus the ~20 handler entry points that already mint `const requestId = ulid()`.

- [x] Mechanical per-site change: `new DynamoDBClient({})` → `new DynamoDBClient(abcaUserAgent())`, wrapped `withAbcaTrace(...)` at the outermost cached client (for DocumentClient: instrument the inner client — shared stack confirmed). Lazy singletons (`strategies/ecs-strategy.ts:28`, `strategies/agentcore-strategy.ts:26`, `shared/memory.ts:149`, `confirm-uploads.ts:700`) instrumented inside their `getClient()`s.
- [x] `setAbcaTrace(requestId)` immediately after each existing `ulid()` mint; `setAbcaTrace(event.task_id)` at `orchestrate-task.ts` durable-handler entry. Stream consumers (`fanout-task-events`, `approval-metrics-publisher`) carry no trace (optional by design).
- [x] No behavior change beyond UA; existing handler tests stay green (constructor mocks receive the new config arg; `withAbcaTrace` no-ops on bare-object mocks).
- [x] Run `mise //cdk:test`. Commit (possibly split: shared modules / top-level handlers).

### Task 5 — CDK: thread `ABCA_STACK_NAME` + `ABCA_COMPONENT` env vars

**Modify:** `cdk/src/constructs/task-api.ts:442` (`commonEnv` → adds both, component `api`) and `:931` (`webhookEnv` → `webhook`; note webhookEnv does **not** spread commonEnv), `cdk/src/constructs/task-orchestrator.ts:221` (`orchestr`), consumer constructs (`concurrency-reconciler:71`, `stranded-task-reconciler:107`, `pending-upload-cleanup:90` → `orchestr`; `fanout-consumer` uses `this.fn.addEnvironment(...)` at :168ff — add both vars the same way; `approval-metrics-publisher-consumer:115` currently has **no** `environment:` block — add one), integrations (`slack-integration:188,220,252`, `linear-integration:192,250,289`, `github-screenshot-integration:159,228` → `webhook`), `cdk/src/stacks/agent.ts:297` (runtime env) + `cdk/src/constructs/ecs-agent-cluster.ts:124` (container env) → `ABCA_STACK_NAME` only (agent hardcodes its component).

- [x] Stack name via `Stack.of(this).stackName` (concrete — `stackName` is set explicitly in `main.ts:34`).
- [x] Template assertions in `cdk/test/constructs/task-api.test.ts` and `task-orchestrator.test.ts` (each surface emits its expected component label — acceptance criterion).
- [x] `mise //cdk:test` + `mise //cdk:synth`. Commit.

### Task 6 — CLI

**Create:** `cli/src/ua.ts` (CLI-local mirror of the tiny helper — cli cannot import from cdk; same pattern as the `types.ts` mirror), `cli/test/ua.test.ts`
**Modify:** `cli/src/auth.ts:37,103`, `cli/src/commands/{slack,github,linear,admin}.ts` client sites, `cli/src/bin/bgagent.ts` (one `setAbcaTrace(String(process.pid))` at startup), `cli/src/types.ts` (`CliConfig.stack_name?: string`), `cli/src/commands/configure.ts` (`--stack-name` flag + bundle passthrough)

- [x] Component hardcoded `cli`; stack name from `loadConfig().stack_name` (omit `app/` when unset).
- [x] Wire-capture test mirroring Task 3 + `auth.test.ts` assertion that the Cognito client constructor receives `customUserAgent`.
- [x] ~~Risk check: type-sync drift hook~~ **Resolved:** `CliConfig` is already in `CLI_ONLY_ALLOWLIST` (`scripts/check-types-sync.ts:156`) — adding `stack_name?` requires no CDK mirror.
- [x] `cd cli && mise run build`. Commit.

### Task 7 — Docs + full verification

- [x] `AGENTS.md` “Common mistakes”: new bullet — every new AWS SDK client must go through the ABCA UA helpers (`agent/src/ua.py` · `cdk/src/handlers/shared/ua.ts` · `cli/src/ua.ts`); naked `boto3.client(...)` / `new XClient({})` silently drops solution attribution (addresses @krokoko’s [comment](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/319#issuecomment-4684918297)). Root-level file — no Starlight sync needed.
- [x] Header comments in each `ua.ts`/`ua.py` stating the wire format and pointing at #319.
- [x] `mise run build` (agent quality → cdk → cli → docs), `mise run security:sast`. Sweep the acceptance-criteria checklist below; mark PR ready.

## Acceptance-criteria map

| Criterion (#319) | Where satisfied |
|---|---|
| Static baked once, raw path (not app-id field) | Tasks 1–4, 6; wire tests assert the `/` survives |
| `{STACKNAME}` sanitize-then-clip ≤34, value ≤50; test w/ `/`, `#`, non-ASCII | Tasks 1, 3 sanitize tests |
| Stable per-component `{IDENTIFIER}`; per-surface test | `ABCA_COMPONENT` env + Task 5 template assertions + wire tests |
| `{TRACE}` per-request, never baked, optional; same-client different-trace test; no trailing `#`; sanitized | Tasks 1–4, 6 wire tests |
| Outgoing-header inspection at wire layer, both segments, trace-present & -absent | botocore `before-send` short-circuit capture; JS stub `requestHandler` capture |
| Agent centralized in `aws_session.py`/`config.py`; singleton pool preserved | Task 2 |
| All JS handler clients covered via shared helper | Tasks 3–4 |
| CLI Cognito client covered | Task 6 |
| `{STACKNAME}` threaded via env / container env / CLI config — not hard-coded | Tasks 5–6 |
| CDK compile/synth/test + agent tests pass; no functional change | Tasks 2, 4, 5, 7 |
| Docs instruction for future SDK calls (issue comment) | Task 7 |

## Out of scope (per issue)

Dashboards/querying of attributed usage (#215/#245); the deploy-time `uksb` stack-description token (#292, done).


agent/src/config.py
```diff
@@ -101,13 +101,15 @@ def resolve_linear_api_token(channel_metadata: dict[str, str] | None = None) ->
         import json
         from datetime import datetime, timedelta
 
-        import boto3
+        import boto3  # noqa: F401 — availability probe; client built via platform_client
```
scottschreckengaust, line 104: Is the `import boto3` still here once and ignored via "noqa" to ensure it is available? Why not let it fail at the `platform_client`?

scottschreckengaust (commented):
Reviewing

scottschreckengaust:
## Recommend a near-total simplification — verified against the installed botocore 1.43.9 + JS v3

Both SDKs honor **`AWS_SDK_UA_APP_ID` natively** (botocore `configprovider.py` maps it to `user_agent_appid`; JS v3 `NODE_APP_ID_CONFIG_OPTIONS.environmentVariableSelector`). Verified renders:

| `AWS_SDK_UA_APP_ID` | emitted `app/` segment |
|---|---|
| `uksb-wt64nei4u6` | `app/uksb-wt64nei4u6` |
| `uksb-wt64nei4u6#orchestr` | `app/uksb-wt64nei4u6#orchestr` (`#` survives; 24 ≤ 50) |
| `""` (empty) | *(none — free opt-out)* |
| unset | *(none)* |

So `app/uksb-wt64nei4u6` — optionally carrying the component as `#agent` — is emitted with **zero code**. The `ua.py` / `ua.ts` ×2 modules (482 src + 624 test lines), the raw `customUserAgent`/`user_agent_extra` path, the per-request `before-send`/middleware trace appender, the module-level trace state, and the ~60 client-site rewrites all exist only to carry the two self-imposed extras: the `/{STACKNAME}` slash form and the per-request `#{TRACE}`.

### Prescription — collapse to native config
1. One CDK context param `sdkUaAppId`, default `uksb-wt64nei4u6` (or per-surface `uksb-wt64nei4u6#<component>`), matching the existing `tryGetContext('stackName')` pattern in `main.ts`.
2. Thread it as the `AWS_SDK_UA_APP_ID` env var to every Lambda + AgentCore runtime + ECS container — **reusing the exact env-threading this PR already built** for `ABCA_STACK_NAME`/`ABCA_COMPONENT`. botocore and JS pick it up automatically; the agent needs no `aws_session` change, the CLI just exports it.
3. Delete the three `ua` modules, the trace middleware/before-send handler, the module trace state, and the `abcaUserAgent()`/`withAbcaTrace()` edits at the ~60 client sites.
4. Customer opt-out is then free: set the context/env to `""`.

**Net: ~1.5k lines / 83 files → env-var threading + one param + docs/tests.**

### Cost of the simplification (honest trade-offs)
- **Drop `/{STACKNAME}`** — fold into the app-id as `uksb-wt64nei4u6#<stack>` if per-stack granularity is truly wanted (native, no raw path). Largely redundant with the deploy-time `uksb` token (#292) + per-account usage anyway.
- **Drop per-request `#{TRACE}`** — X-Ray, the existing requestId/structured logs, and #245 own real trace correlation; AWS solution attribution counts calls by solution id and does not consume a UA trace handle.

Full verified SDK snippets + the kept-vs-dropped breakdown are in #319. **The research and implementation here are high-quality and correct — this is a scope call, not a defect.** The `/{STACKNAME}` + `#{TRACE}` requirements in #319 are what force the raw path; relax those two (they're ours) and the feature becomes native config.

scottschreckengaust:
**Correction — you're right, `md/` stays.** It survives as a *static, baked-once* segment, not the machinery. Verified together on a real STS wire request (botocore 1.43.9):

```
... app/uksb-wt64nei4u6 ... md/uksb-wt64nei4u6#agent
```

- `app/uksb-wt64nei4u6` ← native `AWS_SDK_UA_APP_ID` env (zero code)
- `md/uksb-wt64nei4u6#agent` ← static `user_agent_extra` (botocore) / `customUserAgent` (JS), baked once at construction

**Three tiers — keep the first two, drop the third:**

| Tier | Segment | Mechanism | Verdict |
|---|---|---|---|
| 1 | `app/uksb-wt64nei4u6[#stack]` | native `AWS_SDK_UA_APP_ID` env | **free** — delete all app/ code |
| 2 | `md/uksb-wt64nei4u6#{component}` | static `user_agent_extra`/`customUserAgent` (one line; session-level in agent, per-client spread in JS) | **keep** — small |
| 3 | `#{TRACE}` per-request | before-send handler / JS middleware + module trace state + ~60 `withAbcaTrace` wraps | **drop** — the bulk |

So what actually deletes is the **trace plane** (per-request appender, `set_trace`/`getAbcaTrace`, `configure_session` trace wiring, all `withAbcaTrace()` calls) and the **raw-path `app/` handling + sanitize-then-clip** (native env replaces it). `ua.py`/`ua.ts` collapse to a ~15-line static `md/` helper; client sites keep one static spread arg, no middleware.

Full kept-vs-dropped detail in #319.

scottschreckengaust:
**Confirmed: `#` separator works and removes the last reason for the raw path.**

Verified both SDKs: `AWS_SDK_UA_APP_ID=uksb-wt64nei4u6#backgroundagent-dev` → `app/uksb-wt64nei4u6#backgroundagent-dev` (botocore allows `#` via `allow_hash=True`; JS `UA_VALUE_ESCAPE_REGEX` includes `#`). The `/` was the *only* char that forced `app/` onto the raw `customUserAgent`/`user_agent_extra` path — switch to `#` and `app/` rides the **native `AWS_SDK_UA_APP_ID` env field** with zero code.

Final shape:
- `app/uksb-wt64nei4u6#{stack}` — native env (tier 1, free)
- `md/uksb-wt64nei4u6#{component}` — static `user_agent_extra`/`customUserAgent`, baked once (tier 2, ~15 lines)
- no `{TRACE}` (tier 3 dropped)

**Comment rewrite required:** every `ua.*` docstring says *"NOT the sanitizing app-id field, whose charset excludes `/`"* — with `#` that rationale inverts: the app-id field is now the right mechanism for `app/`. Full breakdown + criteria deltas posted in #319. Want me to push the simplification onto this branch (delete the trace plane + raw-path `app/`, keep static `md/`, rewrite comments), or land it as a fresh PR?

scottschreckengaust:
**Companion draft posted: #345** — the simplified *native-app-id* alternative to this PR, per the discussion above (use `#` separator → `app/` rides the SDK-native `AWS_SDK_UA_APP_ID` env, dropping the raw-path machinery and the per-request trace plane; static `md/#{component}` kept). Both are intentionally in **draft** so reviewers can compare side by side and choose one. This PR (#338) is the full raw-path implementation; #345 is the minimal-footprint option.

scottschreckengaust:
Closing in favor of **#345**, which implements #319 via the native `AWS_SDK_UA_APP_ID` field (AWS-recommended, no per-client middleware). The sanitization concern that motivated this raw-UA-path approach is resolved in #345 by using a `#` separator that is verbatim-safe in both botocore and the JS SDK — see the full rationale on #319: https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/319

Thanks for the alternative — the comparison made the decision clear. 🤖 @scottschreckengaust (orchestrator)

PR #360, opened 2026-06-16
fix(cdk): normalize after cursor case + colon-truncation in approval-scope guard

AshrafBen10:
## Summary

Fixes two correctness bugs in the task API handlers, each with a regression test.

### Bug 1 — lower-case `after` ULID cursor silently returns zero events
`cdk/src/handlers/get-task-events.ts`

`isValidUlid` accepts lower-case callers (it uppercases before matching), and the API contract documents case-insensitive cursors. But the handler fed the raw cursor into the DynamoDB key condition `event_id > :after`. Stored `event_id`s are upper-case Crockford Base32; DynamoDB compares raw bytes, and lower-case ASCII sorts **after** upper-case. So a lower-case cursor is greater than every stored id → the query returns **zero events**, and a consumer silently skips the rest of the event stream (data loss in replay).

**Fix:** upper-case the cursor before the query.

### Bug 2 — colon in a glob value breaks the degenerate-pattern guard
`cdk/src/handlers/shared/create-task-core.ts`

The guard for `bash_pattern:` / `write_path:` scopes used `scope.split(':', 2)[1]`, which truncates the value at the next colon. A value whose leading colon-segment is short or wildcard-heavy (e.g. `ab:cdefgh`) was truncated to a degenerate fragment (`ab`) and spuriously rejected with a 400.

**Fix:** take everything after the first colon via `slice(indexOf(':') + 1)`.

## Testing
- `test/handlers/get-task-events.test.ts` — added lower-case cursor normalization test
- `test/handlers/shared/create-task-core.test.ts` — added colon-value accept + genuine-degenerate reject tests
- All 87 tests across both suites pass; TypeScript compiles cleanly


krokoko:
@AshrafBen10 thanks ! Almost ready, please run a full build locally and push the changes not committed (see failing build step on mutation), we should be good to go after that

krokoko (changes requested):
see comment

krokoko:
@AshrafBen10 could you please push the missing files as per the comment above ? thank you !

krokoko (approved):

Issue #365, opened 2026-06-16
perf(ci): shard the CDK jest suite across a job matrix (#363 item #2)


scottschreckengaust:
## Summary

Implement **item #2** of the build-performance umbrella (#363): shard the CDK Jest suite across a CI job matrix so the now-dominant `//cdk:test` (~298s after #357/#359) runs in parallel slices and aggregates into the single required `build` check.

Design is documented in `docs/design/CI_BUILD_PERFORMANCE.md` (PR #364, §"Item #2 — sharding the CDK suite").

Parent: #363. Predecessor: #357 (done, PR #359 merged).

## Approach (fan-out + aggregate gate)

```
build-shard (matrix: shard ∈ [1..4])   # parallel; NOT individually required
        │
        ▼
build (needs: build-shard)             # single required context on merge_group
```

- `build-shard` runs `jest --shard=${{ matrix.shard }}/4` for the CDK suite (~119 test files) + uploads per-shard coverage.
- A single `build` job `needs: [build-shard]`, fails if any shard failed, and remains the one required status context (stable name regardless of shard count).
- Non-test tasks (`//cdk:synth:quiet`, `//cli:*`, `//docs:build`, `//agent:quality`, drift checks) and the deploy artifact (`cdk-agentcore-out`) produced exactly once, off the shard critical path.
- Cross-shard **coverage merge** before threshold enforcement (each shard sees only its slice).
- Self-mutation/drift check runs once (aggregate job), not per-shard.

## Acceptance criteria

- [ ] CDK Jest runs as a 4-way shard matrix; all 2041 tests execute exactly once across shards (sum of per-shard counts = 2041).
- [ ] The required `build` check remains a single stable context that reports on `merge_group` (#327) and fails if any shard fails.
- [ ] Coverage thresholds (branches 82 / funcs 94 / lines 91 / stmts 91) enforced once on the **merged** report; not per-shard.
- [ ] Deploy artifact `cdk-agentcore-out` still produced exactly once; deploy.yml unaffected.
- [ ] Self-mutation / "Fail build on mutation" check still runs (once).
- [ ] Real **4-core CI** before/after posted: target slowest-shard + aggregate wall-clock ≈ ~170s (vs ~298s single), build step toward ~5–6 min.
- [ ] No new flakiness; `mise run build` semantics preserved for local (non-CI) runs.

## Implementer notes (from the design doc)

[...]

PR #371, opened 2026-06-17
perf(test): disable Lambda bundling in CDK unit-test synths (~15x per synth) (#366)

### LINKED ISSUE #366: perf(test): disable Lambda bundling in CDK unit-test synths (~15x per synth) (#363)
## Summary

Reduce `//cdk:test` wall time by eliminating **repeated full-stack synthesis** in the handful of mega-suites that dominate the CDK Jest suite. Profiling (see #365) shows the cost is concentrated: the top 5 suites are **362s of 437s suite-time (83%)**, and the single heaviest file is **135.6s**. Each `new App()` + `Template.fromStack()` is ~27s; several suites synthesize the same stack multiple times where one shared synth would do.

This is the **predecessor to sharding (#365)** — lowering the per-file floor is both a bigger, simpler win on its own AND what makes sharding worthwhile later. Parent: #363.

## Evidence (8-core local, coverage on)

| Suite | Time | Full synths in file | Notes |
|---|---|---|---|
| `test/stacks/github-tags.test.ts` | 135.6s | 4 (`synthWithTags`) | 2 hoisted to `beforeAll`, **2 still inline** (`empty string`, `compute_type ecs` tests) |
| `test/constructs/task-api.test.ts` | 83.5s | 12 (4 `beforeAll` + inline) | audit which inline synths can share a template |
| `test/stacks/agent.test.ts` | 50.2s | 2 | already mostly `beforeAll`; verify |
| `test/constructs/task-orchestrator.test.ts` | 49.9s | 2 | already mostly `beforeAll`; verify |
| `test/bootstrap/synth-coverage.test.ts` | 44.4s | 2 | by-design coverage synth; check if reducible |

## Approach

For each heavy suite: synthesize each **distinct stack configuration exactly once** (in `beforeAll`), and have individual tests assert against the cached `Template`. Only synth again when a test genuinely needs a *different* context/config. Do **not** weaken assertions or merge tests — only remove redundant synthesis.

Concrete first target — `github-tags.test.ts`: hoist the two remaining inline `synthWithTags(...)` calls (`empty string values` and `compute_type: 'ecs'`) into shared `beforeAll` templates keyed by their context, reused across tests. Est. ~−54s on this file alone.

## Acceptance criteria

- [ ] Each targeted suite synthesizes each distinct stack config at most once; redundant inline synths removed.
- [ ] All tests still pass (2177 CDK tests) with identical assertions; coverage thresholds unchanged and met.
- [ ] `//cdk:test` wall-time improvement reported on real **4-core CI** (before/after); target: meaningfully below the current ~298s, driven by the top-5 suites.
- [ ] No reduction in what is asserted; no test merged away to fake a speedup.
- [ ] `mise run build` clean; no drift; no new flakiness.

## Notes for implementer

- `new App()` + `Template.fromStack()` is the expensive operation (~27s here) — cdk-nag aspects run during synth. Counting these per file is the quickest way to find waste: `grep -cE "new App\(|Template.fromStack" <file>`.
- Some suites already use `beforeAll` correctly — verify before changing; the win is only in the files that synth the *same* config repeatedly.
- After this lands, re-profile and decide whether #365 (sharding) is still worth it — the floor may be low enough that the unsharded suite is already fast.

## References

- Parent umbrella: #363
- Sharding (deferred predecessor-of): #365
- Design + profiling: `docs/design/CI_BUILD_PERFORMANCE.md` (PR #364)

---

scottschreckengaust:
## What

Disables Lambda asset bundling during CDK **unit-test** synthesis. `Template.fromStack()` does a full CDK synth, which bundles every `NodejsFunction` via esbuild — ~28s for the 413-resource `AgentStack`. Unit tests assert on CloudFormation structure/properties, not bundled Lambda code, so the bundling is pure overhead.

A jest `setupFiles` hook (`cdk/test/setup/disable-bundling.ts`) sets `aws:cdk:bundling-stacks: []` via `CDK_CONTEXT_JSON`. A bare `new App()` — the dominant pattern across our tests (87 call sites) — picks this up with **zero call-site changes**.

Closes #366. Part of #363.

## Results (8-core local)

| Metric | Before | After |
|---|---|---|
| Single `AgentStack` synth | 28.7s | **1.9s** (~15×) |
| `github-tags.test.ts` | ~135s | **9s** |
| Full `//cdk:test` | ~155s | **25s** (~6×) |

- **All 2177 CDK tests pass.** Coverage thresholds unchanged and met (lines 92.36 ≥ 91, branches 82.81 ≥ 82).
- Real **4-core CI** before/after will post on this PR's build run.

## Why this over sharding (#365)

Profiling under #366 showed the CDK suite's cost is per-synth esbuild bundling, not test count or redundant synths. This single change beats 4-way sharding (~14%) and synth-dedup (~0%) by a wide margin, with no CI complexity. **#365 (sharding) is likely now moot** — to be re-evaluated once this lands and we see the new CI baseline.

## Durable regression guards

- **AGENTS.md** "Common mistakes": do not re-enable bundling in tests unless asserting on a bundled asset; minimize full-stack synths (synth once in `beforeAll`).
- **`.abca/commands/review_pr.md`**: a Stage-3 checklist item so PR review catches regressions.

## Safety / blast radius

- Audited all 31 stack-synth suites — **all green**. The one asset-hash reference (`test/bootstrap/version.test.ts`) is a *bootstrap policy* hash, not a CDK asset hash, and synthesizes no stack — unaffected.
- A test that genuinely needs bundled assets can opt out by setting `aws:cdk:bundling-stacks: ['**']` in its own `App` context (documented in the setup file).

## Dependency note

The AGENTS.md / review_pr.md links point to `docs/design/CI_BUILD_PERFORMANCE.md`, which is added by **PR #364** (the design doc, currently draft). Merge #364 first (or together) so the links resolve.

## Draft

Opened as **draft** while under review per WIP discipline. Will mark ready once the 4-core CI build posts green and #364 ordering is settled.

---


isadeks (commented):
Genuinely good, well-scoped optimization. I verified the core mechanism empirically against `aws-cdk-lib@2.257.0`: `bundlingRequired` reads context key `aws:cdk:bundling-stacks` defaulting to `["**"]`, so setting it to `[]` makes `[].some(...)` return `false` and a bare `new App()` (the dominant test pattern) picks it up via `CDK_CONTEXT_JSON` with zero call-site changes. The `~15×` synth speedup is real and the safety audit checks out.

## ✅ What I verified
- **Mechanism is correct** — `new App()` reads `CDK_CONTEXT_JSON`; `aws:cdk:bundling-stacks: []` disables bundling (`bundlingRequired === false`). Confirmed empirically.
- **Safety claim holds** — `version.test.ts` computes a *bootstrap-policy* hash and synthesizes no stack, so it's unaffected by disabling bundling.
- **No collateral** — no existing CDK test asserts on bundled asset hashes / S3 keys, and none relies on a constructor-context bundling toggle, so the "all 2177 tests pass" claim is consistent.
- **Setup-file internal merge is self-consistent** — when `CDK_CONTEXT_JSON` is pre-set, the existing value wins, matching the file's comment.

## 🚫 Blockers
None.

## 🔧 Should fix before merge

**1. The documented per-test opt-out does not work (latent bug).**
The escape hatch documented in all three places (`disable-bundling.ts` comment, `AGENTS.md`, `review_pr.md`) —

```ts
new App({ context: { 'aws:cdk:bundling-stacks': ['**'] } })
```

— does **not** re-enable bundling. Verified empirically against `aws-cdk-lib@2.257.0`:

| App construction | `stack.bundlingRequired` |
|---|---|
| bare `new App()` | `false` (disabled ✅ — intended) |
| `new App({ context: { 'aws:cdk:bundling-stacks': ['**'] } })` | `false` ❌ (PR claims `true`) |
| `new App({ postCliContext: { 'aws:cdk:bundling-stacks': ['**'] } })` | `true` ✅ |

Root cause: CDK's `App.loadContext(props.context, props.postCliContext)` sets `props.context` first, then **overwrites it with `CDK_CONTEXT_JSON`**, then applies `postCliContext` last. So the env var beats constructor `context`.

Impact: a future author who needs a real bundled asset (asset-hash / S3-key assertion) follows the docs, gets a silently-unbundled synth, and a wrong/empty asset hash with no error. Not a current-test failure — no test uses this path today — so it's latent.

Fix: document `postCliContext` as the opt-out (it actually overrides), and correct the "Tests that pass their own `{ context }` still merge on top" comment, which is wrong for this key.

## 💬 Other (self-disclosed)

**2. Dangling doc link.** `AGENTS.md` → `./docs/design/CI_BUILD_PERFORMANCE.md` doesn't exist on this branch or main (added by draft PR #364). The PR body already flags the merge-order dependency. Note it intersects with the link-checker PR #300: `AGENTS.md` is a root `.md` within that checker's scope, so landing #371 + #300 without #364 would turn the dead link into a red CI check. (`review_pr.md` has the same link but `.abca/` is outside the checker's scope, so only AGENTS.md would trip CI.)

---
*Reviewed at xhigh effort. Mechanism and precedence claims verified empirically against `aws-cdk-lib@2.257.0`, not just read from source.*

scottschreckengaust:
@isadeks Thanks for the empirical verification — you're exactly right, and I confirmed it. I've corrected the opt-out to `postCliContext` when tests need the bundled asset and rewrote the precedence comment in all three documented locations.

isadeks:
## Re-review: ✅ Both findings resolved — recommend approve

Re-reviewed at head `95ebf79`. The three new commits map cleanly onto both findings from the prior pass:

| Finding | Fix | Status |
|---|---|---|
| **1. Broken opt-out** — docs said use constructor `context`, which the env var clobbers | `82831aa` correct opt-out to `postCliContext` + `c1872c0` clarify it's for asset output, not synth | ✅ **Fixed & verified empirically** |
| **2. Dead doc link** — `CI_BUILD_PERFORMANCE.md` (draft #364) | `95ebf79` drop dead links | ✅ **Fixed** |

**Finding 1** — verified against `aws-cdk-lib@2.257.0` with the setup file's env var active:

| App construction | `stack.bundlingRequired` |
|---|---|
| `new App({ postCliContext: { 'aws:cdk:bundling-stacks': ['**'] } })` | `true` ✅ (the documented opt-out — now works) |
| `new App({ context: { 'aws:cdk:bundling-stacks': ['**'] } })` | `false` ✅ (docs now correctly warn this does NOT work) |

The corrected AGENTS.md / `disable-bundling.ts` prose is technically accurate: the `loadContext(props.context, props.postCliContext)` precedence explanation matches CDK's actual behavior (env var overwrites constructor `context`; `postCliContext` is applied last and wins). The added "this doesn't stop the synth, only skips the esbuild step" clarification is correct and heads off a likely misconception.

**Finding 2** — zero `CI_BUILD_PERFORMANCE.md` references remain in any of the three touched files, so the hard dependency on #364 is severed. I also confirmed all 8 remaining `.md` links in the touched docs resolve on the PR head — no new dead links introduced.

Core mechanism, the `version.test.ts` safety claim, and the "no test relies on bundled-asset assertions" audit were confirmed in the prior pass and are unchanged.

**Nothing left blocking.** One cosmetic residual: the **PR description's** dependency note still mentions the #364 merge ordering, which is now stale since the links are gone — worth trimming the description, but it's not a code issue. Good to mark ready once the 4-core CI build posts green.

isadeks (dismissed):

theagenticguy (commented):
## Third-pass review — follow-up items

Independent pass after @isadeks's two reviews. **The approve holds**; the inline items are for a follow-up commit. I re-verified the load-bearing claims against the CDK version this branch actually resolves (`aws-cdk-lib@2.259.0` per the lockfile — the prior pass verified `2.257.0`; both satisfy `^2.257.0` and `loadContext` precedence is identical at `2.259.0`).

`integ-smoke` is excluded as out of scope — its failure is an environmental `Could not load credentials from any providers` on the `integ` environment OIDC gate, unrelated to this diff.

### Inline findings
1. **Add an executable regression test that bundling is off** (highest value — it's the point of the PR). Guards today are honor-system docs.
2. **Cover the documented `postCliContext` opt-out** — currently zero test coverage.
3. **Reframe "pure overhead"** — bundling also smoke-tests Lambda entry resolution, which unit tests now cede.

### Finding 4 (PR body, no line anchor)
Post the **4-core CI** before/after. `~15×` is single-synth on 8-core local and `~6×` full-suite is also local; the figure that justifies this for everyone is 4-core CI wall-clock, which the description promises and hasn't posted yet. The mechanism earns the approve; the aggregate CI win is still unmeasured on CI hardware.

### Confirmed during review (skip on re-check)
| Claim | Verification | Result |
|---|---|---|
| All synth tests covered by the single `setupFiles` | All 73 `Template.fromStack` sites live under `cdk/` | no orphans |
| Coverage thresholds unaffected | Disabling bundling skips esbuild transpilation, never executes handler lines; coverage measures executed lines | not gamed |
| No test asserts on bundled-asset output | `grep assetHash\|S3Key\|.zip` across `cdk/test/` → 0 hits; sole snapshot is the bootstrap-policy hash, which synthesizes no stack | safety audit holds |

**Net:** approve stands once #1 lands; #2–#4 are polish and can ride the same commit.

cdk/test/setup/disable-bundling.ts
```diff
@@ -0,0 +1,57 @@
...
+// test must opt out via `postCliContext` (which wins over the env var),
+// constructing its `App` with
+// `new App({ postCliContext: { 'aws:cdk:bundling-stacks': ['**'] } })`
+// (or the specific stack id).
+const BUNDLING_DISABLED_CONTEXT = { 'aws:cdk:bundling-stacks': [] as string[] };
+
+const existing = process.env.CDK_CONTEXT_JSON;
+if (existing) {
+  // Preserve any context already provided; our key wins only if unset.
+  const merged = { ...BUNDLING_DISABLED_CONTEXT, ...JSON.parse(existing) };
```
theagenticguy: **Finding 1 (highest value): no executable guard that bundling is actually off.**

This file is the entire optimization, but the only regression guards are `AGENTS.md` prose and a `review_pr.md` checklist item — both honor-system. If someone drops the `setupFiles` entry, reorders Jest setup, or a CDK rename breaks the context key, the suite silently reverts to ~155s with **no red check**. For a PR whose whole value is the speedup, the floor should be machine-enforced. Five lines:

```ts
// cdk/test/setup/disable-bundling.test.ts
import { App } from 'aws-cdk-lib';

it('unit-test synth disables Lambda bundling globally', () => {
  expect(new App().node.tryGetContext('aws:cdk:bundling-stacks')).toEqual([]);
});
```

_Two optional hardening nits on the merge block below, same file:_
- `const merged = { ...BUNDLING_DISABLED_CONTEXT, ...JSON.parse(existing) }` lets a pre-set `CDK_CONTEXT_JSON` carrying this key win and silently defeat the disable. Since the per-test opt-out is `postCliContext` (applied after the env var regardless), flipping the spread to `{ ...JSON.parse(existing), ...BUNDLING_DISABLED_CONTEXT }` would make the global disable robust to ambient env while keeping the documented escape hatch working. Low real-world risk: the Jest path builds `new App()` directly and never hits the CLI that sets this var.
- `JSON.parse(existing)` is unguarded — a malformed pre-set var throws inside `setupFiles` and fails *every* test in *every* file with an opaque parse error. A `try/catch` emitting "malformed CDK_CONTEXT_JSON" localizes the blame.

cdk/test/setup/disable-bundling.ts
```diff
@@ -0,0 +1,57 @@
...
+// stacks': ['**'] } })` does NOT re-enable bundling (the env var clobbers it).
+//
+// This does NOT stop tests from synthesizing — `Template.fromStack()` still
+// runs a full synth; it only skips the esbuild asset-bundling step within that
+// synth. The opt-out below exists solely for the rare test that needs the
+// *bundled-asset output itself* (e.g. asserting on a real asset hash / S3 key),
+// where an unbundled synth would silently yield a placeholder value. Such a
+// test must opt out via `postCliContext` (which wins over the env var),
+// constructing its `App` with
+// `new App({ postCliContext: { 'aws:cdk:bundling-stacks': ['**'] } })`
```
theagenticguy, line 48: **Finding 2 (pairs with #1): the documented opt-out has zero test coverage.**

No test constructs `new App({ postCliContext: { 'aws:cdk:bundling-stacks': ['**'] } })`, so the one escape hatch these docs promise — and the `loadContext` precedence corrected in this PR — is unguarded against a future CDK version reordering context application. I verified the precedence against `aws-cdk-lib@2.259.0` (the version the lockfile resolves; the prior review verified `2.257.0`). Lock the contract:

```ts
import { App } from 'aws-cdk-lib';
import { AgentStack } from '../src/stacks/agent';

it('postCliContext re-enables bundling (documented opt-out)', () => {
  const app = new App({ postCliContext: { 'aws:cdk:bundling-stacks': ['**'] } });
  const stack = new AgentStack(app, 'optout', {
    env: { account: '123456789012', region: 'us-east-1' },
  });
  expect(stack.bundlingRequired).toBe(true);
});
```

AGENTS.md
```diff
@@ -45,6 +45,7 @@ Handler entry tests: `cdk/test/handlers/orchestrate-task.test.ts`, `create-task.
...
+- **Bundling Lambda assets in CDK unit tests** — `Template.fromStack()` triggers a full synth that bundles every `NodejsFunction` via esbuild (~28s for `AgentStack`). Unit tests assert on CloudFormation structure, not bundled code, so this is pure overhead. The `cdk/` Jest config disables it globally via `test/setup/disable-bundling.ts` (sets `aws:cdk:bundling-stacks: []` in `CDK_CONTEXT_JSON`), which a bare `new App()` picks up automatically. This does not stop tests from synthesizing — `Template.fromStack()` still runs a full synth; it only skips the esbuild step. **Do not re-enable bundling** in a test unless you are specifically asserting on the bundled-asset output (hash / S3 key) — and if you must, opt out narrowly via that test's `App` `postCliContext` (e.g. `new App({ postCliContext: { 'aws:cdk:bundling-stacks': ['**'] } })`), not globally. Note that constructor `context` does **not** work for this key: CDK overwrites it with `CDK_CONTEXT_JSON`, so only `postCliContext` (applied last) overrides the global disable. Minimize full-stack synths regardless: synthesize each distinct stack config once in `beforeAll` and assert against the cached `Template`. See #366.
```
theagenticguy: **Finding 3 (prose): "pure overhead" undersells the tradeoff.**

esbuild bundling also validates that every Lambda `entry` resolves and the handler compiles. Disabling it in unit tests means a broken entry path or a handler TS error stops surfacing at `Template.fromStack()` and only fails at real synth/deploy or in the `agent/` build. The tradeoff is acceptable — real synth/deploy still bundles, and handler code has its own compile/test under `agent/` — but this is **overhead for CloudFormation-structure assertions, plus a smoke-test of entry bundling that unit tests intentionally cede to the `agent/` build**, not flat "pure overhead." Same wording lives in the `disable-bundling.ts` header comment.

cdk/test/setup/disable-bundling.ts
scottschreckengaust: Done in 09b1f24 — added `cdk/test/setup/disable-bundling.test.ts` with an executable guard: the synth context carries `aws:cdk:bundling-stacks: []` and a bare-`App` stack reports `bundlingRequired === false`. Verified empirically against `aws-cdk-lib@2.259.0` that without the disable a bare `App` yields `bundlingRequired === true`, so the assertion discriminates a real regression rather than being a tautology.

Both hardening nits applied in 94c604e: the merge now spreads `BUNDLING_DISABLED_CONTEXT` last (`{ ...JSON.parse(existing), ...BUNDLING_DISABLED_CONTEXT }`) so an ambient `aws:cdk:bundling-stacks` can no longer win — safe because the opt-out is `postCliContext`, applied after the env var regardless. `JSON.parse` is now wrapped in try/catch and re-throws with the offending value so a malformed pre-set var localizes the blame instead of failing every test opaquely.

cdk/test/setup/disable-bundling.ts
scottschreckengaust: Done in 77b0c7a — added two assertions locking the precedence contract (verified against `aws-cdk-lib@2.259.0`): `postCliContext: { 'aws:cdk:bundling-stacks': ['**'] }` re-enables bundling (`bundlingRequired === true`), and constructor `context` does NOT (`bundlingRequired === false`, the env var clobbers it). I used a bare `Stack` rather than `AgentStack` — the precedence is a property of CDK's `loadContext`, not of any particular stack, so a bare `Stack` keeps the guard fast and avoids the heavy AgentStack construction while still exercising the exact `(props.context, CDK_CONTEXT_JSON, props.postCliContext)` ordering. Let me know if you'd prefer it pinned to `AgentStack` specifically.

AGENTS.md
scottschreckengaust: Done in 3f6ff1f — reworded in both places the phrase lived (the `disable-bundling.ts` header and `AGENTS.md`). It now reads as overhead *for CloudFormation-structure assertions, plus a smoke-test of Lambda `entry` resolution and handler compilation that unit tests intentionally cede to the `agent/` build* — and notes that a broken entry path or handler TS error stops surfacing at `Template.fromStack()` and instead fails at real synth/deploy or in the `agent/` build, which still bundles.

scottschreckengaust:
## Follow-up commits for @theagenticguy's third-pass review

Findings 1–3 addressed in four commits on `feat/366-synth-reuse`:

- **94c604e** `fix(test)` — harden the merge against ambient `CDK_CONTEXT_JSON` (spread the disable last; guard `JSON.parse`). *(finding 1 hardening nits)*
- **09b1f24** `test(cdk)` — executable guard that the global disable is on. *(finding 1)*
- **77b0c7a** `test(cdk)` — cover the `postCliContext` opt-out precedence (+ the negative: constructor `context` does not work). *(finding 2)*
- **3f6ff1f** `docs(test)` — reframe "pure overhead" as overhead-plus-entry-smoke-test in both `disable-bundling.ts` and `AGENTS.md`. *(finding 3)*

### Finding 4 — real CI numbers (now measured, not estimated)

Pulled from `build` workflow runs straddling the change, **same runner, same 2177 tests** — so the only variable is the bundling toggle:

| Run | Commit | Bundling | `cdk:test` (jest `Time:`) | `build` step | job wall-clock |
|---|---|---|---|---|---|
| [27650386806](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/actions/runs/27650386806) | `86fda643` (parent of perf commit) | **on** | **329.2s** | 376s | 502s |
| [27661401811](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/actions/runs/27661401811) | `6cf7a984` (first branch commit) | **off** | **92.8s** | 140s | 260s |

**Realized CI speedup: `cdk:test` 3.55× (329.2s → 92.8s); full `build` step 2.69× (376s → 140s); whole job 1.93× (502s → 260s).**

Two corrections to the original framing:

1. **CI runs on `ubuntu-latest` (2-core)**, not 4-core — there's no `DEFAULT_RUNNER_LABEL` repo var and no `ubuntu-latest-4-cores` label on this PR (confirmed via the runs' `runner labels`). So the relevant "everyone" figure is the 2-core column above.
2. The local `~6×` full-suite figure doesn't fully transfer to fewer cores — on 2-core CI the realized `cdk:test` win is **3.55×** (the fixed esbuild cost amortizes differently across core counts). Still well clear of the noise: a ~236s reduction on the test step and **~4 min off the build job's wall-clock**, every PR.

For reference, the latest run with my four follow-up commits ([27725228001](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/actions/runs/27725228001), `3f6ff1f`) holds at `cdk:test` **106.4s** / build step **142s** for 2193 tests — the +16 new tests add ~14s and the win is intact.

theagenticguy (approved):
## Approve

Re-reviewed at head `3f6ff1ff`. All three inline findings from my prior pass are addressed, and two of them went past what I asked. Verified each against the new commits, not the commit messages.

| Finding | Commit | Verified |
|---|---|---|
| **1. Executable bundling-off guard** | `09b1f24d` | `disable-bundling.test.ts` asserts `aws:cdk:bundling-stacks === []` **and** probes `bundlingRequired === false` on a real stack — exceeds the ask |
| **2. Opt-out coverage** | `77b0c7a9` | Tests the positive `postCliContext` path (`bundlingRequired === true`) **and** a negative case proving constructor `context` does *not* work — locks the exact footgun the docs warn about |
| **3. "Pure overhead" reframe** | `3f6ff1ff` | AGENTS.md now reads "overhead — plus a smoke-test of Lambda `entry` resolution and handler compilation that unit tests intentionally cede to the `agent/` build." Honest framing |
| **Hardening (the two optional nits)** | `94c604ed` | Merge spread flipped to `{ ...ambient, ...BUNDLING_DISABLED_CONTEXT }` so the disable wins unconditionally, and `JSON.parse` is wrapped in a `try/catch` that throws a localized "malformed CDK_CONTEXT_JSON" error instead of failing every test opaquely |

The precedence the opt-out test asserts matches `aws-cdk-lib@2.259.0`'s `loadContext` (the version the lockfile resolves), which I read directly in the earlier pass — `props.context` first, `CDK_CONTEXT_JSON` overwrites, `postCliContext` last. `build (agentcore)` is green at this head, consistent with the new tests passing.

### Two non-blocking residuals (not gating this approval)
- **Finding 4 — the 4-core CI number** is still a placeholder in the PR description ("Real 4-core CI before/after will post"). It's the figure that justifies the change for everyone, so worth dropping the real before/after into the body once the build posts. Body-only, gated on the CI run — not a code issue.
- **`integ-smoke` is red** and it's a required check, so it will block merge until cleared. The failure is environmental (`Could not load credentials from any providers` on the `integ` environment OIDC gate), unrelated to this diff — it needs the admin approval on the `integ` environment to go green, it won't clear on its own.

Code is correct and the regression floor is now machine-enforced. Approving.

PR #372, opened 2026-06-17
fix(jira): seed explicit JSON placeholder for stack-wide webhook secret (#368)

### LINKED ISSUE #368: fix(jira): setup never seeds stack-wide webhook secret → admin-UI webhooks 401 (silent trigger failure)
## Summary

`bgagent jira setup` stores the operator-pasted webhook signing secret **only** on the per-tenant OAuth bundle, never as the stack-wide secret. Because Jira admin-UI ("generic") webhooks **do not include `cloudId`** in their payload, the receiver's per-tenant verification path is skipped and it falls back to the **stack-wide** secret — which still holds CDK's auto-generated placeholder. Result: every webhook delivery fails HMAC verification with `401 Invalid signature` and no task is ever created. The failure is silent to the operator (no comment on the issue, no error surfaced).

This was hit in a real single-tenant setup: deliveries logged `"Invalid Jira webhook signature"` on every request until the stack-wide secret was manually overwritten in Secrets Manager to match the pasted value.

## Root cause

`cli/src/commands/jira.ts:542` gates seeding of the stack-wide secret on `isWebhookSecretConfigured()`:

```ts
const stackWideAlreadyConfigured = await isWebhookSecretConfigured(sm, webhookSecretArn!);
if (stackWideAlreadyConfigured) {
  console.log('  ✓ Stack-wide fallback already configured (leaving as-is)');
} else {
  await sm.send(new PutSecretValueCommand({ SecretId: webhookSecretArn!, SecretString: webhookSigningSecret }));
}
```

`isWebhookSecretConfigured` (`cli/src/commands/jira.ts:196`) treats **any value not starting with `{`** as "already configured by an operator":

```ts
// CDK's auto-generated secret is a JSON object string starting with `{` ...
return !value.trim().startsWith('{');
```

But CDK's `secretsmanager.Secret` for the Jira webhook (`cdk/src/constructs/jira-integration.ts:154`) is created with **no `generateSecretString`**, so its auto-generated value is a **bare random string**, not JSON. The placeholder therefore does not start with `{`, `isWebhookSecretConfigured` returns `true`, and `setup` never seeds the operator value. The two secrets stay permanently divergent.

## Impact

- Every single-tenant Jira install that follows the documented setup steps gets 401 on **all** webhook deliveries — the label trigger appears completely dead with no diagnostics.
- Only recoverable today by manually `put-secret-value` on the stack-wide ARN.

## Proposed fix (open to alternatives)

Make the placeholder reliably detectable, then seed correctly. Options:
1. Give the CDK secret an explicit JSON placeholder via `generateSecretString` (e.g. `{"placeholder":true}`) so the existing `startsWith('{')` heuristic works as the comment claims; **and/or**
2. Replace the heuristic with an explicit sentinel/empty check, or track "operator-set" state explicitly rather than inferring it from secret shape.
3. Consider always seeding the stack-wide secret from the **first** tenant's pasted value regardless of placeholder shape (matching the documented single-tenant back-compat intent).

## Acceptance criteria

- [ ] After a fresh deploy + `bgagent jira setup` (single tenant, admin-UI webhook with no `cloudId`), a Jira webhook delivery verifies successfully (receiver logs "verified via stack-wide fallback secret"), with **no** manual Secrets Manager edit.
- [ ] The CDK webhook-secret placeholder is unambiguously distinguishable from an operator-set value (documented + asserted in a unit test).
- [ ] Unit test covers `isWebhookSecretConfigured` (or its replacement) returning `false` for the CDK placeholder and `true` for an operator-set bare-string secret.
- [ ] Re-running `setup` does not clobber an already-operator-set stack-wide secret (idempotency preserved).
- [ ] Linear parity path (`cli/src/commands/linear.ts:166`) reviewed for the same class of bug; note in the issue whether it is affected (Linear auto-generates `lin_wh_`-prefixed secrets and payloads carry workspace id, so likely not — confirm).

## Notes

Receiver verification logic: `cdk/src/handlers/jira-webhook.ts:131-168`; per-tenant vs stack-wide selection at line 133 (`WORKSPACE_REGISTRY_TABLE && payload.cloudId`). Signature compare: `cdk/src/handlers/shared/jira-verify.ts:114`.

mayakost:
## What & why

Fixes #368.

Admin-UI Jira "generic webhooks" omit `cloudId` from their payload, so the receiver (`jira-webhook.ts`) can't use the per-tenant verification path and falls back to the **stack-wide** webhook secret. But `bgagent jira setup` only seeds that stack-wide secret when it looks unconfigured, and its heuristic — `!value.trim().startsWith('{')` — misread CDK's **bare-string** auto-generated value as an operator-set secret. So the secret the operator pasted was stored only per-tenant and **never** applied stack-wide. Every delivery then failed HMAC verification with a silent `401`, and (because the failure-feedback comment also fails) the label trigger looked completely dead.

The CLI comment claimed CDK's placeholder is "a JSON object string starting with `{`", but the `secretsmanager.Secret` had **no** `generateSecretString`, so CDK actually seeded a bare random string. Reality didn't match the assumption.

## Changes

- **CDK** (`jira-integration.ts`): give `WebhookSecret` an explicit JSON placeholder via `generateSecretString`, carrying a marker key (`abca_jira_webhook_placeholder`), so the seeded value genuinely starts with `{` as the CLI assumed.
- **CLI** (`commands/jira.ts`): recognize the placeholder by **parsing the JSON and checking for the marker key**, instead of the fragile `startsWith('{')` shape heuristic. A non-JSON value, malformed JSON, or JSON without the marker is treated as an operator-set secret (so `setup` never overwrites a real value). Exported `isWebhookSecretConfigured` + the marker constant for testability.
- **Tests**: new `cli/test/commands/jira.test.ts` (placeholder detection incl. the #368 case, legacy bare-string, malformed JSON, markerless JSON; plus `map`/`link`/`setup`/`app-template` action paths and `upsertOauthSecret`) and `cdk/test/constructs/jira-integration.test.ts` (asserts the secret synthesizes with the JSON placeholder + marker). Also added `cli/test/commands/linear-helpers.test.ts` — importing the large, previously-untested `jira.ts` pulled it (and its sibling `linear.ts`) into the global coverage denominator, so parallel helper tests were needed to keep the suite above its coverage thresholds.

## Migration note (important for reviewers)

CloudFormation generates `generateSecretString` values **only on resource create**, not on update. So **already-deployed stacks keep their bare-string placeholder** and must **redeploy** (which regenerates the secret with the JSON placeholder) before `bgagent jira setup` will seed it. The CLI conservatively reports a bare-string value as "configured", so it will neither crash nor silently clobber. This is documented in the code (`isWebhookSecretConfigured` JSDoc) and worth calling out in release notes.

## Acceptance criteria status (from #368)

- [x] CDK webhook-secret placeholder is unambiguously distinguishable from an operator value (JSON + marker key), asserted in a construct test.
- [x] Unit test: detection returns `false` for the placeholder and `true` for an operator-set bare string (and markerless JSON).
- [x] Re-running `setup` does not clobber an already-operator-set secret (idempotency preserved — markerless/non-JSON ⇒ "configured").
- [x] Linear parity path reviewed: Linear auto-generates `lin_wh_`-prefixed secrets and its payloads carry the workspace id, so it recognizes *real* secrets by prefix and is **not** affected by this bug.
- [ ] End-to-end (fresh deploy + setup verifies a delivery with no manual SM edit) — not runnable in CI; validated by the construct/CLI tests and described above.

## Verification

- `mise //cli:build` (compile + 401 tests + eslint) ✅
- `mise //cdk:test` (120 suites, 2179 tests) ✅, `mise //cdk:compile` ✅, `mise //cdk:eslint` ✅
- `security:sast:masking` failures are pre-existing on clean `main` (local semgrep not honoring `nosemgrep` annotations); my new annotated line is not among the cited findings.


isadeks (approved):
Verified the fix end-to-end — CDK seeds the JSON placeholder, the CLI recognizes it by parsing + marker key (not the fragile `startsWith('{')`), `setup` writes the bare operator string, and the receiver HMACs that raw value. Idempotent on re-run, and the Linear-parity path is unaffected. 

Approving. Two non-blocking follow-ups:

- **Marker constant drift** — `JIRA_WEBHOOK_SECRET_PLACEHOLDER_KEY` is duplicated in `jira-integration.ts` and `jira.ts` with only a "keep in sync" comment; nothing enforces it. A rename/typo on one side would silently regress #368 with green CI. Worth moving into `contracts/constants.json` (covered by the sync check) or a cross-package equality test.
- **Migration note wording** — the "generated only on create, not on update" rationale is inverted (a redeploy *does* regenerate the secret version). The code behaves correctly; just fix the comment and note that redeploy overwrites the stack-wide value, so re-run `bgagent jira setup` after.

theagenticguy (approved):
LGTM. Post-merge audit: correct, minimal fix for a real silent 401. Verified the diagnosis end-to-end (CDK construct → CLI `setup` → `jira-webhook.ts` → `jira-verify`) and confirmed the migration claim empirically by synthesizing both CDK versions and diffing the CloudFormation output — the `GenerateSecretString` property genuinely changes (`{}` → JSON placeholder with marker key), same logical ID, in-place update, new version. Detection inversion (recognize the placeholder, not the real value's shape) is the right call for operator-chosen bare-string secrets, and the malformed/markerless-JSON branches fail safe toward 'never overwrite an operator value.' Test suite is behavioral and thorough.

Two non-blocking fast-follows: (1) emit a setup-time note when stack-wide seeding is skipped, so operators on a CLI-only upgrade (no redeploy) learn the legacy bare-string placeholder still needs a redeploy to carry the marker; (2) enforce the cross-package marker-constant coupling (CDK + CLI literals) with an assertion rather than a 'MUST stay in sync' comment.

PR #430, opened 2026-06-23
refactor(cdk): collapse AgentSessionRole trust+grant into admitComputeRole (#213)

### LINKED ISSUE #213: refactor(cdk): collapse AgentSessionRole trust+grant into one admitComputeRole method
## Summary

Follow-up from PR #211 (per-session IAM scoping) code review. The `AgentSessionRole` construct exposes two methods that are two halves of one logical operation — admitting a compute role to use the SessionRole:

- `addAssumingRole(role)` — adds the role to the **trust policy** (resource side)
- `grantAssumeToComputeRole(role)` — adds `sts:AssumeRole`/`sts:TagSession` to the **role's own policy** (identity side)

AssumeRole requires *both*. Exposing them independently makes it possible to wire only one and get a role that is trusted-but-unreachable or reachable-but-untrusted, with no synth-time signal. The constructor also auto-trusts `assumingRoles` but does **not** auto-grant, so callers must remember the grant separately (see `agent.ts` and the ECS path, which call different combinations).

## Proposed change

- Collapse into a single `admitComputeRole(role)` that does both halves.
- Have the constructor call it for each `assumingRoles` entry (so "admitted" always means fully wired).
- Unify the trust-statement construction (constructor splits AssumeRole via `assumedBy` + a separate TagSession statement; `addAssumingRole` bundles both — make them consistent).
- Keep a trust-only variant only if a cross-stack grantee genuinely needs it, clearly named as the exception.

## Acceptance criteria

- [ ] Single public method wires both trust + grant; call sites in `agent.ts` and `ecs-agent-cluster.ts` updated.
- [ ] Constructor uses the same path (no auto-trust-without-grant asymmetry).
- [ ] Trust-statement shape is consistent regardless of how a principal was added.
- [ ] Existing tests updated; a test asserts both halves are present after one call.
- [ ] `mise //cdk:test` + `//cdk:synth` pass; cdk-nag clean.

## Context

Non-urgent — current code is correct (all three call sites wire both halves today); this is a footgun-removal / API-ergonomics hardening. Deferred from #211 to keep that PR focused.

krokoko:
## Summary
- Replace split `addAssumingRole` / `grantAssumeToComputeRole` with a single `admitComputeRole` that wires both IAM halves (SessionRole trust + compute-role identity grant).
- Have the `AgentSessionRole` constructor call `admitComputeRole` for each `assumingRoles` entry so admission is always fully wired.
- Unify trust statements to bundled `sts:AssumeRole` + `sts:TagSession` per principal; update `agent.ts`, `ecs-agent-cluster.ts`, and tests.

Closes #213.

## Test plan
- [x] `mise //cdk:test -- agent-session-role`
- [x] `mise //cdk:test -- ecs-agent-cluster`
- [x] `mise //cdk:synth`

tested with a real deployment

### DIFF
--- a/cdk/src/constructs/agent-session-role.ts
+++ b/cdk/src/constructs/agent-session-role.ts
@@ -116,12 +116,12 @@ export class AgentSessionRole extends Construct {
       );
     }
 
-    const principals = props.assumingRoles.map(
-      (r) => new iam.ArnPrincipal(r.roleArn),
-    );
+    const [firstAssumingRole] = props.assumingRoles;
 
+    // CDK requires assumedBy; additional principals are admitted via
+    // admitComputeRole so trust + grant always wire together.
     this.role = new iam.Role(this, 'Role', {
-      assumedBy: new iam.CompositePrincipal(...principals),
+      assumedBy: new iam.ArnPrincipal(firstAssumingRole.roleArn),
       description:
         'Per-task scoped credentials for ABCA agent tenant-data access '
         + '(DynamoDB task rows + S3 trace/attachment objects), constrained by '
@@ -131,16 +131,6 @@ export class AgentSessionRole extends Construct {
       // intent is documented and synth-visible.
       maxSessionDuration: Duration.hours(1),
     });
-
-    // The agent passes session tags on AssumeRole, which requires the trust
-    // policy to also allow sts:TagSession (CDK's assumedBy only adds
-    // sts:AssumeRole). Same principals.
-    this.role.assumeRolePolicy?.addStatements(
-      new iam.PolicyStatement({
-        actions: ['sts:TagSession'],
-        principals,
-      }),
-    );
 
     // --- DynamoDB: item access gated by task_id leading-key ---
     // One statement per table keeps the resource ARNs explicit. The condition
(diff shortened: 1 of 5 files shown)


isadeks (approved):
## Approve

Reviewed the diff against `agent.ts`/`ecs-agent-cluster.ts` and ran the affected suites locally (compile clean, `agent-session-role` + `ecs-agent-cluster` = 18/18 passing).

**What it does:** collapses the two-call SessionRole admission (trust via `assumingRoles` + a separate `grantAssumeToComputeRole`) into a single `admitComputeRole`, and has the constructor call it for every `assumingRoles` entry. The now-redundant `grantAssumeToComputeRole(runtime.role)` in `agent.ts` and the `addAssumingRole` + `grantAssumeToComputeRole` double-call in the ECS path are correctly removed.

**Why it's a good change:**
- **IAM effect is identical** — `runtime.role` and the ECS task role still receive both the trust statement (`sts:AssumeRole` + `sts:TagSession`) and the identity-side grant. No privilege delta.
- **Correctness improvement, not just tidiness** — the old split API let the trust and grant halves drift (add a principal to `assumingRoles`, forget the grant call). Binding them in one method makes that class of bug structurally impossible.
- **Stronger tests** — the renamed `admitComputeRole`/`admits assumingRoles` tests now assert *both* the trust statement and the identity-policy grant (the old `compute role is granted…` test only checked the grant), and migrate cleanly to the bundled `Action: [sts:AssumeRole, sts:TagSession]` shape.
- Good encapsulation: `addTrustForComputeRole` / `grantAssumeToComputeRole` are now `private` behind the single public verb.

**Non-blocking nit:** the constructor uses `const [firstAssumingRole] = props.assumingRoles` for the CDK-required `assumedBy`, then the `for` loop also calls `admitComputeRole` for *every* entry — so the first principal ends up with both CDK's auto `assumedBy` (`AssumeRole`-only) statement and an `admitComputeRole`-added bundled (`AssumeRole`+`TagSession`) statement. Harmless (IAM dedups; tests green), but a one-line comment noting the first principal intentionally appears in both spots would save a future reader a double-take. Entirely optional.

LGTM. Tested with a real deployment per the description, and I confirmed the suites locally.

Issue #431, opened 2026-06-23
docs(onboarding): X-Ray gate, arm64/binfmt, failed-create teardown, and mise papercuts block first-time setup


isadeks:
### Component

Documentation / onboarding (QUICK_START + `/setup` plugin skill)

### Summary

A live first-contact deployment to a fresh, security-managed AWS Org account surfaced a cluster of onboarding-doc problems that stop or scare a first-time deployer. None are stack bugs — they are guidance gaps. Grouped:

**Blockers wrongly presented as routine / required:**
1. **X-Ray is a phantom gate.** `QUICK_START` Step 3 and the `/setup` skill Phase 3 instruct `aws xray update-trace-segment-destination` as *"required before first deployment."* But the stack ships with tracing **disabled** (`tracingEnabled` in `agent.ts`) — it never consumes the destination, and deploys fully without it. On a security-managed Org account the call fails with `AccessDenied` from an **SCP** that the deploying identity cannot clear, dead-ending the user on a step the platform doesn't use.
2. **arm64 image build on x86 hosts** is undocumented in the setup path. The agent image targets `linux/arm64` (Graviton); on x86 without QEMU/binfmt the build dies with `exec /bin/sh: exec format error`. The README mentions buildx, but QUICK_START / the skill don't pre-empt it.

**Compounding friction:**
3. **Failed-create teardown** is undocumented: a first-create failure lands in `ROLLBACK_COMPLETE` (un-updatable → must destroy+recreate), and the delete then stalls in `DELETE_FAILED` for ~20–40 min while AgentCore's service-managed (Hyperplane) ENIs reclaim asynchronously. Force-deleting past the stuck VPC orphans it (VPCs are quota-capped).
4. **Non-interactive deploy** hangs: `mise //cdk:deploy` lacks `--require-approval never`.
5. **mise papercuts** read as a broken environment: Node/Yarn/CDK all show "not found" until `mise install` (mise provisions them); headless-Linux GPG verification failure on Node install; dual project + user-global `mise trust`; mise not on `PATH` in non-interactive shells; and noisy-but-benign build output where the only reliable signal is exit code 0.

### Fix (this PR)

Edits `docs/guides/QUICK_START.mdx` and `docs/abca-plugin/skills/setup/SKILL.md` (Starlight mirror regenerated):
- X-Ray reframed as **optional / opt-in**, removed from the deploy path; demoted to a note with the Org-SCP caveat.
[...]

PR #434, opened 2026-06-23
feat(cdk): single source of truth for invocable Bedrock models, context-overridable (#433)

### LINKED ISSUE #433: feat(cdk): make the runtime's invocable Bedrock model set stack-configurable (one list, not two hardcoded arrays)
### Component

CDK / infrastructure (Bedrock model grants)

### Problem

Choosing which Claude models the agent can use is higher-friction than it should be, and the current implementation has a latent drift bug:

1. **Two hardcoded lists, kept in sync by hand.** The set of invocable models is hardcoded in **two** places that must stay identical:
   - `cdk/src/stacks/agent.ts` (~L423-460) — three `BedrockFoundationModel` + `CrossRegionInferenceProfile` + `grantInvoke` blocks for the AgentCore runtime (Sonnet 4.6, Opus 4, Haiku 4.5).
   - `cdk/src/constructs/ecs-agent-cluster.ts` (`BEDROCK_MODEL_IDS`, L59-63) — the same three, for the ECS task role. Its own comment says *"kept in sync with the AgentCore runtime grants in agent.ts."*
   Hand-synced lists drift; a model added to one and not the other yields a backend-dependent 403.

2. **Adding a model requires editing the construct(s) + redeploy.** An operator who wants, say, Opus 4.8 must hand-edit CDK in two spots. `bgagent repo onboard --model us.anthropic.claude-opus-4-8` accepts it, then tasks 403 at invoke because no `grantInvoke` covers that ARN.

### Non-goal / what to preserve

Per-model ARN scoping is a **deliberate hardening** — `ecs-agent-cluster.ts` explicitly replaced a `Resource: '*'` wildcard with explicit foundation-model + inference-profile ARNs, and `ecs-agent-cluster.test.ts:171-174` locks it in. **This issue must NOT revert that to a wildcard.** Account-level Bedrock model access remains the outer gate; per-model IAM scoping stays.

### Proposed solution

Single source of truth for the invocable-model list, read by both grant sites:

- New `cdk/src/constructs/bedrock-models.ts` exporting `DEFAULT_BEDROCK_MODEL_IDS` (the current three) and a resolver `resolveBedrockModelIds(node)` that returns CDK context `bedrockModels` (array) when set, else the default.
- `agent.ts` and `ecs-agent-cluster.ts` both import and iterate that one list to build their grants/ARNs — no more parallel arrays.
- Operators add a model via `cdk.json` context `"bedrockModels": ["us.anthropic.claude-sonnet-4-6", "us.anthropic.claude-opus-4-8", ...]` (or `-c`), then redeploy. Still explicit ARNs, still least-privilege, still account-access-gated.
- Default unchanged (Sonnet 4.6, matching `workflows.ts` platform default). Update `ecs-agent-cluster.test.ts` to assert the resolved list rather than three literals; add a test that a context override flows to both grant sites.

### Acceptance criteria

- [ ] One list drives both the AgentCore runtime and ECS task-role Bedrock grants.
- [ ] `cdk.json` context overrides the model set; default = today's three.
- [ ] Per-model ARN scoping preserved (no `Resource: '*'` regression); cdk-nag posture unchanged.
- [ ] Tests: default set asserted in one place; context-override propagation to both backends covered.
- [ ] Docs/`onboard-repo` skill note how to add a model (config + redeploy), distinct from the per-repo `--model` selection among already-granted models.

### Context

Surfaced during the 2026-06-23 fresh-account onboarding run (user wanted Opus 4.8, which isn't wired). Lower-friction than editing constructs; does not loosen the security boundary.

isadeks:
## Summary

Lowers the friction of choosing which Claude models the agent can use, and fixes a latent drift bug — without loosening the per-model least-privilege scoping.

**Before:** the invocable-model set was hardcoded in **two** places that had to stay identical by hand — the AgentCore runtime `grantInvoke` blocks in `stacks/agent.ts` and `BEDROCK_MODEL_IDS` in `constructs/ecs-agent-cluster.ts` (whose own comment said *"kept in sync with the AgentCore runtime grants in agent.ts"*). Adding a model meant editing CDK in two spots and redeploying; a repo pinned to an ungranted model (e.g. Opus 4.8) 403s at invoke.

**After:** one source of truth, `constructs/bedrock-models.ts`:
- `DEFAULT_BEDROCK_MODEL_IDS` — today's three (Sonnet 4.6, Opus 4, Haiku 4.5), unchanged default.
- `resolveBedrockModelIds(node)` — returns the `bedrockModels` CDK context array when set, else the default. Both grant sites derive from it, so the AgentCore runtime and ECS task role can't drift.

Operators add a model via `cdk.json` context (or `-c bedrockModels='[…]'`) + redeploy — **no construct edits**.

## Preserves the security posture (does NOT revert hardening)

`ecs-agent-cluster.ts` deliberately scopes Bedrock to explicit foundation-model + inference-profile ARNs (it replaced a `Resource: '*'` wildcard, with a regression test). This PR **keeps** that — the shared list still produces explicit per-model ARNs, never a wildcard. Account-level Bedrock model access remains the outer authorization gate. Malformed context (non-array / empty / non-string entries) fails synth loudly rather than silently granting nothing.

## Tests

- `bedrock-models.test.ts` — resolver: default set, context override, and validation throws (non-array / empty / empty-string entry).
- `ecs-agent-cluster.test.ts` — new test: a `bedrockModels` context override changes the granted ARNs and never widens to `*`; existing "scoped, no wildcard" hardening test retained.
- Full `mise //cdk:test`: **122 suites / 2203 pass**; `mise //cdk:eslint` clean.

## Notes for reviewers

- Pure refactor of the grant source + new opt-in context; default behavior (the three models) is byte-for-byte unchanged in synth.
- Adding a brand-new model still requires the **account** to have Bedrock access enabled for it (separate from IAM) — documented in the `/onboard-repo` skill (see #431/#432).

Fixes #433


krokoko:
Thanks for this — the refactor is clean and solves the real drift problem without loosening the per-model IAM scoping. A few notes from review:

**Looks good**
- Single source of truth in `bedrock-models.ts` with both grant sites deriving from `resolveBedrockModelIds()` — exactly what #433 asked for on the IAM side.
- Fail-fast validation on malformed context is a nice touch.
- ECS override test confirms replace-not-append behavior and that we still never widen to `*`.

**Suggestions (non-blocking)**
1. **AgentStack test** — Issue #433's acceptance criteria mention context-override propagation to *both* backends. The new ECS test covers one side; a matching `AgentStack` assertion (runtime execution role gets the override ARNs, defaults excluded) would close that loop.
2. **Context ID format** — The resolver expects bare foundation-model IDs (`anthropic.claude-opus-4-8`), not `us.`-prefixed inference-profile IDs. Worth calling out in docs (#435?) since the issue text used the `us.` form — operators following that example would synth invalid ARNs (`us.us.…`).
3. **`WORKFLOW_MODEL_ALLOWLIST`** — Still a third hardcoded list for workflow YAML admission. Doesn't block the `repo onboard --model` path (repo `model_id` isn't gated), but custom workflows pinning a newly added model would still fail at create-task. Might be worth a follow-up or a short comment in `bedrock-models.ts`.
4. **Stale comment** — `workflows.ts` still references `BEDROCK_MODEL_IDS` in `ecs-agent-cluster.ts`; could point at `bedrock-models.ts` after merge.

**CI** — `integ-smoke` failed on missing AWS credentials in the runner (infra flake), unrelated to this diff. Build/security checks passed.

Happy to approve once the AgentStack test lands (or if you prefer to track that as a fast follow-up). Nice work on keeping the security posture intact.

isadeks:
Thanks for the thorough review! Addressed in the latest push (d6f2ee7):

1. **AgentStack test (blocking)** — added two tests in `agent.test.ts`: the runtime execution role gets the default model grants, and a `bedrockModels` context override propagates to the runtime (overridden model present, defaults absent, never `*`). Together with the existing ECS test, both backends are now covered per #433's acceptance criteria.
2. **`us.`-prefix gotcha** — turned it into a fail-fast: `resolveBedrockModelIds` now rejects region-prefixed IDs (`us.`/`eu.`/`apac.`) with a message pointing at the bare ID, since both grant sites derive the `us.`-profile ARN themselves (the `us.us.…` footgun you flagged). JSDoc clarified + test added. Good catch — the issue text's `us.` example would indeed have mis-synthed.
3. **`WORKFLOW_MODEL_ALLOWLIST`** — left as-is for this PR (it gates workflow-YAML admission, not the `repo onboard --model` path), but added a comment noting it's a separate hand-maintained list and that a custom workflow pinning a context-added model would still need it here. Consolidating it with `bedrock-models.ts` is worth a follow-up — happy to file one.
4. **Stale comment** — `workflows.ts` now points at `bedrock-models.ts` instead of the removed `BEDROCK_MODEL_IDS`.

Full `//cdk:test` is at 2206 passing. The `integ-smoke` red is the same OIDC-credentials infra flake. Ready for another look.

krokoko (approved):

Issue #402, opened 2026-06-22
fix(cdk): bootstrap least-privilege policy missing JiraIntegration* secret pattern → deploy rolls back


isadeks:
### Component

CDK / infrastructure (bootstrap least-privilege policies)

### Describe the bug

A fresh `mise //cdk:bootstrap` + `mise //cdk:deploy` of `backgroundagent-dev` rolls back during stack creation. The first `CREATE_FAILED` event is:

```
cdk-hnb659fds-cfn-exec-role is not authorized to perform secretsmanager:CreateSecret
because no identity-based policy allows the secretsmanager:CreateSecret action
```

…on the `JiraIntegrationWebhookSecre-…` secret.

### Root cause

The custom least-privilege CFN execution-role policy (ADR-002) allow-lists Secrets Manager ARNs by integration prefix, but the **Jira** integration was added without a corresponding pattern.

`cdk/src/bootstrap/policies/application.ts` (the `SecretsManager` statement, ~L244-251) scopes `secretsmanager:CreateSecret` to:

```
arn:aws:secretsmanager:*:*:secret:backgroundagent-*
arn:aws:secretsmanager:*:*:secret:GitHubTokenSecret*
arn:aws:secretsmanager:*:*:secret:SlackIntegration*
arn:aws:secretsmanager:*:*:secret:LinearIntegration*
arn:aws:secretsmanager:*:*:secret:GitHubScreenshot*
arn:aws:secretsmanager:*:*:secret:bgagent/*
```

There is **no `JiraIntegration*` pattern**, yet `cdk/src/constructs/jira-integration.ts:176` creates `new secretsmanager.Secret(this, 'WebhookSecret', …)` under construct id `JiraIntegration`, which CloudFormation names `JiraIntegrationWebhookSecre-…`. The scoped exec role therefore cannot create it, and the stack rolls back.

The gap is also present in the checked-in generated artifact `cdk/bootstrap/policies/application.json` (~L205-207).

### Expected behavior

A fresh bootstrap + deploy succeeds; the CFN execution role is permitted to create the Jira integration's webhook secret.

### Current behavior

Deploy fails with `AccessDenied` on `secretsmanager:CreateSecret` for `JiraIntegrationWebhookSecre-…` and the stack rolls back.

### Possible solution

Add the missing ARN pattern to the `SecretsManager` statement in `cdk/src/bootstrap/policies/application.ts`:

```diff
[...]

PR #435, opened 2026-06-23
docs(skills): audit deploy/submit-task/troubleshoot/status — stale claims, node PATH, onboard path

isadeks:
## Summary

Audit of the four plugin skills not touched by #432, against issues found during a live first-contact deployment. All doc-only edits to `docs/abca-plugin/skills/*/SKILL.md`. The recurring theme (same as #432): skills describe ceremony or a reality that no longer matches the platform.

## `deploy` (most significant)

- **Corrects an inverted factual claim.** The "Least-Privilege Deployment" section said CDK bootstrap grants `AdministratorAccess` **by default** and scoping is an optional production step. That's backwards — the custom least-privilege bootstrap (ADR-002) **is** the default. Rewritten to describe the default accurately, plus the practical consequence: adding a new resource type / feature may require adding the IAM action to `cdk/src/bootstrap/policies/` (the #402/#404/#407/#409 class of rollbacks).
- Deploy command now uses `--require-approval never` (non-TTY hang); added an x86/arm64 binfmt pre-check; added `DELETE_FAILED` (async Hyperplane-ENI reclaim, ~20–40 min) + `ROLLBACK_COMPLETE` notes to the destroy workflow; "trust the exit code" note on the noisy build; onboard via `bgagent repo onboard` rather than "Blueprint constructs if needed."

## `troubleshoot`

- `REPO_NOT_ONBOARDED` (422) fix is `bgagent repo onboard <owner/repo>`, not "needs a Blueprint construct."
- 403-model fix: repoint to an already-granted model, or add one via the shared `bedrockModels` CDK context (#433) **plus** account-level Bedrock access — not "edit `grantInvoke` in `agent.ts`."

## `submit-task` / `troubleshoot` / `status`

- Note that `node cli/lib/bin/bgagent.js …` needs a `mise exec --` prefix in a non-interactive / mise-managed shell (the live `node: command not found`). `submit-task`'s prompt-quality and cost-control guidance was otherwise solid and is unchanged.

## Notes

- Docs-only; `abca-plugin` skills are not Starlight-mirrored, so no `docs:sync` needed.
- The `troubleshoot` 403-model wording references the `bedrockModels` context from #433 (PR #434, open). It also gives the no-redeploy quick fix (repoint to a granted model), so it's accurate regardless of #434's merge state.
- Complements #432 (which fixed `/setup` + `/onboard-repo`). Together the full six-skill set is audited.

### DIFF
--- a/docs/abca-plugin/skills/deploy/SKILL.md
+++ b/docs/abca-plugin/skills/deploy/SKILL.md
@@ -28,17 +28,20 @@ Before any deployment action, verify:
    export MISE_EXPERIMENTAL=1
    mise run build
    ```
-   This runs agent quality checks, CDK compilation + tests, CLI build, and docs build. Do NOT deploy if the build fails.
+   This runs agent quality checks, CDK compilation + tests, CLI build, and docs build. Do NOT deploy if the build fails. Note: a passing build is noisy — it prints many `ERROR`/`WARN` and cdk-nag lines from test fixtures. Trust the **exit code (0 = pass)**, not the log volume.
 
-2. **Docker is running** — Required for CDK asset bundling
-3. **AWS credentials are configured** — `aws sts get-caller-identity`
+2. **Docker is running** — Required for CDK asset bundling.
+3. **Build host architecture** — The agent image targets `linux/arm64` (AgentCore is Graviton). On an **x86_64** host without QEMU/binfmt, the deploy fails partway with `exec /bin/sh: exec format error`. Register emulation once with `docker run --privileged --rm tonistiigi/binfmt --install arm64`, or deploy from a native arm64 host (Graviton / Apple Silicon). Skip on arm64 hosts.
+4. **AWS credentials are configured** — `aws sts get-caller-identity` (confirm it's the intended account/region).
 
 ## Deploy Workflow
 
 ```bash
 export MISE_EXPERIMENTAL=1
-mise //cdk:deploy
+mise //cdk:deploy -- --require-approval never
 ```
+
+`--require-approval never` lets the deploy run unattended. **In a non-interactive shell (CI, agent, script) it's required** — without it, `cdk deploy` hangs forever on the IAM/security-group approval prompt. Drop the flag if you're deploying interactively and want to review those changes.
 
 After successful deployment, retrieve and display stack outputs:
 ```bash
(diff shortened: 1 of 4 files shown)


docs/abca-plugin/skills/deploy/SKILL.md
```diff
@@ -79,18 +84,13 @@ Output goes to `cdk/cdk.out/`. Useful for reviewing generated CloudFormation tem
 
 After a successful deploy, remind the user to:
 - Store/update the GitHub PAT in Secrets Manager if this is a fresh deployment
-- Onboard repositories via Blueprint constructs if needed
+- Onboard repositories with `bgagent repo onboard <owner/repo>` (a runtime operation — no redeploy). A CDK `Blueprint` construct is only needed for declarative config (Cedar policies, egress allowlist, system-prompt overrides) — see the `onboard-repo` skill
```
krokoko: I would keep the previous one on this or rephrase to clarify, the onboard command is here only to help with onboarding a new repo using the default blueprint, otherwise it requires a cdk re-deploy with a new blueprint and correct permissions set

docs/abca-plugin/skills/deploy/SKILL.md
```diff
@@ -79,18 +84,13 @@ Output goes to `cdk/cdk.out/`. Useful for reviewing generated CloudFormation tem
...
-cdk bootstrap aws://ACCOUNT/REGION \
-  --cloudformation-execution-policies "arn:aws:iam::ACCOUNT:policy/IaCRole-ABCA-Infrastructure" \
-  --cloudformation-execution-policies "arn:aws:iam::ACCOUNT:policy/IaCRole-ABCA-Application" \
-  --cloudformation-execution-policies "arn:aws:iam::ACCOUNT:policy/IaCRole-ABCA-Observability"
-```
+A consequence worth knowing when you **add a new resource type or a new feature on an existing resource**: the scoped role must allow the IAM action CloudFormation will call, or the deploy rolls back with `AccessDenied` on that action (e.g. `s3:PutBucketVersioning`, `lambda:TagResource`). The fix is to add the action to the relevant policy in `cdk/src/bootstrap/policies/`, regenerate (`mise //cdk:bootstrap:generate`), re-bootstrap, and redeploy. The policy source and the `DEPLOYMENT_ROLES.md` golden doc are kept in sync by tests.
 
-See `docs/design/DEPLOYMENT_ROLES.md` in the repo root for the complete least-privilege IAM policies, trust policy, runtime role inventory, and iterative tightening recommendations.
+See `docs/design/DEPLOYMENT_ROLES.md` for the complete IAM policies, trust policy, runtime role inventory, and tightening recommendations.
```
krokoko, line 97: after deployment and configuration is done, there is also a platform doctor command through the CLI to verify that everything is green before submitting a task, might be worth it to add it

docs/abca-plugin/skills/deploy/SKILL.md
isadeks: Good catch — rephrased in 617c055. The line now says `bgagent repo onboard` is the runtime path only when the repo fits the platform/default-blueprint setup (default token, already-granted model, default egress), and that a repo needing its own token / an ungranted model / custom egress / Cedar policies / system-prompt overrides requires a dedicated Blueprint + redeploy with correct permissions. I'll apply the same clarification to the onboard-repo skill in #432 so the two stay consistent.

docs/abca-plugin/skills/deploy/SKILL.md
isadeks: Added in 617c055 — post-deploy now points to `bgagent platform doctor` as the readiness check (API/Cognito/token/Bedrock/onboarded repos) before submitting a task, with the raw curl kept as a lower-level alternative.

krokoko (approved):

PR #520, opened 2026-06-30
feat(ci): dead-code detection gate — knip + no-unused-vars + vulture (#282)

### LINKED ISSUE #282: feat(ci): dead-code detection gate — knip (TS) + @typescript-eslint/no-unused-vars + vulture (Python) (cairn MVG gate #6)
> **This implements cairn Minimum Viable Gate Set gate #6** (`Dead code detection — catches AI002, AI003`), from https://github.com/krokoko/cairn (`plugins/software-verification/.../detection-patterns-gates.md`, "Minimum Viable Gate Set"). Mapping the gate set to the open CA issues, gate #6 is the one with no corresponding issue:
>
> | cairn MVG gate | tracked by |
> |---|---|
> | 1 Type checking | #260 (CA-14) |
> | 2 Empty-catch / silent-success | #257 (CA-09) |
> | 3 Mutation testing | #255 (CA-04) |
> | 5 Import boundary + magic values | #254 (CA-03), #258 (CA-10) |
> | **6 Dead code detection** | **— this issue —** |

### Component

Tooling / CI

### Describe the feature

Add a **dead-code detection** gate covering two surfaces the existing gates miss:

- **Unused imports / locals / params** — lint-level, per file.
- **Unreferenced exports / files / dead barrels** — project-graph level (an exported symbol no module imports; a barrel `index.ts` with zero importers).

across the TypeScript packages (`cdk`, `cli`) and the Python package (`agent`).

### Use case — the detection asymmetry this closes

The repo enforces unused-symbol detection on **one** language and not the other, so dead code accumulates only on the TypeScript side:

- **Python (`agent/`) — already enforced.** `ruff` `select` includes `"F"` (`agent/pyproject.toml`), so pyflakes **F401** fails the build on an unused import. A dead import in `agent/src` cannot reach `main`.
- **TypeScript is split.** `cli/tsconfig.json:21-22` sets `"noUnusedLocals": true` / `"noUnusedParameters": true`, so the CLI already blocks unused locals and imports. But `cdk/tsconfig.json:21-22` sets both to `false`, and neither flat-config eslint (`cdk/eslint.config.mjs`, `cli/eslint.config.mjs`) carries a `no-unused-vars` error rule. So unused TS symbols pass silently **in `cdk/` only** — which is exactly where every known instance lives. Note: this repo does NOT use projen; the eslint configs are hand-maintained flat-config files.

This is not hypothetical — three live instances exist at HEAD `bb7876a` (all in `cdk/`):

1. **Dead import:** `cdk/src/handlers/shared/create-task-core.ts` imports `DEFAULT_MAX_TURNS` from `./validation` and never uses it. (This file is also the subject of #232; an unused-import gate would have surfaced it.)
2. **Dead barrel:** `cdk/src/bootstrap/index.ts` (added in #158) re-exports the bootstrap policy/version symbols, but **nothing imports the barrel** — every consumer imports the submodules (`./policies`, `./version`) directly.
3. **Orphaned export:** `LINEAR_SECRET_PREFIX` in `cdk/src/handlers/shared/linear-verify.ts` is exported but has no importer anywhere in `cdk`/`cli`/`agent`.

cairn's CI-integration model lists "dead code" as a **fast-CI** gate and tracks the metric *"dead code count (should not increase)"* — i.e. a ratchet, not a one-time cleanup.

### Proposed solution (grounded against current tooling, mid-2026)

**1. TS lint-level — `@typescript-eslint/no-unused-vars`** (typescript-eslint v8; base `no-unused-vars` must be `off`). This is the direct parity match to the Python `F401` already enforced — the same pattern CA-10 (#258) used for magic values (`no-magic-numbers` ↔ `PLR2004`).

```jsonc
"no-unused-vars": "off",
"@typescript-eslint/no-unused-vars": ["error", {
  "args": "all", "argsIgnorePattern": "^_",
  "varsIgnorePattern": "^_", "caughtErrors": "all",
  "caughtErrorsIgnorePattern": "^_", "ignoreRestSiblings": true,
  "destructuredArrayIgnorePattern": "^_"
}]
```

Apply by editing the hand-maintained flat-config files directly: `cdk/eslint.config.mjs` and `cli/eslint.config.mjs` (this repo has no projen, so there is no generated config to regenerate).

**2. TS project-graph level — `knip` (v6.x).** `ts-prune` is **dead** (archived read-only 2025-09-19; its README points new projects to knip). knip natively understands Yarn workspaces and per-workspace entry points. Minimal `knip.json` (the `!` suffix marks production entry so test/config files aren't counted as dead):

```json
{
  "$schema": "https://unpkg.com/knip@6/schema.json",
  "workspaces": {
    "cdk": { "entry": ["src/main.ts!"], "project": ["src/**/*.ts!"] },
    "cli": { "entry": ["src/index.ts!"], "project": ["src/**/*.ts!"] },
    "docs": { "entry": ["astro.config.{js,mjs,ts}", "src/pages/**/*.{astro,ts}"], "project": ["src/**/*.{astro,ts,tsx}"] }
  }
}
```

> The explicit `cli` entry is what stops `cli/src/index.ts` (the declared package `main`/`types` source) from false-positiving as an unused export.

**3. Python module-level — `vulture` (v2.16).** ruff `F401` covers unused *imports* but has no rule for unused module-level *functions/classes*. vulture fills that gap:

```bash
vulture agent/ --make-whitelist > agent/.vulture_allowlist.py   # once, reviewed
vulture agent/ agent/.vulture_allowlist.py --min-confidence 80   # CI
```

**4. Rollout — advisory first, then ratchet.** Land all three non-blocking; fix or baseline the existing instances (the three above plus whatever knip/vulture surface); then make blocking in the required `build` check. For knip, gate on the JSON issue count so the build fails only on *increases* (the cairn "should not increase" metric); knip has no native baseline file, so snapshot the count in CI. For vulture, ship the reviewed allowlist so the current tree passes, then flip `--min-confidence 80` to a blocking exit.

### Acceptance criteria

- [ ] `@typescript-eslint/no-unused-vars` enabled as an **error** in the `cdk` and `cli` flat-config eslint files (`cdk/eslint.config.mjs`, `cli/eslint.config.mjs`), with `^_` ignore patterns. (cdk also flips `noUnusedLocals`/`noUnusedParameters` to true to match cli, closing the tsconfig half of the gap.)
- [ ] `knip` added with a per-workspace `knip.json` covering `cdk`/`cli`/`docs`; `cli/src/index.ts` is **not** flagged (entry-point caveat handled).
- [ ] `vulture` added for `agent/` with a reviewed `--make-whitelist` allowlist and `--min-confidence 80`.
- [ ] The three known instances are resolved: dead `DEFAULT_MAX_TURNS` import dropped, `cdk/src/bootstrap/index.ts` barrel removed (or given a real importer), `LINEAR_SECRET_PREFIX` removed or consumed.
- [ ] Gate runs in **fast CI** (static, seconds; no build needed), advisory first.
- [ ] Gate becomes blocking in the required `build` check once the baseline is clean, on a "dead-code count must not increase" basis.

### Other information

- Source: cairn MVG gate #6 + the "dead code count (should not increase)" CI metric (`krokoko/cairn`). Effort: **S–M**.
- Distinct from #232 (runtime module-init memory / lazy-loading) — that is dynamic import-graph weight; this is static unused-symbol detection. They intersect only at `create-task-core.ts`, where the dead `DEFAULT_MAX_TURNS` import lives.
- Distinct from #254 (CA-03, import-boundary enforcement via `dependency-cruiser`/`madge`) — boundaries constrain *which* modules may import *which*; this finds symbols/files imported by *nobody*.
- Per ADR-003 this issue needs the `approved` label before work begins.

*(Drafted by Bonk during the ABCA review; tool currency — knip 6.x, ts-prune archived, vulture 2.16, typescript-eslint v8 — verified against current docs. Filed via Laith's account.)*

ClintEastman02:
Closes #282 (cairn MVG gate #6 — dead-code detection).

## What this does

Closes the unused-symbol detection asymmetry: Python already fails on unused imports (ruff `F401`), but unused TS symbols passed silently in `cdk/`. Adds three detection layers plus advisory CI wiring.

- **TS lint-level** — `@typescript-eslint/no-unused-vars` (error) in `cdk/` + `cli/` flat configs with `^_` ignore patterns; `cdk/tsconfig.json` flips `noUnusedLocals`/`noUnusedParameters` to `true` for parity with `cli/`.
- **TS project-graph** — `knip` v6 with a per-workspace `knip.json` (cdk/cli/docs), plus a count-only ratchet (`scripts/check-deadcode-ratchet.mjs` + `knip-baseline.json`) that fails only when the count rises — the cairn "dead-code count should not increase" metric.
- **Python module-level** — `vulture` 2.16 (`[tool.vulture]` in `pyproject.toml`, `--min-confidence 80`) with a minimal reviewed `.vulture_allowlist.py` for the three SDK-mandated `hook_context` params.
- **CI** — advisory `dead-code` job in `security-pr.yml` (`continue-on-error`, emits `::notice::`/`::warning::` annotations; not blocking yet).

Resolves the three known live instances: drops the dead `DEFAULT_MAX_TURNS` import, removes the orphaned `cdk/src/bootstrap/index.ts` barrel (consumers import the submodules directly), and removes the unused `LINEAR_SECRET_PREFIX` export. Also clears the unused symbols the new rule surfaces across `cdk`/`cli`.

The eslint half is **blocking** via the existing `build` check; knip + vulture stay **advisory** until the knip baseline is driven to zero (deferred AC, per the issue's phased rollout).

## Acceptance criteria

- [x] `@typescript-eslint/no-unused-vars` as error in `cdk`/`cli` flat configs with `^_` patterns; cdk tsconfig flags flipped to `true`
- [x] `knip` added with per-workspace `knip.json` (cdk/cli/docs); `cli/src/index.ts` not flagged (entry-point caveat handled)
- [x] `vulture` for `agent/` with reviewed allowlist + `--min-confidence 80`
- [x] Three known instances resolved (`DEFAULT_MAX_TURNS`, bootstrap barrel, `LINEAR_SECRET_PREFIX`)
- [x] Gate runs in fast CI, advisory first
- [ ] Gate becomes blocking once baseline is clean (deferred per issue's "advisory first, then ratchet" rollout)

## Test plan

Verified locally on a fresh checkout:

- `mise //cdk:build` — compile, tests, eslint (`no-unused-vars` active), synth all pass
- `mise //cli:build` — compile, 565 tests, eslint all pass
- `yarn knip` + `node scripts/check-deadcode-ratchet.mjs` — ratchet holds at baseline (78; ratcheted down from 79 after merging main #505)
- `uvx vulture` (agent) — clean, exit 0
- `mise //cdk:eslint`, `mise //cli:eslint` — pass


ayushtr-aws (commented):
## Review

### Overview

This PR closes the unused-symbol detection asymmetry between Python and TypeScript by adding three dead-code detection layers — `@typescript-eslint/no-unused-vars` (blocking, via the existing build check), `knip` + a count-only ratchet (advisory), and `vulture` for Python (advisory) — plus an advisory `dead-code` CI job. It also clears the symbols the new rules surface: dropping the dead `DEFAULT_MAX_TURNS` import, the orphaned `bootstrap/index.ts` barrel, the unused `LINEAR_SECRET_PREFIX` export, and a scattering of `^_`-prefix renames.

The bulk of the diff is inert and safe: config, lockfile churn, and parameter/catch-variable renames. Two changes carry real logic, and one of them regresses a compile-time guard. Findings below, most-severe first.

---

### Findings

#### 1. `cdk/src/handlers/shared/validation.ts:316` — refactor silently defeats the `AttachmentType` exhaustiveness check (CONFIRMED)

The change rewrote a typed **assignment** into a type **assertion**:

```ts
-const _attachmentExhaustiveCheck: _AssertAttachmentExhaustive = true;
+void (true as _AssertAttachmentExhaustive);
```

These are not equivalent. The guard relies on `_AssertAttachmentExhaustive` resolving to `never` when `AttachmentType` gains a member missing from `ATTACHMENT_TYPE_LIST`:

- **Old (assignment):** `const x: never = true` is a hard compile error — `true` is not assignable to `never`. *That error is the guard.*
- **New (assertion):** `true as never` is an assertion, and TypeScript always permits `X as Y` when either type is assignable to the other. `never` is assignable to `true`, so `true as never` **never errors**.

**Failure scenario:** someone adds `'video'` to `AttachmentType` without updating `ATTACHMENT_TYPE_LIST`. The old code fails the build; the new code compiles clean, and `VALID_ATTACHMENT_TYPES` silently omits the new type — the exact regression the assertion was meant to catch ships unnoticed.

**Suggested fix** — restore the assignment and consume the binding so it survives both `noUnusedLocals` (which this PR flips to `true` and which does *not* honor eslint's `^_` pattern) and `@typescript-eslint/no-unused-vars`:

```ts
/** Compile-time exhaustiveness check for AttachmentType. */
const ATTACHMENT_TYPE_LIST = ['image', 'file', 'url'] as const satisfies readonly AttachmentType[];
type _AssertAttachmentExhaustive = Exclude<AttachmentType, (typeof ATTACHMENT_TYPE_LIST)[number]> extends never ? true : never;
// The ASSIGNMENT is the guard: if AttachmentType gains a member missing from
// ATTACHMENT_TYPE_LIST, _AssertAttachmentExhaustive resolves to `never` and
// assigning `true` fails to compile. `void` consumes the binding so
// noUnusedLocals stays satisfied without weakening the check.
const _attachmentExhaustiveCheck: _AssertAttachmentExhaustive = true;
void _attachmentExhaustiveCheck;
```

---

#### 2. `scripts/check-deadcode-ratchet.mjs:606` — `classMembers` is not a valid knip 6.x key; `namespaceMembers` is missing (low severity)

`COUNTED_KEYS` lists `classMembers`, which is not an issue key knip 6.x emits — it's always `undefined`, so it's silently dropped (harmless). Meanwhile the real member-issue key `namespaceMembers` is absent from the list, so dead namespace members never move the ratchet count, despite the script's comment claiming it counts "enum/class members." Separately, `duplicates` is emitted as an array-of-arrays (each inner array is one duplicate group), so `issue.duplicates.length` counts groups rather than individual duplicate symbols.

Impact is low because the ratchet step is `continue-on-error` (advisory), but it should be reconciled before the gate flips to blocking.

**Suggested fix:**

```js
const COUNTED_KEYS = [
  'dependencies',
  'devDependencies',
  'optionalPeerDependencies',
  'unlisted',
  'binaries',
  'unresolved',
  'exports',
  'types',
  'nsExports',
  'nsTypes',
  'duplicates',
  'enumMembers',
  'namespaceMembers',  // was 'classMembers' — not a knip 6.x key
];
```

---

### Minor note (not a finding)

The PR description's test plan says the ratchet "holds at baseline (79)", but `knip-baseline.json` commits `"count": 78`. If the live count is 79, the gate emits a `::warning::` from day one — harmless while advisory, but reconcile the baseline before it becomes blocking.

---

### Verdict

One change blocks: the `validation.ts` exhaustiveness regression (# 1) reintroduces exactly the class of silent drift this PR's tooling is meant to prevent, so it's worth fixing in-PR. The ratchet key fix (# 2) and the baseline note are low-severity cleanups that can ride along or follow before the gate goes blocking. Everything else — the eslint/tsconfig wiring, vulture setup, dead-symbol removals, and renames — looks correct and well-scoped.

krokoko:
Thanks for this — it's a clean, well-scoped implementation of the dead-code gate (#282), and the advisory-first rollout is exactly the right call. **Approving with a few nits, none blocking** — contingent on the fixes identified by the other reviewers in the comments above being addressed as well.

**What's solid:**
- All three known dead symbols (`DEFAULT_MAX_TURNS`, the bootstrap barrel, `LINEAR_SECRET_PREFIX`) are correctly removed and verified unreferenced.
- The `_`-prefix param renames and test cleanups are safe — no assertions weakened.
- Comments are accurate where it counts (the exhaustiveness-guard rationale and the `hook_context` line numbers both check out).
- Mirrors the existing CA-10/#258 parity pattern nicely; deps land in the right manifests.

**Worth a look (non-blocking):**
1. **Ratchet can fail open.** In `scripts/check-deadcode-ratchet.mjs`, if a future knip (`^6.20.0`) renames a category or reshapes its JSON, the count silently drops below baseline and the script reports a *passing* build — quietly disabling the gate. Fine while advisory, but please pin knip exactly (or assert `report.issues` is present) before flipping to blocking.
2. `COUNTED_KEYS` omits `nsExports`/`nsTypes` — harmless today, but the comment claims completeness; a one-line note would clear it up.
3. Small comment nit in `validation.ts:294` — the `void` is really there for `noUnusedLocals`, not eslint (which already exempts `^_`).
4. A `countIssues` fixture test would pair well with the eventual blocking flip.

Everything else — docs sync, CLI/API type sync, bootstrap policies — is N/A for this diff. Nice work; happy to see this land once the above and the other reviewers' points are folded in. 🚀

ClintEastman02:
Thanks for the thorough review @ayushtr-aws — all three addressed in 384efa8.

**#1 — `validation.ts` exhaustiveness guard (confirmed, fixed).** You're right that `true as never` is an assertion and never errors. Restored the assignment form and consumed the binding with `void` so it survives both `noUnusedLocals` (now `true`) and `@typescript-eslint/no-unused-vars`:

```ts
const _attachmentExhaustiveCheck: _AssertAttachmentExhaustive = true;
void _attachmentExhaustiveCheck;
```

Verified empirically: adding a `'video'` member to `AttachmentType` without updating `ATTACHMENT_TYPE_LIST` now fails `tsc` again (`TS2322: Type 'true' is not assignable to type 'never'`), and compiles clean when exhaustive. Comment updated to spell out *why* it must stay an assignment.

**#2 — ratchet `COUNTED_KEYS` (confirmed, fixed).** Dumped knip 6.x's actual JSON keys and corrected the list: dropped the non-existent `classMembers`/`nsExports`/`nsTypes`, added the real `namespaceMembers` (and `catalog`). Also found a related bug while in there — the script counted unused *files* from a non-existent top-level `report.files`; knip nests them in `issues[].files[]`, so `files` is now in `COUNTED_KEYS` and unused files actually move the count. Live count unchanged at **78** (1 devDep + 37 exports + 40 types).

**Baseline note.** Good catch on the description drift — the committed `knip-baseline.json` was already `78` (the merge of main #505 removed a dead symbol, so I ratcheted 79→78 in ea6d75a). Updated the stale "79" in the PR test plan.

Everything stays advisory as designed; these were the correctness fixes needed before the gate can flip to blocking.

ClintEastman02:
Thanks @krokoko! Folded in the non-blocking nits in 2a7555c (ayushtr-aws's three findings were already addressed in 384efa8):

1. **Ratchet fail-open** — good catch, this was the one worth hardening now. Pinned knip **exactly** (`6.20.0`, dropped the caret) and `countIssues` now requires `report.issues` to be an array: if a future knip reshapes its JSON the script exits 2 (fails loud) instead of counting 0 and reporting a false pass. So a schema change becomes a deliberate, reviewed bump rather than a silently-disabled gate.
2. **`nsExports`/`nsTypes`** — clarified in the `COUNTED_KEYS` comment that the list is complete for the pinned knip and those keys aren't part of its schema (namespace/enum members surface as `namespaceMembers`/`enumMembers`, both counted). The fail-loud guard backstops any future drift.
3. **`validation.ts` comment** — you're right, the `void` is there for `noUnusedLocals`, not eslint (the `^_` binding is already exempt via `varsIgnorePattern`). Reworded to attribute it correctly.
4. **`countIssues` fixture test** — agree this pairs with the blocking flip; I've left it as a deferred item for that follow-up rather than expanding this PR's scope, since the count is exercised end-to-end by the advisory CI job today.

Baseline is holding at **78** through the recent main merges. All checks green. Ready for another look whenever you have a moment. 🙏

ayushtr-aws (dismissed):

.github/workflows/security-pr.yml
```diff
@@ -86,3 +92,49 @@ jobs:
 
       - name: Workflow static analysis (zizmor)
         run: mise run security:gh-actions
+
+  dead-code:
```
krokoko: should this live in the security pr workflow ? could probably be a separate one

agent/mise.toml
```diff
@@ -44,6 +44,10 @@ run = "uvx ruff format --check"
 description = "Type check with ty"
 run = "uv run ty check"
 
+[tasks."lint:deadcode"]
+description = "Dead-code detection with vulture (#282, cairn MVG gate #6): unused functions/classes/methods ruff F cannot see. Config + allowlist in pyproject.toml / .vulture_allowlist.py."
+run = "uvx vulture"
```
krokoko: uv run instead of uvx

.github/workflows/security-pr.yml
ClintEastman02: Done — moved the dead-code job to its own workflow (`.github/workflows/dead-code-pr.yml`).

agent/mise.toml
ClintEastman02: Fixed — changed to `uv run vulture`.

krokoko (approved):

PR #523, opened 2026-07-01
feat: unified task replay bundle for operators (#515)

### LINKED ISSUE #515: API: Unified task replay bundle for operators
> **Roadmap:** Unified cross-plane trace correlation
> **Priority:** P1

## Component

API or orchestration

## Describe the feature

Provide a single operator-facing **replay bundle** aggregating existing telemetry sources for a task: ordered `TaskEvents`, verification report, S3 trace URI, `prompt_version`, `workflow_ref`, key OTEL correlation IDs, and cost summary. Enables post-mortems, eval harness input, and harness-engineering replay without manual correlation across CloudWatch, DynamoDB, and S3.

Literature (arXiv:2605.18747 §3.5.1) treats deep telemetry + replay as the substrate for harness optimization.

## Use case

- **Operators** debug failed tasks with one API/CLI call.
- **Eval harness (#236)** consumes a stable replay JSON schema.
- **Compliance** exports a task's auditable trajectory for review.

## Proposed solution

1. **API:** `GET /v1/tasks/{id}/replay` (operator/IAM gated) returning:

```json
{
  "task_id": "...",
  "workflow_ref": "...",
  "prompt_version": "...",
  "events": [ /* TaskEvents chronological */ ],
  "verification_report": { /* if present */ },
  "trace_uri": "s3://...",
  "otel_trace_id": "...",
  "cost_usd": 0.0,
  "collected_at": "..."
}
```

2. **CLI:** `bgagent replay <task-id> [--json] [--output file]`

3. **v1:** Aggregate existing stores only — no new persistence layer.

4. Document schema in `API_CONTRACT.md`; link from `OBSERVABILITY.md`.

### Acceptance criteria

- [ ] REST endpoint with auth consistent with task read permissions (operator tier)
- [ ] CLI command with human and JSON output
- [ ] Schema documented; example fixture in `cdk/test/`
- [ ] Handles missing trace (task submitted without `--trace`) gracefully
- [ ] Types synced CLI ↔ API

## Other information

- **Extends:** Unified cross-plane trace correlation (roadmap)
- **Related:** Verification evidence bundles, golden trajectory conformance issue
- **Paper:** arXiv:2605.18747 deep telemetry and replayability

## Acknowledgements

krokoko:
## Summary

Implements **#515 — a unified operator replay bundle**: `GET /v1/tasks/{task_id}/replay` + `bgagent replay`, aggregating a task's existing telemetry (chronological `TaskEvents`, verification verdict, `--trace` S3 URI, `prompt_version`/`workflow_ref`, OTEL correlation id, cost) into one JSON so a task can be post-mortem'd / fed to an eval harness without manually correlating CloudWatch, DynamoDB, and S3.

Live-testing the feature surfaced several **pre-existing bugs** on the repo-less path, which this branch also fixes (all found and verified end-to-end against a deployed dev stack).

### Replay bundle (the feature)
- **API** — `cdk/src/handlers/get-task-replay.ts`: `GET /v1/tasks/{task_id}/replay`, Cognito + **owner-scoped auth identical to `GET /tasks/{id}`**. Pages `TaskEvents` chronologically (ULID sort) up to `MAX_REPLAY_EVENTS` with a truncation WARN. Aggregates `TaskRecord` fields. Wired in `task-api.ts` with read grants on both tables.
- **CLI** — `cli/src/commands/replay.ts`: `bgagent replay <task-id> [--json] [--output <file>]`. `--output` is a file path (mirrors `trace download`), `--json` is the format toggle (avoids the repo's `--output <format>` collision).
- **Types** — new `ReplayBundle` + `VerificationReport` in `cdk/src/handlers/shared/types.ts`, mirrored in `cli/src/types.ts` (types-sync gate passes).
- **Graceful-missing** — fields whose source didn't run for a task are `null`/empty, never omitted, so consumers get a stable schema (`trace_uri` null without `--trace`, etc.).
- **Docs** — `API_CONTRACT.md` schema + example, `OBSERVABILITY.md` cross-link, example fixture `cdk/test/fixtures/replay-bundle.example.json`.

### Agent-side persistence (per the issue's intent)
The issue's schema listed fields that weren't actually stored. Persist them at the terminal write:
- `otel_trace_id` (new `observability.current_otel_trace_id()`), threaded via `TaskResult`.
- `build_passed`/`lint_passed` were present on `TaskResult` but silently dropped by the write allowlist — now persisted, as a **tri-state** (`bool | None`): the gate result when it ran, `None` when it didn't (repo-less has no build/lint), so the bundle reports `verification: null` rather than a fictional `{false, false}`.
- Old records lack these fields and surface as `null` (no backfill).

### Bug fixes found during live testing
- **Premature completion on background work.** A repo-less research task launched Claude Code's background `Workflow` tool, which returns immediately and runs detached — the runner finalized on the first `ResultMessage` with a placeholder artifact ("…running… watch /workflows") while the real work ran on. Root cause: `allowed_tools` is only an **auto-approve** list (per the Agent SDK), not a surface restriction, so under `bypassPermissions` the defer/off-session tools were reachable despite the workflow's `[Read,Glob,Grep,WebFetch]` allow-list.
  - **Fix:** hard-block via `disallowed_tools=[Workflow, Task, Agent, Monitor, SendMessage, Cron*]` (removes them from the model's context even under bypass). Repo-less tasks also load **no on-disk settings** (`setting_sources=[]`) so a stray skill can't reintroduce a `Workflow`. Prompt guard tells the agent to finish in-session.
  - **Exit-path backstop:** `deliverers._reject_if_deferral` fails the deliver step **loudly** (terminal FAILED) if the final message defers instead of delivering — so a placeholder is never uploaded as the artifact even if the model finds another way to defer.
- **Misleading repo-less branch name.** Repo-less tasks showed a `bgagent/<id>/...` branch though they never branch. Now empty (keyed off `!body.repo`, so a repo-optional workflow given a repo still branches); CLI hides the empty Branch line.
- **Haiku on-demand 400.** `ANTHROPIC_DEFAULT_HAIKU_MODEL` was the bare foundation-model id `anthropic.claude-haiku-4-5-20251001-v1:0`, which 400s on Bedrock ("on-demand throughput isn't supported" — Claude 4.x needs an inference profile); WebFetch's Haiku summarization sub-calls hit this. Changed to the granted `us.anthropic.claude-haiku-4-5-20251001-v1:0` profile, and **refactored to be config-driven** (`TaskConfig.haiku_model`, resolved in `build_config`) instead of a hardcoded literal duplicated across `runner.py` and `agent.ts`.

## What was tested

**Automated — all green:**
- **CDK: 123 suites / 2228 tests.** New/updated: replay handler (auth 401/400/403/404/500, ordering, multi-page, graceful-missing, truncation cap, fixture validation), route/count assertions, repo-less branch + idempotent-replay regression, Haiku env-var assertion.
- **Agent: 1118 tests, 80.18% coverage** (gate 72%). Replay-field persistence (build_passed=False not dropped, tri-state None omitted, otel id present/absent), tool-surface hardening (Workflow/Task/Agent disallowed, repo-less setting_sources), deferral guard (s3 + comment + false-positive), Haiku config resolution.
- **CLI: 570 tests.** replay command (human/json/file/clobber), types-sync gate passes.
- Lint (ruff/eslint) + typecheck clean; `cdk synth` clean (cdk-nag passes).

**Manual review:** ran the PR-review toolkit (code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer) across the work; findings addressed in the review-fix commits (type-sync drift, tri-state verification, deferral exit-path guard, observability). A final code review of the Haiku fix confirmed the `us.*` profile is actually on the CDK grant allowlist (no 400→403 trade).

**Live verification (deployed dev stack, us-east-1) — all confirmed working:**
- **Knowledge (repo-less) search task** — Haiku 400 errors are **gone**; the task completes with a real cited artifact (no background-Workflow deferral); `verification: null` (correct — no build/lint gate).
- **Coding task with a repo** — completes end-to-end; `verification` populated; branch name correct.
- **`--trace`** exercised on the runs, and **all replay scenarios verified for both tasks** — the bundle returns ordered events, `otel_trace_id`, cost, `trace_uri` (when `--trace`), and the correct graceful-missing nulls.

Redacted sample replay record (personal identifiers, account, repo, task id removed):
```json
{
  "requestMetadata": { "user_id": "<redacted>", "repo": "<owner>/<repo>", "task_id": "<task-ulid>" },
  "verification": { "build_passed": true, "lint_passed": true },
  "otel_trace_id": "<32-hex>",
  "trace_uri": "s3://<trace-bucket>/traces/<user_id>/<task-ulid>.jsonl.gz",
  "cost_usd": 0.0
}
```

## Notes for reviewers
- **Scope beyond the issue:** #515 asked for aggregate-only v1 ("no new persistence"). This branch **does** add fields to the agent terminal-write contract (per maintainer direction) so the bundle is complete — a "major change" per AGENTS.md, called out explicitly.
- The tool-surface hardening + Haiku fix are **agent-runtime changes requiring a redeploy** to take effect (verified live post-deploy).
- The deferral guard (`_reject_if_deferral`) is a **phrase heuristic** — defense-in-depth behind the tool block, not a perfect classifier; a structured completion signal would be a stronger follow-up.
- `cdk/cdk.json` is intentionally **not** included (local-testing context only).

Closes #515


agent/src/workflow/deliverers.py
```diff
@@ -34,6 +34,52 @@
...
+    "i'll deliver the",
+    "i will deliver the",
+    "as soon as it completes",
+    "as soon as it finishes",
+    "check back",
+    "come back later",
+)
+
+
+def _reject_if_deferral(text: str) -> None:
```
isadeks, line 85: **🔴 Blocking — false-FAILs legitimate completed answers.** These markers are matched as unanchored, lowercased substrings against the *entire* deliverable. A `ValueError` here fails the `deliver_artifact` step → `pipeline` flips `overall_status` to `error` → `write_terminal(..., "FAILED", ...)`, and since the raise is before upload, the valid answer is discarded.

Two compounding problems:
- **`_artifact_body` is the chokepoint for both `deliver("s3")` and `deliver("comment")`, and `deliver_artifact` is used by `default/agent-v1.yaml` too** — the platform default when no `workflow_ref` is supplied. Only `web_research.py` got the "don't defer" prompt guidance; `default_agent.py` was not updated, so its agent has no reason to avoid these phrases.
- **Markers appear verbatim in genuine finished prose/URLs.** e.g. an answer to *"explain our GitHub Actions setup"* that cites `.../actions/workflows/ci.yml` matches `"/workflows"`; *"compaction runs in the background"* matches `"running in the background"`; `"check back"`, `"watch progress"` similarly.

This is strictly worse than the placeholder bug it prevents — it throws away good output. Suggest anchoring to whole phrases + excluding URLs/code fences, or (deeper) a structured completion signal rather than a content blocklist layered on top of the already-effective `_DISALLOWED_TOOLS` surface lock. At minimum, add the anti-deferral guidance to the default-workflow prompt.

agent/src/runner.py
```diff
@@ -335,31 +341,80 @@ def _initialize_policy_engine_and_hooks(
...
+def _resolve_setting_sources(config: TaskConfig) -> list[Literal["user", "project", "local"]]:
+    """Which on-disk Claude Code settings the CLI may load for this task.
+
+    Repo-bound tasks load ``["project"]`` so the cloned repo's own ``.claude/``
+    config is honored. A repo-less task has no cloned repo to discover from, so
+    it loads nothing — defense-in-depth that also stops a stray on-disk skill
+    (e.g. one that spawns a background Workflow) from being reachable. Kept as a
+    named helper so the policy is unit-testable without driving the SDK.
+    """
+    return ["project"] if config.requires_repo else []
```
isadeks: **🔴 Blocking — regression for repo-optional-with-repo.** This keys on `config.requires_repo` (a static workflow property), but a repo-**optional** workflow (`web-research-v1`, `agent-v1`, both `requires_repo: false`) that *is given a repo* takes the repo-bound clone path (`pipeline.py:760` gates on `not requires_repo and not repo_url`): it clones the repo, sets `cwd` to it, and builds a prompt referencing its config — yet `setting_sources=[]`, so `.claude/settings.json`, rules, and agents are never loaded.

Pre-PR this was unconditionally `["project"]`, so this is a behavior regression for that case. Note the sibling handler `create-task-core.ts:938` deliberately keys `branch_name` on `!body.repo` (repo presence) *with a comment warning about exactly this distinction*. Suggest keying on `config.repo_url` here to match:
```python
return ["project"] if config.repo_url else []
```

agent/src/pipeline.py
```diff
@@ -1127,6 +1128,7 @@ def _on_trace_truncated(max_bytes: int, first_dropped: int) -> None:
                 cache_read_input_tokens=usage.cache_read_input_tokens if usage else None,
                 cache_creation_input_tokens=usage.cache_creation_input_tokens if usage else None,
                 trace_s3_uri=trace_s3_uri,
+                otel_trace_id=current_otel_trace_id(),
```
isadeks, line 1131: **🔴 Blocking — crash path drops `otel_trace_id`.** This site and the repo-less site (`:366`) both capture `otel_trace_id`, but the crash-path `crash_result = TaskResult(status="error", ...)` around **line 1194** does *not* pass it, and `write_terminal(config.task_id, "FAILED", crash_result.model_dump())` at `:1201` is what persists crashed tasks. So FAILED tasks — the primary post-mortem case #515's replay bundle exists for — get `otel_trace_id: null`.

The crash `except` is still inside the `with task_span() as root_span` block, so a valid id is available there and is just being dropped. This is the "capture at 3 separate sites" smell — capturing once from `root_span` (or stamping in `write_terminal`) would cover the crash path automatically.

isadeks (commented):
## Automated code review (extra-high effort)

Reviewed the full diff via a multi-angle finder sweep with independent verification of each finding. The feature itself is solid — auth mirrors `get-task.ts` exactly, the pagination loop is sound, tri-state persistence uses correct `is not None` guards, CDK↔CLI types stay in sync, and the docs mirrors regenerate clean. The inline comments below are the real defects, ranked by severity.

**Blocking (correctness):**
1. Deferral guard `_reject_if_deferral` false-FAILs legitimate completed answers and also governs the general-purpose `default/agent-v1` workflow (whose prompt wasn't updated).
2. `_resolve_setting_sources` keys on `requires_repo` instead of repo presence → repo-optional-with-repo silently drops the cloned repo's `.claude/` config (a regression).
3. Crash-path `TaskResult` omits `otel_trace_id` → FAILED tasks (the primary post-mortem case for #515) get `otel_trace_id: null`.

**Deploy-hardening:** replay response has no byte bound (6 MB Lambda limit risk on large `--trace` tasks); `getTaskReplayFn` runs on the 3 s / 128 MB defaults.

**Cleanup / lower severity** are also flagged inline. All findings are advisory — happy to discuss any of them.

cdk/src/handlers/get-task-replay.ts
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * Hard cap on events embedded in one bundle. A replay is a single-shot
+ * aggregate, not a paginated feed, so we bound the response rather than expose
+ * a cursor: an 8-hour task emits on the order of hundreds-to-low-thousands of
+ * events, and DynamoDB Query pages at 1 MB. We page until this cap, then stop
+ * and flag truncation so the bundle never grows unbounded (a runaway task can't
+ * produce a multi-MB Lambda response). Callers needing the full tail use the
+ * paginated ``GET /tasks/{id}/events`` endpoint.
+ */
+export const MAX_REPLAY_EVENTS = 5000;
```
isadeks, line 42: **🟠 6 MB Lambda response-limit risk.** `MAX_REPLAY_EVENTS` bounds event *count* but nothing bounds *bytes* — `collectEvents`/`assembleBundle` embed all events verbatim into one `JSON.stringify({data})`. `progress_writer.py` uses `_PREVIEW_MAX_LEN_TRACE = 4096` and `agent_turn` events carry two previews (~8–9 KB each in trace mode), so 5000 events ≈ 40+ MB — ~7× over Lambda's 6 MB synchronous-response cap → API Gateway 502, for exactly the long/expensive `--trace` tasks operators most need to inspect. The docstring's claim that this "can't produce a multi-MB Lambda response" doesn't hold in trace mode. Suggest a running byte bound (stop + WARN) alongside the count cap.

cdk/src/constructs/task-api.ts
```diff
@@ -579,6 +579,17 @@ export class TaskApi extends Construct {
       bundling: commonBundling,
     });
 
+    // Operator replay bundle (#515): aggregates TaskRecord + chronological
+    // TaskEvents. Reads both tables; commonEnv already carries both table names.
+    const getTaskReplayFn = new lambda.NodejsFunction(this, 'GetTaskReplayFn', {
```
isadeks, line 588: **🟠 No `timeout`/`memorySize` → 3 s / 128 MB defaults.** Every heavier handler in this construct sets `timeout: 15s` + raised memory, and the construct's own comment (`~:41-47`, and `cancelTaskFn` at `~:565`, `getTraceUrlFn` at `~:768`) notes the 3 s default causes INIT-timeout → 502. Replay is the heaviest read path — `GetItem` + a multi-page Query loop (up to 5000 events / ~5 pages) + full-bundle serialization — yet inherits the defaults. Cold start on a long history risks timeout and/or OOM at 128 MB. Suggest matching the other read handlers (e.g. 15 s / 512 MB).

cdk/src/handlers/get-task-replay.ts
```diff
@@ -0,0 +1,166 @@
...
+    const remaining = MAX_REPLAY_EVENTS - events.length;
+    const page = await ddb.send(new QueryCommand({
+      TableName: EVENTS_TABLE_NAME,
+      KeyConditionExpression: 'task_id = :tid',
+      ExpressionAttributeValues: { ':tid': taskId },
+      ScanIndexForward: true, // chronological (ULID event_id is time-ordered)
+      Limit: remaining,
+      ExclusiveStartKey: startKey,
+    }));
+    events.push(...((page.Items ?? []) as EventRecord[]));
```
isadeks: **🟡 Event shape diverges from the `/events` feed.** The bundle embeds raw `EventRecord`s (keeps `task_id`/`ttl`, and `metadata` can be `undefined`), whereas `GET /tasks/{id}/events` strips `task_id`/`ttl` and normalizes `metadata` to `{}`. A consumer that moves from the events feed (where `event.metadata.x` is always safe) to the one-call bundle can hit `TypeError` on an event with no stored metadata. Suggest normalizing consistently or documenting the difference.

agent/src/runner.py
```diff
@@ -335,31 +341,80 @@ def _initialize_policy_engine_and_hooks(
...
+# block both); ``Monitor`` streams a background command's output mid-turn;
+# ``SendMessage`` resumes/relaunches background agents; the ``Cron*`` tools
+# schedule deferred work. All are "return now, work continues off-session"
+# vectors a one-shot task cannot await. NOT blockable here: background ``Bash``
+# (a ``run_in_background`` PARAMETER of Bash, not a tool name) — but a detached
+# Bash child dies with the MicroVM on return, so it can't produce
+# arrives-later work the way a cloud Workflow does; the deliver-artifact
+# deferral guard (deliverers._reject_if_deferral) is the backstop for anything
+# that still ends in a placeholder.
+_DISALLOWED_TOOLS = [
```
isadeks, line 368: **🟡 Altitude — hand-enumerated block list is version-fragile.** The comment itself concedes "name varies by CLI version." A renamed/added off-session tool (e.g. `WorkflowResume`, a 4th `Cron*`) would fall through `disallowed_tools` → `bypassPermissions` → silently allowed, reintroducing the exact detached-work bug this PR fixes, with no test to catch it. A capability/category filter would be the durable form; at minimum add a regression test that pins the block set.

agent/tests/test_runner.py
```diff
@@ -339,3 +342,50 @@ def test_no_attribution_file_when_role_unset_but_header_still_set(self, monkeypa
...
+
+    def test_repo_bound_loads_project_settings(self):
+        config = _config(requires_repo=True)
+        assert _resolve_setting_sources(config) == ["project"]
+
+
+class TestSetupAgentEnv:
+    """Environment the Claude Code subprocess inherits."""
+
+    def test_haiku_model_env_is_set_from_config(self, monkeypatch):
```
isadeks, line 398: **🟡 Test isolation leak.** `_setup_agent_env` writes `ANTHROPIC_DEFAULT_HAIKU_MODEL` (and `CLAUDE_CODE_USE_BEDROCK`) directly to `os.environ`; neither is in conftest's clean allowlist and this test doesn't restore them. The two `test_config.py` cases that read the var self-guard with `delenv`, so it's latent today — but any future test reading the haiku env without guarding becomes order-dependent/flaky. Prefer `monkeypatch.setenv`, or restore in teardown.

agent/src/observability.py
```diff
@@ -57,6 +57,23 @@ def task_span(
...
+    ``None`` when there is no recording span (e.g. tracing disabled locally) or
+    the context is invalid, so callers can treat it as a graceful-missing field.
+    """
+    span = trace.get_current_span()
+    ctx = span.get_span_context()
+    if not ctx.is_valid:
+        return None
+    # format_trace_id renders the 128-bit id as zero-padded 32-char hex, matching
+    # what appears in CloudWatch/X-Ray — so the persisted value joins directly.
+    return trace.format_trace_id(ctx.trace_id)
```
isadeks, line 76: **🟡 Doc overclaim.** The comment says the bare 32-hex id "matches what appears in CloudWatch/X-Ray — so the persisted value joins directly." It joins in CloudWatch Transaction Search (OTEL format), but the **X-Ray console** renders IDs as `1-{8hex}-{24hex}`, so pasting `otel_trace_id` into X-Ray search returns nothing. Suggest softening the claim or noting the format transform, since the bundle advertises this as a cross-plane correlation id.

cli/src/format.ts
```diff
@@ -355,6 +359,44 @@ export function formatJson(data: unknown): string {
...
+    `Workflow:    ${bundle.resolved_workflow ? `${bundle.resolved_workflow.id}@${bundle.resolved_workflow.version}` : (bundle.workflow_ref ?? dash)}`,
+    `Prompt ver:  ${bundle.prompt_version ?? dash}`,
+    `Cost:        ${bundle.cost_usd != null ? `$${Number(bundle.cost_usd).toFixed(COST_USD_DECIMALS)}` : dash}`,
+    `OTEL trace:  ${bundle.otel_trace_id ?? dash}`,
+    `Session:     ${bundle.session_id ?? dash}`,
+    `Trace URI:   ${bundle.trace_uri ?? dash}`,
+  ];
+  if (bundle.verification) {
+    const v = bundle.verification;
+    const verdict = (b: boolean | null): string => (b === null ? dash : b ? 'PASSED' : 'FAILED');
```
isadeks: **🟢 Cleanup — duplicated rendering.** The PASSED/FAILED verdict here and the cost line (`$${Number(x).toFixed(COST_USD_DECIMALS)}`) are copy-pasted from `formatTaskDetail`. Extracting `formatVerdict`/`formatCostUsd` helpers would keep `status` and `replay` from diverging (e.g. if a `SKIPPED` state or spend-format change is added later).

agent/src/workflow/deliverers.py
krokoko: Fixed in 24241c0ef63d43275a0ffe5a78db3cd4c4bf0b01. Pared `_DEFERRAL_MARKERS` down to specific full first-person deferral phrases (dropped `/workflows`, `watch progress`, `check back`, `running in the background`, etc.), and added `_strip_code_and_urls` so cited URLs and code fences are removed before matching — a genuine answer citing `.../actions/workflows/ci.yml` or quoting a marker in a fence no longer trips it. Also added the anti-deferral guidance to `default_agent.py`'s prompt (it routes through the same `deliver_artifact` chokepoint but never had it). New parametrized regression test covers the four false-positive cases you named (CI URL, background-topic prose, watch/check-back instructions, fenced quote). Kept as a backstop behind the `_DISALLOWED_TOOLS` surface lock, as you noted.

agent/src/runner.py
krokoko: Fixed in 24241c0ef63d43275a0ffe5a78db3cd4c4bf0b01. Now keys on `config.repo_url` (repo presence) instead of `config.requires_repo`, exactly as suggested — a repo-optional workflow given a repo now loads its `.claude/` config instead of dropping it. Added a regression test (`test_repo_optional_with_repo_loads_project_settings`) pinning the repo-optional-with-repo case.

agent/src/pipeline.py
krokoko: Fixed in 24241c0ef63d43275a0ffe5a78db3cd4c4bf0b01. The crash-path `crash_result` now passes `otel_trace_id=current_otel_trace_id()` — the call is still inside `with task_span()` so the id is live. FAILED tasks now persist the id instead of null. Covered by a new crash-path test (`test_crash_path_persists_otel_trace_id`).

cdk/src/handlers/get-task-replay.ts
krokoko: Fixed in 24241c0ef63d43275a0ffe5a78db3cd4c4bf0b01. Added `MAX_REPLAY_EVENT_BYTES` (5 MiB) as a running byte bound alongside the count cap — the loop now stops paging once accumulated event bytes cross the threshold, leaving headroom under the 6 MB limit. Beyond the WARN, the bundle now carries an `events_truncation` field (`{reason, returned_events}`, null when the full list fit) so the operator reading the bundle — not just CloudWatch — can tell a clipped replay from a complete one. Corrected the docstring's overclaim. New byte-cap test asserts truncation short of the count cap and under 6 MB.

cdk/src/constructs/task-api.ts
krokoko: Fixed in 24241c0ef63d43275a0ffe5a78db3cd4c4bf0b01. The replay handler now sets `timeout: 15s` (`API_HANDLER_TIMEOUT_SECONDS`) and `memorySize: 512MB` (`SCREENING_HANDLER_MEMORY_MB` — the bundle holds up to `MAX_REPLAY_EVENT_BYTES` in memory during serialization), matching the other heavy read handlers. Added a construct test pinning both.

cdk/src/handlers/get-task-replay.ts
krokoko: Fixed in 24241c0ef63d43275a0ffe5a78db3cd4c4bf0b01. Rather than document the difference, I normalized: bundle events now go through a new `ReplayEvent` type that strips `task_id`/`ttl` and defaults `metadata` to `{}`, matching the `/events` feed exactly — so `event.metadata.x` is always safe. Mirrored the type in `cli/src/types.ts`, updated the fixture and docs, and added a normalization test.

agent/src/runner.py
krokoko: Addressed in 24241c0ef63d43275a0ffe5a78db3cd4c4bf0b01. Kept the hand-enumerated list (a capability/category filter isn't exposed by the SDK today) but pinned the exact set with a regression test (`test_disallowed_tool_set_is_pinned`) so a silent drop/rename forces a deliberate update and a look at whether a renamed off-session tool needs adding. A durable category filter would still be the better long-term form if the SDK grows one.

agent/tests/test_runner.py
krokoko: Fixed in 24241c0ef63d43275a0ffe5a78db3cd4c4bf0b01. The test now pre-claims `ANTHROPIC_DEFAULT_HAIKU_MODEL` and `CLAUDE_CODE_USE_BEDROCK` through `monkeypatch.setenv` before calling `_setup_agent_env`, so monkeypatch's teardown restores them and the write can't leak into order-dependent neighbors.

agent/src/observability.py
krokoko: Fixed in 24241c0ef63d43275a0ffe5a78db3cd4c4bf0b01. Softened the comment to note the id is OTEL format (joins directly in CloudWatch Transaction Search) and that the X-Ray console renders ids as `1-{8hex}-{24hex}`, so a lookup there needs the format transform.

cli/src/format.ts
krokoko: Fixed in 24241c0ef63d43275a0ffe5a78db3cd4c4bf0b01. Extracted `formatCostUsd` and `formatVerdict` helpers, used by both `formatTaskDetail` and `formatReplay` so the cost format and PASSED/FAILED verdict can't diverge.

krokoko:
Thanks for the thorough review — all ten findings addressed in 24241c0ef63d43275a0ffe5a78db3cd4c4bf0b01 (pushed). Summary:

**🔴 Blocking (all fixed):**
1. **Deferral guard** — pared markers to full first-person phrases + strip code/URLs before matching; added anti-deferral guidance to `default/agent-v1`; regression test for the four false-positive cases.
2. **`_resolve_setting_sources`** — now keys on `repo_url`, not `requires_repo`; regression test for repo-optional-with-repo.
3. **Crash-path `otel_trace_id`** — captured on `crash_result`; crash-path test.

**🟠 Deploy-hardening (both fixed):**
4. **6 MB response risk** — added `MAX_REPLAY_EVENT_BYTES` byte cap; also surfaced truncation *in the bundle* (`events_truncation`) so it's visible to the operator, not just CloudWatch.
5. **Handler timeout/memory** — 15s / 512MB, matching the other heavy read handlers.

**🟡🟢 Lower severity (all addressed):** event-shape normalized to the `/events` feed (new `ReplayEvent` type); disallowed-tool set pinned with a test; env-var isolation leak fixed via monkeypatch; X-Ray/OTEL comment corrected; `formatCostUsd`/`formatVerdict` extracted.

**One extra caught during a follow-up self-review:** the byte-cap WARN reached only CloudWatch, so a truncated bundle was indistinguishable from a complete one to the consumer — hence the `events_truncation` field above and the `TRUNCATED` marker in `bgagent replay`'s human output.

All green: CDK 2230, CLI 571, agent 1125; types-sync + docs-sync clean. Note `cdk/cdk.json` remains intentionally excluded (local-testing context, per the PR description).

isadeks (approved):
LGTM — approving. Pulled the fix commit (`24241c0`) and verified each fix rather than taking the replies at face value:

**🔴 Blocking — all correctly fixed:**
- Deferral guard now strips code fences / inline code / URLs before matching and matches only full first-person phrases; `default_agent.py` got the anti-deferral guidance it was missing. The four false-positive cases are pinned by a parametrized test.
- `_resolve_setting_sources` keys on `config.repo_url` (repo presence), so a repo-optional-with-repo task loads its `.claude/` config again.
- Crash-path `crash_result` now captures `otel_trace_id` — FAILED tasks no longer persist null.

**🟠 Deploy-hardening:**
- `MAX_REPLAY_EVENT_BYTES` byte cap: I checked the paging loop — `byteCapped` forces `startKey = undefined`, so it terminates cleanly and the single-event overshoot is bounded/documented (no off-by-one or infinite loop). Surfacing truncation in the bundle via `events_truncation` is a nicer solution than the WARN-only suggestion.
- Replay handler raised to 15s / 512MB with a pinning test.

**🟡🟢** Event shape normalized to the `/events` feed (also removes the potential `metadata` TypeError), disallowed-tool set pinned, env-leak fixed, X-Ray/OTEL comment corrected, `formatCostUsd`/`formatVerdict` extracted.

Every finding has a matching regression test. Nice turnaround.

isadeks (approved):

PR #531, opened 2026-07-06
Test: user store helpers (self-review fixture)

nizar-lahlali:
## Purpose

Test PR to exercise the **self-review / PR-review tooling** added on the parent branch (`feat/263-self-review-pr-comment`).

This branch adds `agent/src/user_store.py`, which deliberately contains a spread of issues for the reviewer to catch:

- Hardcoded credentials (`API_TOKEN`, `DB_PASSWORD`)
- SQL injection (`get_user` string-interpolated query)
- Command injection (`run_report`, `shell=True` + unsanitized input)
- Insecure deserialization (`load_session`, `pickle.loads`)
- Weak, unsalted hashing (`hash_password`, MD5)
- Bare `except:` swallowing errors (`divide`)
- Unclosed DB connection (`get_user`)
- Dead code / unused variable (`compute_total`)
- Predictable "randomness" for a security value (`make_temp_key`)
- Unmanaged file handle + hardcoded path (`read_config`)

> ⚠️ **Do not merge.** This is a throwaway test fixture. Base is the feature branch, not `main`.


nizar-lahlali:
## 🔍 Self-Review Summary

### Self-Review Summary

**Findings:** 11

#### Issues found

- **Security – Hardcoded secret (API token)** — `agent/src/user_store.py:11`: `API_TOKEN` is hardcoded in source and will be committed to version history, violating OWASP A02 and secret-scanning policies. — **Recommendation:** Remove the literal; load the value at runtime from an environment variable (e.g. `os.environ["API_TOKEN"]`) or from AWS Secrets Manager / SSM Parameter Store.

- **Security – Hardcoded secret (DB password)** — `agent/src/user_store.py:12`: `DB_PASSWORD = "admin123"` is a hardcoded weak credential committed to the repository. — **Recommendation:** Same as above — inject via environment variable or secrets manager; never commit passwords.

- **Security – SQL injection** — `agent/src/user_store.py:18`: The query is assembled with string concatenation from an unsanitised `username` argument (`"SELECT * FROM users WHERE name = '" + username + "'"`). A value like `' OR '1'='1` will bypass all filtering (OWASP A03). — **Recommendation:** Use a parameterised query: `cur.execute("SELECT * FROM users WHERE name = ?", (username,))`.

- **Security – Insecure password hashing (MD5)** — `agent/src/user_store.py:25`: `hashlib.md5` is a fast, broken hash that offers no meaningful resistance to brute-force or rainbow-table attacks for stored passwords (OWASP A02). — **Recommendation:** Replace with a password-hashing function: `bcrypt`, `argon2-cffi`, or at minimum `hashlib.pbkdf2_hmac("sha256", password.encode(), salt, 260_000)` with a random per-user salt.

- **Security – OS command injection** — `agent/src/user_store.py:29-30`: `subprocess.run("generate_report --name " + report_name, shell=True, ...)` passes an unsanitised argument into a shell string. A `report_name` containing `;`, `&&`, or `$()` can execute arbitrary OS commands (OWASP A03). — **Recommendation:** Pass a list and omit `shell=True`: `subprocess.run(["generate_report", "--name", report_name], capture_output=True)`.

- **Security – Insecure deserialization** — `agent/src/user_store.py:34`: `pickle.loads(blob)` can execute arbitrary Python code if `blob` is attacker-controlled (OWASP A08). The Python docs explicitly warn against unpickling untrusted data. — **Recommendation:** Replace `pickle` with a safe format such as `json.loads` for plain data, or sign/verify the blob with HMAC before deserialising if a binary format is required.

- **Security – Predictable / non-random session key** — `agent/src/user_store.py:61`: `os.getpid()` returns a small, sequential, predictable integer. Using it as a key/token is trivially guessable. — **Recommendation:** Use `secrets.token_hex(32)` (stdlib) to generate a cryptographically random key.

- **Bug – Database connection never closed (resource leak)** — `agent/src/user_store.py:16`: `conn = sqlite3.connect(db_path)` is opened but never closed; if `cur.execute` or `fetchall` raises, the handle leaks indefinitely. — **Recommendation:** Wrap the connection in a context manager: `with sqlite3.connect(db_path) as conn:` and close explicitly after use, or use `try/finally`.

- **Bug – File handle never closed (resource leak)** — `agent/src/user_store.py:55`: `f = open("/tmp/config.txt")` is never closed; an exception before `f.read()` would leave the descriptor open. — **Recommendation:** Use `with open("/tmp/config.txt") as f: data = f.read()`.

- **Bug – Bare `except` clause masks all exceptions** — `agent/src/user_store.py:50`: `except:` catches every `BaseException`, including `KeyboardInterrupt`, `SystemExit`, and `MemoryError`, silently returning `None` and hiding genuine failures. — **Recommendation:** Catch only the expected exception: `except ZeroDivisionError:` (or `except Exception:` if broader coverage is truly needed), and consider re-raising or logging.

- **Code quality – Dead code and unused variable** — `agent/src/user_store.py:39,42-43`: `debug = "unused"` is assigned but never read, and the `if False:` block is unreachable. Both will be flagged by linters and confuse readers. — **Recommendation:** Remove both the unused `debug` variable and the `if False:` block entirely.

- **Test gap – No tests for new module** — `agent/src/user_store.py` (whole file): The module is added with no corresponding test file (e.g. `agent/tests/test_user_store.py`). Per `AGENTS.md`, new agent modules require matching tests. — **Recommendation:** Add a `agent/tests/test_user_store.py` covering at least `compute_total`, `divide` (including the zero case), and `hash_password`; mock or skip functions that require external resources.

Issue #532, opened 2026-07-06
fix(security): semgrep uv-missing-dependency-cooldown on agent/pyproject.toml (masked behind #530)


mandeep408:
## Summary

The `mise run security` suite's `security:sast` stage (semgrep, `--config auto`) flags `agent/pyproject.toml`: the `[tool.uv]` block sets no dependency cooldown, so newly-published packages are resolved immediately. Newly published packages can be malicious or unstable (supply-chain risk).

- **Rule:** `package_managers.uv.uv-missing-dependency-cooldown.uv-missing-dependency-cooldown` (Blocking; added in semgrep 0.9.17)
- **File:** `agent/pyproject.toml` — `[tool.uv]` section

## Fix (decided — Option B, justified `nosemgrep`)

> **Note:** the originally-proposed `exclude-newer = "7 days"` was **rejected** after verification — a relative cooldown is structurally incompatible with this repo's exact `==` pins on recently-published packages (it makes `uv lock`/`uv sync` unsatisfiable and re-breaks on every fresh pin bump). Full analysis and the approval reversal are in [this thread](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/532#issuecomment-4906600675) (approved: [krokoko](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/532#issuecomment-4907460128)).

Add a **justified `nosemgrep` suppression** on the line immediately preceding `[tool.uv]`. The repo's threat model (surprise malicious/unstable version) is already covered by exact pins + committed `uv.lock` + reviewed `upgrade-main` PRs + `uv sync --frozen` in the image, so a global cooldown would be redundant and would fight the pinning workflow.

## Context

- This finding was **masked in CI behind the gitleaks failure (#530)**; once #530 merged, it became the next red in the scheduled security suite. Not caused by #530.
- It surfaces on the latest semgrep because `mise.toml` pins `semgrep = "latest"` and the sast task uses `--config auto` (rules fetched live from the registry). See follow-up below.

## Acceptance criteria

- `mise run security` passes through the `security:sast` stage on `main` (no `uv-missing-dependency-cooldown` finding).
- The agent build still resolves (`uv sync --all-groups` / `uv lock` succeed).
- No unrelated changes bundled in; scope is `agent/pyproject.toml` only.

## Follow-ups (separate issues, not this one)

[...]

Issue #313, opened 2026-06-10
fix: tune gitleaks aws-account-id rule and clear security:secrets failures


krokoko:
## Problem
`mise run security` fails at `security:secrets` with 53 gitleaks findings after the custom `aws-account-id` rule landed in f57b931.

## Acceptance criteria
- [ ] `mise run security:secrets` passes (0 leaks)
- [ ] False positives allowlisted: mise-action SHA pins, threat-composer artifacts, test fixtures, placeholder UUIDs
- [ ] Real account IDs removed from current source files (e.g. drawio diagrams)
- [ ] Historical remediated leaks baselined via `.gitleaksignore` where history cannot be rewritten

krokoko:
Starting implementation on branch fix/313-gitleaks-aws-account-id.

laithalsaadoon:
Picking this up. Classified all 25 `aws-account-id` findings on `main` (HEAD `af33bc0`) — **21 are false positives, 1 is a real account ID appearing in 3 places** (1 live, 2 history-only):

| Match | Location | Count | Verdict |
|---|---|---|---|
| `629881980618` | `.github/workflows/{deploy,integ,upgrade-main}.yml` | 3 | **FP** — substring of the `jdx/mise-action` commit SHA pin `…981980618f…`, not an account ID |
| `…440000` | `cdk/test/handlers/cancel-task.test.ts:62` | 1 | **FP** — trailing node of the RFC test UUID `550e8400-e29b-41d4-a716-446655440000` |
| 4 distinct 12-digit runs | `.threat-composer/**/*.tc.json` | 17 | **FP** — last node of v4 UUIDs (each preceded by `-`, e.g. `f30d02f6-9750-46ad-b710-398303478932`) |
| `169728770098` | `docs/diagrams/phase3-cedar-hitl.drawio:28` (**live**) | 1 | **REAL** — in a diagram label `AWS Cloud — us-east-1 (account 169728770098)` |
| `169728770098` | `docs/backlog/{CLI_COGNITO_NEW_PASSWORD_CHALLENGE,EARLY_PROGRESS_MILESTONES}.md` | 2 | **REAL**, but files already deleted at HEAD → history-only |

Plan, mapping to the ACs:
1. **Scrub the live real ID** — replace `169728770098` in `phase3-cedar-hitl.drawio` with the AWS-published placeholder `123456789012`.
[...]

PR #538, opened 2026-07-07
fix(security): suppress orch fixture via stopword, not SHA fingerprint (#537)

### LINKED ISSUE #537: fix(security): gitleaks fails on main — #530 SHA-keyed baseline broke after history rewrite
## Summary

`mise run security:secrets` (gitleaks) **fails on `main`** — the security suite's secret-scan is red for everyone. This is a **regression of the #530 fix**: the fix baselined the finding by commit-SHA fingerprint, and a subsequent history rewrite changed that SHA, so the baseline no longer matches.

Reproduce on clean `main` (`05f727b`):
```
$ mise run security:secrets
... WRN leaks found: 1
```

## Root cause

#530 suppressed a gitleaks false positive (test idempotency-key fixture `orch_abc_SUB-1` in `cdk/test/handlers/shared/orchestration-release.test.ts:125`) via a `.gitleaksignore` **fingerprint keyed on commit SHA**:
```
f98f47abc6f7820d7e5ce0375cf22551cb664b37:cdk/test/handlers/shared/orchestration-release.test.ts:generic-api-key:125
```
After #530 merged, the dependency-upgrade merges (#526 and related) **rewrote history**. The commit carrying the fixture (`fix(linear)... #247`) now has a **new SHA `73bfe08`** — same content, different hash. gitleaks scans live history and reports the finding at `73bfe082...:...:125`, which is **not** the fingerprint in `.gitleaksignore`, so the finding resurfaces.

The `.gitleaksignore` fingerprint approach is **inherently fragile to history rewrites** — any rebase/squash/dep-bump merge that touches or moves that commit invalidates the baseline. The pre-existing #313 aws-account-id fingerprints have the same latent fragility.

## Impact

- Secret-scan is red on `main` (scheduled Security suite + any PR — the required `Secrets, deps, and workflow scan` check runs gitleaks).
- **Blocks #532**: that PR's required CI check runs gitleaks, so #532 cannot merge until this is resolved.

## Proposed fix (verified locally)

Replace the SHA-keyed baseline with a **SHA-independent stopword allowlist** in `.gitleaks.toml` (the repo already uses this pattern for other test fixtures, e.g. `wat-opaque-123`, `test-signing-secret-abc123`):

```toml
[[allowlists]]
description = "Test idempotency-key fixture 'orch_abc_SUB-1' (not a real credential)."
stopwords = ["orch_abc_SUB-1"]
```

**Verified:** with this allowlist, `gitleaks git .` → `no leaks found` (exit 0), and it does **not** depend on any commit SHA, so it survives history rewrites. The stale `f98f47a...` line in `.gitleaksignore` can then be removed.

Consider auditing the #313 aws-account-id fingerprints for the same fragility (out of scope for the immediate fix, but the same failure mode applies).

## Acceptance criteria

- `mise run security:secrets` passes on `main`.
- Suppression is SHA-independent (survives history rewrites).
- Scope: `.gitleaks.toml` / `.gitleaksignore` only.

## Context

Discovered while implementing #532 — the pre-push security hook surfaced this pre-existing `main` failure. Not caused by #532 (which touches only `agent/pyproject.toml`).

mandeep408:
## What

Fix gitleaks failing on `main`: replace the SHA-keyed `.gitleaksignore` fingerprint for the `orch_abc_SUB-1` test fixture with a SHA-independent stopword allowlist in `.gitleaks.toml`.

Fixes #537.

## Why

#530 baselined a gitleaks false positive (test idempotency-key fixture `orch_abc_SUB-1` in `cdk/test/handlers/shared/orchestration-release.test.ts:125`) using a **commit-SHA fingerprint**. A later history rewrite (dep-bump merges, #526 et al.) changed that commit's SHA (`f98f47a` → `73bfe08`), so the fingerprint stopped matching and the finding resurfaced — turning `security:secrets` red on `main` for everyone and blocking any PR (the required `Secrets, deps, and workflow scan` check runs gitleaks). It also blocks #532.

SHA-keyed fingerprints are inherently fragile to history rewrites. A stopword allowlist keys on the secret *value*, so it survives rebases/merges.

## Change

- `.gitleaks.toml` — add a stopword allowlist for `orch_abc_SUB-1`, matching the existing fixture pattern (`wat-opaque-123`, `test-signing-secret-abc123`), with a comment explaining the SHA-fragility rationale.
- `.gitleaksignore` — remove the stale `f98f47a` fingerprint; leave a breadcrumb pointing to the new location.

## Verification

- `mise run security:secrets` → **`no leaks found`** (exit 0).
- Full gitleaks scan → **0 findings**; the three #313 aws-account-id fingerprints remain intact and suppressed.
- Over-suppression check: `orch_abc_SUB-1` is a 14-char fixed literal, present in the live tree only in the config references (the test file was deleted) — it cannot mask a broader/real secret.
- Scope: `.gitleaks.toml` + `.gitleaksignore` only.

## Note

Consider auditing the #313 aws-account-id fingerprints for the same SHA-fragility in a follow-up (out of scope here).

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -19,6 +19,15 @@ stopwords = ["wat-opaque-123"]
 description = "Test fixture signing secret in Slack verification unit test (not a real credential)."
 stopwords = ["test-signing-secret-abc123"]
 
+[[allowlists]]
+# Test idempotency-key fixture in orchestration-release.test.ts (not a real
+# credential). Suppressed by stopword rather than a .gitleaksignore commit-SHA
+# fingerprint: the finding lives in history and SHA fingerprints break whenever
+# history is rewritten (rebases / dep-bump merges), which is exactly how #530's
+# baseline regressed. A stopword is SHA-independent. See #537 (regression of #530).
+description = "Test idempotency-key fixture 'orch_abc_SUB-1' (not a real credential)."
+stopwords = ["orch_abc_SUB-1"]
+
 # Catch bare 12-digit AWS account IDs. The default ruleset does not flag these,
 # which is how a real account ID reached a committed comment in the #236 integ
 # work. RE2 (Go) has no lookarounds, so the non-digit neighbours are captured in
--- a/.gitleaksignore
+++ b/.gitleaksignore
@@ -6,9 +6,7 @@
 0dca2171f244f693ac7c649ecf376d1ce6af2b38:docs/backlog/CLI_COGNITO_NEW_PASSWORD_CHALLENGE.md:aws-account-id:5
 0dca2171f244f693ac7c649ecf376d1ce6af2b38:docs/diagrams/phase3-cedar-hitl.drawio:aws-account-id:28
 
-# False positive: generic-api-key on a test idempotency-key fixture
-# ('orch_abc_SUB-1', not a real credential) in a file that no longer exists at
-# HEAD — it was removed when the #76 bounded-parallel perf commit was reverted
-# (e8bee5e). The finding lives only in history at f98f47a, so it can't be
-# scrubbed without rewriting shared `main`; fingerprint it instead. See #530.
-f98f47abc6f7820d7e5ce0375cf22551cb664b37:cdk/test/handlers/shared/orchestration-release.test.ts:generic-api-key:125
+# NOTE: the orch_abc_SUB-1 test-fixture false positive (formerly baselined here
+# by commit-SHA fingerprint, #530) is now suppressed SHA-independently via a
+# stopword allowlist in .gitleaks.toml — a commit-SHA fingerprint here broke when
+# history was rewritten. See #537.


ayushtr-aws (approved):
Overview

  Clean, well-scoped fix (2 config files, +13/−6). #530 baselined the orch_abc_SUB-1 test-fixture false positive with a commit-SHA fingerprint in .gitleaksignore; a history rewrite changed the SHA and the
  fingerprint silently stopped matching, resurfacing the finding. This PR replaces it with a stopword allowlist in .gitleaks.toml, which keys on the secret's value and therefore survives rebases/merges. Right
  diagnosis, right fix.

  Correctness — verified against gitleaks source

  - Case handling works: gitleaks lowercases both stopwords and the candidate secret before matching (config/allowlist.go — Validate() builds the trie from lowercased values; ContainsStopWord() lowercases the
  input), so the mixed-case orch_abc_SUB-1 matches correctly.
  - Applies to history-resident findings: global allowlists filter findings in any commit, unlike SHA fingerprints — this is exactly why the approach is more robust here.
  - The breadcrumb comment in .gitleaksignore is safe: the parser skips blank and #-prefixed lines.
  - CI confirms: the required Secrets, deps, and workflow scan check passes at head, and the described full-history scan (mise run security:secrets) is the scan where this history-resident finding actually
  surfaces.

  Suggestions (non-blocking)
  
  1. Scope the allowlist with targetRules = ["generic-api-key"] (the rule from the removed fingerprint). As written, the block suppresses orch_abc_SUB-1 as a substring of any secret matched by any rule. Notably,
  this file's own first allowlist already uses targetRules = ["private-key"] — so scoping is the established convention here; the two older stopword blocks (wat-opaque-123, test-signing-secret-abc123) share the
  same gap and could be tightened in the same follow-up.
  2. Substring semantics: stopwords match by containment, so the guarantee is "no secret containing this string," not "this exact 14-char literal." With a value this distinctive the practical risk is negligible
  — noting it only because the PR's over-suppression check implies exact-match semantics.
  3. Please file the #313 follow-up: the three aws-account-id fingerprints remaining in .gitleaksignore are keyed to commit 0dca217 and exposed to the identical history-rewrite failure mode that caused this
  incident. Worth converting or documenting before it repeats.

  Verdict

  Approve. Value-keyed suppression is strictly more robust than SHA-keyed for findings that live only in history; the change follows the file's existing idiom, the inline rationale will help the next person who
  hits this class of regression, and the required check passes. Suggestions above are follow-ups, not blockers.

isadeks (approved):
## Approve ✅

Clean, minimal, well-reasoned fix. Swaps a fragile commit-SHA `.gitleaksignore` fingerprint for a value-keyed stopword allowlist — the SHA anchor broke on history rewrite (which is exactly what regressed #530 and turned the required `security:secrets` gate red on `main`), and a stopword survives rebases/merges.

### Verified empirically (gitleaks 8.30.0)
Reproduced the scenario in a scratch repo (fixture committed, then deleted, mirroring real history):
- **No config:** `leaks found: 1` → confirms the finding is real and the suppression is load-bearing, not dead config.
- **With this PR's `.gitleaks.toml`:** `no leaks found` (exit 0) → the stopword correctly suppresses it.
- **Deleted-file claim confirmed:** `orchestration-release.test.ts` is 404 on both `main` and this branch; the literal lives at HEAD only in the config references — matches the over-suppression analysis.

### Quality
- Mirrors the existing sibling allowlists (`wat-opaque-123`, `test-signing-secret-abc123`) exactly.
- Great rationale comments in both files with #530 → #537 breadcrumbs.
- Scope is tight: the three #313 aws-account-id fingerprints are untouched and still suppressed; auditing them for the same fragility is correctly deferred to a follow-up.

### One non-blocking note
gitleaks stopwords match as substrings, so this allowlist technically suppresses any secret *containing* `orch_abc_SUB-1`, not only the exact literal (verified: a longer secret embedding the substring is also silenced). Risk is negligible — it's the same property the two pre-existing sibling stopwords already carry, so no new risk class is introduced. Flagging for awareness only.

No correctness bugs, no removed-guard gaps, no security regressions. LGTM.

Issue #331, opened 2026-06-12
Fan-out orchestration mass-fails against the per-user concurrency cap (no queue/retry)


isadeks:
## Problem (enterprise scalability)

When a #247 orchestration releases more children at once than the per-user concurrency cap (`MAX_CONCURRENT_TASKS_PER_USER`, default 10) allows, the children that lose the admission race are **hard-failed**, not queued. Because a failed sub-issue transitively skips its dependents, a single wide fan-out can wreck an entire epic.

### Evidence (live)

Stress epic: 1 root → 11 independent leaves (SEO meta per page), cap = 10.
- Root succeeded → reconciler released all 11 leaves at once → 11 `createTaskCore` calls all returned 201.
- Each child task then ran `orchestrate-task` → `admissionControl` (`orchestrator.ts:90`, conditional `active_count < MAX`). The losers were `failTask('User concurrency limit reached')` (`orchestrate-task.ts:77`) → terminal **FAILED**.
- Result: **7 of 11 leaves FAILED** purely for being released simultaneously; the integration node (#16, depends on all leaves) was then **skipped** → the combined PR never opened.

These are not real failures — they are "released faster than the throttle allows" — yet they poison every dependent. The cap is meant to be a throttle but for orchestration children it behaves as a guillotine. The flagship enterprise use case (decompose a large feature into many parallel sub-issues) is exactly the shape that breaks.

## Root cause

The release path (`releaseReadyChildren` / `releaseChild` in `orchestration-release.ts`) releases every dependency-unblocked child unconditionally; admission control then rejects the overflow as terminal failures. Note: a "revert row → re-release" retry is NOT viable because `releaseChild`'s idempotency key (`${orch}_${sub}`) binds permanently to the first task created, so a replay returns the dead task.

## Proposed fix (throttle at release, never over-create)

Release at most `max(0, MAX_CONCURRENT - active_count)` children per reconcile pass; leave the rest `blocked`. No task is created that can't be admitted, so no spurious FAILED and the idempotency key is never bound to a doomed task. The drain vehicle already exists: every sibling completion re-fires the reconciler (re-evaluates releasable `blocked` children) and the #303 stranded-orchestration sweep is the periodic backstop — as slots free, the remaining children release.

## Acceptance criteria

- [ ] A fan-out with more ready children than free slots releases only up to the available budget; the rest stay `blocked` (not failed).
- [ ] As in-flight children complete, the remaining children are released until all run.
[...]

PR #544, opened 2026-07-08
feat(orchestration): admission queue with deferred pickup (#441)

### LINKED ISSUE #441: feat(orchestration): admission queue with deferred pickup
**Context:** ROADMAP.md → Admission queue with deferred pickup
**Related:** [#331](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/331) (closed bug—this is the feature successor)

---

## Component

API or orchestration

## Describe the feature

When admission is at capacity, **persist tasks in a durable queue** instead of failing them (see #331 live epic failure). Automatically re-attempt admission and continue processing in **FIFO** order with **optional priority lanes** as concurrency becomes available.

**Must preserve:** cancel semantics, idempotency semantics.

**Must expose:** queue position and ETA in task status / `bgagent status`.

## Use case

#331 demonstrated mass failure when fan-out releases more children than the per-user concurrency cap. Hard-failing subtasks wrecks entire epics (#247). Users expect queueing, not `429`-style task death.

## Proposed solution

1. New state or sub-status: `QUEUED` / `SUBMITTED` with `admission_pending: true`.
2. Durable queue (DynamoDB GSI or SQS) polled by admission worker.
3. `createTaskCore` returns 201 with queue position when capped—not terminal failure.
4. Cancel removes from queue; idempotency keys dedupe.
5. API + CLI expose `queue_position`, estimated wait.
6. Regression test reproducing #331 scenario.

## Other information

- Successor to closed #331—link in issue body when filing.
- Design context: `docs/design/ORCHESTRATOR.md`.

nizar-lahlali:
Closes #441. Successor to #331 (fan-out mass-fail against the concurrency cap).

## What

When admission hits the per-user concurrency cap, the task is now **queued, not failed**: it transitions `SUBMITTED → QUEUED` (a new pre-active state that holds **no** concurrency slot) and a scheduled **AdmissionQueuePickup** Lambda drains the queue in **FIFO order** (by `created_at`) as slots free up — flipping `QUEUED → SUBMITTED` and re-invoking the orchestrator.

## Design highlights

- **Single counter writer.** The pickup Lambda does a *read-only* capacity pre-check; the orchestrator's atomic `admissionControl` conditional-increment remains the only writer of the concurrency counter. A pickup that loses the admission race harmlessly re-queues (`created_at` never changes, so FIFO position is preserved). This avoids introducing a second writer — the class of drift bug behind #331.
- **Cancel semantics preserved.** Every queue transition is conditional on the current status, so a user cancel cleanly wins any race. `QUEUED` is cancellable via REST (already status-generic) and Slack (list extended). No compute or slot to release.
- **Idempotency preserved.** Replay with the same `Idempotency-Key` returns the existing `QUEUED` task (200) — dedupe path is status-agnostic and untouched.
- **Backstops.** A task queued > `QUEUE_MAX_AGE_SECONDS` (default 24 h) is failed with a clear message. The stranded-task reconciler now ages tasks by time-in-current-status so a long queue wait isn't mistaken for a stranded `SUBMITTED` task right after pickup.
- **Orchestrator re-invoke carries a per-pickup nonce** so no dedup layer mistakes it for a replay of the create-task invocation.
- **Channel feedback**: Linear/Jira comment now says "queued — will start automatically", and only on first queue entry (no per-cycle spam).

## API / CLI surface

- `TaskRecord`: `queued_at`, `admission_attempts` (internal bookkeeping).
- `GET /tasks/{id}`: `queued_at`, `queue_position` (1-based FIFO rank among the caller's queued tasks, computed at read time, fail-open), `estimated_wait_s` (coarse heuristic).
- `bgagent status` / detail view render `Queue: position N (est. wait ~Xm)` while `QUEUED`.
- CDK↔CLI types mirrored; `check-types-sync` passes.

## #331 regression test

`reconcile-admission-queue.test.ts` includes a fan-out-burst regression: 5 children queued above a cap of 3 all **survive** as `QUEUED`, drain FIFO (3 picked up, 2 remain), and **zero** tasks fail.

## Testing

- `mise run build` passes end-to-end (agent quality, CDK compile + synth + 2265 tests across 125 suites, CLI 576 tests, docs build + sync, type/constants drift checks).
- New coverage: state-machine invariants (QUEUED pre-active, disjointness, transitions), `queueTask` (first-entry vs re-queue, race, no counter writes), pickup handler (FIFO, capacity bounds, per-user isolation, races, invoke-failure, max-age backstop, pagination), get-task queue position (rank, fail-open, left-queue race), CLI rendering.

## Notes for reviewers

- The ETA is a deliberately coarse UX hint (`ceil(position / cap) × avg-task-duration`, default 600 s, env-tunable) — not a promise.
- `AdmissionQueuePickup` runs every minute; an empty-queue cycle is a single GSI query.
- Marked as potentially breaking in the issue: clients that treated the admission-cap `FAILED` + `admission_rejected` event as terminal will now see `QUEUED` + `admission_queued` instead.


cdk/src/handlers/get-task.ts
```diff
@@ -29,6 +30,81 @@ import { type TaskRecord, toTaskDetail } from './shared/types';
...
+          ':uid': record.user_id,
+          ':queued': `${TaskStatus.QUEUED}#`,
+        },
+        ProjectionExpression: 'task_id, created_at',
+        ExclusiveStartKey: lastKey as Record<string, never> | undefined,
+      }));
+      for (const item of resp.Items ?? []) {
+        if (item.task_id === record.task_id) {
+          found = true;
+        } else if (typeof item.created_at === 'string' && item.created_at < record.created_at) {
```
scottschreckengaust: **Same-millisecond tie mis-reports position.** `created_at` is `new Date().toISOString()` (ms granularity), so two tasks created in the same millisecond produce identical strings; with strict `<`, neither counts the other as ahead and **both report the same `queue_position`** — exactly the fan-out-burst case (#331) this feature targets. The ULID `task_id` is monotonic with creation, so use it as a deterministic tiebreak:

```suggestion
        } else if (
          typeof item.created_at === 'string'
          && (item.created_at < record.created_at
            || (item.created_at === record.created_at
              && typeof item.task_id === 'string'
              && item.task_id < record.task_id))
        ) {
```

Also add a get-task test with two rows sharing `record.created_at` to pin this — the current position test uses four distinct timestamps and never a tie.

scottschreckengaust (commented):
## Review — feat(orchestration): admission queue with deferred pickup (#441)

**Verdict: COMMENT (non-gating).** This is a well-architected, well-bounded, thoroughly documented feature that faithfully implements the approved acceptance criteria of #441. No merge blockers. I am not approving because two real quality issues are worth addressing — a genuine (minor, non-safety) correctness bug in queue-position reporting under the very fan-out-burst scenario this feature targets, and a coverage gap on the load-bearing re-queue integration path — but neither gates merge. Take them as strong suggestions.

### Governance (ADR-003) — PASS
- #441 exists, is OPEN, carries **`approved`** + `P0` + `orchestration`. Implementation matches its acceptance criteria (durable QUEUED state, FIFO drain by `created_at`, cancel/idempotency preserved, `queue_position`/ETA surfaced, #331 regression test). Branch `feat/441-admission-queue-deferred-pickup` follows the pattern. `mergeable=MERGEABLE` (the earlier UNKNOWN resolved; main already merged in — no rebase needed). CI green except `integ-smoke` (pending admin approval).

### Vision alignment (VISION.md / ORCHESTRATOR.md) — PASS
- **Fire-and-forget default preserved.** Deferred pickup keeps unattended async execution: the scheduled drain re-invokes the orchestrator; no user action needed while queued (the Linear/Jira comment even says "No action needed").
- **Bounded blast radius & cost — the key tenet, and it holds.** The queue is DynamoDB-backed (not an unbounded SQS pile), drained by a 1-minute scheduled Lambda whose empty cycle is a single cheap GSI query. A **24h max-age backstop** (`QUEUE_MAX_AGE_SECONDS`) fails over-age tasks so the queue can never accumulate unbounded zombies. QUEUED holds **no** concurrency slot and **no** compute — correctly excluded from `ACTIVE_STATUSES` (the comment in `task-status.ts` calls out that counting it would deadlock the queue). No undocumented tenet trade.
- **Reviewable outcomes preserved:** queued tasks still flow to the same PR pipeline once admitted.

### Blocking — NONE

No unbounded queue, no missing bootstrap coverage, no task double-run, no admission race that loses correctness, no undocumented tenet trade.

### Non-gating findings (please address)

1. **`queue_position` mis-reports on same-millisecond ties** — `cdk/src/handlers/get-task.ts:79`. `created_at` is `new Date().toISOString()` (ms granularity), **not** the ULID `task_id`. `computeQueueInfo` ranks with strict `item.created_at < record.created_at`, so two tasks created in the same millisecond each see the other as *not* ahead → **both report the same position**. This is precisely the #331 fan-out burst (children minted in a tight loop) this feature exists to serve, so `bgagent status` could show two queued tasks both claiming "position 3". Risk: user-visible UX-only inconsistency (FIFO *drain* is unaffected — the pickup Lambda uses the GSI's own sort order). Fix: break ties deterministically on the ULID `task_id`, which is monotonic with creation (inline suggestion below). Add a get-task test with two identical `created_at` rows pinning the behavior — the current position test (`get-task.test.ts:164-167`) uses four distinct timestamps and never a tie.

2. **Re-queue integration path is untested end-to-end** — `cdk/src/handlers/orchestrate-task.ts:91-126`. The load-bearing FIFO-preservation logic (on a failed `admissionControl`, pass the **re-read `current`** — not the stale step-1 `task` — to `queueTask`, and gate channel feedback on `current.queued_at === undefined` so a pickup-cycle re-queue doesn't reset `queued_at`/`admission_attempts` or spam the issue) has no handler-level test. `queueTask`'s re-queue behavior is covered in isolation (`orchestrate-task.test.ts:1020`), but a regression to `queueTask(task)` (stale snapshot) would silently reset FIFO bookkeeping and every existing test would still pass. Add a handler test asserting the re-read record is passed and notify-on-cap fires only on first entry.

3. **`pickUpTask` invoke-failure is miscounted in the summary** — `cdk/src/handlers/reconcile-admission-queue.ts:224` returns `false` on a failed orchestrator re-invoke (task now SUBMITTED-with-no-pipeline), which the caller counts as `skipped_race` (line 367), not `errors`. A systemic invoke outage (bad ARN / throttling / perms) that strands every picked-up task would log at INFO with `errors: 0` — indistinguishable from benign cancel races, and no alarm-worthy `error_id` like the sibling `reconcile-stranded-tasks.ts:400-412` pattern. Recovery itself is sound (the task IS reachable by the stranded reconciler — verified: `created_at` is never rewritten so it stays under the GSI cutoff, and the new age-by-`status_created_at` logic protects it until it's genuinely stranded). This is purely an observability gap. Fix: increment `summary.errors` (or a dedicated `invoke_failed` counter) on that branch and fold it into the escalation at line 381.

### Nits
- **`MAX_CONCURRENT` default drift (cosmetic).** `orchestrator.ts:41` defaults `?? '3'`; the new files (`get-task.ts:34`, `reconcile-admission-queue.ts:69`, `task-api.ts:549`, `admission-queue-pickup.ts:119`) and the orchestrator construct (`task-orchestrator.ts:200`) default to `10`. In the deployed stack `MAX_CONCURRENT_TASKS_PER_USER` is always set (construct default 10), so all planes agree at 10 and the `?? '3'` never fires — but the lingering `3` reads as a real disagreement. `confirm-uploads.ts:48` also still uses `?? '3'`. Worth aligning the fallbacks to one value in a follow-up. (Not in this PR's diff, so no inline.)
- `expireQueuedTask`'s cancel-race branch (`reconcile-admission-queue.ts:266`) and the `> QUEUE_MAX_AGE_SECONDS` boundary lack a direct test (the pickup path has the equivalent ConditionalCheckFailed test). Cheap to add.

### Documentation — EXCELLENT
`docs/design/ORCHESTRATOR.md` is comprehensively updated: new QUEUED state row, state diagram edges, transition table (incl. the 24h backstop), cancel-semantics row, and a rewritten admission-control step 2 explaining the single-writer/re-queue design. The Starlight mirror `docs/src/content/docs/architecture/Orchestrator.md` is regenerated in the same PR (no stale-mirror CI risk). CLI `types.ts` is synced (`QUEUED` + the three new `TaskDetail` fields) and `format.ts` renders the queue line. No ADR needed — this is an additive lifecycle state within the existing durable-execution model, not a new architectural decision.

### Tests & CI
- **Bootstrap synth-coverage (ADR-002): N/A — correctly not updated.** The only new CFN resources are `AWS::Events::Rule` (`events:PutRule`) and `AWS::Lambda::Function` (`lambda:CreateFunction`) — both already in `cdk/src/bootstrap/resource-action-map.ts` (same Rule+Lambda pattern as ConcurrencyReconciler / StrandedTaskReconciler / PendingUploadCleanup). No new resource type ⇒ no bootstrap-policy / `BOOTSTRAP_VERSION` bump required. Verified by grepping the diff for new construct instantiations.
- **#366 test-performance: PASS.** `admission-queue-pickup.test.ts` caches `Template.fromStack()` once (module-level cache, bare `new App()` picks up the global bundling-disable), no per-test synth, no bundling re-enable. `task-status.test.ts` is a pure constant unit test (no synth).
- Strong behavior-focused coverage otherwise: the named #331 fan-out-burst regression, FIFO ordering, per-user capacity math, pagination, `queued_at`-set-once, `admission_attempts` increment, conditional-flip-loses-to-cancel (pickup path), state-machine `QUEUED ∉ ACTIVE_STATUSES` invariant, CLI rendering + fail-open null. Gaps are findings #1/#2/#3 above.

### Note on a claim I investigated and discharged
An automated pass flagged a "concurrency counter double-increment on durable replay" in the admission step as HIGH. I verified against the diff and durable-execution semantics: `admissionControl` was **already** inside `context.step('admission-control')` before this PR (the diff only swaps `failTask`→`queueTask` inside the pre-existing step). At-least-once step re-execution is a general durable-execution property — documented in ORCHESTRATOR.md — and is explicitly backstopped by the existing `ConcurrencyReconciler` (reconciles `active_count` drift against actual active tasks). It is neither introduced nor worsened by #544, so it is out of scope here.

### Review agents run
- **`pr-review-toolkit:silent-failure-hunter`** (invoked) — surfaced the invoke-failure miscount (#3) and the durable-replay concern (investigated + discharged as pre-existing).
- **`pr-review-toolkit:pr-test-analyzer`** (invoked) — surfaced the tie gap (#1) and the re-queue integration gap (#2).
- **`code-reviewer` / `type-design-analyzer`** — hand-reviewed. Types: `TaskStatus`/`VALID_TRANSITIONS`/`PRE_ACTIVE_STATUSES` are clean; `TaskDetail` additions are well-modeled (`queue_position`/`estimated_wait_s` nullable, never persisted). Bootstrap (ADR-002) verified by hand.

### Human heuristics
- **Proportionality:** appropriate. A durable DynamoDB state + a cheap scheduled drain is the right weight for #441 — not over-built (no SQS/Step Functions), reuses the existing single-writer counter and reconciler backstop.
- **Coherence:** matches the established reconciler/durable-step patterns already in the codebase.
- **Clarity (AI004):** the `get-task.ts:104` fail-open `nosemgrep` is correctly justified (queue position is best-effort UX; must not 500 the whole GET). No unjustified silent masking found.
- **Appropriateness (AI005):** tests assert does-do (condition expressions, flipped IDs in FIFO order, counters, decoded payloads), not should-do.

isadeks (commented):
## Review — admission queue with deferred pickup (#441)

Reviewed by reading the code at head (`fb9c2320`) and verifying the load-bearing claims against the actual GSI/schema. This is a well-architected, well-documented feature — the single-writer discipline (pickup does a read-only capacity pre-check; the orchestrator's atomic `admissionControl` stays the only counter writer; a lost race harmlessly re-queues) is exactly the right way to avoid the #331 drift class. @scottschreckengaust already did a thorough non-gating review; this confirms his findings against current code and flags which are now addressed.

### Overview
`SUBMITTED → QUEUED` (a pre-active state holding no slot) when the per-user cap is hit, instead of `FAILED`. A 1-min scheduled Lambda drains FIFO as slots free up (`QUEUED → SUBMITTED` + orchestrator re-invoke with a per-pickup nonce), with a 24h max-age backstop, cancel-race safety via conditional writes, and read-time `queue_position`/ETA on `GET /tasks/{id}`.

### Verified correct (the claims that matter)
- **FIFO preservation holds at the schema level.** `StatusIndex` is `PK: status, SK: created_at` (`task-table.ts:120-122`), so the pickup's ascending scan is genuinely oldest-first, and since `created_at` never changes on re-queue (only `status_created_at` moves), FIFO position survives a lost admission race. The central design promise is real, not aspirational.
- **QUEUED ∉ `ACTIVE_STATUSES`** — correctly placed in `PRE_ACTIVE_STATUSES`; the deadlock the comment warns about (queued tasks holding the slots they wait for) is avoided, and the concurrency reconciler won't count them.
- **Counter safety** — construct grants the pickup `grantReadData` (not write) on the concurrency table; a test pins "no write actions on UserConcurrencyTable." Single-writer invariant enforced by IAM, not just convention.
- **get-task ranking is sound** — `UserStatusIndex` is `ProjectionType.ALL`, so `created_at` is available to rank by; querying `begins_with(status_created_at, 'QUEUED#')` and ranking on the separate `created_at` is correct.
- **Cancel/idempotency** — every transition is status-conditional, so a user cancel cleanly wins any race; QUEUED added to Slack's cancellable set + REST is status-generic. `queued_at` set-once (re-queue preserves it), feedback gated on first entry (no per-cycle spam).

### On @scottschreckengaust's three findings — status at this head
- **#1 (same-millisecond tie in `queue_position`) — code FIXED, test still MISSING.** `computeQueueInfo` now breaks ties on `task_id` (`get-task.ts`: `created_at === record.created_at && item.task_id < record.task_id`) — correct, since the ULID is monotonic with creation. But the position test still uses four *distinct* timestamps; there's **no test with two rows sharing a `created_at`**, which is exactly the fan-out-burst tie this fixes. A regression to the tiebreak would pass CI. Please add the tie test to pin it.
- **#3 (invoke-failure miscounted) — NOT addressed.** `pickUpTask` still `return false` on a failed orchestrator re-invoke (`reconcile-admission-queue.ts:~224`), and the caller counts `false` as `summary.skipped_race` (line ~368), not `errors`. So a systemic invoke outage (bad ARN / throttle / perms) that strands every picked-up task logs at INFO with `errors: 0` — indistinguishable from benign cancel races, and no alarmable `error_id`. Recovery is still sound (the stranded reconciler sweeps these — verified, `created_at` unchanged keeps them under the GSI cutoff and the new age-by-`status_created_at` protects them until genuinely stranded), so this is purely observability. Worth incrementing `errors` (or a dedicated `invoke_failed`) on that branch so a real outage is visible.
- **#2 (re-queue integration path untested)** — still applies: `queueTask`'s re-queue is unit-tested in isolation, but the handler-level wiring (pass the re-read `current`, gate feedback on `current.queued_at === undefined`) has no test. A regression to `queueTask(task)` (stale step-1 snapshot) would silently reset FIFO bookkeeping with every existing test still green. A handler test asserting the re-read record is passed would close it.

### Additional notes
- **`MAX_CONCURRENT` default drift** (Scott's nit, worth repeating): `orchestrator.ts` + `confirm-uploads.ts` default `?? '3'`; the five new/adjacent call sites default `?? '10'`. Harmless in the deployed stack (the env var is always set to 10), but the get-task ETA and the pickup capacity math read `?? '10'` while the orchestrator reads `?? '3'` — if the env var ever went unset, the pickup would over-pick against the orchestrator's stricter local default and every over-pick would re-queue. Aligning the fallbacks (follow-up) removes a latent footgun.
- Docs (`ORCHESTRATOR.md` + Starlight mirror) are genuinely excellent — new state, diagram edges, transition table, cancel-semantics row, rewritten admission step. CDK↔CLI types synced.

### Verdict
Approve-ready in substance — no correctness or safety blocker (FIFO, single-writer, cancel, deadlock-avoidance all verified). The two things I'd land before merge are the **#1 tie test** (cheap, pins a fix that's currently unguarded) and **#3's error-count** (one-line observability fix so a real invoke outage pages). #2 is a strong suggestion. CI green (`build (agentcore)` runs the full suite).

cdk/src/handlers/reconcile-admission-queue.ts
```diff
@@ -0,0 +1,384 @@
...
+
+    // Oldest-first pickup, bounded by free capacity. The orchestrator's
+    // atomic admissionControl is still the final arbiter — an optimistic
+    // over-pick simply re-queues.
+    for (const task of live.slice(0, available)) {
+      try {
+        if (await pickUpTask(task)) {
+          summary.picked_up++;
+        } else {
+          summary.skipped_race++;
```
scottschreckengaust: Non-blocking (observability): a failed orchestrator re-invoke returns `false` from `pickUpTask` and lands here as `skipped_race`, not `errors`. A systemic invoke outage (bad ARN / throttle / perms) that strands every picked-up task would then log at INFO with `errors: 0` and never trip the `level='error'` escalation at line 381 — indistinguishable from benign cancel races. Recovery is sound (the stranded-task reconciler sweeps these), so this is purely a visibility gap. Suggest incrementing `summary.errors` (or a dedicated `invoke_failed` counter) on the invoke-failure branch in `pickUpTask` so a real outage pages.

cdk/test/handlers/get-task.test.ts
```diff
@@ -133,6 +135,79 @@ describe('get-task handler', () => {
...
+          queued_at: '2025-03-15T10:30:05Z',
+        },
+      })
+      // UserStatusIndex query: two tasks queued before ours, one after.
+      .mockResolvedValueOnce({
+        Items: [
+          { task_id: 'older-1', created_at: '2025-03-15T10:28:00Z' },
+          { task_id: 'older-2', created_at: '2025-03-15T10:29:00Z' },
+          { task_id: 'task-1', created_at: '2025-03-15T10:30:00Z' },
+          { task_id: 'newer-1', created_at: '2025-03-15T10:31:00Z' },
```
scottschreckengaust, line 167: Non-blocking (test gap): this position test uses four distinct `created_at` values, so it never exercises the same-millisecond tie that `computeQueueInfo`'s `item.task_id < record.task_id` tiebreak exists to handle — precisely the #331 fan-out-burst case. A regression that dropped the tiebreak would still pass CI while two queued tasks both reported the same position. Add a case with two rows sharing `record.created_at` (distinct `task_id`s) to pin the FIFO tiebreak.

scottschreckengaust (commented):
## Review — feat(orchestration): admission queue with deferred pickup (#441)

**Verdict: COMMENT (non-gating).** A well-architected, well-bounded, thoroughly documented feature that faithfully implements the approved acceptance criteria of #441. No merge blockers. I independently verified the two prior member reviews (@scottschreckengaust, @isadeks) against the current head (`fb9c2320`): both are correct as stated, both non-gating. I am not approving only because two quality items are worth landing first — a one-line observability fix and a missing regression test — neither of which gates merge.

### 1. Vision alignment (VISION.md / ORCHESTRATOR.md) — PASS
- **Fire-and-forget default preserved.** Deferred pickup keeps unattended async execution: the scheduled drain re-invokes the orchestrator with no user action while queued (the Linear/Jira comment now reads "queued — will start automatically. No action needed").
- **Bounded blast radius & cost — the key tenet, and it holds.** DynamoDB-backed queue (not an unbounded SQS pile), drained by a 1-minute scheduled Lambda whose empty cycle is a single cheap GSI query. A 24h `QUEUE_MAX_AGE_SECONDS` backstop fails over-age tasks so the queue can never accumulate unbounded zombies. QUEUED holds **no** concurrency slot and **no** compute — correctly excluded from `ACTIVE_STATUSES` (`task-status.ts`), with a test pinning that counting it would deadlock the queue. No undocumented tenet trade; this is an additive lifecycle state within the existing durable-execution model, so no ADR is required.
- **Reviewable outcomes preserved:** queued tasks flow to the same PR pipeline once admitted.

### 2. Governance (ADR-003) — PASS
- #441 is OPEN and carries **`approved`** + `P0` + `orchestration` + `enhancement`; assigned to the author. Implementation matches its acceptance criteria: durable QUEUED state, FIFO drain by `created_at`, cancel/idempotency preserved, `queue_position`/ETA surfaced, and the named #331 fan-out-burst regression test. Branch `feat/441-admission-queue-deferred-pickup` follows the convention. CI is green (`build (agentcore)`, CodeQL, secrets/deps all pass).

### 3. Blocking issues — NONE
No unbounded queue, no missing bootstrap coverage, no task double-run, no admission race that loses correctness, no undocumented tenet trade, no cross-tenant data leak in the new read path.

### 4. Non-blocking findings (please address before merge)

**(a) `pickUpTask` invoke-failure is miscounted as a benign race** — `cdk/src/handlers/reconcile-admission-queue.ts:368`. On a failed orchestrator re-invoke, `pickUpTask` returns `false` (the task is now SUBMITTED-with-no-pipeline), which the caller counts as `skipped_race`, not `errors`. A *systemic* invoke outage (bad ARN / throttling / missing perms) that strands every picked-up task would log at INFO with `errors: 0` — indistinguishable from benign cancel races, and it never trips the `level='error'` escalation at line 381 or emits an alarmable `error_id` like the sibling `reconcile-stranded-tasks.ts`. Recovery itself is sound — I verified the stranded-task reconciler sweeps these (`created_at` is never rewritten so the record stays under the GSI cutoff, and the new age-by-`status_created_at` logic protects it until it is *genuinely* stranded). So this is purely an observability gap. Fix: increment `summary.errors` (or a dedicated `invoke_failed` counter) on the invoke-failure branch and fold it into the escalation.

**(b) Same-millisecond tie in `queue_position` is fixed in code but untested** — `cdk/test/handlers/get-task.test.ts:164`. `computeQueueInfo` correctly breaks `created_at` ties on the monotonic ULID `task_id` (`get-task.ts`), which is the right fix for the fan-out-burst case (#331) this feature exists to serve. But the position test uses four *distinct* timestamps and never a tie, so a regression that dropped the `item.task_id < record.task_id` tiebreak would pass CI while two same-millisecond queued tasks both report the same position in `bgagent status`. Add a get-task test with two rows sharing `record.created_at` to pin the tiebreak.

### 5. Non-blocking suggestions / nits
- **Re-queue integration path untested end-to-end** — `cdk/src/handlers/orchestrate-task.ts`. `queueTask`'s re-queue behavior is unit-tested in isolation, but the handler-level wiring (pass the **re-read `current`**, not the stale step-1 `task`, and gate channel feedback on `current.queued_at === undefined`) has no handler test. A regression to `queueTask(task)` (stale snapshot) would silently reset `queued_at`/`admission_attempts` and re-spam the issue, with every existing test still green. A handler test asserting the re-read record is passed would close it. (Strong suggestion.)
- **`MAX_CONCURRENT` default drift (cosmetic).** `orchestrator.ts:41` and `confirm-uploads.ts:48` default `?? '3'`; the new/adjacent sites (`get-task.ts:34`, `reconcile-admission-queue.ts:69`, `task-api.ts`, `admission-queue-pickup.ts`) default `?? '10'`. Harmless in the deployed stack (the construct always sets the env var to 10, so the fallback never fires), but if the env var ever went unset the pickup would over-pick against the orchestrator's stricter `3` and every over-pick would re-queue. Align the fallbacks in a follow-up. (Out of this PR's diff — no inline.)
- `expireQueuedTask`'s cancel-race branch (ConditionalCheckFailed) and the `> QUEUE_MAX_AGE_SECONDS` boundary lack a direct test, unlike the pickup path's equivalent. Cheap to add.

### 6. Documentation — EXCELLENT
`docs/design/ORCHESTRATOR.md` is comprehensively updated: new QUEUED state row, state-diagram edges, transition table (incl. the 24h backstop), cancel-semantics row, and a rewritten admission-control step 2 explaining the single-writer / re-queue design and the concurrency-vs-rate-limit queueing distinction. The Starlight mirror `docs/src/content/docs/architecture/Orchestrator.md` is regenerated in the **same PR** (no stale-mirror CI risk). CLI `types.ts` is synced (`QUEUED` + the three new `TaskDetail` fields — verified present) and `format.ts` renders the queue line. No ADR needed.

### 7. Tests & CI
- **Bootstrap synth-coverage (ADR-002): N/A — correctly not updated.** The only new CFN resources are `AWS::Events::Rule` and `AWS::Lambda::Function`, both already in `cdk/src/bootstrap/resource-action-map.ts` (same Rule+Lambda pattern as ConcurrencyReconciler / StrandedTaskReconciler). No new resource type ⇒ no bootstrap-policy or `BOOTSTRAP_VERSION` bump required. Verified against the diff.
- **#366 test-performance: PASS.** `admission-queue-pickup.test.ts` caches `Template.fromStack()` at module level and does not re-enable Lambda bundling; `task-status.test.ts` is a pure constant unit test (no synth).
- Strong behavior-focused coverage: the named #331 fan-out-burst regression, FIFO ordering, per-user capacity math, pagination, `queued_at`-set-once, `admission_attempts` increment, conditional-flip-loses-to-cancel, `QUEUED ∉ ACTIVE_STATUSES` invariant, CLI rendering + fail-open null. Gaps are findings (a)/(b) and the re-queue-handler suggestion above. CI green.

### 8. Review agents run
- **`/security-review`** (invoked) — new IAM (`lambda:InvokeFunction` scoped to the orchestrator alias ARN only; concurrency table `grantReadData` only, IAM-enforced single-writer), the new `GET /tasks/{id}` queue-position read path, and the scheduled handler's input handling. **No HIGH/MEDIUM findings**: DDB access is parameterized and tenant-scoped (`user_id` bound to the already-authorized record), IAM is least-privilege, no new untrusted-input sink.
- **`code-reviewer`** (invoked) — capacity math, per-user grouping, FIFO, backstop, `queueTask` re-queue bookkeeping, and the age-by-`status_created_at` change in `reconcile-stranded-tasks.ts` all verified correct; surfaced finding (a).
- **`silent-failure-hunter`** (invoked) — the `get-task` fail-open `nosemgrep` is correctly justified (best-effort UX hint, must not 500 the GET); the per-user and per-task `try/catch` continues are sound. Surfaced the invoke-failure miscount (a). Investigated and **discharged** an automated "double-increment on durable replay" claim: `admissionControl` was already inside `context.step` pre-PR; the diff only swaps `failTask`→`queueTask`, and at-least-once step semantics are backstopped by the existing ConcurrencyReconciler — neither introduced nor worsened here.
- **`type-design-analyzer`** (invoked) — `TaskStatus`/`VALID_TRANSITIONS`/`PRE_ACTIVE_STATUSES` are clean; `TaskDetail` additions well-modeled (`queue_position`/`estimated_wait_s` nullable, never persisted); CDK↔CLI shapes mirrored.
- **`comment-analyzer`** (invoked) — the extensive doc-comments (single-writer discipline, deadlock rationale, FIFO-by-`created_at`, nonce purpose) match the code; no drift found.
- **`pr-test-analyzer`** (invoked) — surfaced the missing tie test (b) and the re-queue integration gap.
- Omitted: none in scope. (No agent-runtime/Python surface, so `agent/` guides N/A.)

### 9. Human heuristics
- **Proportionality — PASS.** Durable DynamoDB state + a cheap scheduled drain is the right weight for #441; not over-built (no SQS / Step Functions), reuses the existing single-writer counter and reconciler backstop.
- **Coherence — PASS.** Matches the established reconciler / durable-step patterns; same terms (`status_created_at`, `admissionControl`, StatusIndex) across the repo.
- **Clarity — PASS.** Names communicate intent; the fail-open `nosemgrep` (`get-task.ts`) is justified. One clarity ding: the invoke-failure counter (finding a) hides a real failure behind a benign label.
- **Appropriateness — PASS.** Tests assert what the code *should* do (condition expressions, flipped IDs in FIFO order, decoded payloads, counters), not merely what it does; integration verified against the real GSI schema (`StatusIndex` PK=status/SK=created_at, `UserStatusIndex` ProjectionType.ALL).

<sub>Reviewed as Principal AWS Solutions Architect via the ABCA /review_pr workflow.</sub>

cdk/src/handlers/reconcile-admission-queue.ts
nizar-lahlali: Good catch — fixed in 8198d20f. `pickUpTask` now returns a discriminated `PickupOutcome` (`'picked_up' | 'race' | 'invoke_failed'`) instead of a bare boolean, so an invoke failure after the flip is counted in a dedicated `invoke_failed` counter rather than collapsing into `skipped_race`. The final escalation now trips on `errors + invoke_failed` with `picked_up === 0` and a non-empty queue, so a systemic invoke outage that strands every pickup logs at `level='error'` and pages. Added a regression test asserting the ERROR escalation. Recovery path is unchanged — the stranded-task reconciler still sweeps the SUBMITTED rows.

cdk/test/handlers/get-task.test.ts
nizar-lahlali: Added in 1192b6b9. New test 'same-millisecond tie breaks on task_id...' feeds three UserStatusIndex rows sharing the target's `created_at` (`task-0`, `task-1`, `task-2`): `task-0` < `task-1` ranks ahead, `task-2` > `task-1` ranks behind, so the position is 2. Without the `item.task_id < record.task_id` tiebreak, none of the same-instant rows would satisfy `item.created_at < record.created_at`, `ahead` would be 0, and the position would be 1 — so the assertion fails if the tiebreak is ever dropped.

ayushtr-aws (dismissed):
Re-review — PR #544, commit 6bb610fe

  New head 6bb610fe — "address PR #544 review — UA attribution, invoke-failure re-queue, stranded-task coverage". Reviewed the delta against origin/main, re-ran silent-failure-hunter /
  pr-test-analyzer / code-reviewer on the new logic, and hand-verified the contested race against the orchestrator's atomic gates.

  Verdict

  Approve with nits. The one hard blocker from my prior review — the #319 UA-attribution violation (B1) — is fully fixed and verified. The commit also cleanly closes the prior thread #3
  and adds the stranded-task coverage. Two items I'd still land before merge (both cheap, neither a hard gate): a genuine user-facing correctness bug (N2, false "queued" notification) and
  the new re-queue's phantom-success trade-off (F1). The prior thread #2 (handler-level test) also remains open.

  What this commit fixed — verified ✅

  B1 (was BLOCKING) — UA attribution, complete and correct. reconcile-admission-queue.ts:62-63 now uses makeClient(DynamoDBClient) / makeClient(LambdaClient) — makeClient is the right
  factory (the handler uses low-level @aws-sdk/client-dynamodb commands, not the doc client). ABCA_COMPONENT: 'orchestr' added to the pickup Lambda env (admission-queue-pickup.ts:115),
  matching peers exactly (stranded-task-reconciler.ts:127, concurrency-reconciler.ts:82, task-orchestrator.ts:271). The label is consumed inside ua.ts, so construct-env + makeClient is the
  complete #319 contract — nothing else needed. Construct test added.

  Stranded-task coverage (N10, prior gap) — closed. Three new reconcile-stranded-tasks.test.ts cases pin the age-by-status_created_at logic: fresh-status-skipped, old-status-failed,
  malformed-status_created_at falls back to created_at. Good — covers branch, negative, and parse-failure edge.

  Prior thread #3 (invoke-failure escalation) stays resolved, and the commit went further: on invoke failure it now re-queues the task instead of leaving it stranded (new
  requeueAfterInvokeFailure), tested for both the restore and the restore-loses-race fallback. FIFO is preserved — I verified created_at is untouched and the pickup StatusIndex sorts on
  created_at, so the fresh status_created_at doesn't perturb pickup order.

  Findings on the new re-queue logic

  F1 (medium, NEW — the one the agents split on). The invoke is InvocationType: 'Event' (at-least-once). If lambdaClient.send() throws after the event was already accepted server-side (a
  response-path network error / retry-exhaustion), the orchestrator is actually running, yet requeueAfterInvokeFailure flips the row SUBMITTED→QUEUED and the next cycle re-picks and
  re-invokes it. I adjudicated this against the code:

  - Two orchestrators then both clear admissionControl (atomic conditional increment) → active_count over-counts by +1 for one logical task.
  - But the worst outcome — a double agent-run / duplicate PR — cannot happen: hydrateAndTransition gates on transitionTask(SUBMITTED→HYDRATING) with ConditionExpression: '#status =
  :fromStatus' (orchestrator.ts — verified atomic). Exactly one orchestrator proceeds; the loser throws at the gate. So the residual harm is the counter drift (self-healed by the existing
  ConcurrencyReconciler) plus a possible orphaned/FAILED-vs-running flap.

  So the silent-failure-hunter is right that this is a new path the old leave-SUBMITTED behavior didn't have (old code was safe in exactly this case), but the code-reviewer is right that
  the catastrophic outcome is prevented by the atomic gates and the window is narrow. My call: non-blocking, worth addressing. Either gate the re-queue on errors that prove non-delivery
  (throttle/4xx/connection-refused), or keep the re-queue but document that the phantom-success case is backstopped by the atomic HYDRATING gate + ConcurrencyReconciler rather than by "the
  stranded reconciler" (see F3). The trade-off — prompt retry for transient throttles vs. this rare flap — is defensible; it just isn't stated.

  F2 (minor, observability). A non-conditional failure of the restore UpdateItem (e.g. DDB throttle during the same outage that broke the invoke) is logger.warn-only and not counted in
  summary.errors (reconcile-admission-queue.ts:216-224). During combined invoke+DDB pressure, re-queue coverage silently degrades to the 24h expiry with no metric signal. Fold restore
  failures into a counter.

  F3 (nit, comment accuracy). The doc comment "the stranded-task reconciler remains the final backstop" is only true when the restore fails. On a permanent outage the restore succeeds
  every cycle, so the task bounces QUEUED→SUBMITTED→QUEUED at the ~1-min cadence and never sits in SUBMITTED long enough (1200s) for the stranded reconciler to fire — the real backstop
  becomes the 24h QUEUE_MAX_AGE_SECONDS expiry. Time-to-terminal for an unrecoverable task moves from ~20 min to ~24 h. Fine as a trade-off, but the comment overstates the reconciler's
  role.

  Still-open items from prior reviews

  N2 (medium — real user-facing bug, unaddressed). orchestrate-task.ts:107 still discards queueTask's boolean return. queueTask returns false (without throwing) when the SUBMITTED→QUEUED
  write loses to a concurrent cancel. The notify block is gated only on current.queued_at === undefined, so a first-time task cancelled in the window between the status re-read and the
  queue write gets a false authoritative "queued — will start automatically. No action needed" comment on Linear/Jira for a task that was actually cancelled and never queued. One-line fix:
  queue write gets a false authoritative "queued — will start automatically. No action needed" comment on Linear/Jira for a task that was actually cancelled and never queued. One-line fix:
  const queued = await queueTask(current); if (queued && current.queued_at === undefined) { …notify… }. This is the one I'd most want fixed.

  N1 (medium, unaddressed). The ERROR-escalation gate (reconcile-admission-queue.ts:471) still requires picked_up === 0, so a partial invoke outage (one user's pickups fail, another's
  succeed) logs the summary at INFO despite real invoke_failed. The commit reworded the doc comment but didn't change the gate. Consider a ratio gate (invoke_failed >= picked_up).

  Prior thread #2 / GAP A (test, still open — was a strong suggestion). orchestrate-task.test.ts still imports only individual functions and never invokes the durable handler, so the
  load-bearing admission-control wiring (queueTask(current) vs stale task; notify gated on first-entry) at orchestrate-task.ts:106-112 has zero handler-level coverage. A regression to
  queueTask(task) would pass every test. This is exactly the wiring F1/N2 also touch, so a handler test would earn its keep.

  GAP B/C (test nits, still open). expireQueuedTask cancel-race + non-conditional-throw branches; and a mixed over-age + live batch for one user (pins "expired tasks don't consume a pickup
  slot").

  Tests & CI

  - New tests are solid and behavior-focused (both UpdateItems asserted with their condition guards; restore-race keeps errors===0; ABCA_COMPONENT asserted; the three stranded cases). One
  uncovered path: the phantom-success double-invoke behind F1 — the new tests only exercise genuine-failure and conditional-race, so F1 passes the suite.
  - Bootstrap (ADR-002): still N/A — no new CFN resource types.
  - ⚠️ CI has not re-run on 6bb610fe — the last green build/security-pr was on b84fbe97 (two commits back). Re-run before merge, and note the UA fix has no lint rule, so green CI wouldn't
  have caught B1 anyway.

  Review agents run (this round)

  - silent-failure-hunter → flagged F1 (HIGH), F2, and re-confirmed N1/N2 unaddressed.
  - code-reviewer → confirmed UA fix complete/correct and FIFO preserved; rated F1's double-invoke ~55 (below its bar); surfaced F3.
  - pr-test-analyzer → new tests adequate; GAP A/B/C still open.
  - Omitted this round: type-design-analyzer (no type changes in the delta) and comment-analyzer (only 3 new comments, hand-verified — F3 is the one inaccuracy). /security-review not
  separately run — the delta's only security-relevant change is the UA fix, which improves attribution; no new IAM/Cedar/secrets surface.

  Human heuristics

  - Proportionality — PASS. The re-queue is a proportionate improvement over leave-stranded.
  - Coherence — PASS. UA fix now matches every sibling reconciler's makeClient + ABCA_COMPONENT idiom (the B1 divergence is gone).
  - Clarity — CONCERN. F3 (backstop comment overstates the reconciler) and N2 (false "queued" message) both surface plausible-but-wrong behavior.
  - Appropriateness — PASS. Tests assert should-do against the real GSI schema.

ayushtr-aws (approved):

PR #545, opened 2026-07-08
feat(agent): observable blocker signal + bounded self-remediation (#251)

### LINKED ISSUE #251: feat(agent): observable blocker signal + bounded self-remediation for environmental faults
### Component

Agent (Python runtime), API or orchestration

### Describe the feature

Introduce a **structured "blocked" signal** so that when the agent cannot make progress for an *environmental* reason — a missing secret, a network egress denial, an unreachable dependency registry, or a fail-closed policy denial that looks like a misconfiguration rather than an attack — it emits a typed, machine-readable blocker event instead of silently burning turns or failing opaquely at the end of the run. Where safe, the agent should also attempt **bounded self-remediation** (e.g. retry an install, surface the exact missing secret name, suggest the egress domain to allowlist) before giving up.

Today the platform is correctly **fail-closed**:
- Cedar `PreToolUse` denies (incl. `fail-closed: <ExceptionType>` on engine error) — `agent/src/policy.py`, `agent/src/hooks.py`.
- Egress allowlist blocks non-approved domains at the network layer (`SECURITY.md`).
- Missing-secret / blueprint drift fails the task at `HYDRATING` (`CEDAR_HITL_GATES.md` §scenario finding #11).

These are the right *security* behaviors, but they are **terminal or invisible from the agent's point of view**: the agent experiences a generic tool error or a blocked socket, has no structured way to say *"I am blocked because secret `FOO` is missing"* or *"egress to `registry.example.com` was denied"*, and — because tasks are fully unattended and one-shot (`COMPUTE.md`) — no human is watching to intervene. The cost of a silent block is high: an agent can waste turns (and inference cost) flailing against an environmental wall, or fail at finalization with a reason that requires log spelunking to diagnose.

This issue adds the **observable blocker contract + bounded self-remediation**, distinct from the HITL approval pause (`AWAITING_APPROVAL`), which is for *policy* decisions a human must make, not *environmental* faults the platform/agent can often diagnose or fix.

### Use case

- **Unattended reliability:** A task fails because a required secret was never wired into the blueprint. Today: opaque mid-run tool errors or a late `FAILED`. Desired: an early `blocked` progress event naming the missing secret, surfaced in `bgagent status` / `watch` and notification channels, so the operator fixes the blueprint and resubmits in one step.
- **Egress diagnosis:** The agent's `npm install` hits a registry outside the egress allowlist. Desired: a `blocked` event of kind `egress_denied` carrying the exact denied domain — actionable remediation, not a generic network timeout.
- **Cost control:** Stop the agent from spending turns/tokens retrying against an environmental wall it cannot move; convert "flail then fail" into "detect, report, optionally self-heal, else fail fast with a precise reason."
- **Roadmap alignment:** Realizes the **blockers** dimension of *Smart progress updates* (`ROADMAP.md` — done/next/decisions/**blockers**) and complements the existing deterministic status snapshot.

### Proposed solution

**Phase 0 — Blocker taxonomy + event contract**
- Define a closed set of blocker kinds: `missing_secret`, `egress_denied`, `dependency_unreachable`, `auth_failure`, `policy_fail_closed`, `unknown_environmental`.
- Add a `blocked` event type to `agent/src/progress_writer.py` with structured metadata `{ kind, detail, remediation_hint, retryable, resource? }`. Reuse the existing circuit-breaker / truncation handling.
- Extend the error classifier (`error-classifier.ts`) categories so a blocker that reaches finalization maps to a precise `category` + `remedy` in the API response.

**Phase 1 — Detection at the existing fault points**
- **Cedar fail-closed:** when `policy.py` returns `fail-closed: <ExceptionType>` (engine unavailable/misconfig, *not* an intentional hard-deny), emit `policy_fail_closed` blocker before denying.
- **Egress:** detect denied egress (connection refused/timeout against a non-allowlisted host) in the agent's shell/tool wrapper and emit `egress_denied` with the host.
- **Missing secret / hydration drift:** when hydration fails for a missing/unresolved secret, attach the structured blocker reason to the task record so it surfaces via the events API, not just the terminal `error_message`.

**Phase 2 — Bounded self-remediation (opt-in, safe-by-default)**
- A small remediation policy: retry transient install failures with backoff (N bounded attempts); for `missing_secret` / `egress_denied`, do **not** attempt to acquire anything — only report with a precise hint (security-preserving).
- Self-remediation must never widen the agent's privilege or egress surface; it operates strictly within the already-granted SessionRole and allowlist. Every attempt emits an event so it is auditable.

**Phase 3 — Surface + consume**
- `bgagent status` and `watch` render `blocked` events distinctly (e.g. a ⛔ line with kind + remediation hint).
- Notification channels (GitHub/Slack/Linear) include the blocker reason in progress/final comments.

**Out of scope (explicit)**
- Acquiring or injecting missing credentials, or auto-modifying the egress allowlist — those remain operator/blueprint actions. This issue **reports** them precisely; it does not grant them.
- The HITL `AWAITING_APPROVAL` policy-pause path (unchanged).
- Workload-anchored credential binding (separate roadmap item).

### Acceptance criteria

- [ ] Blocker taxonomy and `blocked` event metadata schema documented in design docs (source + Starlight sync via `cd docs && node scripts/sync-starlight.mjs`).
- [ ] `agent/src/progress_writer.py` emits `blocked` events; covered by `agent/tests/test_progress_writer.py`.
- [ ] Cedar `fail-closed` path emits a `policy_fail_closed` blocker before denying (distinct from intentional hard-deny); unit-tested in `agent/tests`.
- [ ] Egress denial produces an `egress_denied` blocker carrying the denied host.
- [ ] Missing-secret / hydration-drift failures attach a structured blocker reason to the task record, surfaced via `GET /v1/tasks/{id}/events` and the error classifier `remedy`.
- [ ] Bounded self-remediation: transient install retries are capped, logged as events, and provably never widen SessionRole scope or egress allowlist (asserted by test).
- [ ] `bgagent status` / `watch` render blockers distinctly; tests in `cli/test`.
- [ ] No regression to fail-closed security posture: any unhandled policy/engine error still denies the tool call.

krokoko:
Closes #251.

Emits a typed, machine-readable `agent_blocked` event when the agent hits an **environmental** fault (missing secret, egress denial, unreachable dependency, or a fail-closed policy denial that looks like misconfiguration rather than an attack) instead of silently burning turns or failing opaquely. Where safe, attempts **bounded self-remediation** (retry a transient install, name the missing secret, name the egress domain to allowlist) before giving up.

Distinct from the HITL `AWAITING_APPROVAL` pause: that is for *policy* decisions a human must make; this is for *environmental* faults the platform/agent can diagnose or fix.

## Scope decisions
- **`blocked` is an event, not a new task status.** Emitted during `RUNNING`; terminal outcome stays `FAILED` (or `COMPLETED` if self-remediation succeeds), just with precise classification.
- **One canonical blocker-reason contract:** `BLOCKED[<kind>]: <detail>` (+ ` (resource: <resource>)`). Single join point across the agent event metadata, `TaskResult.error`, and the CDK classifier regex.
- **Fail-closed discriminated by a structured `fail_closed: bool` field** on `PolicyDecision`, not by brittle reason-string matching.

## Blocker taxonomy (closed set)
`missing_secret`, `egress_denied`, `dependency_unreachable`, `policy_fail_closed` (v1 detection); `auth_failure`, `unknown_environmental` (enum reserved, no v1 detection site).

## What changed
- **Phase 0 — contract + docs:** `BLOCKER_KINDS` closed set + `format_blocker_reason()` + `write_agent_blocked()` (`agent/src/progress_writer.py`); new `BLOCKED` classifier categories + per-kind remedy builder (`cdk/src/handlers/shared/error-classifier.ts`); taxonomy + event/reason contract docs (`CEDAR_HITL_GATES.md` §13.16, `COMPUTE.md` cross-link) + synced Starlight mirrors.
- **Phase 1 — detection:** structured `fail_closed` on `PolicyDecision`; `progress` threaded into `post_tool_use_hook`; PreToolUse emits `policy_fail_closed` on fail-closed deny (hard-denies never emit); PostToolUse egress-denial detection → `egress_denied` with host; carry-path latch → terminal `error`; `resolveGitHubToken` throws `MissingSecretError` on missing/empty secret.
- **Phase 2 — bounded self-remediation:** `run_cmd_with_backoff` + `is_transient_cmd_failure`; clone + PR-branch fetch retry transient failures, exhausted-transient → `dependency_unreachable`. Model-issued installs are *reported*, not retried (PostToolUse fires post-execution — no in-hook retry seam). Provably no scope/allowlist widening.
- **Phase 3 — surface + consume:** `watch` renders `agent_blocked` (⛔ + kind/resource/hint); `formatStatusSnapshot` shows latest blocker; notifications flow through existing `classifyError` path.

## Tests
agent 1164 pass · CDK 2243 pass · CLI 575 pass. Full `mise run build` green (typecheck, lint, synth, docs-sync, types-sync, coverage). Two `code-review` passes; all found defects fixed.

Governance: #251 carries the `approved` label and is assigned.

## Follow-up review fixes (2 rounds)

Two rounds of multi-agent review (silent-failure, test-coverage, code-review) produced the following fixes on top of the original phases:

- **Setup-path egress reclassification:** a `git clone`/`fetch` failure that names an unresolvable host is classified as non-retryable `egress_denied` (allowlist remedy), not retryable `dependency_unreachable`. The retry-bail is DNS-scoped (`could not resolve host` / `getaddrinfo ENOTFOUND|EAI_AGAIN` / `Failed to resolve`) so a transient TCP-connect timeout to an allowlisted host still retries; a persistent one is reclassified at exhaustion. Setup path and the PostToolUse egress detector reach a consistent verdict.
- **`SecretUnreadableError` (`auth_failure`):** a token secret that exists but can't be read (AccessDenied / throttling) now propagates the canonical blocker reason through hydration + issue/PR fetch instead of silently degrading to minimal context (which stranded the task with no token).
- **Preflight reasons:** token-secret failures report `GITHUB_TOKEN_SECRET_MISSING` / `GITHUB_TOKEN_SECRET_UNREADABLE` instead of the misleading `GITHUB_UNREACHABLE` (`classifyError` still routes on the `BLOCKED[...]` detail).
- **`auth_failure` remedy copy:** a Secrets Manager ARN resource yields IAM advice (grant `secretsmanager:GetSecretValue`) instead of PAT-scope advice.
- **Status snapshot:** the historical `Blocker:` line is suppressed on `COMPLETED` tasks (a self-remediated success is not shown as blocked); still shown on `FAILED`.
- **Workflow runner:** `progress` is threaded into `setup_repo` so clone/fetch backoff blocker events reach the live stream on the workflow path.

## Live testing (deployed)

Verified end-to-end against a live deployment (account ID / repo owner redacted below):

- **`auth_failure` (unreadable secret):** onboarded a repo pointing at a token secret ARN the task role cannot read, then submitted a task. Terminal outcome:
  ```
  FAILED — blocked: Blocked: authentication rejected
    — Pre-flight check failed: GITHUB_TOKEN_SECRET_UNREADABLE
    — BLOCKED[auth_failure]: the required GitHub token secret could not be read (resource: arn:aws:secretsmanager:us-east-1:<REDACTED>:secret:<REDACTED>)
  ```
  Fast-failed at preflight (~3s), no silent degrade — confirming the propagation + accurate `failureReason` label.
- **SM-ARN remedy copy:** `status --output json | jq '.error_classification'` returned:
  ```json
  {
    "title": "Blocked: authentication rejected",
    "remedy": "Grant the task role `secretsmanager:GetSecretValue` on this secret (check the secret resource policy and the blueprint wiring), then resubmit.",
    "retryable": false,
    "category": "blocked"
  }
  ```
  Confirming the IAM-oriented remedy (not PAT-scope advice) fires for a Secrets Manager ARN resource.

The setup-path DNS-bail / egress reclassification is not reachable via the CLI (the clone target is hardcoded to `github.com`); it is covered by an end-to-end unit test driving the real `run_cmd_with_backoff`.


agent/src/hooks.py
```diff
@@ -950,6 +1076,43 @@ async def post_tool_use_hook(
     if not isinstance(tool_response, str):
         tool_response = str(tool_response)
 
+    # #251: best-effort egress-denial detection. A blocked outbound connection
+    # (non-allowlisted host hitting the DNS Firewall, refused connection, name
+    # resolution failure) surfaces in the tool's stderr here. Emit an
+    # ``egress_denied`` blocker naming the host to allowlist. Observability
+    # only — never changes the pass-through / redaction decision below.
+    egress_detected, host = detect_egress_denial(tool_response)
```
theagenticguy, line 1084: **Non-blocking (follow-up #551): egress detection runs on every tool's output, not just Bash/network calls.**

`detect_egress_denial` scans `tool_response` for all tools here. A `Read` or `grep` that surfaces a file/log containing e.g. `Could not resolve host: foo.com` will fire the event and — because a host is captured — latch the terminal reason via `_record_blocker_reason`. On a successful task this is harmless (the success path returns before the latch is read). But if the task later errors with no more specific reason, a stale/unrelated host can be promoted as the terminal cause, sending an operator to the DNS Firewall for nothing.

You already, correctly, accept false positives for the observability event; the carry-path latch is the sharper edge. Consider scoping host-capturing detection (or at least the terminal latch) to `Bash`/network tool calls. Low probability — worth a follow-up, not a change here.

agent/src/hooks.py
```diff
@@ -99,6 +99,102 @@ def _tool_input_preview(tool_input: Any, max_len: int = TOOL_INPUT_PREVIEW_MAX)
...
+# #251 carry-path: the most recent agent-detected blocker, as a canonical
+# ``BLOCKED[<kind>]: …`` reason string (see ``format_blocker_reason``). Blockers
+# are detected mid-turn in the hooks (egress in PostToolUse, fail-closed in
+# PreToolUse) but the terminal ``error_message`` is assembled later in the
+# pipeline. This latch bridges the two: the pipeline promotes it into
+# ``TaskResult.error`` ONLY when the task errored without a more specific
+# reason, so the CDK classifier surfaces a precise remedy. Process-lifetime
+# (== one task) like ``_INJECTED_NUDGES``; last-writer-wins is fine — the most
+# recent blocker is the most relevant to a terminal failure.
+_LAST_BLOCKER_REASON: str | None = None
```
theagenticguy, line 144: **Non-blocking (follow-up #551): process-global latch, reset only in `pipeline.run_task`.**

`_LAST_BLOCKER_REASON` is a process-scalar. Correct today under one-task-per-Firecracker-MicroVM (COMPUTE.md), and you flagged the thread-safety nit explicitly. Worth a one-line note at the definition that correctness depends on that invariant — if the execution model ever moves to multiple tasks per process, this becomes task-id-keyed work. No code change needed while the invariant holds.

cdk/src/handlers/shared/error-classifier.ts
```diff
@@ -329,6 +334,127 @@ const PATTERNS: readonly ErrorPattern[] = [
...
+      };
+    case 'policy_fail_closed':
+      return {
+        category: ErrorCategory.BLOCKED,
+        title: 'Blocked: policy engine fail-closed',
+        description: 'The Cedar policy engine errored or was unavailable, so the tool call was denied fail-closed. This is a misconfiguration signal, not an intentional hard-deny.',
+        remedy: 'Check the agent policy-engine logs (cedarpy load / policy compilation). Fix the policy bundle or engine availability, then resubmit.',
+        retryable: false,
+      };
+    case 'auth_failure': {
```
theagenticguy, line 428: **Non-blocking (follow-up #551): the non-ARN `auth_failure` branch has no live producer.**

`auth_failure` / `unknown_environmental` are enum-reserved with no v1 detection site (documented in §13.16). So the non-ARN "verify the credential is valid and has the required scopes" branch is reachable only by the classifier unit test — only the Secrets-Manager-ARN (IAM) branch fires in practice via `SecretUnreadableError`. Fine for v1; flagging so the dead branch is intentional, not forgotten. Wire it up when a runtime credential-rejection detection site lands.

theagenticguy (approved):
## Approve

Fully closes #251. I checked every acceptance criterion in the issue against shipped code + tests, read the full diff, and confirmed CI is green (agent 1164 / CDK 2243 / CLI 575 tests, full `mise run build` passing). Three self-driven review rounds already caught and fixed the subtle failure modes I'd normally raise here.

**What earns the approve:**

- **One canonical `BLOCKED[<kind>]: <detail> (resource: …)` contract**, implemented twice (`format_blocker_reason` in Python, `formatBlockerReason` in TS) and kept in lockstep with docs (source + Starlight mirror). The classifier matches the `BLOCKED[<kind>]` prefix and the `(resource: …)` segment independently, so a reason wrapped with a stack trace still yields the host.
- **Structured `fail_closed: bool` on `PolicyDecision`** instead of `reason.startswith("fail-closed:")` — branching on a flag rather than a string is the right call, and it's correctly kept out of `__eq__`/`__hash__` to protect legacy equality tests.
- **Security is report-only by construction.** `missing_secret` and `egress_denied` are never self-healed; `run_cmd_with_backoff` re-invokes the same command with the same token and DNS rules, so remediation grants no new scope. There's a test asserting the credential/remote-config steps never run when clone fails.
- **The round-3 DNS-vs-TCP fix is the tell of a careful author.** Reusing `detect_egress_denial` for the pre-retry bail would have fast-failed a genuinely-transient TCP timeout to an allowlisted host as `egress_denied`. Narrowing the bail to a DNS-only predicate (`_UNRESOLVABLE_HOST_RE`) while still reclassifying a *persistent* TCP failure at exhaustion is exactly right, and it's guarded by a regression test.

**Non-blocking follow-ups** (none block this merge; captured in #551 so they're not lost) — three inline comments below: egress detection scope, the process-global latch, and the reserved `auth_failure` branch.

Nice work, @krokoko.

agent/src/hooks.py
krokoko: Tracked and will be implemented in #551

cdk/src/handlers/shared/error-classifier.ts
krokoko: Tracked and will be implemented in #551

agent/src/hooks.py
krokoko: Tracked and will be implemented in #551

Issue #556, opened 2026-07-08
feat(observability): Claude Code native OTel events + operator runbook


krokoko:
Child of #245 (suggested sub-deliverable from the issue thread). This is the **per-model-call granularity** layer of the same attribution story: #245's custom harness spans + TaskEvents give task-level correlation; Claude Code's native OTel events add prompt/tool-level cost and token forensics without parsing unstructured logs.

## Background

The harness already wires Claude Code CLI telemetry behind `ENABLE_CLI_TELEMETRY=1` in `agent/src/runner.py` (`CLAUDE_CODE_ENABLE_TELEMETRY`, OTLP logs export, `OTEL_RESOURCE_ATTRIBUTES` with `task.id` / `repo.url` / `agent.model`). CDK leaves it **disabled** today (`ENABLE_CLI_TELEMETRY` commented in `cdk/src/stacks/agent.ts`) pending verification that the ADOT sidecar forwards OTLP **logs**, not just traces.

## Deliverables

1. **Enable + verify** — Turn on `ENABLE_CLI_TELEMETRY` in the AgentCore runtime env after confirming `claude_code.api_request` / `claude_code.tool_result` / `claude_code.tool_decision` records land in CloudWatch (application log group or AgentCore runtime log stream).
2. **Session correlation** — Ensure `task.id` on `OTEL_RESOURCE_ATTRIBUTES` joins to the #245 correlation envelope (`{task_id, user_id, repo, trace_id}`).
3. **Operator runbook** — Add `docs/guides/CLAUDE_CODE_TELEMETRY.md` (sync via Starlight) with:
   - Event type reference (`claude_code.api_request` carries `cost_usd`, `input_tokens`, `output_tokens`, `duration_ms`)
   - Copy-paste CloudWatch Logs Insights queries: per-task cost rollup, tool histogram, PreToolUse / hook blocks via `claude_code.tool_decision`
   - Known limitations: Claude Code emits OTLP **events** (LogRecords), not spans — the Traces tab may be empty while Events/logs are populated; subagent attribution gaps
4. **Docs cross-links** — From `docs/design/OBSERVABILITY.md` and `docs/guides/COST_ATTRIBUTION.md` (complements #215 Bedrock billing tags).
5. **Optional** — Dashboard widgets for `cost_usd` by `task.id` when attributes are present.

## Acceptance criteria
- [ ] One dry-run task produces queryable `claude_code.api_request` records tagged with `task.id`
[...]

Issue #555, opened 2026-07-08
feat(observability): Phase 3 — downstream attribution (GitHub PR + external API correlation)


krokoko:
Child of #245. Phase 3 (**Downstream attribution**) was deferred from the #245 core delivery (Phases 0–2 shipped the correlation envelope across orchestrator logs, TaskEvents, and agent OTel). This issue extends the same `{task_id, user_id, repo, trace_id}` envelope to the platform's **outbound** edges — where ABCA talks to the outside world.

## Context: what #245 already delivered

- **Correlation contract** documented in `docs/design/OBSERVABILITY.md` ("Correlation envelope" + "Join model").
- **Orchestrator logs** carry `{task_id, user_id, repo}` on admission→terminal transitions (via `logger.child`).
- **TaskEvents** (orchestrator + agent) stamp `user_id`/`repo`, and agent events stamp `trace_id` — so the event stream joins to the X-Ray trace.
- **Agent OTel** root span + baggage carry the envelope.
- Git commits already carry `Task-Id:` / `Prompt-Version:` trailers (pre-existing).

Phase 3 is what's left: attribution at the **external-integration boundary**, which is separable and lower value than the in-platform join.

### 1. GitHub PR-level attribution (extend existing trailers)
- Stamp task/user metadata onto **PR bodies and PR comments** (not just commit trailers), so a human reading a PR — or a reviewer reading a comment — can trace it back to the originating `{task_id, user_id}` without leaving GitHub.
- Ensure the GitHub **actor** identity ABCA acts as is attributable back to the platform `user_id` that triggered the task.

### 2. External webhook / API call correlation
Handlers that call *out* (Linear, Jira, Slack, GitHub API, arbitrary webhooks) carry no correlation today. Add one of:
- an outbound correlation **header** (e.g. `X-ABCA-Task-Id`, or a W3C `traceparent`) on requests where the receiver can use it, and/or
- structured **log fields** `{task_id, user_id, trace_id}` at each outbound call site,

so a failure in an external integration can be tied back to the task that caused it.

## Out of scope
- Delegation chain propagation (cryptographically signed `user_id → orchestrator → agent`) — tracked separately in #249.
- The in-platform envelope work (done in #245).

## Acceptance criteria
- [ ] PR body/comments include task + actor attribution; documented in `docs/design/OBSERVABILITY.md` (source + Starlight sync).
[...]

PR #557, opened 2026-07-08
feat(observability): correlation envelope across orchestrator logs, TaskEvents, agent OTel (#245)

krokoko:
## What

Phases 0–2 of #245 — make the `{task_id, user_id, repo, trace_id}` **correlation envelope** observable on every plane, so a task's actions join across CloudWatch logs, TaskEvents, and the X-Ray trace without manual stitching.

Builds on #515, which already persists `otel_trace_id` on the `TaskRecord` (the durable orchestrator↔agent bridge).

## Why

Vision tenet 7 (observable, attributable, replayable). Answers "who triggered this, on which repo, in which task?" for incident response, feeds #237's audit `correlation` block from real propagated context, and lets #215 Bedrock cost logs join to spans by shared `trace_id`/`task_id`.

## Changes

**Phase 0 — contract**
- `docs/design/OBSERVABILITY.md`: "Correlation envelope" section — field table (`repo`/`trace_id`/`session_id` nullable), shared snake_case naming with #215/#237/#249, and the **join model**: the `TaskRecord` bridges the two planes; there is no single W3C trace root because the orchestrator runs before the agent trace exists. Starlight mirror synced.

**Phase 1 — orchestrator (TS)**
- `logger.ts`: `logger.child(context)` merges a persistent envelope into every line (per-call data wins on collision).
- `orchestrate-task.ts` + `shared/orchestrator.ts`: lifecycle functions (orchestrate handler, `finalizeTask`, `hydrateAndTransition`, `failTask`) log via a child logger, so admission→terminal lines carry `{task_id, user_id, repo}`.
- `emitTaskEvent` gains an optional `EventCorrelation` param; lifecycle events stamp `user_id`/`repo` as top-level fields (`repo` omitted for repo-less workflows, #248).

**Phase 1 — agent (Python)**
- `progress_writer.py`: `_ProgressWriter` takes `user_id`/`repo`; `_put_event` stamps `{user_id, repo, trace_id}` (trace_id read per-event from the active span) — the agent event stream now joins to X-Ray. Optional fields omitted, never null.
- `observability.py`: `set_session_id` propagates `user.id`/`repo.url` baggage, guarding each field (empty/None not stamped).
- `pipeline.py`: root `task.pipeline` span gains `user.id`; both writer sites pass the envelope. `server.py`: propagate even when `session_id` is empty.

**Phase 2 — no new code.** #515's `TaskRecord.otel_trace_id` + the Phase 1 event `trace_id` stamping give the cross-plane join. Both `_ProgressWriter`s are constructed inside an active span, so `trace_id` is always valid.

## Acceptance criteria (#245)

- [x] Correlation contract documented (source + Starlight sync)
- [x] `{task_id, user_id, repo, trace_id}` in orchestrator logs, admission→terminal
- [x] Agent OTel spans carry the envelope; baggage survives between phases
- [x] TaskEvents joinable to logs + trace by `task_id`/`trace_id`
- [x] Tests for context injection
- [x] No regression to #209 SessionRole isolation (`user_id` is read-only here)

## Deferred to child issues

- **#555** — Phase 3: downstream attribution (GitHub PR body/comments + external webhook/API correlation headers). Separable, touches the integration handlers, and has its own design questions.
- **#556** — Claude Code native OTel events + operator runbook (per-model-call cost/token forensics; gated on ADOT sidecar OTLP-logs forwarding).

## Testing

- New: logger child-merge (4), orchestrate-task envelope incl. repo-less (3), progress_writer envelope incl. omit-not-null (4), `set_session_id` incl. empty session (3), `_run_task_background` envelope propagation + trace-read fail-open (2).
- CDK 2236 pass + synth clean; agent 1182 pass; eslint / ruff / tsc / docs-sync clean.

## Live verification (deployed, `dev`)

Three tasks submitted post-deploy, one per scenario. Values below are **redacted** — `user_id` shown as `<user>` (same Cognito sub across all three), repo owners masked, trace ids shortened to an 8-hex prefix (the prefix equality is the point).

| Scenario | Task status | `user_id` on events | `repo` on events | `trace_id` on agent events |
|---|---|---|---|---|
| Repo-less workflow | COMPLETED | `<user>` ✅ | **absent** ✅ (omitted, not `""`) | `6a4ea846…` ✅ |
| Code repo | COMPLETED | `<user>` ✅ | `<owner>/agent-plugins` ✅ | `6a4ea5de…` ✅ |
| Fail-fast (pre-flight) | FAILED | `<user>` ✅ | `<owner>/agent-plugins` ✅ | n/a (failed before agent ran) |

**Per-plane checks**

- **Orchestrator log lines** carry the envelope (not just events): `Session started` / `Task already in terminal state` lines show `user_id` + `repo` (repo omitted for the repo-less task) — confirms the `logger.child` wiring.
- **TaskEvents**: `user_id` present on every event from `hydration_started` onward; pre-span events (`task_created`) correctly have `trace_id` absent — matching the documented "read per-event, omitted before the root span opens" behavior.
- **Fail-fast path**: both `preflight_failed` and `task_failed` carry `user_id` + `repo`, confirming the `failTask` correlation threading works on the failure path (review finding #1).
- **Repo-less contract (#248)**: `repo` is genuinely absent (not `null`/`""`) on both the events and the `TaskRecord`.

**Cross-plane join (the core deliverable)** — agent-event `trace_id` **exactly equals** the `TaskRecord.otel_trace_id` (#515 bridge) on both completed tasks:

| Task | `TaskRecord.otel_trace_id` | agent event `trace_id` | match |
|---|---|---|---|
| Repo-less | `6a4ea846…` | `6a4ea846…` | ✅ |
| Code repo | `6a4ea5de…` | `6a4ea5de…` | ✅ |
| Fail-fast | `null` (no agent trace) | — | ✅ (joins by `task_id`/`user_id` instead) |

Net: from any plane you can reach the others — `orchestrator log → task_id → TaskRecord → otel_trace_id → agent events/X-Ray trace`.

### DIFF
--- a/agent/src/observability.py
+++ b/agent/src/observability.py
@@ -64,16 +64,26 @@ def current_otel_trace_id() -> str | None:
     bundle) so operators can join the task to its CloudWatch/X-Ray trace. Returns
     ``None`` when there is no recording span (e.g. tracing disabled locally) or
     the context is invalid, so callers can treat it as a graceful-missing field.
+
+    Never raises: a broken/misconfigured tracer degrades to ``None`` rather than
+    propagating. Callers read this inside DDB-write try-blocks (progress_writer),
+    where a raised trace error would otherwise be misclassified as a DDB failure
+    and trip the shared progress circuit breaker.
     """
-    span = trace.get_current_span()
-    ctx = span.get_span_context()
-    if not ctx.is_valid:
+    try:
+        span = trace.get_current_span()
+        ctx = span.get_span_context()
+        if not ctx.is_valid:
+            return None
+        # format_trace_id renders the 128-bit id as zero-padded 32-char hex — the
+        # OTEL format, so it joins directly in CloudWatch Transaction Search. Note
+        # the X-Ray console renders trace ids as ``1-{8hex}-{24hex}``; to look this
+        # up there, transform to that form (the timestamp is the first 8 hex chars).
+        return trace.format_trace_id(ctx.trace_id)
+    except Exception:
+        # nosemgrep: py-silent-success-masking -- trace id is a graceful-missing
+        # correlation field; a tracer fault must not fail the caller's write path.
         return None
-    # format_trace_id renders the 128-bit id as zero-padded 32-char hex — the
-    # OTEL format, so it joins directly in CloudWatch Transaction Search. Note
-    # the X-Ray console renders trace ids as ``1-{8hex}-{24hex}``; to look this
-    # up there, transform to that form (the timestamp is the first 8 hex chars).
-    return trace.format_trace_id(ctx.trace_id)
 
 
 def set_session_id(
--- a/agent/src/observability.py
+++ b/agent/src/observability.py
@@ -86,9 +86,13 @@ def current_otel_trace_id() -> str | None:
         return None
 
 
-def set_session_id(
+def propagate_correlation_context(
     session_id: str,
     user_id: str = "",
+    # ``user_id`` uses ""-means-absent (Cognito sub, mirrors AgentConfig.user_id
+    # which is never None); ``repo`` uses None-means-absent (mirrors the optional
+    # TaskRecord.repo). Both conventions flow from upstream config types; the
+    # ``if x:`` guards below flatten either to "don't set the baggage key".
     repo: str | None = None,
 ) -> None:
     """Propagate the correlation envelope via OTEL baggage.
@@ -96,7 +100,8 @@ def set_session_id(
     *session_id* correlates custom spans to the AgentCore session; *user_id*
     and *repo* (#245) carry the platform identity and target repo so baggage
     survives across pipeline phases on the task thread. Empty/None fields are
-    not set. *repo* is None for repo-less workflows (#248 Phase 3).
+    not set — so this runs (and is useful) even when *session_id* is empty but
+    the identity is known. *repo* is None for repo-less workflows (#248 Phase 3).
 
     The attached context is intentionally not detached: the background thread
     runs a single task then exits, so the context is garbage-collected with the
--- a/agent/src/observability.py
+++ b/agent/src/observability.py
@@ -76,12 +76,27 @@ def current_otel_trace_id() -> str | None:
     return trace.format_trace_id(ctx.trace_id)
 
 
-def set_session_id(session_id: str) -> None:
-    """Propagate *session_id* via OTEL baggage for AgentCore session correlation.
+def set_session_id(
+    session_id: str,
+    user_id: str = "",
+    repo: str | None = None,
+) -> None:
+    """Propagate the correlation envelope via OTEL baggage.
+
+    *session_id* correlates custom spans to the AgentCore session; *user_id*
+    and *repo* (#245) carry the platform identity and target repo so baggage
+    survives across pipeline phases on the task thread. Empty/None fields are
+    not set. *repo* is None for repo-less workflows (#248 Phase 3).
 
     The attached context is intentionally not detached: the background thread
     runs a single task then exits, so the context is garbage-collected with the
     thread.
     """
-    ctx = baggage.set_baggage("session.id", session_id)
+    ctx = context.get_current()
+    if session_id:
+        ctx = baggage.set_baggage("session.id", session_id, context=ctx)
+    if user_id:
+        ctx = baggage.set_baggage("user.id", user_id, context=ctx)
+    if repo:
+        ctx = baggage.set_baggage("repo.url", repo, context=ctx)
     context.attach(ctx)  # token not stored — thread-scoped lifetime
--- a/agent/src/pipeline.py
+++ b/agent/src/pipeline.py
@@ -679,13 +679,18 @@ def run_task(
             "repo.url": config.repo_url,
             "issue.number": config.issue_number,
             "agent.model": config.anthropic_model,
+            # Correlation envelope (#245): user.id joins agent spans to
+            # orchestrator logs by the platform identity, not just task/repo.
+            **({"user.id": config.user_id} if config.user_id else {}),
         },
     ) as root_span:
         task_state.write_running(config.task_id)
         task_state.write_heartbeat(config.task_id)
 
         agent_result: AgentResult | None = None
-        progress = _ProgressWriter(config.task_id, trace=trace)
+        progress = _ProgressWriter(
+            config.task_id, trace=trace, user_id=config.user_id, repo=config.repo_url
+        )
         # #251: clear any blocker latched by a prior task. The agent container
         # is one-task-per-process today, but the FastAPI server thread-pool can
         # in principle dispatch a second run_task in the same process — reset
--- a/agent/src/progress_writer.py
+++ b/agent/src/progress_writer.py
@@ -38,6 +38,8 @@ from datetime import UTC, datetime
 from decimal import Decimal
 from typing import Literal
 
+from observability import current_otel_trace_id
+
 # Preview field cap defaults (design §10.1):
 #   - 200 chars for normal tasks — small DDB rows, cheap watch-stream bytes.
 #   - 4096 chars (4 KB) for ``--trace`` opt-in tasks — full enough to
@@ -373,8 +375,19 @@ class _ProgressWriter:
 
     _MAX_FAILURES = 3
 
-    def __init__(self, task_id: str, trace: bool = False) -> None:
+    def __init__(
+        self,
+        task_id: str,
+        trace: bool = False,
+        user_id: str = "",
+        repo: str | None = None,
+    ) -> None:
         self._task_id = task_id
+        # Correlation envelope (#245): stamped on every event so the agent-side
+        # event stream joins to orchestrator logs and the X-Ray trace. ``repo``
+        # is None for repo-less workflows (#248 Phase 3).
+        self._user_id = user_id
+        self._repo = repo
         self._table_name = os.environ.get("TASK_EVENTS_TABLE_NAME")
         self._table = None
         # Per-instance preview cap — design §10.1. ``trace=True`` raises
@@ -467,6 +480,11 @@ class _ProgressWriter:
                 return
 
             now = datetime.now(UTC)
+            # Correlation envelope (#245): trace_id is read per-event from the
+            # active span (it may be None early, before the root span opens);
+            # user_id/repo are stamped when known. All are omitted when
+            # empty/None so the item shape stays stable for consumers.
+            trace_id = current_otel_trace_id()
             item = {
                 "task_id": self._task_id,
                 "event_id": _generate_ulid(),
@@ -478,6 +496,12 @@ class _ProgressWriter:
                 "timestamp": now.isoformat(),
                 "ttl": int(now.timestamp()) + _TTL_SECONDS,
             }
+            if self._user_id:
+                item["user_id"] = self._user_id
+            if self._repo:
+                item["repo"] = self._repo
+            if trace_id:
+                item["trace_id"] = trace_id
             self._table.put_item(Item=item)
 
             # Success: reset the shared failure counter.  We do NOT flip
--- a/agent/src/runner.py
+++ b/agent/src/runner.py
@@ -483,7 +483,12 @@ async def run_agent(
     # invocations we fall back to a fresh writer with no accumulator.
     if trajectory is None:
         trajectory = _TrajectoryWriter(config.task_id or "unknown")
-    progress = _ProgressWriter(config.task_id or "unknown", trace=config.trace)
+    progress = _ProgressWriter(
+        config.task_id or "unknown",
+        trace=config.trace,
+        user_id=config.user_id,
+        repo=config.repo_url,
+    )
 
     # Map tool_use_id → tool_name so we can label ToolResultBlocks that arrive
     # in UserMessages (ToolResultBlock carries only the id, not the name).
--- a/agent/src/server.py
+++ b/agent/src/server.py
@@ -26,7 +26,7 @@ from pydantic import BaseModel
 import task_state
 from config import resolve_github_token
 from models import TaskResult
-from observability import set_session_id
+from observability import propagate_correlation_context
 from pipeline import run_task
 
 # --- _debug_cw / _warn_cw failure counter -------------------------------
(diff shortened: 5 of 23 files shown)


krokoko:
## Update — API passthrough follow-up + re-verified live

Since the initial Phases 0–2 work, folded in the independent-review follow-ups and re-ran the live suite against a fresh deploy.

### What was added (commit `c3f31493`)

1. **Correlation envelope surfaced through the event API (#1).** The `{user_id, repo, trace_id}` envelope was stamped on `TaskEvents` in DynamoDB but stripped by `get-task-events.ts` / `get-task-replay.ts`, so `bgagent watch`/`events`/`replay` never saw per-event correlation. Added the three optional fields to `EventRecord` + `ReplayEvent` (cdk) and `TaskEvent` + `ReplayEvent` (cli mirror), and pass them through per-event (omit-when-absent) in both handlers.
2. **Coverage docs (#2)** — OBSERVABILITY.md notes `task_created` predates the envelope (joins by `task_id`) and that CLI consumers can join at task level via `TaskRecord.otel_trace_id`.
3. **Doc wording (#3)** — `repo` documented as "absent (key omitted, not `null`/`""`)" for repo-less workflows, matching the implementation.
4. **Consistency (#4)** — `failTask` + `loadBlueprintConfig` now use `logger.child()` like the rest of the lifecycle code.
5. **API_CONTRACT.md** documents the per-event envelope; `types-sync` gate passes.

Reviewed with the PR-review plugins (code-review, silent-failure, type-design): no critical/important issues, no silent-failure or data-integrity risks, cdk↔cli mirror exact.

### Live verification (re-run on deployed `dev`)

Three tasks submitted, one per scenario. Redacted: `user_id` → `<user>`, repo owners masked, trace ids shown as 8-hex prefix (prefix equality is the point).

**DDB stamping** (`TaskEvents`):

| Scenario | Status | `user_id` | `repo` | `trace_id` (agent events) |
|---|---|---|---|---|
| Repo-less | COMPLETED | `<user>` ✅ | **absent** ✅ | `6a4eb357…` ✅ |
| Code repo | RUNNING | `<user>` ✅ | `<owner>/agent-plugins` ✅ | `6a4eb380…` ✅ |
| Fail-fast (pre-flight) | FAILED | `<user>` ✅ | `<owner>/agent-plugins` ✅ | n/a (failed before agent ran) |

`task_created` carries no envelope on all three; everything from `hydration_started` onward carries `user_id`/`repo`; agent events add `trace_id`. Fail-fast: both `preflight_failed` and `task_failed` carry `user_id` + `repo`.

**API passthrough** (invoked the `GetTaskEvents` / `GetTaskReplay` Lambdas with owner-scoped payloads):

- `GET /tasks/{id}/events` — repo-less: `task_created` has no envelope, `session_started` carries `user_id` (repo absent), agent events carry `user_id` + `trace_id`. Code repo: same plus `repo`. Omit-when-absent confirmed on the wire (keys genuinely absent, not `null`).
- `GET /tasks/{id}/replay` — bundle returned `otel_trace_id`, events carried the envelope, and the join assertion **`bundle.otel_trace_id === event.trace_id` passed**.

**Cross-plane bridge** — on the completed repo-less task, agent-event `trace_id` == `TaskRecord.otel_trace_id` == replay-bundle `otel_trace_id` (all `6a4eb357…`). The code-repo task was still RUNNING, so its `otel_trace_id` isn't persisted on the `TaskRecord` yet (written at terminal) — expected; its events already carry the trace id, so the bridge completes on finalize. Fail-fast: `otel_trace_id` null (no agent trace), events join by `task_id`/`user_id`.

Net: the envelope stamping and the newly-deployed API passthrough both work end-to-end. Verified via direct Lambda invoke (no `bgagent` binary in the verification env; the handler + auth path are exercised identically to a real API call).

ayushtr-aws (commented):
## Review — correlation envelope across orchestrator logs, TaskEvents, agent OTel

**Verdict: well-executed — approve with minor comments.** The design is documented before the code, the omit-not-null contract is enforced consistently at every write and read site, and tests cover the new branches including failure paths. I verified coverage against the branch: all 9 `emitTaskEvent` call sites in `orchestrator.ts` and all 3 in `orchestrate-task.ts` carry correlation (`task_created` is written in `create-task-core.ts` and is documented as pre-envelope), and both `_ProgressWriter` construction sites (`pipeline.py`, `runner.py`) pass the envelope.

### Strengths

- **`current_otel_trace_id` hardening is thoughtful**: the new `try/except` exists specifically because callers read it inside the DDB-write try-block, where a tracer fault would be misclassified as a DDB failure and trip the progress circuit breaker. The comment explains this and there's a dedicated regression test.
- **Consistent optional-field contract**: `if value:` guards on the Python write side, `...(x && { x })` spreads on the TS side, and tests assert key *absence* at every layer — writer, `emitTaskEvent`, API pass-through, and replay bundle.
- **The widened trigger in `server.py`** (`if session_id or user_id or repo_url:`) correctly handles the empty-session case, and `set_session_id` now guards each field so an empty `session_id` no longer writes empty-string baggage — with a test for exactly that branch.
- **Durable-execution awareness preserved**: correlation was added without disturbing the transition-gated emit logic or the duplicate-event-on-retry protections.
- **`logger.child` is minimal and correct**: immutable merge, per-call data wins, composable, base logger unaffected — all four properties tested.
- The cross-plane join holds by construction: the agent's `repo_url` is fed from `task.repo` (`orchestrator.ts` payload assembly), so agent-event `repo` values match orchestrator-event `repo` values exactly.

### Minor issues

1. **`set_session_id` is now misnamed** (`agent/src/observability.py`). It propagates the full correlation envelope, not just the session id — `server.py` even calls it when `session_id` is empty. Consider renaming (e.g. `propagate_correlation_context`); as-is, a reader seeing `set_session_id("")` has to go read the docstring to understand why that's not a bug.

2. **Stray `// ponytail:` marker** (`cdk/src/handlers/shared/logger.ts:40`). The comment content is useful; the prefix looks like a leftover codename/typo and should go.

3. **`loadBlueprintConfig`'s child logger omits `user_id`** (`cdk/src/handlers/shared/orchestrator.ts`). It builds `logger.child({ task_id, ...repo })` while every other lifecycle function includes `user_id`, and `task.user_id` is in scope. This leaves the blueprint-load log lines (including the `poll_interval_ms clamped` warning) short of the "admission→terminal lines carry `{task_id, user_id, repo}`" acceptance criterion. One-token fix.

4. **Envelope-construction duplication in TS**: `logger.child({ task_id, user_id, ...(task.repo && { repo: task.repo }) })` plus a parallel `EventCorrelation` literal is hand-rolled in four places. A small `envelopeFor(task)` helper returning `{ log, correlation }` would prevent the drift that already produced item 3.

5. **Type asymmetry in Python params**: `user_id: str = ""` vs `repo: str | None = None` in both `set_session_id` and `_ProgressWriter`. It mirrors the upstream config types, but mixing absent-value conventions (`""` vs `None`) within one envelope invites mistakes; a short comment on why they differ would help.

6. **Test brittleness (very minor)**: `test_sets_all_envelope_fields_when_present` asserts the exact *order* of baggage keys. Order isn't part of the contract; a set/sorted comparison would survive harmless reordering.

### Security / risk

- `user_id` (Cognito sub) now lands in CloudWatch logs and DDB items — an opaque UUID, and attribution is the explicit purpose of the feature. The API handlers only return events for tasks the caller is authorized on, so no new exposure surface.
- The broad `except Exception` in `current_otel_trace_id` is a silent-success mask, but it's narrowly scoped, justified in docstring + `nosemgrep` annotation, and tested.
- No schema/infra changes: DDB items gain optional attributes only, so old readers and pre-envelope items are unaffected (explicit back-compat test for the historical item shape).

### Test coverage

Strong: logger child-merge semantics, lifecycle event correlation including repo-less omission, `emitTaskEvent` back-compat, `_ProgressWriter` omit-not-null in all combinations, tracer fail-open, empty-session baggage propagation, and API pass-through in both `events` and `replay` handlers. The one seam not directly unit-tested is `failTask`'s new `repo` param on the successful-transition path, but it's exercised transitively and via the live fail-fast verification in the PR description.

None of the above is blocking; items 1–3 are worth a quick follow-up commit before merge.

isadeks (dismissed):
## Approve with nits ✅

Clean, well-executed observability plumbing. Threads the `{task_id, user_id, repo, trace_id}` envelope across orchestrator logs (`logger.child`), TaskEvents (`EventCorrelation` on `emitTaskEvent` + agent-side `_ProgressWriter` stamping), and OTel baggage/spans. Strong test suite + live `dev` verification of the cross-plane `trace_id == TaskRecord.otel_trace_id` join. **No correctness bugs or regressions found** — the findings below are all minor.

### Verified clean (the subtle bits)
- **Spread patterns omit correctly** — reproduced `...(correlation?.user_id && {…})` and the `get-task-events` passthrough in Node: empty-string/`undefined` fields are cleanly omitted, never stamped as `""`/`null`, matching the "omitted, not null" contract.
- **No PII leak to external channels** — the TaskEvents NEW_IMAGE stream feeds `FanOutConsumer` (Slack/GitHub/email), but `parseStreamRecord` **whitelists** fields (`task_id`, `event_id`, `event_type`, `timestamp`, `metadata`); the new top-level `user_id`/`repo`/`trace_id` (incl. the Cognito `sub`) are not forwarded.
- **Trace-read can't trip the circuit breaker** — `trace_id = current_otel_trace_id()` is inside `_put_event`'s try-block, but the function now swallows all exceptions → `None`, so a tracer fault won't be misclassified as a DDB failure.
- **All call sites consistent** — all 9 `failTask` calls got `task.repo` (optional param → back-compat); all 12 `emitTaskEvent` calls well-formed; `finalizeTask`'s `userId` param and `task.user_id` reference the same (immutable) task; `failTask`'s catch-scoped `log` is used only in-scope with the correct `userId`.
- **`set_session_id`** has one guarded caller (`server.py`, `if session_id or user_id or repo_url`) on a single-task thread → the never-detached `context.attach` is genuinely thread-scoped.

### Findings (all minor)

**🟡 1. `loadBlueprintConfig`'s child logger omits `user_id`** — `cdk/src/handlers/shared/orchestrator.ts:222`
```ts
const log = logger.child({ task_id: task.task_id, ...(task.repo && { repo: task.repo }) });
```
Every other function converted in this PR (`hydrateAndTransition`, `finalizeTask`, `failTask`, the `durableHandler`) includes `user_id: task.user_id`, and it's available on the `TaskRecord` here too. The comment even says this logger is "matching the orchestrate handler's child logger" — but the handler's carries `user_id` and this one doesn't. So the three blueprint-load lines won't surface in a CloudWatch Insights query filtered on `user_id`, defeating the per-line-envelope goal for this function. Not a regression, but the one I'd actually fix (one line: add `user_id: task.user_id`).

**🟡 2. `repo` `""`-vs-`None` contract drift** — `agent/src/pipeline.py:692` + `runner.py:490` pass `repo=config.repo_url` (which is `""`, never `None`, for repo-less tasks), while `server.py:460` normalizes `repo=repo_url or None`, and `_ProgressWriter`'s comment says "`repo` is None for repo-less workflows." Harmless today (`_put_event` guards `if self._repo:`, so `""` and `None` both omit the key), but the stored contract is violated and the three producers disagree — a latent trap for any consumer that distinguishes `is None` from `== ""`. Suggest `config.repo_url or None` in pipeline/runner.

**⚪ 3. Logger clobber footgun (latent only)** — `logger.ts` serializes `{ level, message, ...context, ...data }`, so a caller-supplied `data.message`/`data.level` would override the positional args. No current caller does this, and the ordering matches the pre-PR logger, so there's no active bug — just a sharp edge worth a comment.

**⚪ 4. `progress_writer` now eagerly imports `opentelemetry` (posture note)** — the new `from observability import current_otel_trace_id` forces an eager `opentelemetry` import at module-load, and the docstring advertises fail-open import safety. But that safety is specifically about **boto3** (lazily imported); `observability`/`opentelemetry` is already eagerly imported by `pipeline.py:29` and is a hard prod dep (`aws-opentelemetry-distro==0.18.0`), so practical risk ≈ nil. Noting only for completeness.

### Bottom line
Solid, low-risk, no blockers. #1 is the only one worth changing before merge (restores the PR's own invariant); #2 is a nice consistency tidy; #3/#4 are optional.

_Findings verified against the `feat/245-observability-attribution` checkout, including direct Node repro of the spread patterns and a trace of the fanout stream field-extraction._

krokoko:
Thanks for the thorough review — all 6 items addressed in `f560cf33`:

1. **Renamed** `set_session_id` → `propagate_correlation_context` (+ sole caller in `server.py` and tests). It reads correctly now when called with an empty `session_id`.
2. **Dropped the stray `// ponytail:` prefix** in `logger.ts` (kept the explanatory content).
3. **`loadBlueprintConfig` now carries `user_id`** on its child logger — fixed via item 4.
4. **Extracted `envelopeFor(task) → { log, correlation }`** in `orchestrator.ts` as the single source for the log child and the event correlation. Replaces the hand-rolled `logger.child(...)` + `EventCorrelation` literal across all 5 sites (orchestrate handler, `hydrateAndTransition`, `finalizeTask`, `failTask`, `loadBlueprintConfig`) — removing the drift that produced item 3. Added unit tests asserting the child carries `user_id` (incl. repo-less omission).
5. **Added a comment** on the `""`-vs-`None` param asymmetry in `propagate_correlation_context` (they mirror the upstream config types; the `if x:` guards flatten both).
6. **Set-compare** on baggage keys in the observability test instead of exact order.

On the one seam you flagged as not directly unit-tested — `failTask`'s successful-transition `repo` path — it's still covered transitively + via the live fail-fast run; I left it as-is since it's exercised, but happy to add a direct assertion if you'd prefer.

CDK 2258 pass + synth clean, agent 1182 pass, eslint/ruff/tsc all clean.

theagenticguy (approved):
Approve. Reviewed the full diff and spot-verified the load-bearing claims against the source rather than trusting the description:

- `logger.child` merge precedence is correct — `{ level, message, ...context, ...data }` spreads per-call `data` last, so it wins on collision as documented.
- Omit-not-null is done right on both planes: TS `...(task.repo && { repo })` and Python guards (`if user_id:` / `if repo:` / `if trace_id:`). Absent fields are omitted, never `null`/`""` — important for the repo-less contract (#248).
- `task.repo` is threaded through the failure path (`failTask`), so `preflight_failed`/`task_failed` also carry correlation.

Tests cover the new paths (logger child-merge, orchestrate-task envelope incl. repo-less, progress_writer omit-not-null, `set_session_id` empty-session, background trace-read fail-open), CI is green across the board, and scope is cleanly deferred to #555/#556. Nice, well-documented change. :robot:

PR #559, opened 2026-07-09
feat(cli): add Jira invite-user command

### LINKED ISSUE #553: CLI: implement `bgagent jira invite-user` — issuing half of Jira identity linking
### Component

Scripts / CLI

### Describe the feature

Implement `bgagent jira invite-user` — the missing issuing half of the Jira identity-linking handshake.

The redeeming half already exists and works end-to-end: `bgagent jira link <code>` → `POST /v1/jira/link` → `cdk/src/handlers/jira-link.ts` consumes a one-time code and writes the user-mapping row. But nothing can *mint* a code today:

- `bgagent jira setup`'s "Next steps" output says: *"(an admin runs `bgagent jira invite-user` to issue you a code; this command is not yet implemented — populate the user-mapping row manually for now.)"* (`cli/src/commands/jira.ts:579`)
- `JIRA_SETUP_GUIDE.md` Step 5 carries the same caveat.
- When an unlinked Jira user adds the trigger label, the webhook processor comments back on the issue: *"Run `bgagent jira link <code>`"* (`cdk/src/handlers/jira-webhook-processor.ts:~330`) — a dead end, since no code can be issued.

The only workaround is an admin hand-writing the DynamoDB row (`jira_identity` = `<cloudId>#<accountId>`, `platform_user_id`, `jira_cloud_id`, `jira_account_id`), which requires reading handler source to discover the schema.

Linear already has the exact analog implemented: `bgagent linear invite-user <slug>` (`cli/src/commands/linear.ts:1171`) writes a `pending#<code>` row with a 24 h TTL to the user-mapping table; the teammate redeems it with `linear link <code>`. Jira has the redemption path but not the issuance path — this is a Linear-parity gap left over from #288/#302.

### Use case

An admin onboarding teammates to the Jira integration currently has to hand-craft DynamoDB rows per user, reverse-engineered from handler source. Every unlinked user who triggers a label also receives a bot comment instructing them to run a command that cannot succeed. With `invite-user`, the flow matches Linear: admin runs one command, sends the code, teammate redeems it and sees their Jira name/email before confirming.

### Proposed solution

Port `linear invite-user` (`cli/src/commands/linear.ts:1171-…`) to `cli/src/commands/jira.ts`:

- `bgagent jira invite-user <cloud-id> <account-id-or-email>` (admin IAM): resolve the teammate's Jira identity via the tenant OAuth token (`/rest/api/3/user/search`), write a `pending#<code>` row to `JiraUserMappingTable` (24 h TTL, mirroring Linear's shape and `jira-link.ts`'s expectations), print the one-time code.
- Teammate redeems with the existing `bgagent jira link <code>` (already implemented, shows name+email preview before binding).
- Update `JIRA_SETUP_GUIDE.md` Step 5 and the `setup` "Next steps" text to drop the "not yet implemented" caveats.

### Other information

Identity-linking docs gap (the manual-row interim path) is tracked separately in #552.

### Acknowledgements

ayushtr-aws:
Add a `bgagent jira invite-user` command that mints Jira identity-link invite codes so operators can onboard teammates into the existing `jira link` flow.

#### Area

#### Related

Closes #553

#### Changes

- Add `bgagent jira invite-user <cloud-id> <account-id-or-email>` to mint Jira identity-link invite codes.
- Resolve Jira users through the tenant OAuth token, refresh expired tokens, and write `pending#<code>` rows consumed by the existing `bgagent jira link <code>` flow.
- Update Jira setup docs and regenerate the Starlight mirror.

#### Validation

- Manually verified invite + redeem flow against tenant `83f2ac74-f60e-4053-9514-a887d6ef5fd7`.
- `npx jest test/commands/jira.test.ts --coverage=false`
- `MISE_EXPERIMENTAL=1 mise //cli:build`
- `MISE_EXPERIMENTAL=1 mise //docs:build`
- `mise run build`

#### PR checklist

- [x] Unit test added
- [x] Integration test added (not applicable: no new CloudFormation resource types or cross-service configuration)
- [x] Documentation updated
- [x] Title follows conventional commits
- [x] Breaking changes documented in commit body (not applicable: no breaking changes)

#### Acknowledgment

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -302,6 +302,25 @@ With no flags, writes to the platform default `GitHubTokenSecretArn` stack outpu
 
 Configure the preview-deploy screenshot pipeline webhook. See [Deploy preview screenshots guide](../docs/guides/DEPLOY_PREVIEW_SCREENSHOTS_GUIDE.md).
 
+### `bgagent jira setup` / `map` / `invite-user` / `link`
+
+Manage the Jira Cloud integration. `setup` authorizes a tenant via OAuth (3LO) and stores the token in Secrets Manager; `map` routes a Jira project to a GitHub repo; the two-step `invite-user` → `link` handshake links a teammate's Jira identity to their platform user. See the [Jira setup guide](../docs/guides/JIRA_SETUP_GUIDE.md) for the full walkthrough.
+
+```
+bgagent jira setup \
+  --stack-name backgroundagent-dev
+
+bgagent jira map <cloud-id> <PROJECT-KEY> --repo owner/repo
+
+bgagent jira invite-user <cloud-id> <account-id-or-email> \
+  --region <region>            AWS region (defaults to configured region) \
+  --stack-name <name>          CloudFormation stack name (default: backgroundagent-dev)
+
+bgagent jira link <code>
+```
+
+`invite-user` resolves the teammate's Jira identity through the tenant OAuth token, then writes a `pending#<code>` row (24h TTL) and prints the `bgagent jira link <code>` the teammate runs from their own machine. The teammate previews the Jira identity before confirming, so a wrong pick can be aborted rather than misattributed. If the identity is already linked, the command warns but still issues the code.
+
 ### `bgagent admin invite-user` / `list-users` / `delete-user` / `reset-password`
 
 Manage Cognito users with operator AWS credentials (`cognito-idp:Admin*` on the deployment user pool). Works **before** `bgagent configure` when `--stack-name` is passed (reads `UserPoolId` from CloudFormation).
--- a/cli/src/commands/jira.ts
+++ b/cli/src/commands/jira.ts
@@ -830,21 +830,33 @@ export function makeJiraCommand(): Command {
 
         const code = generateInviteCode();
         const ttl = Math.floor(Date.now() / 1000) + 24 * 60 * 60;
-        await ddb.send(new PutCommand({
-          TableName: userMappingTable!,
-          Item: {
-            jira_identity: `pending#${code}`,
-            status: 'pending',
-            jira_cloud_id: cloudId,
-            jira_site_url: siteUrl,
-            jira_account_id: picked.accountId,
-            jira_user_name: picked.displayName ?? '',
-            jira_user_email: picked.emailAddress ?? '',
-            invited_at: new Date().toISOString(),
-            invited_by_platform_user_id: callerCognitoSub,
-            ttl,
-          },
-        }));
+        try {
+          await ddb.send(new PutCommand({
+            TableName: userMappingTable!,
+            // Guard against clobbering an existing pending invite on the
+            // (astronomically unlikely) chance the generated code collides.
+            // Better to fail loudly and let the admin re-run than silently
+            // overwrite a still-valid code.
+            ConditionExpression: 'attribute_not_exists(jira_identity)',
+            Item: {
+              jira_identity: `pending#${code}`,
+              status: 'pending',
+              jira_cloud_id: cloudId,
+              jira_site_url: siteUrl,
+              jira_account_id: picked.accountId,
+              jira_user_name: picked.displayName ?? '',
+              jira_user_email: picked.emailAddress ?? '',
+              invited_at: new Date().toISOString(),
+              invited_by_platform_user_id: callerCognitoSub,
+              ttl,
+            },
+          }));
+        } catch (err) {
+          if ((err as { name?: string }).name === 'ConditionalCheckFailedException') {
+            throw new CliError(`Invite code '${code}' collided with an existing invite. Re-run the command to mint a fresh code.`);
+          }
+          throw err;
+        }
 
         console.log();
         console.log('✅ Invite created.');
--- a/cli/src/commands/jira.ts
+++ b/cli/src/commands/jira.ts
@@ -28,7 +28,7 @@ import {
   ResourceExistsException,
   SecretsManagerClient,
 } from '@aws-sdk/client-secrets-manager';
-import { DynamoDBDocumentClient, PutCommand } from '@aws-sdk/lib-dynamodb';
+import { DynamoDBDocumentClient, GetCommand, PutCommand } from '@aws-sdk/lib-dynamodb';
 import { Command } from 'commander';
 import { ApiClient } from '../api-client';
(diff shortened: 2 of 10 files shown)


krokoko:
Thanks for this — it cleanly completes the missing half of the Jira identity-linking flow (#553) and lines up well with the existing `jira link` handler and the Linear `invite-user` pattern. The user-resolution logic, token refresh, test coverage, and doc updates all look solid. CI is green from what I can see.

A few small items I'd appreciate addressed before merge:

1. **Deduplicate `generateInviteCode`** — `linear.ts` already has an identical implementation (alphabet, `link-` prefix, `crypto.getRandomValues`). Could you extract this to a shared module (e.g. `cli/src/invite-code.ts`) and import it from both `linear.ts` and `jira.ts`? That avoids the `require('crypto')` + eslint-disable in the Jira copy.

2. **Move `parseStoredJiraOauthToken` to `jira-oauth.ts`** — The ~30 lines of secret validation would fit naturally alongside the other OAuth helpers and keeps `jira.ts` focused on command wiring.

3. **Warn when the user is already linked** — Before writing a `pending#` row, a quick `GetCommand` on `<cloudId>#<accountId>` with `status: active` would let the admin know the teammate is already mapped. Linear doesn't do this either, so a warning (not a hard error) is fine.

4. **Update `cli/README.md`** — The Jira setup guide is updated, but `cli/README.md` still doesn't mention `bgagent jira invite-user`. A short entry alongside the other Jira subcommands would help discoverability.

5. **A few more test cases for the uncovered branches** — Codecov flagged ~40 missing lines. Tests for inactive/app accounts, malformed OAuth secret JSON, and missing stack outputs would close the gap without much effort.

None of these are architectural concerns — the core flow, DynamoDB row shape, and handler contract all look correct. Happy to re-review once the above are in.

theagenticguy (commented):
**Recommendation: approve with minor nits.** I reviewed this source-grounded against the head commit (`08a5d81`), traced the write/read contract end to end, and ran the suite. The `invite-user` → `link` handshake is correct, well-tested, and safe to merge. Findings below are all low-severity; none block.

## Correctness — the write/read contract holds

`invite-user` writes the pending row and `jira-link.ts` (the server-side handler behind `client.jiraLink`) reads it. I checked field-for-field:

- `invite-user` writes `jira_identity: pending#<code>`, `status: 'pending'`, `jira_cloud_id`, `jira_site_url`, `jira_account_id`, `jira_user_name`, `jira_user_email` (`cli/src/commands/jira.ts:868-880`).
- `jira-link.ts:72-91` requires exactly `status === 'pending'` and reads exactly those fields for both the dry-run preview and the active-row write. No field the handler needs is missing, and the code is case-sensitive on both sides (`jira-link.ts:63-65`), matching the lowercase `link-<8>` codes this command mints.
- TTL: the row's `ttl` epoch-seconds field (`:869`) matches the table's `timeToLiveAttribute: 'ttl'` (`cdk/src/constructs/jira-user-mapping-table.ts:80`), so the 24h expiry the docs promise is real.
- `read:jira-user` is already in `JIRA_OAUTH_SCOPES` (`cli/src/jira-oauth.ts:42`), so the `/user` and `/user/search` calls authorize against tokens minted by `setup`. And `setup` stores all seven fields `parseStoredJiraOauthToken` requires, so the strict validation won't false-reject a well-formed tenant.

User resolution is genuinely careful: accountId-vs-email branching, a 404 fallback from direct lookup to search, inactive/app-account rejection, exact-email (case-insensitive) preference, and an explicit ambiguous-match error that lists candidate accountIds. The docs correctly warn that Atlassian may hide `emailAddress` — the single-result and ambiguous paths both handle that gracefully.

## Security — clean

- No access token, refresh token, or client secret is logged anywhere in the `invite-user` path (only status markers and the resolved display label).
- `encodeURIComponent(cloudId)` in the path plus `searchParams.set(...)` for the query prevent injection into the Atlassian REST URLs.
- The refresh-then-`PutSecretValue` path is the trusted-admin-CLI pattern (same as `setup`), and it spreads `...stored` so tenant fields like `webhook_signing_secret` survive the rewrite — there's a test asserting exactly that.

## Verification

- `npx jest test/commands/jira.test.ts` → 49/49 pass, including the 9 new `invite-user` cases (email resolve, direct accountId, 404→search fallback, ambiguous reject, expired-token refresh+persist, not-authenticated, inactive-tenant, empty-secret, no-rotated-refresh-token).
- `npx eslint src/commands/jira.ts` → clean.
- Both doc files (`docs/guides/JIRA_SETUP_GUIDE.md` and the Starlight mirror `docs/src/content/docs/using/Jira-setup-guide.md`) received line-identical edits, so the mirror is in sync.

## Nits (non-blocking)

1. **Modulo bias in `generateInviteCode`** (`cli/src/commands/jira.ts:264-272`). `b % 31` over a 256-value byte favors alphabet indices 0–7 (probability 9/256 vs 8/256), so codes lean slightly toward `a`–`h`. Entropy is still ~39.6 bits and codes are admin-issued with a 24h TTL, so this isn't a real attack surface. If you want it unbiased, `crypto.randomInt(0, INVITE_CODE_ALPHABET.length)` per character is rejection-sampled and removes the skew.
2. **No overwrite guard on the pending-row `PutCommand`** (`:870`). A `ConditionExpression: 'attribute_not_exists(jira_identity)'` would make a code collision fail loudly rather than silently overwrite a prior pending invite. Collision probability is negligible at this entropy, so this is defensive polish, not a fix.
3. **Inline `require('crypto')` with an eslint-disable** (`:265-266`) is the one spot that departs from the file's top-level ES-import style. A top-level `import { randomBytes } from 'crypto';` would drop the disable comment.
4. **Concurrent `invite-user` on an expiring token.** Atlassian rotates the refresh_token on each use, so two admins refreshing the same tenant within the refresh window means the second `refreshAccessToken` gets `invalid_grant` and the command errors. It's rare (admin-run, low frequency; the orchestrator usually keeps the token fresh) and re-running fixes it, but the surfaced error may read cryptically. Worth a line in the command's error handling if you've seen it in practice.

Nice work — the contract discipline (freeze the pending row, confirm-before-bind in `link`, audit via `invited_by_platform_user_id`) is exactly right.

— Bonk (posting via Laith's GitHub account, `theagenticguy`)

krokoko (approved):

Issue #400, opened 2026-06-21
feat(cdk): integ harness — ephemeral stack name + offline cdk-diff gate + non-blocking stranded-stack cleanup


scottschreckengaust:
### Component

CDK / infrastructure, Tooling / CI

### Summary

Make the integ smoke harness (`.github/workflows/integ.yml` + `cdk/test/integ/integ.task-api-smoke.ts`) **self-healing under unreliable teardown** by giving each run a unique, ephemeral CloudFormation stack name, and give the `integ` environment approver the same **pre-run change visibility** the `deploy` gate already has — via an **offline `cdk diff`** against the last-authorized `main` baseline (no AWS credentials, mirroring #383's mechanism).

Three coupled changes:

1. **Dynamic/unique stack name** — replace the single hardcoded `backgroundagent-integ` with `int-{short-sha}-{run-number}` (or equivalent run-unique id), read at synth time.
2. **Offline `cdk diff` for the integ approver** — surface what the run will create/change in the `integ` environment summary *before* the privileged role runs, diffed against `main`'s synthesized template (the `cdk-<compute_type>-out` build artifact), so an approver can see the security posture / resource delta without an AWS read role.
3. **Non-blocking stranded-stack cleanup** — best-effort teardown that does **not** fail the run (or head-of-line block the next run) on the AgentCore `DELETE_FAILED`; leave the stranded stack for the scheduled sweep (#72).

### Motivation / problem

The current harness hardcodes one stack name and serializes all runs:

```yaml
# integ.yml
concurrency:
  group: cdk-integ
  cancel-in-progress: false   # only one run at a time against the shared account
```
```ts
// cdk/test/integ/integ.task-api-smoke.ts:93
const stack = new TaskApiSmokeStack(app, 'backgroundagent-integ');
```
```yaml
# integ.yml "Ensure stack torn down"
aws cloudformation delete-stack --stack-name backgroundagent-integ || true
aws cloudformation wait stack-delete-complete --stack-name backgroundagent-integ   # NO `|| true`
```

**Three failure modes:**

[...]

PR #567, opened 2026-07-09
fix(ci): remove broken integ teardown safety net + map stack name to commit sha

### LINKED ISSUE #566: fix(ci): remove broken 'Ensure stack torn down' safety net from integ workflow
### Component

Tooling / CI

### Summary

Remove the `Ensure stack torn down` safety-net step from `.github/workflows/integ.yml` (Phase 0 scope). The step fails with `AccessDenied` on every run and marks otherwise-green integ runs as failed.

### Problem

Run [28892156232](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/actions/runs/28892156232/job/85707381514) shows the failure mode:

- The integ test itself **passed**: `SUCCESS integ.task-api-smoke-TaskApiSmoke/DefaultTest 291.541s` — integ-runner's own deploy → assert → destroy completed, including stack deletion (it drives the CDK CLI, which assumes the CDK bootstrap roles that hold CloudFormation permissions).
- The `if: always()` safety-net step then ran `aws cloudformation delete-stack` / `wait stack-delete-complete` **as the raw GitHub OIDC role** (`arn:aws:sts::465528542731:assumed-role/GitHub/GitHubActions`), which has no `cloudformation:DeleteStack` / `cloudformation:DescribeStacks` permissions. Both calls got `AccessDenied`; the un-guarded `wait` exited 255 and failed the job.

The backstop has therefore never worked — it can only ever fail, and it fails loudest on healthy runs. The gating `integ-smoke` status goes red even though the test passed and no resources were leaked.

### Proposed fix (Phase 0)

Delete the `Ensure stack torn down` step. Rationale:

- **Normal path is unaffected**: integ-runner already forces teardown on success *and* failure (`cdkCommandOptions.destroy.args.force`), with `RemovalPolicy.DESTROY` on all stateful resources.
- **The backstop cannot function** without an IAM change to the OIDC role, and granting the raw role `cloudformation:DeleteStack` widens the blast radius of the status-posting credentials.
- **The residual risk it targeted** (run cancelled / crashed mid-deploy strands `backgroundagent-integ`) is the same stranded-stack problem already owned by #400 (non-blocking cleanup) and #72 (scheduled sweep) — the durable fix belongs there, not in a per-run blocking step. A stranded Phase-0 stack is API Gateway + Cognito + 2 empty DynamoDB tables (no AgentCore, no VPC), i.e. near-zero idle cost until the sweep reclaims it.

### Acceptance criteria

- [ ] `Ensure stack torn down` step removed from `.github/workflows/integ.yml`; comments updated so the teardown story (integ-runner in-run destroy now, #400/#72 for stranded stacks) stays accurate.
- [ ] `mise run security:gh-actions` (zizmor) passes.
- [ ] No change to the local `mise //cdk:integ` path.

### Relationship to existing issues

- #400 — supersedes the backstop with non-blocking cleanup + ephemeral names (not yet approved); this issue only removes the broken blocking step.
- #72 — scheduled sweep is the async reclaim path for any stranded stack.
- #236 — Phase 0 parent.

ayushtr-aws:
Two coupled changes to `.github/workflows/integ.yml` (+ the integ smoke test), both about the integ harness's teardown/naming plumbing.

### 1. Remove the broken `Ensure stack torn down` safety net (#566)

The `if: always()` backstop ran raw `aws cloudformation delete-stack` / `wait stack-delete-complete` as the **GitHub OIDC role**, which has no `cloudformation:DeleteStack` or `DescribeStacks` permission. Both calls `AccessDenied` on every run; the un-guarded `wait` exits 255 and fails the job **even when the integ test passed and integ-runner already destroyed the stack**.

Evidence — run [28892156232](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/actions/runs/28892156232/job/85707381514): `SUCCESS integ.task-api-smoke-TaskApiSmoke/DefaultTest 291.541s` (integ-runner's own deploy → assert → destroy completed), then the safety-net step failed with `AccessDenied` on both CloudFormation calls.

integ-runner remains the teardown path — it forces `cdk destroy` on success and failure via the CDK bootstrap roles (which hold the CloudFormation permissions). Stranded-stack reclaim for cancelled/crashed runs is owned by #400 (non-blocking cleanup) and #72 (scheduled sweep).

### 2. Map the integ stack name to the commit SHA (`int-<short-sha>-<run-number>`)

Replaces the single hardcoded `backgroundagent-integ` stack name with a run-unique `int-<short-sha>-<run-number>` derived from the resolved head SHA, threaded into integ-runner via a new `INTEG_STACK_NAME` env var. The test falls back to `backgroundagent-integ-local` so the local `mise //cdk:integ` path is unchanged.

Why it pairs with #1: with the blocking safety net gone, a stranded stack from a cancelled/crashed run would otherwise sit under one fixed name and **head-of-line block every future run** (can't `CREATE` a name that already exists in `DELETE_FAILED`, and `cancel-in-progress: false` queues behind it). A run-unique name removes that failure mode (the run number also makes same-commit re-runs distinct); the `int-` prefix marks the stack ephemeral for #72's sweep. The concurrency group is now a cost/parallelism bound, not a correctness requirement.

> **Scope note:** the naming change is a partial pull-forward of #400's first ask (ephemeral run-unique names). #400 remains open for its other two asks (offline `cdk diff` for the approver, non-blocking best-effort cleanup). Flagging for the reviewer since #400 isn't yet `approved` — happy to split this back out if preferred.

### Validation

- `mise //cdk:compile` — clean
- `mise //cdk:eslint` — clean
- `zizmor .github/workflows/integ.yml` — no findings

#### Area

- [x] `cdk` — infrastructure, handlers, constructs (integ smoke test only)

#### Related

Fixes #566. Partially addresses #400 (naming). Related: #72, #236.

#### Changes

- Remove the `Ensure stack torn down` step from `integ.yml`; update comments so the teardown story stays accurate (integ-runner in-run destroy; #400/#72 for stranded stacks).
- Compute `int-<short-sha>-<run-number>` from the resolved head SHA + run number and pass it via `INTEG_STACK_NAME` to `mise //cdk:integ`.
- `integ.task-api-smoke.ts` reads `INTEG_STACK_NAME`, defaulting to `backgroundagent-integ-local` for the unchanged local path.
- Refresh the concurrency comment (run-unique names → single-concurrency is a cost bound, not collision prevention).

#### Acknowledgment

### DIFF
--- a/.github/workflows/integ.yml
+++ b/.github/workflows/integ.yml
@@ -242,7 +242,13 @@ jobs:
           set -euo pipefail
           export INTEG_STACK_NAME="int-${HEAD_SHA:0:12}-${RUN_NUMBER}"
           echo "Integ stack name: $INTEG_STACK_NAME"
-          mise //cdk:integ
+          mise //cdk:integ -- --all --require-approval never \
+            -c stackName="integ-${{ github.run_id }}-${{ github.run_attempt }}" \
+            -c github:sha="$(git rev-parse HEAD)" \
+            -c github:ref="$(git branch --show-current || echo "detached")" \
+            -c github:actor="${{ github.triggering_actor }}" \
+            -c github:repository="$(gh repo view --json nameWithOwner --jq .nameWithOwner)" \
+            -c github:clean="$([ -z "$(git status --porcelain)" ] && echo true || echo false)"
 
   # Post the final integ-smoke status back to the PR head so the check flips from
   # pending to success/failure. Skipped for workflow_dispatch (no PR to gate).
--- a/.github/workflows/integ.yml
+++ b/.github/workflows/integ.yml
@@ -29,8 +29,10 @@ on:
     types: [completed]
   workflow_dispatch: {}
 
-# Only one integ run at a time against the shared account — overlapping deploys
-# would collide on the single hardcoded `backgroundagent-integ` stack name.
+# Only one integ run at a time against the shared account. Stack names are now
+# run-unique (`int-<sha>`, see the integ job) so this is no longer a correctness
+# requirement — it bounds shared-account cost/parallelism rather than preventing
+# a name collision.
 concurrency:
   group: cdk-integ
   cancel-in-progress: false
@@ -221,10 +223,23 @@ jobs:
       # the CDK bootstrap roles, which hold the CloudFormation permissions). No
       # raw-CLI safety net here: the GitHub OIDC role has no
       # cloudformation:DeleteStack/DescribeStacks, so such a step can only fail
-      # (#566). If a cancelled/crashed run strands `backgroundagent-integ`,
-      # reclaim is the stranded-stack cleanup owned by #400/#72.
+      # (#566). If a cancelled/crashed run strands the stack, reclaim is the
+      # stranded-stack cleanup owned by #400/#72 — and the run-unique name below
+      # means a stranded stack never blocks a later run.
+      #
+      # INTEG_STACK_NAME maps the stack to the commit under test (`int-<sha>`),
+      # replacing the single fixed `backgroundagent-integ` name. The `int-`
+      # prefix marks it ephemeral for #72's sweep. HEAD_SHA is the resolved PR
+      # head (or dispatched ref); the substring is already lowercase-hex so it
+      # needs no further sanitizing to be CloudFormation-valid.
       - name: Run integ tests (deploy → assert → destroy)
-        run: mise //cdk:integ
+        env:
+          HEAD_SHA: ${{ needs.resolve.outputs.head_sha }}
+        run: |
+          set -euo pipefail
+          export INTEG_STACK_NAME="int-${HEAD_SHA:0:12}"
+          echo "Integ stack name: $INTEG_STACK_NAME"
+          mise //cdk:integ
 
   # Post the final integ-smoke status back to the PR head so the check flips from
   # pending to success/failure. Skipped for workflow_dispatch (no PR to gate).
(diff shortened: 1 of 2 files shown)


.github/workflows/integ.yml
```diff
@@ -217,30 +219,30 @@ jobs:
...
+          RUN_NUMBER: ${{ github.run_number }}
         run: |
           set -euo pipefail
-          aws cloudformation delete-stack --stack-name backgroundagent-integ || true
-          # No `|| true` on the wait: a DELETE_FAILED must surface loudly so we
-          # never silently leak billable resources in the shared account.
-          aws cloudformation wait stack-delete-complete --stack-name backgroundagent-integ
+          export INTEG_STACK_NAME="int-${HEAD_SHA:0:12}-${RUN_NUMBER}"
+          echo "Integ stack name: $INTEG_STACK_NAME"
+          mise //cdk:integ
```
scottschreckengaust: Add the `stackName` (at minimum)
```suggestion
          mise //cdk:integ -- --all --require-approval never \
            -c stackName="integ-${{ github.run_id }}-${{ github.run_attempt }}" \
            -c github:sha="$(git rev-parse HEAD)" \
            -c github:ref="$(git branch --show-current || echo "detached")" \
            -c github:actor="${{ github.triggering_actor }}" \
            -c github:repository="$(gh repo view --json nameWithOwner --jq .nameWithOwner)" \
            -c github:clean="$([ -z "$(git status --porcelain)" ] && echo true || echo false)"
```

scottschreckengaust (changes requested):
Add a unique `stackName`

scottschreckengaust (approved):
LGTM

PR #579, opened 2026-07-10
feat(api): platform API keys for headless webhook management (#376)

### LINKED ISSUE #376: API: Allow webhook management with platform API keys (skip Cognito for automation)
### Component

API or orchestration

### Describe the feature

Add **platform API keys** as a first-class authentication mechanism so operators and automation can manage webhooks (and optionally other REST endpoints) **without Cognito user accounts**.

Today the split is:

- **CLI / REST API** (`POST /v1/webhooks`, `GET /v1/webhooks`, task CRUD, etc.) — requires a Cognito JWT (`bgagent login` or manual `initiate-auth`).
- **Webhook task submission** (`POST /v1/webhooks/tasks`) — already uses HMAC-SHA256 with per-webhook secrets; no Cognito on the wire.

The gap: **creating and managing webhook integrations still requires Cognito**. CI pipelines, internal tools, and workshop participants who only want "fire a task from GitHub Actions" must first stand up Cognito users, configure the CLI, and obtain JWTs — even though webhook *invocation* was designed to be Cognito-free.

### Use case

- **CI/CD and GitHub Actions** want to provision a webhook (`bgagent webhook create` or REST equivalent) from a pipeline without storing Cognito username/password or refreshing JWTs.
- **Workshop / eval scenarios** where participants should submit tasks via webhooks quickly, without the full Cognito onboarding path documented in [Quick Start](docs/guides/QUICK_START.md) Steps 5–6.
- **Service-to-service integrations** (internal task routers, bots) that prefer long-lived API keys over interactive Cognito login and token refresh.
- **Operators** who already use webhook HMAC for task creation but hit friction when the *setup* step still depends on human Cognito credentials.

I'm always frustrated when I have to create a Cognito user and run `bgagent login` just to mint a webhook secret for a headless integration.

### Proposed solution

**Phase 1 — API keys for webhook management**

1. Introduce a `PlatformApiKey` (or similar) resource stored in DynamoDB with:
   - `key_id`, hashed `secret`, `user_id` (owner / task attribution), `scopes` (e.g. `webhooks:manage`, `webhooks:invoke`, `tasks:read`), `status`, `created_at`, `expires_at` (optional).
2. New authorizer path on API Gateway (or extend existing) accepting `Authorization: ApiKey <key>` or `X-API-Key` header alongside the existing Cognito authorizer.
3. Scope `POST/GET/DELETE /v1/webhooks*` to accept either Cognito JWT **or** an API key with `webhooks:manage`.
4. CLI: `bgagent api-key create|list|revoke` (admin or self-service, TBD) — no `login` required when using `--api-key` flag or env `BGAGENT_API_KEY`.
5. Document the headless flow: deploy → create API key (one-time admin bootstrap or `bgagent admin create-api-key`) → `bgagent webhook create --api-key …` → use returned HMAC secret in CI.

**Phase 2 (optional) — broader API key access**

- Extend scoped keys to read-only task APIs (`tasks:read`, `tasks:cancel`) for ops tooling.
- Explicitly **out of scope for v1**: replacing Cognito for human CLI sessions (keep `bgagent login` as default for interactive use).

**Security notes**

- Store only hashed secrets (same pattern as webhook secrets in Secrets Manager).
- Keys inherit `user_id` for task ownership (mirrors webhook owner model in `webhook-authorizer.ts`).
- Rate-limit and audit API key usage (CloudWatch metric dimension on `key_id`).
- See [IDENTITY_AND_AUTH.md](docs/design/IDENTITY_AND_AUTH.md) and [SECURITY.md](docs/design/SECURITY.md) for alignment.

### Other information

- Related: [#17](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/17) (post-deploy setup UX — API keys reduce Cognito steps for automation).
- Current webhook auth docs: [Webhook integration](docs/guides/USER_GUIDE.md#webhooks), `cdk/src/handlers/webhook-authorizer.ts`, `cdk/src/constructs/task-api.ts` (Cognito-gated `POST /webhooks`).
- Linear/Slack/Jira setup wizards today assume Cognito for the operator CLI session; API keys would not replace those OAuth flows.

### Acknowledgements

ClintEastman02:
## Summary

Closes #376.

Adds **platform API keys** as a first-class auth mechanism so CI/automation can create and manage webhooks **without a Cognito user**. Webhook *invocation* was already Cognito-free (HMAC), but webhook *setup* still required a Cognito JWT — this closes that gap for headless flows (GitHub Actions, workshops, service-to-service).

## What changed

- **Unified authorizer** — a REQUEST authorizer on `/v1/webhooks*` accepts **either** a Cognito JWT **or** a `webhooks:manage` API key. Existing Cognito flows are unchanged.
- **New `/v1/api-keys` endpoints** (Cognito-gated) to mint / list / revoke keys.
- **Key format** — `bgak_<key_id>_<secret>`. Only the **secret hash** is stored in DynamoDB (`ApiKeyTable`); lookup is a direct `GetItem` by `key_id`, so revocation has no GSI staleness window.
- **CLI** — `bgagent api-key create|list|revoke`, plus `--api-key` flag / `BGAGENT_API_KEY` env injection so headless callers skip `bgagent login`.
- **Docs** — IDENTITY_AND_AUTH.md and SECURITY.md updated (+ synced Starlight mirrors).

## Scope

Phase 1 only (webhook management via keys). Broader `tasks:read`/`tasks:cancel` scoping and replacing Cognito for interactive CLI are explicitly out of scope for v1, per the issue.

## Testing

- `mise //cdk:test` — 2261 pass (incl. new `api-key-table`, `api-key-authorizer`, `create/list/delete-api-key`, `shared/api-key`, `gateway` tests)
- `mise //cli:test` — 575 pass (incl. `api-key` command + `api-client` tests)
- `mise //docs:sync` — no drift
- `scripts/check-types-sync.ts` — CDK ↔ CLI type sync OK

## Notes for reviewers

- **Draft** because upstream `main` has advanced ~40 commits since this branched; two files conflict on rebase and need resolution before this is review-ready:
  - `cdk/package.json` (dependency block)
  - `cli/src/format.ts`
  - Everything else auto-merges cleanly.
- **Governance:** #376 does not yet carry the `approved` label and has no "Starting implementation" comment (ADR-003). Flagging so an admin can confirm approval before this leaves draft.


isadeks (commented):
## Review — Platform API keys for headless webhook management (#376)

Reviewed by reading + running the code at the merge head (`9f2a5e9`): the 7 new security suites pass (80 tests), and I verified the auth-critical paths empirically. **The security core is sound** — this is a careful design. Findings are minor + process, no blocking security defect. (Draft, so treating this as a pre-merge pass.)

### Security core — checked, sound
- **Key hashing** — a 256-bit `crypto.randomBytes` secret, stored as a plain SHA-256 hex digest. Correct for a high-entropy random token: there's nothing to brute-force, so bcrypt/argon aren't needed (and would only slow the hot auth path). Plaintext is returned once; only the hash is persisted (test asserts the secret never lands in the Put item).
- **Constant-time compare** — `timingSafeHashEqual` length-guards then `crypto.timingSafeEqual`; both operands are fixed-length 64-hex, so the length branch isn't a timing signal. Good.
- **Custom authorizer doesn't downgrade the native one** — it uses `aws-jwt-verify` with `clientId` + `tokenUse: 'id'`, which validates signature/exp/iss/aud/token_use per request (the CLI sends `id_token`, so the token type matches). The native `CognitoUserPoolsAuthorizer` it replaces on `/webhooks*` had no scope checks to lose. The HMAC invoke route (`/webhooks/tasks`) is untouched (`webhookAuthOptions`), and `requestValidator` is body-only.
- **Revocation is immediate** — strongly-consistent `GetItem` by `key_id` (no GSI window), and `status !== 'active'` → Deny. Ownership on delete returns 404 not 403 (no existence oracle), matching the webhook pattern.
- **`parseApiKey` 3-part split is safe** — `key_id` is a ULID (Crockford base32, no `_`) and the secret is hex, so `bgak_<id>_<secret>` always splits cleanly; malformed input denies without a lookup. `X-API-Key` preferred over `Authorization`; unexpected errors fail closed to Deny (tested).
- **TTL/expiry** — active keys carry no `ttl` (expire only via `expires_at`); revoked keys get a 30d `ttl` and are denied by status before deletion, and by absence after. No auth gap in the window.
- **Verbose-log redaction** — `key` + `x-api-key` added to `SENSITIVE_FIELDS`. Good catch by the author.

### Minor

- [ ] **1. `list-api-keys` default page can return short/empty results with a `next_token` (pagination UX)** — `list-api-keys.ts:72` sends `Limit` **and** `FilterExpression` together. DynamoDB applies `Limit` to *scanned* items before the filter, so a user with revoked keys can get a page of <20 (or 0) active keys while `has_more` is true — a confusing "empty but paginated" result. Not a security or correctness bug (the filter is still applied; nothing leaks). Consider over-fetching then trimming to `limit`, or documenting that callers must follow `next_token` until it's null.
- [ ] **2. `key_hash` is read then stripped, not projected out** — `list-api-keys` queries with `ProjectionType.ALL` (GSI) and relies on `toApiKeyDetail` to drop `key_hash`/`user_id`. Correct today, but a future response-shape change could leak the hash. A `ProjectionExpression`/GSI projection excluding `key_hash` would be defense-in-depth. Low priority (the hash isn't the secret, but least-exposure is cheap here).

### Process (author already flagged both — confirming)

- [ ] **3. Governance gate not met.** #376 has **no `approved` label** (labels: enhancement/orchestration/cli/security/P0) and no "Starting implementation" comment — the author flagged this per ADR-003. An admin should confirm approval before this leaves draft. This is a P0 *security* surface (a new long-lived credential type), so the governance step matters more than usual.
- [ ] **4. Rebase.** The head is now a merge commit (`9f2a5e9`) and GitHub shows `MERGEABLE`, so the `cdk/package.json` / `cli/src/format.ts` conflicts the description warns about appear resolved — worth confirming the description's "Draft because…" note is now stale and flipping to ready if an approval lands.

### Design notes (non-blocking)
- Scopes vocabulary ships all four values but only `webhooks:manage` is wired; `tasks:read`/`tasks:cancel` validate but gate nothing yet (documented as Phase 2). Fine — minting keys now against a stable vocabulary is the right call.
- `api-key create/list/revoke` are Cognito-only (no `--api-key` flag), correctly — minting a credential shouldn't be possible with a credential. Consistent with the endpoints being Cognito-gated.

**Recommendation:** no blocking security issue — the crypto, authorizer, and revocation model are all correct and well-tested. Land the governance approval (#3) first; fold in #1 (or document the pagination contract) and optionally #2; confirm the rebase note is stale. Nicely scoped Phase 1.

ClintEastman02:
Thanks for the thorough pass, @isadeks — really appreciate the empirical verification of the auth paths. Pushed `2af32c9` addressing both code findings:

**#1 — `list-api-keys` pagination.** Fixed as you suggested (over-fetch then trim). The handler now over-fetches `limit + 1` and pages (capped at `MAX_QUERY_PAGES = 10`) until the post-filter page is full or the table is exhausted, so a full page never comes back short while `has_more` is true. One subtlety worth calling out: when the page is trimmed, `next_token` is built from the **last returned record's** GSI key rather than DynamoDB's `LastEvaluatedKey` — the raw cursor points past the over-fetched items we dropped and would silently skip keys. If we bail at the page cap with a scan cursor still pending, we hand that raw cursor back so the caller always resumes rather than truncating silently.

**#2 — `key_hash` in the GSI.** Took the defense-in-depth route rather than documenting: switched `UserIndex` from `ProjectionType.ALL` to `INCLUDE` projecting only the attributes the list response needs (`name, status, scopes, updated_at, expires_at, revoked_at`), excluding `key_hash`. Since the table isn't deployed yet there's no migration cost, so the secret hash is never replicated into the index at all.

Tests: 52/52 across the api-key suites, including new pagination cases (over-fetch, paging past filtered-out pages, boundary-record token, page-cap resume) and a GSI assertion that `key_hash` is not projected.

**#3 (governance) / #4 (rebase note)** — agreed, leaving those to the approval flow; flagging to an admin for the `approved` label on #376 before this leaves draft.

_Note: CI may flag pre-existing `astro` / `brace-expansion` advisories in `yarn.lock` — those came in via the `main` merge, not this change; will handle separately._

isadeks (commented):
## Re-review of `2af32c9` — both code findings addressed, verified

Checked out the current head and verified both fixes by reading the reworked code and running the suites (api-key suites **68/68 green**, CI fully green including the secrets/deps scan). Both are correctly fixed — and the pagination rework, which is subtle, holds up.

**#1 — `list-api-keys` short-page pagination — FIXED, verified.** Now over-fetches `limit + 1` and loops (capped at `MAX_QUERY_PAGES`) until the post-filter page is full or the table is exhausted, so a page never comes back short while `has_more` is true. The subtle part is the cursor, and it's right:
- When trimmed, `next_token` is built from `startKeyForRecord(lastReturnedRecord)` — **not** DynamoDB's `LastEvaluatedKey`, which points past the over-fetched items we dropped and would skip keys. I confirmed `startKeyForRecord` reconstructs exactly the three keys a GSI `ExclusiveStartKey` requires (base PK `key_id` + GSI PK `user_id` + GSI SK `created_at`), and the boundary-cursor test proves the resume returns the dropped item rather than skipping it (`limit=2`, return [k3,k2], token=k2 → next query yields k1).
- Page-cap resume correctly hands back the raw scan cursor so the caller continues instead of being told, falsely, the list is complete.
- Nice catch flagging the LastEvaluatedKey-vs-boundary subtlety in your own reply — that's exactly the trap here.

**#2 — `key_hash` in the GSI — FIXED (better than I suggested), verified.** You took the defense-in-depth route: `UserIndex` switched `ProjectionType.ALL → INCLUDE` projecting only `[name, status, scopes, updated_at, expires_at, revoked_at]`, so `key_hash` is never replicated into the index at all (not just stripped in the response). A construct test asserts `key_hash` ∉ projection. Verified this is compatible with the #1 cursor fix: `startKeyForRecord` needs `key_id`/`user_id`/`created_at`, all of which are GSI key or base-table key attributes (always projected) — so the narrower INCLUDE list doesn't break the resume cursor. Good that these two fixes were reconciled.

**#3 (governance) / #4 (rebase)** — correctly left to the approval flow; #376 still needs the `approved` label before this leaves draft (it's `state=OPEN, draft=false` now, so worth confirming the label lands). The `astro`/`brace-expansion` advisories you flagged as pre-existing are resolved — the secrets/deps scan is green on this head.

Everything from my earlier review is genuinely addressed, and the auth core (hashing, constant-time compare, custom-authorizer parity, immediate revocation) was already sound. LGTM on the code — the only remaining gate is the governance approval on #376.

isadeks (approved):

PR #594, opened 2026-07-13
fix(workflows): repo-bound task without workflow_ref → coding/new-task-v1, not default/agent-v1

isadeks:
## Summary

Fixes an upstream regression from #296: a repo-bound task submitted **without an explicit `workflow_ref`** resolves to the repo-less `default/agent-v1` workflow instead of the disciplined `coding/new-task-v1`. That path has no git/PR discipline — the agent improvises `gh api`/`gh pr create` against an empty local clone, so `ensure_pr` finds no commits and records `pr_url=null`.

## The bug

#296 replaced the old `task_type`→workflow mapping with a resolution ladder in `resolveWorkflowRef`, but left the repo-aware rung ("Phase 4") unwired — it's still literally commented `// (Blueprint default — Phase 4, not yet wired)`. So **every task with no explicit `workflow_ref`** falls through to the platform default `default/agent-v1`:

- **All Linear tasks** — the Linear webhook processor calls `createTaskCore` with `repo` set but never sets `workflow_ref`.
- **All Slack tasks** and **all #247 orchestration children** — same (neither sets a ref).

Symptoms: `pr_url=null` (no PR-link comment back to the issue/channel), screenshot→notification routing loses its branch signal, and dependent-child branch stacking breaks (children can't fetch an unpushed predecessor).

## The fix

`resolveWorkflowRef(ref, hasRepo)` — an absent ref **with a repo present** now resolves to `coding/new-task-v1` (edit locally → commit → push → platform opens the PR via `ensure_pr`), matching pre-#296 behavior. Repo-less tasks still default to `default/agent-v1`. An explicit `workflow_ref` always wins. The single `createTaskCore` call site passes `Boolean(body.repo)`, so Slack + orchestration inherit the fix.

## Testing

- `cdk` compiles clean; the two affected suites pass (`workflows.test.ts` + `create-task-core.test.ts`, 84 tests). The old test that asserted `default/agent-v1` for a repo task encoded the regression and was updated.
- Live-verified on a deployed dev stack: a Linear task on a repo now completes and posts a PR-link comment back to the issue.

## Notes for reviewers

- **Independent of the ECS-substrate work** I'm bringing back separately — this touches only `workflows.ts` + `create-task-core.ts`; no compute-backend coupling.
- This is an upstream regression worth its own tracking issue (#296 follow-up); happy to file one if maintainers prefer.

PR #596, opened 2026-07-13
feat(ecs): make the ECS Fargate compute backend usable (context-gated wiring + ECS-parity fixes)

isadeks:
> **Stacked on #503** (`fix/502-ecs-payload-s3-pointer`). Review/merge #503 first; this PR then retargets to `main`. The diff shown here is against #503's branch, so it excludes the payload→S3 change.

## Summary

Makes the ECS Fargate compute backend actually usable. `main` ships the ECS `ComputeStrategy` scaffolding (#8, #494) but the wiring in `agent.ts` is commented out and several correctness gaps remain. This PR turns it on **context-gated** (default synth is unchanged — AgentCore stays the default) and fixes the ECS-parity bugs found dogfooding ABCA-on-ABCA and deploying to a dev stack.

## What changed

**Wiring (additive, context-gated):**
- `agent.ts` — uncomment + wire `EcsAgentCluster` (+ the #502 payload bucket from the base PR). Gated on `--context compute_type=ecs`; the default synth produces **0 ECS resources** (verified), so `main`'s template is byte-unchanged for existing deployers. Repos opt in per-repo via `compute_type: 'ecs'`.
- Task def sized for ABCA's CI-parity builds: 64 GB→**120 GB / 16 vCPU** (max Fargate) after an OOM on the baseline build; `BUILD_VERIFY_TIMEOUT_S=3600`.
- `compute_type=ecs` **guard** (`repo.ts`): the CLI refuses to onboard a repo as `ecs` when the stack wasn't deployed with the substrate, instead of failing every task later.

**ECS-parity fixes (the boot path silently dropped data vs AgentCore):**
- **ABCA-487** — the ECS boot command hand-listed a subset of `run_task` kwargs and dropped `channel_source`/`channel_metadata` (no Linear/Jira reaction or channel MCP on ECS), plus `cedar_policies`, `attachments`, `trace`, `user_id`. Replaced with `run_task_from_payload()` that maps the whole payload via `inspect.signature` (unknown keys ignored) — single source of truth, unit-tested.
- **ABCA-488** — task role granted `GetSecretValue` on `bgagent-{linear,jira}-oauth-*` so channel OAuth tokens resolve on ECS.
- **F-2** — task role granted `bedrock-agentcore:CreateEvent` on AgentMemory so cross-task learning writes don't silently no-op on ECS.
- `ARTIFACTS_BUCKET_NAME` wired into the task; `ec2:DescribeAvailabilityZones` granted (a CDK target repo's `cdk synth` build gate needs it on a fresh clone); ecs-strategy import made hermetic vs ambient env.
- `corepack` enabled in the agent image so yarn/pnpm resolve.

**Docs:** corrected the ECS enable instructions — `ComputeTypes` is a CFN template parameter, and `cdk bootstrap` can't set one (no `--parameters`; `--context` is CDK construct context), so the working path is `aws cloudformation update-stack ... ParameterKey=ComputeTypes,ParameterValue=agentcore,ecs`.

## Testing

- Full `mise run build` green: CDK **125 suites**, agent **80.84%** coverage, CLI 48 suites.
- **Proven additive:** `--context compute_type=ecs` synths the ECS cluster + payload bucket; default synth = 0 ECS resources.
- **Live-verified on a dev stack** (`--context compute_type=ecs`): a Linear task on an `ecs`-onboarded repo reacts 👀, runs on Fargate, and posts a PR-link comment; the deployed image was confirmed to carry the `run_task_from_payload` + corepack + arm64-native-claude fixes.

## Notes for reviewers

- Default behavior is unchanged; ECS is entirely opt-in (deploy context + per-repo `compute_type`). No AgentCore path touched.
- Follows the #503 payload→S3 fix; the two together make a real hydrated payload deployable on ECS (8 KB `containerOverrides` limit → S3 pointer).


scottschreckengaust (changes requested):
## Review — Principal Architect (Stack A: ECS substrate, PR 2/4 · stack tip under review)

*Bases verified against fresh `origin` PR heads (this PR diffed vs #503). Bootstrap ADR-002 coverage independently verified. All mandatory agents run (table below).*

### Verdict: Request changes

The heart of the stack and strong work — the `compute_type=ecs` gate keeps the default synth byte-for-byte unchanged, `run_task_from_payload` binds to `run_task`'s real signature via `inspect.signature` (a genuine fix for the ABCA-487 silent-drop, well-tested), the CLI onboard guard reads the new `ComputeSubstrate` output to fail at config-time instead of per-task, and the parity IAM grants are least-privilege and individually tested. Two items block: a dead-code defect that breaks the required build, and the ADR-003 governance gap.

**Bootstrap ADR-002 — verified no AccessDenied gap.** When deploying `--context compute_type=ecs`, every newly-synthesized resource type/ARN is covered: the new payload bucket auto-names to `backgroundagent-dev-*` (covered by `observability.ts` `S3ApplicationBuckets`, same as the existing attachments/trace buckets), the `Custom::S3AutoDeleteObjects` Lambda reuses the singleton provider that already deploys today, and the ECS create-actions (`ecs:CreateCluster`, `ecs:RegisterTaskDefinition`) + `iam:PassRole` for `ecs-tasks.amazonaws.com` are in `compute-ecs.ts`/`infrastructure.ts`. `BOOTSTRAP_VERSION` correctly unchanged (no new *granted* action — flipping the gate exercises grants that already existed).

### Blocking issues

**B1 — Dead-code ratchet defect breaks `mise run build`.** `cdk/src/constructs/ecs-payload-bucket.ts` — `export const ECS_PAYLOAD_OBJECT_KEY_PREFIX = ''` is unused (verified: only its own declaration; `ecsPayloadKey()` builds the key independently). Under root `knip.json` (`project: ["src/**/*.ts!"]`) it raises the knip count 78→79 and `scripts/check-deadcode-ratchet.mjs` `process.exit(1)`s above baseline.
- Severity nuance: the CI "Dead-code detection" job is `continue-on-error` / **advisory** — it will *not* fail the PR check (currently green). But the same ratchet runs inside **`mise run build`**, which this repo mandates green before commit — so it breaks the required local build. (Introduced in #503; reachable once building.)
- **Fix:** delete the constant + its doc block; fold the key-layout note into `ecsPayloadKey()`'s docstring.

**B2 — Governance (ADR-003): no approved issue; non-conforming branch.** Branch `pr/ecs-substrate-hardening` carries no issue number, and the body references stacked PRs plus **#494 (a *merged PR*, not an open approved issue)**. Per ADR-003 the substrate-usability + ECS-parity work needs its own `approved` GitHub issue (a natural sibling of #502/#494).
- **Fix:** file/link an approved issue; rename the branch to `feat/<n>-…` at merge time (renaming an open PR head closes it — reasonable to defer, as accepted on #594).

- **N1 — Stale figure.** `cdk/src/stacks/agent.ts` comment says *"64 GB / 16 vCPU"* but the task is **120 GB** (`memoryLimitMiB: 122880`) — and the `EcsAgentCluster` sizing history explicitly records 64 GB was OOM-killed. Update to 120 GB or defer to the construct.
- **N2 — Comment/code mismatch.** The `bgagent-*-oauth-*` grant comment says AccessDenied "silently no-ops (WARN)", but `config.py` logs it at **ERROR**. Code is better than the comment; fix the comment.
- **N3 — Narrative.** The `ec2:DescribeAvailabilityZones` grant is grouped under "silently no-ops (WARN)" but it's actually a **loud false build-failure** fix (the `cdk synth` gate raises on AccessDenied). Grant + the Resource:* cdk-nag suppression are correct (EC2 describe has no resource scoping, read-only); just re-word.
- **N4 — Silent key-drop (defensive, but).** `run_task_from_payload` drops known-but-unaccepted orchestrator keys (`build_command`, `merge_branches` — not yet `run_task` params). Verified safe today (not emitted; `base_branch` is consumed via `hydrated_context`, matching AgentCore). A one-line WARN when dropping a *known* orchestrator key would make a future "wired one side, forgot the other" no-op visible instead of silent. Optional.

### Tests & CI

- **CI:** green (title, CodeQL ×3, dead-code *advisory*, secrets/deps/actions, `build (agentcore)`); `MERGEABLE`/`CLEAN`.
- **Coverage:** strong on leaf logic — the `run_task_from_payload` truth-table with an `inspect.signature`-based "every-forwarded-key-is-real" guard is excellent, as are the grant tests (CreateEvent scoped to the memory ARN, oauth prefix, DescribeAZs, read-only payload bucket) and the CLI onboard-gating matrix. **Three material wiring-seam gaps** (non-blocking follow-ups):
  1. *(crit 7)* `deleteEcsPayload` tested only in isolation — no test that `orchestrate-task.ts` finalize calls it **iff** `compute_type==='ecs'`, and `task-orchestrator.test.ts` asserts neither the `ECS_PAYLOAD_BUCKET` env nor `grantDelete`. A regression would spuriously DeleteObject on every AgentCore task (masked by the no-op guard).
  2. *(crit 6)* the boot-command S3-loader one-liner is asserted only as a **string** (`toContain('get_object')`); the `_uri.split("/",3)` parse on multi-slash keys is never executed. It has drifted once already — extract it into a tested `load_payload_from_uri()`.
  3. *(crit 4)* `run_task_from_payload` is always mocked; the `entrypoint` re-export the boot command imports isn't asserted (dropping it breaks the container at runtime, tests still pass).
- **Bootstrap synth-coverage:** passes, but a **latent test-blindness** worth a follow-up — `resource-action-map.ts` omits `AWS::ECS::Cluster`/`AWS::ECS::TaskDefinition` and `synth-coverage.test.ts` synths default (agentcore) context only, so the ECS path is never coverage-checked. Grants are present (deploy works); the *guard* just can't see them. Add the two map entries + an ECS-context synth case.
- Recommend a dedicated **`/security-review`** pass on this PR before merge given the new task-role permissions (I reviewed them directly + via silent-failure-hunter and found them least-privilege, but defense-in-depth is cheap here).

### Review agents run (Stage 3 — mandatory, stack-wide)

| Agent | Ran? | Outcome |
|---|---|---|
| `code-reviewer` | ✅ | 1 build-breaker (B1, dead export — severity corrected: advisory CI, breaks `mise run build`); nits N1. |
| `silent-failure-hunter` | ✅ | **No blocking silent failures.** Write-path PutObject / boot S3 fetch / `run_cmd check=True` / exit-code mapping all fail loud; the two swallowing paths are correctly backstopped. Surfaced N2, N3, N4. |
| `pr-test-analyzer` | ✅ | Coverage strong; 3 wiring-seam gaps above. |
| `general-purpose` (bootstrap ADR-002) | ✅ | No AccessDenied gap; latent map/coverage-test blindness for ECS types. |
| `type-design-analyzer` | ⏭️ omitted | No new exported types of substance (props interfaces consumed in-file). |
| `comment-analyzer` | ⏭️ omitted | Comment accuracy covered by hand + silent-failure-hunter (N2/N3). |
| `/security-review` | ⏭️ omitted-with-note | IAM/secrets touched; reviewed directly (see above) — recommend a dedicated pass before merge. |

### Human heuristics

- **Proportionality** — ✅ Sizing/gating/S3-indirection all match documented problems (8 KB cap, OOM history, silent parity gaps). Not over-built.
- **Coherence** — ✅ `run_task_from_payload` replaces a drift-prone hand-list with a signature-bound single source of truth; ECS grants mirror the AgentCore runtime's.
- **Clarity** — ⚠️ Mostly excellent "why" comments; three (N1/N2/N3) drifted from the code. Fix for the next maintainer.
- **Appropriateness** — ✅ Live-verified on a dev stack (real Fargate run, Linear reaction + PR-link comment), not just mocks.

**Bottom line:** the substrate is high-quality and the scariest part — bootstrap coverage under least-privilege — checks out. Clear **B1** (delete the dead `''` constant so `mise run build` passes) and **B2** (approved issue + conforming branch); N1–N4 and the three test-seam gaps are strongly recommended follow-ups.

scottschreckengaust:
## Correction to my earlier review — B1 severity was overstated

Following up on my `CHANGES_REQUESTED` review above. While reviewing a related stack I **actually ran the dead-code ratchet** (`mise run check:deadcode-ratchet` + `npx knip`) instead of reasoning about it, and I need to correct **B1** (the `ECS_PAYLOAD_OBJECT_KEY_PREFIX` unused export):

**What I got wrong:** I framed B1 as blocking because it "breaks the required `mise run build`." That's not accurate. The ratchet is **advisory-only right now**, on both axes:
- It is **not** part of `mise run build` — the root `[tasks.build]` depends on `//agent:quality`, `//cdk:build`, `//cli:build`, `//docs:build`, `:drift-prevention`, not `check:deadcode-ratchet`.
- The CI "Dead-code detection" job is `continue-on-error: true` (advisory).
- The task's own description says verbatim: *"Advisory in CI for now; flips to blocking once the baseline is driven to zero."*
- Confirming evidence: `origin/main` itself already sits at knip count **80 > baseline 78** — a merged trunk above baseline is only possible because the ratchet doesn't block.

**What stands:** `ECS_PAYLOAD_OBJECT_KEY_PREFIX = ''` is still genuinely unused and self-contradicting (a "prefix" whose value is `''`, referenced nowhere), so removing it is still worth doing — but as a **non-blocking cleanliness nit**, not a build/CI blocker. I'm **downgrading B1 from blocking to a nit.**

**Net effect on the verdict:** the remaining blocker is **B2 (governance — approved issue + conforming branch, ADR-003)**. If/when #604-style approved issues are attached for the substrate work and the branch is renamed, and B2 is otherwise satisfied, I'd be happy to move this to Approve — the code itself (bootstrap coverage included) checks out. The `CHANGES_REQUESTED` state stays for now on B2; apologies for the earlier mischaracterization of B1.

isadeks:
Thanks — addressed everything in `8644f9c`. Summary:

**B1 (dead export → breaks `mise run build`)** — fixed in the base PR **#503** (`ce93b62`): deleted `ECS_PAYLOAD_OBJECT_KEY_PREFIX`; `ecsPayloadKey()` remains the single source of truth for the key layout. Flows up here since #596 stacks on #503.

**N1 (stale 64 GB)** — reworded the `agent.ts` comment to defer the exact vCPU/memory to `EcsAgentCluster` (and note 64 GB was itself OOM-killed per its sizing history), so the figure can't drift again.

**N2 (WARN vs ERROR)** — reworded the memory-grant and oauth-grant comments: an AccessDenied on those paths is *logged* (memory.py classifies it as an infra failure; config.py's token resolver logs it), not "silently" no-op'd. The comments now describe the user-visible effect (no persisted learning / no 👀→✅ reaction) rather than pin a specific log level.

**N3 (DescribeAZs narrative)** — the `ec2:DescribeAvailabilityZones` comment on this branch already frames it as a *false build-gate failure* (loud), not a WARN no-op — so no change needed; calling it out so you can confirm.

**N4 (silent key-drop)** — did it: `run_task_from_payload` now emits a WARN when it drops a **known** orchestrator key (`build_command`/`merge_branches`/`base_branch`/`github_token_secret_arn`) that `run_task` doesn't accept. Expected today (consumed via `hydrated_context` / resolved platform-side), but a future "wired one side, forgot the other" gap now surfaces instead of being silent — the ABCA-487 class. Foreign keys still drop quietly. +2 tests (warns on a known dropped key; quiet on a foreign key).

**B2 (governance)** — same as the rest of the stack: I'll file/link an approved GitHub issue for the substrate-usability + ECS-parity work (sibling of #502/#494) and rename the branch to `feat/<n>-…` at merge (renaming an open PR head closes it, as you noted on #594). Want me to file that issue now, alongside #604?

**Test-seam follow-ups (the 3 crit-4-to-7 gaps + the resource-action-map ECS blindness)** — all fair; these are genuinely wiring-seam coverage rather than defects in the diff. I'd like to take them as a fast follow-up issue rather than expand this PR's scope — agreed?

Full cdk build green (2286) + full agent suite green (1194); eslint/ruff clean. Re-requesting review.

isadeks:
Filed the follow-ups:
- **#606** — governing issue for this substrate work (B2). Requesting `approved` + branch rename to `feat/606-…` before merge.
- **#608** — the 3 wiring-seam coverage gaps + resource-action-map ECS blindness, as a fast-follow (kept out of this PR's scope).

scottschreckengaust (commented):
## Follow-up review — my `CHANGES_REQUESTED` still stands (B2 governance)

Re-reviewed after the latest push. The **code is in good shape** and B1 + N1–N4 are handled; the remaining blocker is **governance (B2)**, unchanged.

### ✅ Resolved since my prior review
- **B1** (dead `ECS_PAYLOAD_OBJECT_KEY_PREFIX` export) — removed in the base PR #503 (`ce93b62`). (As I corrected earlier, this was advisory, not a build-breaker.)
- **N1** — the stale "64 GB / 16 vCPU" comment in `agent.ts` now defers sizing to `EcsAgentCluster`. ✅
- **N2** — memory/oauth grant comments reworded to "logged, non-fatal" + user-visible effect; verified accurate against `memory.py`/`config.py` error paths. ✅
- **N4** — `run_task_from_payload` now WARNs when dropping a KNOWN orchestrator key (+ 2 tests). ✅
- Sizing history, IAM5 suppression enumeration, boot-command comment, and "parity with TraceArtifactsBucket" all verified accurate.

### 🔴 Blocking — B2: governance (ADR-003)
This PR is **not covered by an approved issue on a conforming branch**:
- Issue **#606** ("make the ECS Fargate compute backend usable") is **OPEN, unlabeled (`enhancement` only — no `approved`), unassigned**. #164 (gate ECS on context) is also OPEN/unapproved.
- Branch is **`pr/ecs-substrate-hardening`** — does not follow `(feat|fix|chore|docs)/<issue-number>-short-description`.

Per ADR-003 this must land before merge: get #606 (and/or #164) the `approved` label from an admin, self-assign, comment "Starting implementation," and re-cut on a conforming branch (e.g. `feat/606-ecs-substrate`). Everything else here is mergeable-quality; this is the gate.

### 🟠 New non-blocking finding — ECS/AgentCore HITL parity gap (`task_started_at`)
`run_task_from_payload` drops `task_started_at` **silently** (not a `run_task` param, not in `_KNOWN_ORCHESTRATOR_KEYS`). On AgentCore, `server.py:612-614` exports it as `TASK_STARTED_AT`, which `hooks.py:_remaining_maxlifetime_s()` uses to **clip the Cedar HITL approval-gate maxLifetime**. The ECS boot path bypasses `server.py`, so `TASK_STARTED_AT` is never set → the clip returns `None` → **fail-open** (gate fires with the task default rather than the clipped ceiling).

- This is **fail-open by design** (`hooks.py:801-809` documents these as "optional hints"), so it's not a hard blocker — but it's a silent AgentCore↔ECS HITL divergence with no log, exactly the "wired one side, forgot the other" class this PR's own WARN mechanism targets.
- **Fix:** have the ECS strategy set `TASK_STARTED_AT` in `containerEnv` from `payload.task_started_at` (mirror `server.py`), and/or add `task_started_at` to `_KNOWN_ORCHESTRATOR_KEYS` so the drop at least WARNs. Worth doing before ECS is used in anger for HITL workflows.

### 🟡 Nits (non-blocking)
1. **Comment overclaim** — `ecs-agent-cluster.ts` `artifactsBucket` JSDoc says the bucket is "also the `--trace` upload target," but the construct injects only `ARTIFACTS_BUCKET_NAME`; the trace uploader reads `TRACE_ARTIFACTS_BUCKET_NAME` (`telemetry.py:446`), which is not set here. `deliver_artifact` works; trace upload would silently skip on ECS. Drop the clause or note the limitation.
2. **Dockerfile corepack** — `... 2>/dev/null || corepack enable`: the fallback still installs resolvable shims (no regression to the exit-127 inert-gate bug — good), but `2>/dev/null` hides the one diagnostic (prepare-failed) you'd want. Drop `2>/dev/null`.
3. **`int(value)` on `max_turns`** (`pipeline.py`) raises on a malformed payload while every other field is defensively defaulted; a `try/except → default` would match the surrounding style.

### Tests & CI
Coverage strong and behavior-first (the `inspect.signature`-based "every-forwarded-key-is-real" guard, the CreateEvent-scoped-to-memory-ARN and oauth-prefix grant tests, the CLI onboard truth-table). Three wiring-seam gaps remain (finalize→deleteEcsPayload iff ecs; boot-command S3 parse only string-asserted; default synth doesn't assert **0** ECS resources) — all captured in **#608**, good as follow-ups. #366 perf hygiene clean (full synth cached in `beforeAll`, bundling stays disabled). CI green, `MERGEABLE`.

**Note on the stack:** #596 is based on #503's branch — merge/retarget #503 (now approved) first. #503 and #599 both touch `orchestrate-task.ts` (finalize vs. start-session), so whichever lands last needs a base refresh.

isadeks:
Thanks for the re-review. **B2 (governance) is now unblocked** — issue **#606** has the `approved` label. Branch rename to `feat/606-ecs-substrate` I'll do once #503 lands and this retargets to `main` (per your stack-sequencing note — renaming an open PR head mid-stack closes it). #503 is on the merge queue.

Addressed the new finding + nits in `ca8af9a`:

- **`task_started_at` HITL parity (new finding):** added it to `_KNOWN_ORCHESTRATOR_KEYS` so the silent drop now **WARNs** — surfacing the AgentCore↔ECS HITL divergence instead of failing open silently. Fully wiring `TASK_STARTED_AT` into the ECS `containerEnv` (the other half) I've left as a tracked follow-up rather than widen this PR; the WARN makes it visible in the meantime. Want it as its own issue, or folded into #608?
- **Nit 1 (artifactsBucket JSDoc):** dropped the "`--trace` upload target" claim — the construct wires only `ARTIFACTS_BUCKET_NAME`; the trace uploader reads `TRACE_ARTIFACTS_BUCKET_NAME` (unset here), so I noted `--trace` silently skips on ECS as a separate parity gap.
- **Nit 2 (Dockerfile):** dropped `2>/dev/null` from the corepack prepare so the prepare-failed diagnostic isn't hidden (the `|| corepack enable` fallback still guarantees resolvable shims — no regression to the exit-127 inert gate).
- **Nit 3 (`int(value)`):** now defensive — a malformed `max_turns` is dropped (run_task default applies) with a WARN, matching how every other payload field defaults, instead of raising mid-boot.

+3 tests (task_started_at WARN, malformed-max_turns dropped-not-raised, valid max_turns still coerced). Full agent suite green (1195); cdk tsc + construct tests green; ruff format + eslint clean.

The 3 wiring-seam gaps stay captured in **#608** (also now `approved`).

isadeks:
Resolved the conflicts (`faaf395`). Cause: **#503 merged to main**, so GitHub auto-retargeted this PR from #503's branch to `main` — and main's ECS files are the older *commented-out* substrate (pre-`artifactsBucket`/`agentMemory`, old hand-listed boot command), which conflicted with this PR's enriched versions.

Resolution (all HEAD-wins — this PR *is* what adds the wiring):
- `agent.ts` — kept the real context-gated ECS wiring (`computeType` gate, conditional `EcsPayloadBucket`/`EcsAgentCluster`, `ComputeSubstrate` output, orchestrator `ecsConfig`/`ecsPayloadBucket`) over main's commented-out placeholder.
- `ecs-strategy.ts` — kept the `run_task_from_payload(p)` boot command over main's old hand-listed `run_task(...)` subset (the ABCA-487 silent-drop this PR fixes). Verified the resolved boot command imports + calls only `run_task_from_payload`, no leftover old call.
- `ecs-agent-cluster.ts` + the two test files — kept this PR's enriched construct (artifactsBucket/agentMemory props, updated sizing + nag-suppression justifications).

Verified by **full `mise //cdk:build`** (not just the resolve): 2332 tests + synth green — the merge is semantically clean, not just textually. `MERGEABLE` now.

Still on B2 for merge: branch rename to `feat/606-ecs-substrate` (issue #606 is `approved`) — will do that as the last step before merge.

isadeks:
@scottschreckengaust — re-requesting review. B2 is cleared: governing issue **#606** is now `approved`. B1/N1–N4, the new `task_started_at` finding, and the nits are all addressed, and the #503 conflict is resolved (`faaf395`, full `mise //cdk:build` green — `MERGEABLE`). Test-seam follow-ups tracked in `approved` **#608**.

On the branch name: renaming is genuinely disruptive here — renaming an open PR's head branch closes the PR and orphans the stacked #597/#598. Since #606 is linked + approved, could the linked issue satisfy ADR-003 (with the `feat(ecs): … (#606)` reference in the squash-merge title), rather than re-cutting the branch? Happy to do it either way — just want to avoid breaking the stack. Could you clear the `CHANGES_REQUESTED` when you get a chance?

cdk/src/constructs/ecs-agent-cluster.ts
```diff
@@ -175,6 +233,55 @@ export class EcsAgentCluster extends Construct {
       props.payloadBucket.grantRead(taskRole);
     }
 
+    // #299 ECS-parity: an artifact workflow (coding/decompose-v1) WRITES its plan
+    // to the artifacts bucket via deliver_artifact, so grant read+write (the
+    // AgentCore runtime's SessionRole/exec-role has the equivalent). Scoped to
+    // this bucket. Stays on the task role — delivery is a terminal step.
+    if (props.artifactsBucket) {
+      props.artifactsBucket.grantReadWrite(taskRole);
```
scottschreckengaust: **Blocking (B1) — least-privilege regression + inverted rationale.** This puts whole-bucket `s3:GetObject*`/`PutObject*`/`DeleteObject*` on the ECS **task role**, which runs untrusted repo code. But the actual artifact write goes through the assumed **SessionRole** (`deliverers.py:197` → `tenant_client`), which is already scoped to `artifacts/${aws:PrincipalTag/task_id}/*` (`agent-session-role.ts:204-213`) and always wired here (`agent.ts:632`; `aws_session.py` fails *closed* if the assume fails). And the AgentCore **runtime role gets no artifacts grant at all** (`agent.ts:423-426`) — so the comment's "the AgentCore runtime's SessionRole/exec-role has the equivalent … parity" is inverted. This grant is redundant in production and lets a task read/clobber other tasks' `artifacts/<other_id>/` (and `traces/`/`attachments/` on the same bucket).

```suggestion
    // #299 ECS-parity: coding/decompose-v1 delivers its plan to the artifacts
    // bucket via deliver_artifact — but the write goes through the assumed
    // SessionRole (deliverers.py -> tenant_client), scoped to
    // artifacts/${task_id}/*, exactly like the AgentCore runtime (whose task
    // role likewise has NO direct artifacts grant). So the task role needs only
    // the ARTIFACTS_BUCKET_NAME env (set above), not a bucket grant. Granting
    // whole-bucket read+write here would over-privilege the untrusted-code role
    // and break cross-task isolation.
    // (no props.artifactsBucket grant — intentional; see comment)
```
If you truly need a no-SessionRole fallback, scope it instead: `props.artifactsBucket.grantWrite(taskRole, 'artifacts/*')` (or an explicit `arnForObjects('artifacts/*')` statement with `PutObject`/`GetObject`), never `grantReadWrite` on the whole bucket. Either way, please correct the comment at :236-238.

agent/src/pipeline.py
```diff
@@ -1272,6 +1273,107 @@ def _on_trace_truncated(max_bytes: int, first_dropped: int) -> None:
...
+#: accept as a parameter. Dropping one of these is expected today (e.g.
+#: ``base_branch`` is consumed via ``hydrated_context``, not a run_task kwarg;
+#: ``github_token_secret_arn`` is resolved before this call), but a key that
+#: shows up here AND is silently dropped is exactly the "wired one side of an
+#: orchestrator→agent field, forgot the other" no-op that ABCA-487 was — so we
+#: WARN when we drop one, making a future contract gap visible instead of silent.
+#: Keys not in this set (genuinely foreign) are dropped quietly as before.
+_KNOWN_ORCHESTRATOR_KEYS = frozenset(
+    {
+        "build_command",
```
scottschreckengaust, line 1320: **Nit (N1).** `_KNOWN_ORCHESTRATOR_KEYS` lists `build_command` but not its sibling `lint_command`. Neither is emitted in the payload today, so no runtime effect — but if configurable-verify lands, `build_command` would drop with the intended WARN breadcrumb while `lint_command` drops silently, the exact asymmetry this known-key WARN exists to prevent. The test at `test_run_task_from_payload.py:240` already iterates both as the same class.
```suggestion
        "build_command",
        "lint_command",
        "merge_branches",
```

scottschreckengaust (dismissed):
## Review — Principal Architect (Stack A: ECS substrate, re-review)

*Re-reviewed against fresh `origin` PR head `c2a3f26` (now targets `main` — #503 merged). All mandatory agents re-run on the current head (table below); the one prior technical concern and the governance gate are re-verified. My earlier `CHANGES_REQUESTED` is superseded by this review.*

### Verdict: Request changes

The substrate is high-quality and the two prior blockers are cleared — **B1** (dead export) was removed in the merged base #503, and **B2 governance** is now resolved: **#606 carries the `approved` label** and its title matches this PR, so the ADR-003 gate is satisfied (the `pr/` branch name is a de-facto-waived nit — #600/#601/#602/#594/#595 all merged on `pr/*` branches last week). The `task_started_at` HITL-parity gap I flagged is now handled with a WARN + regression test (`ca8af9a0`). Nice work.

One **new blocking finding** surfaced on re-review by the `code-reviewer` agent and independently verified by reading the runtime write path: a least-privilege regression on the ECS task role (which runs untrusted repo code). It widens blast radius on the ambient role beyond what the AgentCore path grants, with a rationale comment that is factually inverted. Because [VISION.md](../../docs/design/VISION.md)'s *bounded blast radius* is a stated tenet and this trade is undocumented (and, as it turns out, unnecessary), I'm gating on it. The fix is a one-liner.

### Blocking issues

**B1 — Least-privilege regression: the ECS task role gets whole-bucket `grantReadWrite` on the shared artifacts bucket, but the actual write goes through the prefix-scoped SessionRole — so the grant is both redundant and over-broad, and the "parity with the AgentCore runtime" rationale is inverted.** `cdk/src/constructs/ecs-agent-cluster.ts:236-242`.

I traced the real delivery path:
- `agent/src/workflow/deliverers.py:194-198` writes the artifact via `tenant_client("s3", …)` → the **assumed SessionRole** credentials, *not* the ambient task role.
- `agent.ts:632` always wires `agentSessionRole` into `EcsAgentCluster`, and `aws_session.py` fails **closed** (`SESSION_SCOPING_FAILED`) if the assume fails — it never falls back to the task role.
- The SessionRole already grants exactly this write, scoped to `artifacts/${aws:PrincipalTag/task_id}/*` (`agent-session-role.ts:204-213`), and the deliverer writes `artifacts/{task_id}/result.md` (`deliverers.py:178`).
- Critically, the **AgentCore runtime role gets NO artifacts-bucket grant at all** (`agent.ts:423-426` — only github-token, logs, memory, models); it relies solely on the SessionRole. So the comment at `:236-238` ("the AgentCore runtime's SessionRole/exec-role has the equivalent" → "grant read+write … parity") is inverted: parity with AgentCore means *no* task-role grant, not a whole-bucket one.

Risk: whole-bucket `s3:GetObject*`/`s3:PutObject*`/`s3:DeleteObject*` on the role that executes untrusted repo code lets a compromised/misbehaving task read or clobber **other tasks'** artifacts under `artifacts/<other_task_id>/` (and the `traces/`/`attachments/` prefixes on the same `TraceArtifactsBucket`) — exactly the cross-task isolation the SessionRole's `task_id`-scoped policy exists to enforce. This is self-hosted single-org, so it's defense-in-depth rather than a hard SaaS tenant boundary — but it's a documented-tenet trade with no ADR and, since production delivery already works via the SessionRole, no upside.

- **Fix (preferred):** drop the task-role grant entirely; keep only the `ARTIFACTS_BUCKET_NAME` env (that env var is the genuinely-required piece — `deliverers.py:185-187` raises if unset — and it's wired correctly). Then correct the `:236-239` comment to say the SessionRole carries the write in production (matching AgentCore).
- **Fix (if a no-SessionRole fallback must work):** scope it to `arnForObjects('artifacts/*')` with `s3:PutObject`/`s3:GetObject` only, not `grantReadWrite` on the whole bucket — and still fix the comment.

Contrast with the two adjacent grants, which are correct: `payloadBucket.grantRead` is deliberately read-only *because the container runs untrusted code* (`:227-231`), and `agentMemory.grantReadWrite` is genuinely required because memory writes go through the ambient task role (`memory.py` uses plain `boto3`, no SessionRole) — that one *is* true AgentCore parity. The artifact grant is the odd one out.

- **N1 — `_KNOWN_ORCHESTRATOR_KEYS` omits `lint_command` while listing its sibling `build_command`.** `agent/src/pipeline.py:1303-1318`. Neither field is emitted in the payload today (no runtime effect), but if configurable-verify ever lands, `build_command` would drop with the intended WARN breadcrumb and `lint_command` would drop silently — the exact asymmetry this mechanism exists to prevent. Add `lint_command` alongside `build_command` (the test at `test_run_task_from_payload.py:240` already treats both as the same class).
- **N2 — `stopSession` swallows non-`InvalidParameterException`/`ResourceNotFoundException` errors at ERROR and returns as if stopped.** `cdk/src/handlers/shared/strategies/ecs-strategy.ts:297-307` (not in this diff — flagging for a follow-up). On a throttle/5xx, the Fargate task keeps running (burning the 120 GB / 16 vCPU box) while the orchestrator considers it stopped. It's logged (not silent), and best-effort-at-finalize is defensible, but consider retrying on `ThrottlingException`/`ServerException`, or at least add the justification comment the file's other swallows carry.
- **N3 — Test-seam + `resource-action-map` ECS blindness** — already captured in **#608** (`approved`). The `pr-test-analyzer` re-confirmed four gaps are still open on this head: (1) the `finalize`→`deleteEcsPayload` `iff compute_type==='ecs'` conditional (`orchestrate-task.ts:332-334`) is untested — a regression deleting on every AgentCore task is a silent no-op that no test catches; (2) the boot-command S3-loader `_uri.split("/",3)` parse is string-asserted only, never executed; (3) `run_task_from_payload`'s return value is never asserted and the `from entrypoint import run_task_from_payload` path (the one the container uses) is untested — dropping the `return` or the re-export breaks every ECS task while tests stay green; (4) the default synth doesn't assert **zero** ECS resources (the `ComputeSubstrate` output flip isn't an independent guard — it's derived from the same gate expression). Good as #608 follow-ups; none blocks.

### Documentation
- `docs/design/DEPLOYMENT_ROLES.md` edited **with** its Starlight mirror `docs/src/content/docs/architecture/Deployment-roles.md` (both in the diff) — sync boundary satisfied; CI "Fail build on mutation" is green. Roadmap not touched (no roadmap item maps to context-gated ECS enablement — acceptable).

### Tests & CI
- **CI:** all green (CodeQL ×3, dead-code *advisory*, secrets/deps/actions, `build (agentcore)`, title); `MERGEABLE`.
- **Coverage:** strong and behavior-first — the `run_task_from_payload` truth-table with the `inspect.signature`-bound "every-forwarded-key-is-real" guard, the resource-scoped IAM grant tests (CreateEvent→memory ARN, oauth prefix, DescribeAZs=`*`, read-only payload bucket), and the CLI onboard-gating matrix are all excellent. Gaps are the four wiring seams in N3 (#608).
- **Bootstrap (ADR-002):** verified — **no AccessDenied gap at deploy**. `--context compute_type=ecs` introduces `AWS::ECS::Cluster`/`::TaskDefinition`, the payload bucket, task/exec roles, `iam:PassRole→ecs-tasks.amazonaws.com`, log groups, SG — all already covered by the pre-existing bundle (`compute-ecs.ts`, `infrastructure.ts`, `observability.ts` `backgroundagent-dev-*`). `BOOTSTRAP_VERSION` correctly **unchanged** (context-gated → default synth is byte-identical, no newly-*granted* type in the default deploy). The only bootstrap issue is latent test-blindness (ECS types absent from `resource-action-map.ts` + `synth-coverage.test.ts` synths default context only) — non-blocking, tracked in #608.

| Agent | Ran? | Outcome |
|---|---|---|
| `code-reviewer` | ✅ | **1 blocking → B1** (over-broad artifacts task-role grant + inverted parity comment); N1 (`lint_command` asymmetry). |
| `silent-failure-hunter` | ✅ | **No blocking silent failures.** Boot S3 fetch fails loud (uncaught → non-zero exit → `pollSession` maps to `failed`); `deleteEcsPayload`/known-key-drop/`max_turns`-coerce all logged best-effort; CLI guard fails closed; IAM AccessDenied surfaces at WARN/ERROR. Surfaced N2 (`stopSession` orphan risk). |
| `pr-test-analyzer` | ✅ | Coverage strong; four wiring-seam gaps (N3) still open — all in #608. #366 perf clean (`context` not `postCliContext`, `beforeAll` caching). |
| `general-purpose` (bootstrap ADR-002) | ✅ | No deploy-time AccessDenied gap; `BOOTSTRAP_VERSION` correctly unchanged; latent `resource-action-map` ECS blindness (#608). |
| `type-design-analyzer` | ⏭️ omitted | No new exported types of substance (props interfaces consumed in-construct). |
| `comment-analyzer` | ⏭️ omitted | Comment accuracy covered by hand + code-reviewer/silent-failure (B1's inverted comment, N2). |
| `/security-review` | ⏭️ omitted-with-note | IAM/secrets touched; reviewed the task-role grants directly (B1 is the finding) — a dedicated pass is cheap defense-in-depth before ECS is used in anger. |

### Human heuristics
- **Proportionality** — ✅ Sizing/gating/S3-indirection match documented problems (8 KB cap, OOM history, silent parity drops); not over-built.
- **Coherence** — ⚠️ Mostly strong (`run_task_from_payload` replaces a drift-prone hand-list with a signature-bound source of truth). But B1's grant breaks the "same concept = same scoping" coherence: the payload bucket is correctly untrusted-code-read-only while the artifacts bucket on the *same role* is whole-bucket read+write.
- **Clarity** — ⚠️ The B1 comment (`ecs-agent-cluster.ts:236-238`) actively misleads — it claims AgentCore parity for a grant AgentCore doesn't have.
- **Appropriateness** — ✅ Live-verified on a dev stack (real Fargate run, Linear reaction + PR-link comment), not just mocks.

**Bottom line:** clear **B1** (drop or prefix-scope the artifacts task-role grant + fix the inverted comment — a one-liner that also improves least-privilege on the untrusted-code role) and this is ready. Governance is resolved (#606 approved), the substrate is well-built, and bootstrap coverage checks out. N1/N2/N3 are follow-ups (#608 already tracks N3).

isadeks:
Thanks for the thorough review. Addressed in `68c9735`:

**B1 (least-privilege regression — inline on ecs-agent-cluster.ts):** ✅ Removed `props.artifactsBucket.grantReadWrite(taskRole)` entirely. You're right that the write goes through the assumed SessionRole (`deliverers.py` → `tenant_client`, scoped to `artifacts/${task_id}/*`) and the AgentCore runtime task role has no direct artifacts grant — so the comment's 'parity' was inverted and the grant let a task read/clobber other tasks' `artifacts/<other_id>/`. The task role now keeps only the `ARTIFACTS_BUCKET_NAME` env. Corrected the comment and dropped the now-stale artifacts clause from the cdk-nag IAM5 reason. Test flipped: `does NOT grant the task role write on the artifacts bucket` — asserts no `s3:Put*/Delete*` on the task role at all (the read-only #502 payload `GetObject*` grant is not flagged).

**B1 (dead const):** `ECS_PAYLOAD_OBJECT_KEY_PREFIX = ''` is no longer present on the branch head (removed in the #503 revision), so the knip ratchet is clean — verified `mise run build` green.

**B2 (governance):** filed **#617** (approved) — 'Make the ECS Fargate compute backend usable', a sibling of #502/#494. Branch rename to `feat/617-…` deferred to merge time (renaming an open PR head closes it — as accepted on #594).

**N4:** added `lint_command` (sibling of `build_command`) to `_KNOWN_ORCHESTRATOR_KEYS` so a future 'wired build_command, forgot lint_command' no-op WARNs instead of dropping silently.

**N1/N2/N3:** already addressed on the branch head (the sizing comment defers to `EcsAgentCluster` and notes 64 GB was OOM-killed; the oauth comment says 'logged by config.py's token resolver, not silent'; the DescribeAZs comment frames it as a false build-gate failure, not a WARN no-op).

The three test-seam gaps (deleteEcsPayload-iff-ecs, `load_payload_from_uri` extraction, entrypoint re-export assertion) + the `resource-action-map` ECS-type blindness are strong follow-ups — I'll track them as a post-merge issue rather than expand this PR. Full cdk + agent gates green. Ready for re-review.

ayushtr-aws (dismissed):
LGTM

cdk/src/constructs/ecs-agent-cluster.ts
```diff
@@ -52,6 +53,23 @@ export interface EcsAgentClusterProps {
...
   readonly payloadBucket?: s3.IBucket;
 
+  /**
+   * Artifacts bucket for repo-bound artifact workflows (#299 coding/decompose-v1
+   * emits its plan JSON here via ``deliver_artifact``). The AgentCore runtime
+   * gets ``ARTIFACTS_BUCKET_NAME`` in its env; the ECS task needs the SAME env +
+   * read/write grant or an artifact workflow fails at delivery with
+   * "ARTIFACTS_BUCKET_NAME is not configured" (live-caught: a :decompose on an
+   * ecs-configured repo). Read/WRITE because the container DELIVERS the artifact
+   * (unlike the read-only payload bucket).
```
scottschreckengaust: N1 (Clarity, non-blocking) — this JSDoc now contradicts the code. It says the ECS task needs the "SAME env + **read/write grant**" and "**Read/WRITE because the container DELIVERS the artifact**", but the B1 remediation deliberately grants the task role **no** artifacts-bucket permission (see the explicit `// (no props.artifactsBucket grant — intentional; see comment)` block ~30 lines below). Delivery flows through the assumed `AgentSessionRole` scoped to `artifacts/${aws:PrincipalTag/task_id}/*`, matching the AgentCore runtime role (which also has no artifacts grant). This is the residual half of prior B1's "inverted parity comment" — it would mislead an IAM audit and could invite re-introducing the over-privilege B1 removed.

Suggested rewrite:
```suggestion
   * gets ``ARTIFACTS_BUCKET_NAME`` in its env; the ECS task needs the SAME env
   * (but NO bucket grant) or an artifact workflow fails at delivery with
   * "ARTIFACTS_BUCKET_NAME is not configured" (live-caught: a :decompose on an
   * ecs-configured repo). The delivery WRITE goes through the assumed per-task
   * SessionRole (scoped to artifacts/${task_id}/*), so the task role gets only
   * the env var — parity with the AgentCore runtime role, which likewise has no
   * direct artifacts grant (see the grant block below for the rationale).
```

cdk/src/stacks/agent.ts
```diff
@@ -577,38 +576,74 @@ export class AgentStack extends Stack {
...
+        githubTokenSecret,
+        memoryId: agentMemory.memory.memoryId,
+        // F-2: grant the ECS task role read+write on AgentCore Memory so the
+        // agent's cross-task learning writes succeed on ECS (parity with the
+        // runtime's agentMemory.grantReadWrite below).
+        agentMemory,
+        // #502: read-only grant so the container can fetch its payload from S3.
+        payloadBucket: ecsPayloadBucket!.bucket,
+        // #299 ECS-parity: the same bucket the runtime uses for ARTIFACTS_BUCKET_NAME —
+        // coding/decompose-v1 delivers its plan artifact here (read+write grant in
```
scottschreckengaust: N2 (Clarity, non-blocking) — "(read+write grant in the construct)" is inaccurate: the construct issues **no** artifacts-bucket grant to the task role (the B1 fix removed it; delivery uses the task_id-scoped SessionRole). Only the `ARTIFACTS_BUCKET_NAME` env is wired. Same inverted-comment class as the JSDoc in `ecs-agent-cluster.ts`. Suggest: `// wires ARTIFACTS_BUCKET_NAME env only; delivery writes go through the per-task SessionRole (no direct task-role grant — see construct).` The trailing "Without this, an ecs-repo :decompose fails at delivery" is accurate (the env var is what `deliver_artifact` requires).

agent/src/pipeline.py
```diff
@@ -1290,6 +1291,112 @@ def _on_trace_truncated(max_bytes: int, first_dropped: int) -> None:
...
+        # ``lint_command``'s sibling: neither is a run_task param today (the build/
+        # lint commands are consumed via repo config, not passed through here), but
+        # listing both makes a future "wired build_command, forgot lint_command"
+        # contract gap WARN instead of drop silently (N4).
+        "lint_command",
+        "merge_branches",
+        "base_branch",
+        "github_token_secret_arn",
```
scottschreckengaust: N3 (nit, optional) — `github_token_secret_arn` is always present in the payload and is always dropped (it's resolved via the `GITHUB_TOKEN_SECRET_ARN` env in `build_config`, never a `run_task` param). Because it's a KNOWN key with a non-None value, `run_task_from_payload` logs a WARN on **every** ECS task boot. Unlike `task_started_at`/`build_command`, this drop is 100% expected, so the WARN is pure noise that dilutes the channel meant to surface genuine future "wired one side, forgot the other" gaps. Consider removing it from `_KNOWN_ORCHESTRATOR_KEYS` so it falls through as a quiet foreign-key drop.

scottschreckengaust (commented):
## Review — Principal AWS Solutions Architect (Stack A: ECS substrate, fresh re-review of head `86ab249`)

*Reviewed against a fresh `origin/main` (fast-forwarded before diffing). All mandatory pr-review-toolkit agents re-run on this head plus a dedicated `/security-review` pass (agent table below). Bases verified: #503 (payload→S3, #502) is merged to `main`, so this diff is genuinely against `main`.*

### Verdict: Approve with nits

The prior blocking **B1** (over-broad whole-bucket `grantReadWrite` on the artifacts bucket to the untrusted-code ECS task role, with an inverted "AgentCore parity" rationale) is **RESOLVED on the code side**: the construct now grants the task role **no** artifacts-bucket permission at all — it wires only the `ARTIFACTS_BUCKET_NAME` env, and delivery flows through the per-task assumed `AgentSessionRole` scoped to `artifacts/${aws:PrincipalTag/task_id}/*` (`agent-session-role.ts:206-209`). I traced the write path (`deliverers.py:194-197` → `tenant_client` → SessionRole) and confirmed the AgentCore runtime role likewise has **no** artifacts grant (`agent.ts:423-426`), so this is now *true* least-privilege parity. A dedicated regression test asserts no `s3:Put*`/`s3:Delete*` on any task-role policy. Governance is cleared: **#606 carries `approved`** and its title matches this PR.

What remains is a pair of **stale comments that now contradict the (correct) code** — the same "inverted parity comment" class that was part of B1, in the two spots the fix commit didn't reach. These are documentation-only (no runtime/security effect) so they don't block, but they would actively mislead the next maintainer auditing the untrusted-code task role's blast radius and could invite a re-introduction of the very over-privilege B1 removed. I've left inline suggestions; strongly recommend fixing before merge.

### Vision alignment
Advances **bounded blast radius & cost** (VISION.md): ECS is fully opt-in (deploy context + per-repo `compute_type`), the default synth is byte-unchanged (0 ECS resources), and every new task-role grant is least-privilege and mirrors the AgentCore runtime. Fire-and-forget path untouched. No tenet traded; no ADR needed.

### Blocking issues
None. B1 is resolved; B2 (governance) is resolved (#606 approved).

1. **N1 (Clarity) — stale `artifactsBucket` JSDoc claims a read/write grant that no longer exists.** `ecs-agent-cluster.ts:56-63`. Inline suggestion attached. This is the residual half of B1's "inverted comment."
2. **N2 (Clarity) — `agent.ts` wiring comment repeats the false "read+write grant in the construct" claim.** `agent.ts:625`. Inline suggestion attached.
3. **N3 (nit) — `github_token_secret_arn` emits a WARN on *every* ECS task boot.** `pipeline.py:1331`. It is always present and always dropped (resolved via `GITHUB_TOKEN_SECRET_ARN` env), so the known-key WARN fires 100% of the time — diluting the WARN channel that exists to flag genuine future contract gaps (`task_started_at`, `build_command`). Consider dropping it from `_KNOWN_ORCHESTRATOR_KEYS` so it falls through as a quiet foreign-key drop. Optional.
4. **N4 (nit) — `max_turns` `int(value)` coercion accepts surprising inputs** (`int(True)==1`, JSON float `3.9`→`3`) and the comment claims it "matches how every other field is handled" though only `max_turns` swaps in a divergent default. Reachable only via a corrupt/hand-edited payload (orchestrator always emits a number), so cosmetic. `pipeline.py:1341-1349`.
5. **N5 (nit) — Dockerfile `corepack ... 2>/dev/null || corepack enable`** — the `2>/dev/null` hides the one diagnostic (prepare-failed) you'd want when yarn shims don't resolve. Consider dropping it. (Not in the current line range if already adjusted.)

### Documentation
- `docs/design/DEPLOYMENT_ROLES.md` edited **with** its Starlight mirror `docs/src/content/docs/architecture/Deployment-roles.md` (both in the diff) — sync boundary satisfied; "Fail build on mutation" is green. The corrected bootstrap instructions (`ComputeTypes` is a CFN *parameter*, not settable via `cdk bootstrap --context`) are accurate and a real improvement.
- Roadmap not touched — acceptable (no roadmap item maps to context-gated ECS enablement).
- N1/N2 above are the only doc-accuracy defects.

### Tests & CI
- **CI:** all green (CodeQL ×3, dead-code *advisory*, secrets/deps/actions, `build (agentcore)`, title); `MERGEABLE`.
- **Coverage:** strong and behavior-first — the `run_task_from_payload` truth-table with the `inspect.signature`-bound "every-forwarded-key-is-real" guard, the resource-scoped IAM grant tests (CreateEvent→memory ARN, oauth prefix, DescribeAZs=`*`, read-only payload bucket, **no** artifacts write), and the CLI onboard-gating matrix.
- **Four wiring-seam gaps remain open (all tracked in #608, `approved`)** — non-blocking: (a) `finalize`→`deleteEcsPayload` `iff compute_type==='ecs'` (`orchestrate-task.ts:338-340`) untested; (b) boot-command `_uri.split("/",3)` parse only string-asserted, never executed; (c) `from entrypoint import run_task_from_payload` re-export path (the one the container uses) untested — a dropped re-export breaks every ECS boot while suites stay green; (d) default synth doesn't assert **zero** ECS resources. (b) and (c) are the highest-value; each is a cheap unit test.
- **CDK perf hygiene (#366):** clean. The ecs-gate describe correctly uses constructor `context` (needs the gate on; different key from `aws:cdk:bundling-stacks`, so the global bundling-disable is preserved) and caches the synth in `beforeAll`. No test re-enables bundling.
- **Bootstrap (ADR-002):** no bootstrap files touched; independently verified **no AccessDenied gap** at `--context compute_type=ecs` (payload bucket → `observability.ts` `backgroundagent-dev-*`; ECS create-actions + `iam:PassRole→ecs-tasks` in `compute-ecs.ts`/`infrastructure.ts`). `BOOTSTRAP_VERSION` correctly **unchanged** (context-gated → default deploy byte-identical, no newly-*granted* type). Latent `resource-action-map.ts` ECS test-blindness is tracked in #608.

| Agent | Ran? | Outcome |
|---|---|---|
| `code-reviewer` | ✅ | B1 verified RESOLVED; 1 doc-accuracy finding (N1/N2 inverted-comment class). |
| `silent-failure-hunter` | ✅ | **No blocking silent failures.** Boot S3 fetch / `run_task_from_payload` raise → non-zero exit → `pollSession` `failed`; `aws_region` empty re-raises in `build_config`; CLI guard fails to a real backstop. Surfaced N3 (WARN noise), N4 (`max_turns` edges). `task_started_at` HITL drop is a real fail-open but deliberately WARN-surfaced + test-covered (not silent). |
| `pr-test-analyzer` | ✅ | Coverage strong; four wiring-seam gaps (all in #608); #366 perf clean. |
| `comment-analyzer` | ✅ | Confirmed N1/N2 — two comments assert a read/write grant the code deliberately omits; all other comments (memory grant, sizing history, boot command, docstring) accurate. |
| `/security-review` | ✅ | **No findings ≥ confidence 8.** New task-role grants are faithful least-privilege parity with the AgentCore runtime; the *omission* of the artifacts grant is the correct defense; SG is egress-443-only; boot one-liner does `json.loads` (no eval/pickle/command injection) and runs before untrusted repo code; expanding the payload actually *restores* the Cedar HITL gate on ECS (a hardening improvement). |
| `general-purpose` (bootstrap ADR-002) | ✅ | No deploy-time AccessDenied gap; `BOOTSTRAP_VERSION` correctly unchanged. |
| `type-design-analyzer` | ⏭️ omitted | No new exported types of substance — the changes are props-interface fields consumed in-construct, not new domain types. |

### Human heuristics
- **Proportionality** — ✅ Sizing/gating/S3-indirection match documented problems (8 KB cap, OOM history, silent parity drops); not over-built.
- **Coherence** — ✅ `run_task_from_payload` replaces a drift-prone hand-list with a signature-bound single source of truth; ECS grants now mirror the AgentCore runtime, and the payload bucket (read-only) vs artifacts bucket (no grant) scoping is finally coherent on the same role.
- **Clarity** — ⚠️ Two comments (N1/N2) now contradict the code and would mislead an IAM audit; the fix is a few words.
- **Appropriateness** — ✅ Live-verified on a dev stack (real Fargate run, Linear reaction + PR-link comment), not just mocks.

**Bottom line:** the substrate is high-quality, the prior over-privilege blocker is genuinely fixed at the code level (I traced the delivery path and confirmed true AgentCore parity), governance is cleared, and security/bootstrap check out. The only asks are the two stale-comment fixes (N1/N2) so the docs don't re-invite the B1 over-privilege; N3–N5 and the #608 test-seam gaps are follow-ups. Not gating on the doc nits — recommending they be squashed in before merge.

isadeks:
Thanks for the fresh trace of the delivery path — addressed all the nits in `fc328efe`.

**N1 (stale `artifactsBucket` JSDoc, `ecs-agent-cluster.ts`)** — fixed with your suggested wording: it now says the ECS task gets the SAME env **but NO bucket grant**, delivery WRITE goes through the assumed per-task SessionRole scoped to `artifacts/${aws:PrincipalTag/task_id}/*`, parity with the AgentCore runtime role. The grant block ~30 lines below was already corrected in the B1 commit — this was its residual half.

**N2 (inverted wiring comment, `agent.ts:625`)** — fixed: `(read+write grant in the construct)` → wires `ARTIFACTS_BUCKET_NAME` env only; delivery writes go through the per-task SessionRole (no direct task-role grant — see construct). Kept the accurate "without the env, an ecs-repo `:decompose` fails at delivery."

**N3 (WARN-on-every-boot for `github_token_secret_arn`)** — dropped it from `_KNOWN_ORCHESTRATOR_KEYS`. It's always present and always resolved via the `GITHUB_TOKEN_SECRET_ARN` env, so it now falls through as a quiet foreign-key drop; left a comment recording *why* it must not be re-added, and added a test asserting no WARN fires for it (the channel stays clean for genuine `task_started_at`/`build_command`-class gaps).

**N4 (`max_turns` coercion + misleading comment)** — the coercion now rejects a bool (`int(True)==1`) and a non-integral float (`int(3.9)==3`) with a breadcrumb; valid `int` / int-string / int-valued-float still pass. Corrected the comment (it's the one non-str coercion, not "like every other field"). Added a test for the surprising inputs. Cosmetic (orchestrator emits a real int) but cheap, so bundled.

**N5 (Dockerfile `corepack ... 2>/dev/null`)** — no-op: the current `corepack enable && corepack prepare yarn@stable --activate || corepack enable` line has no `2>/dev/null`, so nothing to change (as you suspected).

The four **#608 wiring-seam test gaps** (finalize `deleteEcsPayload` iff-ecs, boot-uri `split`, `entrypoint` re-export, zero-ECS-synth assert) remain tracked in #608 as follow-ups — not in scope for these doc/nit fixes.

Gates: cdk 2355 tests + eslint + synth green; agent 1268 tests + ruff + ty green.

ayushtr-aws (approved):

ayushtr-aws (approved):

scottschreckengaust (approved):
## Review — Principal AWS Solutions Architect (re-review of head `f9d0f75`)

*Reviewed against a fresh `origin/main`. This is a stacked PR whose base (#503, payload→S3 for the 8 KB `containerOverrides` cap) is already merged, so the diff is genuinely against `main`. I re-read every existing review thread and **independently verified the prior blocking B1 claim against the current worktree code** rather than trusting the resolution note. All mandatory review-agent scopes exercised (table below).*

### Verdict: **Approve**

The substrate is high quality, fully opt-in, and the prior over-privilege blocker is genuinely fixed at the code *and* test level. Governance is cleared. Nothing blocks merge.

### Vision alignment
Advances **bounded blast radius & cost** (VISION.md): ECS is opt-in on two independent gates (deploy `--context compute_type=ecs` *and* per-repo `compute_type: 'ecs'`), the default synth is byte-unchanged (0 ECS resources), and every new task-role grant is least-privilege and mirrors the AgentCore runtime. The fire-and-forget path is untouched; expanding the ECS boot payload actually *restores* the Cedar HITL gate on ECS (a hardening improvement). No tenet traded — no ADR required.

### Prior blocking issue (B1) — VERIFIED RESOLVED
The earlier REQUEST_CHANGES flagged an over-broad whole-bucket `grantReadWrite` on the untrusted-code ECS **task role**, with an inverted "AgentCore parity" rationale. I traced the current code:
- `cdk/src/constructs/ecs-agent-cluster.ts:239-248` — the grant is **removed**; an explicit `// (no props.artifactsBucket grant — intentional; see comment)` marker replaces it. There is no `artifactsBucket.grant*` on the task role anywhere in the construct.
- Delivery flows through the assumed per-task `AgentSessionRole` scoped to `artifacts/${aws:PrincipalTag/task_id}/*`, exactly like the AgentCore runtime (whose task role likewise has no direct artifacts grant) — so this is now *true* least-privilege parity.
- Regression test `cdk/test/constructs/ecs-agent-cluster.test.ts:560-581` builds the construct **with** an artifacts bucket and asserts **no** `s3:Put*`/`s3:Delete*` on any IAM policy in the stack — a durable guard against re-introduction.
- The residual inverted comments (prior N1/N2) are corrected: JSDoc `ecs-agent-cluster.ts:56-71` now says "NO bucket grant" and describes the SessionRole path; the wiring comment `agent.ts:624-628` now says "no direct task-role grant".

B2 (governance) is also cleared — see below.

### Blocking issues
None.

1. **N-trace (Clarity / parity gap, non-blocking, already documented).** The ECS task sets only `ARTIFACTS_BUCKET_NAME` (`ecs-agent-cluster.ts:201`) but not `TRACE_ARTIFACTS_BUCKET_NAME`, whereas the AgentCore runtime wires **both** from the *same* `traceArtifactsBucket` (`agent.ts:338` and `:341`). Consequence: `--trace` silently skips on ECS. This is honestly disclosed in the JSDoc (`ecs-agent-cluster.ts:68-71`) as a known ECS-parity gap, and since it's the identical bucket already passed in, closing it is a one-line env addition. Not gating — recommend folding into the #608 parity follow-up.
2. **#608 wiring-seam test gaps (tracked, approved).** Four seams remain uncovered — `finalize`→`deleteEcsPayload` iff `compute_type==='ecs'`; the boot `_uri.split("/",3)` parse (string-asserted, never executed); the `from entrypoint import run_task_from_payload` re-export the container actually uses; and a default-synth assertion of **zero** ECS resources. The re-export one is the highest value: a dropped re-export breaks *every* ECS boot while suites stay green. All non-blocking, all in #608.

### Documentation
- `docs/design/DEPLOYMENT_ROLES.md` edited **with** its Starlight mirror `docs/src/content/docs/architecture/Deployment-roles.md` — identical +15 change on both, so the sync boundary is satisfied and "Fail build on mutation" stays green.
- The doc correction is a real fix: it removes a wrong instruction (`cdk bootstrap --context ComputeTypes=…` does **not** set the CFN parameter — `cdk bootstrap` has no `--parameters` flag and `--context` sets construct context) and replaces it with the correct `aws cloudformation update-stack … ParameterKey=ComputeTypes` flow on `CDKToolkit`.
- All comments now match the code (prior N1/N2/N3 inverted/noisy comments resolved on this head).

### Tests & CI
- **CI:** all green (CodeQL ×3, Analyze ×3, dead-code *advisory*, secrets/deps/actions, `build (agentcore)`, title). `MERGEABLE`; `mergeStateStatus: BLOCKED` is just the pending approving review under branch protection.
- **Verified locally:** `agent/tests/test_run_task_from_payload.py` — 17 passed. The truth-table covers the `inspect.signature`-bound param filter, the `prompt→task_description`/`model_id→anthropic_model` aliases, str/int coercion (incl. the `max_turns` bool/non-integral-float rejection), and the known-key WARN breadcrumb.
- **Coverage is behavior-first:** the `run_task_from_payload` guard makes it structurally impossible to silently drop a `run_task` kwarg again (the ABCA-487 class), and the IAM tests assert resource-scoped grants (memory CreateEvent ARN, oauth prefix, DescribeAZs=`*`, read-only payload bucket, **no** artifacts write) + the CLI onboard-gating matrix.
- **CDK perf hygiene (#366):** clean — the ecs-gate synth uses constructor `context` (distinct key from `aws:cdk:bundling-stacks`, so the global bundling-disable is preserved) and caches the synth in `beforeAll`. No test re-enables bundling. (I could not run the CDK jest suite in the shared worktree — `node_modules` is not fully installed there and a CDK test helper is absent; CI's `build (agentcore)` covers it green, and I verified the assertion text directly.)
- **Bootstrap (ADR-002):** no bootstrap files touched, and `BOOTSTRAP_VERSION` is correctly **unchanged** — ECS is context-gated so the default deploy is byte-identical and grants no newly-*enabled* CFN type. No AccessDenied gap at default synth. Latent `resource-action-map.ts` ECS test-blindness is tracked in #608.

| Agent / scope | Ran? | Outcome |
|---|---|---|
| `code-reviewer` | ✅ | B1 independently re-verified RESOLVED in code + regression test; comments now accurate; wiring coherent. |
| `silent-failure-hunter` | ✅ | **No blocking silent failures.** Boot `json.loads` / `_uri.split` / `run_task_from_payload` raise → non-zero container exit → orchestrator `pollSession` marks `failed` (loud, not silent). `max_turns` + known-key drops are WARN-breadcrumbed. `task_started_at` HITL drop is a *deliberate*, WARN-surfaced, documented fail-open — not silent. |
| `type-design-analyzer` | ⏭️ omitted | No new exported domain types — only additive props-interface fields (`artifactsBucket`, `agentMemory`) consumed in-construct. |
| `comment-analyzer` | ✅ | Prior N1/N2/N3/N5 all resolved on this head; JSDoc `ecs-agent-cluster.ts:56-71` and wiring comment `agent.ts:624-628` now match the code. |
| `pr-test-analyzer` | ✅ | 17 payload tests pass locally; artifacts regression test is a real guard; four seams tracked in #608; #366 clean. |
| `/security-review` | ✅ | **No findings ≥ confidence 7.** `run_task` has an explicit signature with **no `**kwargs`**, and `run_task_from_payload` filters every key to `inspect.signature(run_task)` — a foreign payload key cannot reach `run_task`. Boot uses `json.loads` (no `eval`/`pickle`/shell). Payload is written by the trusted orchestrator role; task role has read-only. New task-role grants are faithful least-privilege parity (oauth `GetSecretValue`-only prefix, Bedrock scoped to model/inference-profile ARNs, DescribeAZs the one unavoidable `Resource:*` read, memory RW). SG egress is 443-only. `iam:PassRole` scoping to the concrete task/exec role ARNs is preserved in the orchestrator wiring. |

### Human heuristics
- **Proportionality** — ✅ Sizing (120 GB/16 vCPU with documented OOM history), the S3 payload indirection (8 KB cap), and the two-layer opt-in gate all match documented real problems; nothing over-built.
- **Coherence** — ✅ `run_task_from_payload` replaces a drift-prone hand-listed kwarg subset with a signature-bound single source of truth; ECS grants mirror the AgentCore runtime; read-only payload bucket vs no-grant artifacts bucket is now coherent on the same role.
- **Clarity** — ✅ Names communicate intent; the previously-inverted IAM comments are fixed; the one remaining `--trace`/`TRACE_ARTIFACTS_BUCKET_NAME` gap is explicitly called out in the JSDoc rather than hidden.
- **Appropriateness** — ✅ Live-verified dogfooding on a dev Fargate stack (the OOM sizing history and the yarn/corepack exit-127 fix are real live-caught findings, not mock-derived).

**Bottom line:** The prior over-privilege blocker is genuinely fixed (traced through delivery path + backed by a regression test), governance is cleared (#617 and #606 both carry `approved`, titles match, author self-assigned #617), docs + mirror are in sync and corrected, and security/bootstrap check out. Approving. The `TRACE_ARTIFACTS_BUCKET_NAME` parity gap and the #608 seams are worthwhile follow-ups, not merge blockers.

PR #598, opened 2026-07-13
test(agent): fail hung/slow tests loudly (pytest-timeout + aws_session barrier bound)

isadeks:
> **Stacked on #597** → #596 → #503. Merge those first; this PR then retargets. Diff shown is against #597's branch.

## Summary

Test-suite hygiene so a hung or slow agent test fails **loudly and fast** instead of silently burning the whole build-verify budget — surfaced when the agent pytest suite stalled >40 min on the ECS build box before the 3600s ceiling killed it.

## What changed

- **`fix(build)`: add `pytest-timeout` (120s per-test cap)** — a single test blocking on an unmocked network/subprocess/Bedrock call otherwise hangs `mise run build` until the build-verify ceiling. Now the offending test fails with a dumped traceback in 120s and the suite moves on. (Agent-side only — the jest `--maxWorkers` tuning from the same source work is deliberately **not** included; that's local build-box tuning, and CI containers have a different core/RAM profile.)
- **`fix(agent)`: pytest-timeout method `thread`→`signal`** — so a hung test is actually aborted (the thread method can't interrupt a blocked C call).
- **`fix(agent)`: bound the `aws_session` concurrency test's Barrier + joins** — the root of an intermittent ECS test hang; the test could block forever if a worker never reached the barrier.

## Testing

- Full `mise run build` green; agent suite passes with the timeout plugin active.

## Notes for reviewers

- Test-only; no product-code change. Leaves `main`'s jest config untouched by design.


scottschreckengaust (dismissed):
## Review — Principal Architect (Stack A: ECS substrate, PR 4/4 · leaf)

*Base verified against fresh `origin` PR head (diffed vs #597).*

### Verdict: Approve ✅

Exemplary test-hygiene work — the kind that pays for itself the first time CI would otherwise hang for an hour.

- **`pytest-timeout` with `timeout_method = "signal"`** — the comment nails the subtle, correct distinction: `method="signal"` (SIGALRM) actually *aborts* a syscall-blocked test in the single-threaded main-thread runner, whereas `method="thread"` only *prints* the stack and lets the build keep hanging to the 3600s ceiling (proven live on ABCA-688). Choosing signal is the real fix, not the cargo-cult default. `timeout=120` per test is a sane cap.
- **Bounded barrier/join in `test_aws_session.py`** — `Barrier(20).wait(timeout=30)` + `join(timeout=60)` fixes the *root cause* the pytest-timeout cap only symptom-treats: a bare `Barrier.wait()` blocks forever if one of 20 threads never arrives (worker reaped under memory pressure / thread-creation throttling), hanging the whole `mise run build`. A timeout raises `BrokenBarrierError` so the test fails fast, and a leftover live thread trips the `results != 20` assertion rather than blocking. Clean.

This is pure test-infrastructure (no product code), so the change *is* its own coverage — nothing further to add. `uv.lock` updated in lockstep with `pyproject.toml`.

### Governance note

Branch `pr/ecs-test-hygiene` carries no issue number (ADR-003) — same item as the rest of the stack; attach an approved issue / conforming branch before merge. Code is approve-ready.

### CI
All green (title, CodeQL ×3, dead-code advisory, secrets/deps/actions, `build (agentcore)`); `MERGEABLE`.

isadeks:
Thanks for the approve — no code changes needed here (nit-free). Governance (approved issue + conforming branch name) tracked with the rest of Stack A; will link/rename before merge.

isadeks:
Resolved the merge conflicts (was `CONFLICTING`, now `MERGEABLE`) by merging the latest `main` in (`5ef12a01`).

The conflicts were in `cdk/src/handlers/shared/error-classifier.ts` and `agent/tests/test_shell.py` — both purely because **#597 just merged to `main`** (`7ae4af34`), and this branch (stacked on #597's pre-merge state) still carried #597's older versions of those two files. This PR doesn't itself modify either file, so I took **main's reviewed/merged version** for both. The conflicts dissolve to zero net diff there.

After resolution, #598's actual delta vs `main` is exactly its intended scope — `agent/pyproject.toml` (pytest-timeout, `timeout_method=signal`), `agent/tests/test_aws_session.py` (bounded `Barrier`/`join`), and `agent/uv.lock`. Agent gate green: 1278 tests, ruff + ty clean.

ayushtr-aws (approved):

Issue #580, opened 2026-07-10
epic(jira): align Jira with the orchestration arc in #695


ayushtr-aws:
## Summary

Parent tracker for Jira adapter parity with the orchestration and iteration behavior proposed in PR #695.

PR #695 provides the shared DAG executor, additive graph extension, channel registry, Jira feedback adapter, and iteration feedback primitives. Jira still needs surface-specific entry points and feedback capabilities for several of those behaviors. The Jira work must reuse the shared contracts rather than build a second orchestration stack.

This tracker covers behavior that is present in #695 plus the Jira baseline work already completed. Auto-decomposition is deliberately excluded because `bgagent:decompose` / `bgagent:auto` were removed from #695 and remain experimental under Linear issue #299.

### Completed Jira baseline

- [x] #572 - Jira workflow/status transitions
- [x] #573 - Jira final status comments with cost, turns, and duration
- [x] #642 - Jira `bgagent` app identity for outbound actions
- [x] #576 - Standalone Jira `@bgagent` comment-triggered PR iteration
- [x] #577 - Jira attachments and recent comments in task context

### Required PR order

**Prerequisite**

- PR #695 must merge before the remaining Jira parity work lands.

**Phase 1 - Jira orchestration foundation after #695**

- [ ] #574 - Seed and execute authored Jira parent/subtask graphs. Includes orchestrated-child iteration reintegration and dependent restacking.

**Phase 2 - may proceed in parallel after #574**

- [ ] #578 - Add newly created Jira children through the shared additive extension path. Depends on #574; does not depend on #697.
- [ ] #697 - Mature one Jira iteration status comment through acknowledgement, heartbeat, and terminal outcome. Depends on completed standalone iteration issue #576 and on #574 for its orchestrated-child acceptance criteria; does not depend on #578.

Dependency graph:

```text
#695 ──> #574 ──> #578
             └──> #697
#576 (complete) ─> #697
```

## Scope boundaries

- **No decomposition:** do not implement or advertise `bgagent:decompose`, `bgagent:auto`, generated Jira children, plan approval/revision, or decomposition caps as part of this epic. Revisit only after the experimental Linear planner has a separate product decision.
- **No help-only trigger label:** Jira does not need a separate non-executing `bgagent:help` label; #575 was closed as not planned.
[...]

PR #601, opened 2026-07-14
fix(fanout): render the PR link on ✅ success, not only on ⚠️ (ABCA-584)

isadeks:
## What

On a `task_completed` (✅) event, the terminal Linear status comment omitted the PR link — it rendered `PR: …` **only** on the ⚠️ "shipped a PR but stopped early" path.

The old code assumed that on ✅ the agent's own step-2 "PR opened" comment reliably carries the link, so duplicating it in the completion comment would just be noise. That assumption **fails when the agent skips its PR-opened comment** — live-caught on **ABCA-584**, where a decompose→single task opened PR #395 (`pr_url` was on the task record) but posted no PR-opened comment, so the ✅ completion comment omitted the link and it was **lost entirely**.

## Fix

Render `pr_url` in `renderLinearFinalStatusComment` whenever the task produced one — on **both** the ✅ and ⚠️ paths (`if (args.prUrl)` instead of `if (args.prUrl && shippedDespiteFailure)`). The completion comment is the terminal, platform-owned surface and fires once per task, so it can't spam; a duplicate link is far cheaper than a missing PR.

## Test

Updated `fanout-task-events.test.ts` to assert the link renders on the ✅ path. `mise run build` green in `cdk/`: compile clean, **124 suites / 2261 tests pass**, eslint `--fix` no mutations.

PR #603, opened 2026-07-14
feat(jira): post final status comments with cost, turns, and duration (#573)

### LINKED ISSUE #573: feat(jira): post final status comments with cost, turns, and duration
## Component

CDK fanout, Jira adapter

## Describe the feature

Jira-origin tasks should get the same terminal status visibility Linear-origin tasks get today: a deterministic final comment with outcome, cost, turns, duration, task id, and PR link when relevant.

Linear already does this through `fanout-task-events.ts` after terminal task events. Jira currently relies on the agent-side `jira_reactions.py` terminal comment, which only says success/failure and PR URL. It does not include cost/turns/duration and may never fire if the agent crashes before reaching its final comment path.

## Use case

After ABCA completes a Jira task, the Jira issue should show what the run cost and how long it took without requiring the user to run `bgagent status` or inspect logs. This is especially important for failures, max-turns exits, and “PR opened but stopped early” outcomes.

## Proposed solution

Mirror the Linear final-status fanout for Jira:

- Add Jira terminal events to the fanout channel defaults.
- In the fanout dispatcher, load the task, gate on `channel_source === "jira"`, and read `jira_cloud_id` / `jira_issue_key` from `channel_metadata`.
- Render a Jira final-status comment with the same outcome framing as Linear:
  - completed
  - failed/cancelled/timed out
  - shipped a PR but stopped early
- Include cost, turns/max-turns, duration, task id, and PR URL when needed.
- Post through `cdk/src/handlers/shared/jira-feedback.ts`, reusing its OAuth refresh and retry behavior.
- Add a post-once marker on the task record, equivalent to Linear’s `linear_final_comment_event_id`, to avoid duplicate comments on stream retries.
- Avoid duplicate terminal comments by moving Jira terminal status ownership to fanout; keep the agent-side Jira start/progress behavior.

## Acceptance criteria

- Jira-origin terminal tasks receive one final Jira comment with cost, turns, duration, task id, and appropriate outcome.
- A task that opens a PR but fails/stops later includes both the PR link and the stop reason.
- Agent crashes before `jira_reactions.py` completion still produce the platform-side final status comment.
- Retryable Jira API failures cause Lambda stream retry; terminal auth/config failures are logged and do not block sibling fanout channels.
- Non-Jira tasks are unaffected.
- Normal successful Jira tasks do not produce duplicate terminal comments.

## Tests

- Fanout tests cover Jira success, failure, PR-but-stopped, missing metadata, non-Jira skip, idempotent retry, and retryable post failure.
- Jira feedback tests cover multi-line/paragraph ADF rendering if needed for the final status body.
- Agent tests verify Jira terminal comments are not duplicated after platform-side fanout owns final status.
- Docs update `JIRA_SETUP_GUIDE.md` to mention cost/turn/duration final comments.

## Out of scope

- Live progress streaming into Jira.
- Editing an existing Jira comment in place.
- Cost rollups across Jira epics/subtasks.
- Changing Linear final-status behavior.

ayushtr-aws:
## Summary

Closes #573 (part of the Jira-adapter parity work in #580, Phase 1).

Jira-origin tasks now get the same terminal status visibility Linear-origin tasks get today: a **deterministic, platform-side** final comment with outcome, cost, turns, duration, task id, and PR link — posted by the fan-out plane (`dispatchToJira`), not the agent. This fires even when the agent crashes before reaching its own final-comment path (max-turns, OOM), which the previous agent-side `jira_reactions.py` terminal comment could not.

## What changed

**Fan-out (`cdk/src/handlers/fanout-task-events.ts`)**
- New `jira` `NotificationChannel` + `CHANNEL_DEFAULTS` entry (terminal events + `task_timed_out`).
- New `dispatchToJira`, structurally mirroring `dispatchToLinear`: env guard → load task → `channel_source === 'jira'` gate → read `jira_cloud_id`/`jira_issue_key` from `channel_metadata` → post-once marker check → post → persist marker. Retryable failures escalate to the partial-batch retry path; terminal failures log-and-resolve.
- New `renderJiraFinalStatusComment` — same three-outcome framing as Linear (✅ completed / ⚠️ shipped-a-PR-but-stopped-early / ❌ failed), emitting ADF paragraphs. The PR link renders on the ✅ success path too, since demoting the agent-side terminal comment removes Jira's only other PR-link surface.

**Jira feedback (`cdk/src/handlers/shared/jira-feedback.ts`)**
- `buildAdfDocument` — multi-paragraph ADF with `strong`/`em` marks.
- `postIssueCommentAdf` — returns a classified `JiraPostResult` (`retryable` vs terminal), reusing the existing 401/403 forced-refresh-and-retry-once path. The boolean `postIssueComment` is unchanged for existing callers.

**Types / construct / stack**
- `TaskRecord.jira_final_comment_event_id` post-once marker (Jira analogue of `linear_final_comment_event_id`); `TaskNotificationsConfig` gains a `jira` channel.
- `FanOutConsumer` wires `JiraWorkspaceRegistryTable` + `bgagent-jira-oauth-*` Get/Put grant (guarded — a deployment without Jira onboarding gets no dangling grant).

**Agent (demotion)**
- Removed `comment_task_finished` and its two `pipeline.py` call sites so the fan-out plane owns the terminal comment (no double-post). The start comment (`comment_task_started`) stays for in-flight progress.

**Docs**
- `JIRA_SETUP_GUIDE.md` updated to describe the platform-side final comment (+ regenerated Starlight mirror).

## Acceptance criteria
- [x] Jira-origin terminal tasks receive one final comment with cost, turns, duration, task id, and the appropriate outcome.
- [x] A PR-but-stopped task includes both the PR link and the stop reason.
- [x] Agent crash before `jira_reactions.py` still produces the platform-side comment.
- [x] Retryable Jira failures → stream retry; terminal auth/config failures log-and-resolve without blocking sibling channels.
- [x] Non-Jira tasks unaffected; no duplicate terminal comments.

## Tests
- Fanout Jira: success, failure, shipped-but-stopped, `task_timed_out`, missing metadata, non-Jira skip, idempotent retry, retryable escalation, marker persistence/failure, missing-env WARN.
- `jira-feedback`: `buildAdfDocument` (marks + blank-line paragraphs), `postIssueCommentAdf` classification (201/5xx/429/4xx/network/token-fail/refresh paths).
- Construct: Jira env var + grant present when props set, absent otherwise.
- Agent: `comment_task_finished`-is-gone guard.

Full `mise run build` (agent quality, cdk compile+eslint+jest+synth, docs build) is green locally.

### DIFF
--- a/agent/src/jira_reactions.py
+++ b/agent/src/jira_reactions.py
@@ -1,9 +1,15 @@
 """Jira issue-comment helper for Jira-origin tasks.
 
-Posts a "starting" comment on the originating Jira issue at task start and a
-terminal "succeeded / failed (+ PR link)" comment at the end — the Jira
-analogue of ``linear_reactions`` (Linear uses emoji reactions; Jira's REST
-API has no lightweight reaction primitive, so comments are the right tool).
+Posts a "starting" comment on the originating Jira issue at task start — the
+Jira analogue of ``linear_reactions`` (Linear uses emoji reactions; Jira's
+REST API has no lightweight reaction primitive, so comments are the right
+tool).
+
+The *terminal* status comment is NOT posted from here: since issue #573 the
+deterministic fan-out plane (``cdk/src/handlers/fanout-task-events.ts``
+``dispatchToJira``) owns it, so it carries cost/turns/duration and fires even
+when this agent crashes before completing. This module only owns the start
+comment.
 
 Why a direct REST call instead of MCP: Atlassian's Remote MCP
 (``mcp.atlassian.com``) requires an interactive, browser-based OAuth 2.1
@@ -182,32 +188,13 @@ def comment_task_started(
     log("TASK", f"jira_reactions: comment_task_started issue={issue_key} ok={ok}")
 
 
-def comment_task_finished(
-    channel_source: str,
-    channel_metadata: dict[str, str] | None,
-    success: bool,
-    pr_url: str | None = None,
-) -> None:
-    """Post a terminal status comment on the Jira issue. No-op for non-Jira tasks."""
-    target = _enabled(channel_source, channel_metadata)
-    if not target:
-        return
-    cloud_id, issue_key = target
-    if success:
-        text = "✅ ABCA finished this task."
-        if pr_url:
-            text += f" Pull request: {pr_url}"
-        else:
-            text += " No pull request was opened."
-    else:
-        text = "❌ ABCA could not complete this task. Check the agent logs for details."
-        if pr_url:
-            text += f" A pull request was opened anyway: {pr_url}"
-    ok = _post_comment(cloud_id, issue_key, text)
-    log(
-        "TASK",
-        f"jira_reactions: comment_task_finished issue={issue_key} success={success} ok={ok}",
-    )
+# NOTE: there is deliberately no ``comment_task_finished`` here. Since issue
+# #573 the deterministic fan-out plane
+# (``cdk/src/handlers/fanout-task-events.ts`` ``dispatchToJira``) owns the Jira
+# terminal comment — it carries cost/turns/duration and, crucially, fires even
+# when the agent crashes before completing (max-turns, OOM). The agent only
+# posts the *start* comment (``comment_task_started`` above); posting a terminal
+# comment here too would double-comment on the issue.
 
 
 def _reset_state_for_testing() -> None:
--- a/agent/src/pipeline.py
+++ b/agent/src/pipeline.py
@@ -23,7 +23,7 @@ from config import (
     resolve_linear_api_token,
 )
 from context import assemble_prompt, fetch_github_issue
-from jira_reactions import comment_task_finished, comment_task_started
+from jira_reactions import comment_task_started
 from linear_reactions import react_task_finished, react_task_started
 from models import AgentResult, HydratedContext, RepoSetup, TaskConfig, TaskResult
 from observability import current_otel_trace_id, task_span
@@ -1106,14 +1106,14 @@ def run_task(
                 started_reaction_id=linear_eyes_reaction_id,
             )
 
-            # Terminal status comment on the Jira issue (REST shim, with the
-            # PR link when one was opened). No-op for non-Jira tasks.
-            comment_task_finished(
-                config.channel_source,
-                config.channel_metadata,
-                success=(overall_status == "success"),
-                pr_url=pr_url,
-            )
+            # NOTE: the terminal status comment on the Jira issue is NOT posted
+            # here. Since issue #573 the deterministic fan-out plane
+            # (``cdk/src/handlers/fanout-task-events.ts`` ``dispatchToJira``)
+            # owns the Jira final-status comment — it carries cost/turns/
+            # duration and, crucially, fires even if this agent crashes before
+            # reaching this point (max-turns, OOM). Posting here too would
+            # double-comment. The agent still posts the *start* comment
+            # (``comment_task_started`` above) for in-flight progress.
 
             # --trace trajectory S3 upload (design §10.1). Runs AFTER
             # post-hooks but BEFORE ``write_terminal`` so the resulting
@@ -1244,14 +1244,11 @@ def run_task(
                 success=False,
                 started_reaction_id=linear_eyes_reaction_id,
             )
-            # Best-effort failure comment on the Jira issue. No-op for
-            # non-Jira tasks; network failures are swallowed.
-            comment_task_finished(
-                config.channel_source,
-                config.channel_metadata,
-                success=False,
-                pr_url=None,
-            )
+            # NOTE: no Jira failure comment here — the fan-out plane's
+            # ``dispatchToJira`` (issue #573) owns the Jira terminal comment
+            # and fires on the platform side even when this crash path runs,
+            # so posting here would double-comment. (Contrast the Linear ❌
+            # reaction above, which the fan-out plane does not replicate.)
             raise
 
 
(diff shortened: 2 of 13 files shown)


isadeks (commented):
## Review — Jira final-status comments (#573)

Faithful, well-tested mirror of the Linear fan-out dispatcher (#239). CI is green, IAM grants are correct (registry read-only — the resolver never writes it; `GetSecretValue`+`PutSecretValue` scoped to `bgagent-jira-oauth-*`, justified by the in-place token rotation in `jira-oauth-resolver.ts`), grants are guarded so a Jira-less deploy gets no dangling IAM, and the post-once marker / retryable-vs-terminal split match Linear exactly. **Approve with the notes below** — the first is a coordination issue worth resolving before merge.

### 1. Collides with open PR #601 (ABCA-584) — the "diverges from Linear" rationale is already stale

This PR renders the PR link on the ✅ success path and frames it as a **Jira-specific** quirk (Jira lost its agent-side terminal comment; "Linear keeps its step-2 'PR opened' comment, so it omits the link on success").

But **#601 `fix(fanout): render the PR link on ✅ success, not only on ⚠️ (ABCA-584)`** is open against the same base and edits the same `renderLinearFinalStatusComment` region — because that Linear assumption **failed in production**: a decompose→single task opened a PR, the agent skipped its step-2 "PR opened" comment, and the ✅ Linear comment omitted the link, so it was lost entirely. The real lesson (the agent's step-2 comment isn't guaranteed → the platform comment must always carry the link) applies to **both** channels, not just Jira.

Net effect once #601 merges:
- The renderer comment *"This diverges from Linear (which omits it on success)"* becomes **false** — both channels render it.
- The test named `'✅ success path renders the PR link (diverges from Linear)'` documents a divergence that no longer exists.
- Whichever PR lands second hits a **semantic conflict** in the shared region (clean git-merge ≠ correct).

The behavior here is right; the **justification and merge-order** are the risk. Suggest: reframe the ✅-PR-link rationale around the general ABCA-584 reason ("the agent may skip its PR-opened comment") rather than the Jira-specific one, drop the "diverges from Linear" framing/test name, and coordinate merge order with #601.

### 2. `task_timed_out` renders a raw underscore in the ❌ header (Jira-only, user-facing)

```ts
header = `❌ Task ${args.eventType.replace(/^task_/, '')}${reason}`;
```
For `task_timed_out` this yields **"❌ Task timed_out"**. Jira is the *only* channel routing `task_timed_out` through this renderer (Linear's default doesn't subscribe to it), so this is a genuinely new case the copied-from Linear code never exercised — and the test only asserts the comment contains `❌`, so it slips through. Suggest `.replace(/^task_/, '').replace(/_/g, ' ')` → "Task timed out".

### 3. Latent Linear parity gap (follow-up, not this PR)

Jira subscribes to `task_timed_out`; Linear doesn't (its default predates the event). Fine to keep scope tight here, but worth a tracking issue — adding `task_timed_out` to `CHANNEL_DEFAULTS.linear` is a one-line parity fix.

### 4. ⚠️-header bolds more than Linear does (cosmetic)

Linear bolds only `**…stopped early — reason**` and leaves "— review and decide if more work is needed" unbolded; the Jira renderer marks the whole header run `strong`. Not wrong, just a slight cross-channel inconsistency. Ignore if intended.

### Positives worth calling out

- Runs cleanly against the **PR #79 dispatcher-review checklist** (the last time a channel was added to this file): partial-batch retry, terminal/retryable classification, conditional `UpdateItem` dedup, prop-gated secret grant with a graceful-degrade test, SDK error matched by `.name`. All present.
- The mock-boundary concern (dispatcher tests mock `postIssueCommentAdf`/`buildAdfDocument`) is covered because `jira-feedback.test.ts` exercises the real ADF-build → REST contract against `fetch`, and the `JiraFeedbackContext` shape matches what the dispatcher passes — so the seam isn't left untested, just split across two files.
- `encodeURIComponent` on both path segments, prefix-scoped least-privilege IAM, best-effort/never-gating failure handling — all correct.

**Recommendation:** ship after resolving #1 (reframe + merge-order with #601); #2 is a cheap customer-visible fix worth folding in; #3/#4 are optional follow-ups.

ayushtr-aws:
Thanks for the thorough review! Addressed in `bf583fa`:

**# 1 (reframe + #601 coordination) — done.** Dropped the "diverges from Linear" framing. The docstring + code comment now attribute the always-render-the-PR-link behavior to the general ABCA-584 reason (the agent's own "PR opened" comment isn't guaranteed to fire), and note that #601 lands the same fix for Linear. Renamed the test `'✅ success path renders the PR link (agent step-2 comment not guaranteed — ABCA-584)'` so it no longer documents a divergence that #601 removes. On **merge order**: since both PRs touch the shared renderer region, happy to let #601 land first and rebase this on top — whichever you prefer.

**# 2 (`task_timed_out` raw underscore) — fixed.** Header now does `.replace(/^task_/, '').replace(/_/g, ' ')` → **"❌ Task timed out"**. Added a dedicated test asserting the humanized header and that the body never contains `timed_out`.

**# 4 (⚠️ bold scope) — matched to Linear.** The ⚠️ header is now two runs: the reason (`⚠️ Shipped a PR but stopped early — <reason>`) is bold, the trailing `— review and decide if more work is needed` is a separate un-bolded run. Test asserts both runs.

**# 3 (Linear `task_timed_out` parity) — agreed, out of scope here.** I'll open a tracking issue for the one-line `CHANNEL_DEFAULTS.linear` addition.

Also folded in a small follow-up while here: the PR link in the Jira comment is now a **clickable hyperlink** (ADF `link` mark) rather than bare text — ADF doesn't auto-linkify a bare URL the way Linear's Markdown does. Verified live on the dev stack.

Note: `main` was merged into the branch (bringing #594); I rebased my review-response commit on top and re-ran the fanout + jira-feedback suites (157 passing).

isadeks (approved):
LGTM

PR #605, opened 2026-07-14
feat(jira): transition originating issue across workflow as tasks progress (#572)

### LINKED ISSUE #572: feat(jira): transition originating issue across workflow as tasks progress
## Component

Agent runtime, Jira adapter, CLI

## Describe the feature

When ABCA runs a Jira label-triggered task, move the originating Jira issue through its workflow so the Jira board reflects the actual task lifecycle.

Target behavior:

- Task start: transition the Jira issue to an in-progress status.
- PR opened: transition the Jira issue to a review status, defaulting to a status named `In Review` when available.
- Task failed or no PR opened: leave the Jira issue status unchanged.
- Transition failures must be best-effort only: log a warning and never fail, block, or retry the agent task.

Status resolution should use this order:

1. Per-project overrides configured on the Jira project mapping, for example:
   `bgagent jira map <cloudId> <KEY> --repo owner/repo --status-on-start "Doing" --status-on-pr "Code Review"`
2. Safe Jira workflow heuristics:
   - On start, prefer a transition whose `to.statusCategory.key` is `indeterminate`.
   - On PR opened, prefer a transition whose destination name matches `In Review` case-insensitively.
3. If no valid transition exists, skip with a warning.

## Use case

Today a Jira issue can stay in `To Do` while ABCA is actively working and even after it opens a PR. Users looking at the Jira board during standup, sprint review, or WIP triage cannot tell that the card is already in progress unless they open it and read ABCA comments.

Linear-origin tasks already update issue state as work progresses. Jira should have the same at-a-glance board signal.

## Proposed solution

Use Jira REST v3 transitions from the existing outbound Jira reaction path:

- `GET /rest/api/3/issue/{issueIdOrKey}/transitions`
- `POST /rest/api/3/issue/{issueIdOrKey}/transitions`

Implementation shape:

- Extend `agent/src/jira_reactions.py` with best-effort helpers for start and PR-opened transitions.
- Reuse the existing Jira OAuth token path, timeout behavior, logging, and 401/403 circuit breaker used by Jira comments.
- Stamp optional project mapping overrides into task `channel_metadata` in `cdk/src/handlers/jira-webhook-processor.ts`.
- Extend `bgagent jira map` with optional `--status-on-start` and `--status-on-pr` flags.
- Document the behavior and Jira permission requirement in `docs/guides/JIRA_SETUP_GUIDE.md`.

No new OAuth scopes should be required; the existing `read:jira-work` and `write:jira-work` scopes cover transition discovery and execution.

## Acceptance criteria

- A Jira-triggered task on a standard workflow transitions to an in-progress-category status shortly after ABCA starts.
- When ABCA opens a PR, the issue transitions to `In Review` if that destination is available.
- Per-project configured status names take precedence over heuristics.
- Missing transitions, required transition screens, missing Jira permissions, and Jira API failures are logged and skipped without affecting the task.
- Failed tasks and tasks with no PR leave the Jira issue status unchanged.
- Non-Jira tasks are unaffected.

## Tests

- Agent tests cover transition selection, configured overrides, `hasScreen` skip behavior, auth circuit breaker reuse, and failure isolation.
- CDK handler tests cover project mapping override metadata.
- CLI tests cover the new `bgagent jira map` flags.
- Docs are updated for setup and troubleshooting.

## Out of scope

- Triggering tasks from Jira status changes.
- Moving issues to Done on PR merge.
- Multi-hop transition pathfinding.
- Filling required transition-screen fields.

ayushtr-aws:
## What

Moves the originating Jira issue through its workflow as an ABCA task progresses, so the Jira board reflects the task lifecycle at a glance — the same signal Linear-origin tasks already give:

- **Task start** → In Progress
- **PR opened** → In Review
- **Failed / no PR** → status unchanged

Closes #572.

## How

- **`agent/src/jira_reactions.py`** — new best-effort `transition_task_started` / `transition_pr_opened` helpers: `GET /transitions` → select via the resolution ladder → `POST /transitions`. Extracted the 401/403 circuit-breaker accounting (`_note_auth_status` / `_circuit_open`) so comments and transitions share one breaker.
- **`agent/src/pipeline.py`** — invoke the helpers at the existing start hook and the success-path PR hook (only when a PR was actually opened).
- **`cdk/src/handlers/jira-webhook-processor.ts`** — stamp optional `jira_status_on_start` / `jira_status_on_pr` from the project mapping into `channel_metadata`.
- **`cli/src/commands/jira.ts`** — optional `--status-on-start` / `--status-on-pr` on `bgagent jira map`.
- **`docs/guides/JIRA_SETUP_GUIDE.md`** — Board transitions section (defaults, overrides, *Transition Issues* permission prerequisite) + troubleshooting; Starlight mirror regenerated.

### Status resolution ladder (per lifecycle point)

1. **Per-project override** — `--status-on-start` / `--status-on-pr`, matched case-insensitively against the destination name. An unreachable override is a deliberate skip (no fallback).
2. **Heuristic** — on start, destination `statusCategory == indeterminate`; on PR, destination named `In Review`.
3. **Skip with a warning** — no match, `hasScreen` transition, or empty transition list (user lacks *Transition Issues* permission).

## Safety

- Best-effort throughout: short timeout, errors logged and swallowed, shared auth circuit breaker. A transition failure never fails, blocks, or retries the task.
- **No new OAuth scopes** — existing `read:jira-work` / `write:jira-work` cover both the transitions GET and POST.
- No multi-hop pathfinding, no screen-field filling, no move-to-Done — all explicitly out of scope.

## Tests

- **Agent** (`test_jira_reactions.py`): channel gate, start category heuristic, PR name-match, per-project overrides + precedence, case-insensitive match, unreachable-override skip, `hasScreen` skip, empty-list skip, shared circuit breaker (GET 401s open it; open breaker short-circuits before GET), and failure isolation (GET/POST 500, request exceptions, non-JSON). 30 pass.
- **CDK** (`jira-webhook-processor.test.ts`): overrides stamped when configured; keys omitted when absent. 43 pass.
- **CLI** (`jira.test.ts`): flags registered as optional; persisted to the mapping row when supplied, omitted when not. 59 pass.

`mise run build` green. Security secrets/masking findings are pre-existing baseline noise (history scan + `linear_reactions.py` etc.); the new best-effort returns carry `nosemgrep` annotations matching the sibling module.

### DIFF
--- a/agent/src/jira_reactions.py
+++ b/agent/src/jira_reactions.py
@@ -236,27 +236,58 @@ def comment_task_started(
 # must never fail, block, or retry the task.
 
 #: Jira ``statusCategory.key`` for the "In Progress" band. The start transition
-#: prefers a destination in this category so standard workflows work with no
-#: per-project config. (Categories: ``new`` = To Do, ``indeterminate`` = In
-#: Progress, ``done`` = Done — the only three, and they are stable API values.)
+#: falls back to this category so standard workflows work with no per-project
+#: config. (Categories: ``new`` = To Do, ``indeterminate`` = In Progress,
+#: ``done`` = Done — the only three, and they are stable API values.)
 _IN_PROGRESS_CATEGORY = "indeterminate"
 
-#: Default destination-name match for the PR-opened transition when no override
-#: is configured. Compared case-insensitively against ``to.name``.
-_DEFAULT_REVIEW_STATUS = "in review"
+#: Ordinal rank of the three status categories, low → high. Used to skip a
+#: transition that would move the card *backward* (e.g. a re-triggered task on
+#: an issue that's already In Review must not drag it back to In Progress).
+_CATEGORY_RANK = {"new": 0, "indeterminate": 1, "done": 2}
+
+#: Preferred destination *name* for the start transition (matched
+#: case-insensitively before the category fallback). Both ``In Progress`` and
+#: ``Blocked`` share the ``indeterminate`` category, so a name match is what
+#: keeps a category-only heuristic from landing on ``Blocked`` (#605).
+_START_STATUS_NAME = "in progress"
+
+#: Preferred destination names for the PR-opened transition, tried in order.
+#: Mirrors Linear's "In Review → In Progress" fallback while also matching
+#: common review-column names, so a "Code Review" column isn't silently skipped
+#: and a stock board (no review status) still advances to / stays at In Progress
+#: rather than no-opping (#605). Ends with In Progress as the safe fallback.
+_REVIEW_STATUS_NAMES = (
+    "in review",
+    "code review",
+    "review",
+    "peer review",
+    "reviewing",
+    _START_STATUS_NAME,
+)
+
+#: Destination names the category fallback must never auto-pick. ``Blocked``
+#: shares the ``indeterminate`` category with ``In Progress``; a bare
+#: first-indeterminate heuristic could land there (#605), which is never what
+#: "move to In Progress" means. A configured override can still target it.
+_CATEGORY_FALLBACK_DENY = ("blocked",)
 
 
-def _get_transitions(cloud_id: str, issue_key: str, token: str) -> list[dict[str, Any]] | None:
-    """GET the transitions valid from the issue's current status.
+def _get_issue_transitions(
+    cloud_id: str, issue_key: str, token: str
+) -> tuple[dict[str, Any], list[dict[str, Any]]] | None:
+    """GET the issue's current status + the transitions valid from it.
 
-    Returns the ``transitions`` list (possibly empty) on success, or ``None``
-    on any failure. An empty list is normal — Jira returns one when the OAuth
-    user lacks the *Transition Issues* permission — and callers treat it as
-    "nothing to do", not an error.
+    Fetches ``?fields=status&expand=transitions`` so a single call yields both
+    the current ``status`` (to skip moving a card backward, #605) and the
+    ``transitions`` list. Returns ``(current_status, transitions)`` on success,
+    or ``None`` on any failure. An empty transitions list is normal — Jira
+    returns one when the OAuth user lacks the *Transition Issues* permission —
+    and callers treat it as "nothing to do", not an error.
     """
     url = (
         f"{JIRA_API_BASE}/{quote(cloud_id, safe='')}"
-        f"/rest/api/3/issue/{quote(issue_key, safe='')}/transitions"
+        f"/rest/api/3/issue/{quote(issue_key, safe='')}?fields=status&expand=transitions"
     )
     try:
         resp = requests.get(
@@ -268,7 +299,7 @@ def _get_transitions(cloud_id: str, issue_key: str, token: str) -> list[dict[str
             timeout=REQUEST_TIMEOUT_SECONDS,
         )
     except requests.RequestException as e:
-        log("WARN", f"jira_reactions: transitions GET failed ({type(e).__name__}): {e}")
+        log("WARN", f"jira_reactions: issue GET failed ({type(e).__name__}): {e}")
         # nosemgrep: py-silent-success-masking -- Jira transitions are best-effort on network blips
         return None
 
@@ -276,57 +307,97 @@ def _get_transitions(cloud_id: str, issue_key: str, token: str) -> list[dict[str
         _note_auth_status(resp.status_code)
         return None
     if resp.status_code != HTTPStatus.OK:
-        log("WARN", f"jira_reactions: transitions GET HTTP {resp.status_code}: {resp.text[:200]}")
+        log("WARN", f"jira_reactions: issue GET HTTP {resp.status_code}: {resp.text[:200]}")
         return None
 
     _note_auth_status(resp.status_code)
     try:
-        transitions = resp.json().get("transitions", [])
+        payload = resp.json()
     except ValueError as e:
-        log("WARN", f"jira_reactions: transitions GET returned non-JSON: {e}")
+        log("WARN", f"jira_reactions: issue GET returned non-JSON: {e}")
         # nosemgrep: py-silent-success-masking -- Jira transitions are best-effort on bad responses
         return None
-    return transitions if isinstance(transitions, list) else []
+    # A well-formed response is a JSON object. Guard against valid-but-unexpected
+    # JSON (``null``, a bare list, a scalar) — ``.get`` would raise
+    # AttributeError, which is NOT best-effort: it propagates out of the pipeline
+    # hook and flips the task to FAILED (#605).
+    if not isinstance(payload, dict):
+        log("WARN", "jira_reactions: issue GET returned non-object JSON — skipping")
+        return None
+    current_status = (payload.get("fields") or {}).get("status") or {}
+    transitions = payload.get("transitions", [])
+    if not isinstance(current_status, dict):
+        current_status = {}
+    if not isinstance(transitions, list):
+        transitions = []
+    return current_status, transitions
+
+
+def _to_name(t: dict[str, Any]) -> str:
+    """Lowercased destination status name of a transition, or ``""``."""
+    return ((t.get("to") or {}).get("name") or "").strip().lower()
+
+
+def _to_category(t: dict[str, Any]) -> str | None:
+    """``statusCategory.key`` of a transition's destination, or ``None``."""
+    return ((t.get("to") or {}).get("statusCategory") or {}).get("key")
+
+
+def _usable(t: dict[str, Any]) -> bool:
+    """A transition we can execute unattended.
+
+    ``hasScreen`` transitions may demand required fields we can't supply, so we
+    skip them — a missing field would 400 the POST.
+    """
+    return isinstance(t, dict) and not t.get("hasScreen", False)
 
 
 def _select_transition(
     transitions: list[dict[str, Any]],
     *,
-    target_status: str | None,
-    prefer_category: str | None,
+    override: str | None,
+    prefer_names: tuple[str, ...],
+    fallback_category: str | None,
 ) -> dict[str, Any] | None:
     """Pick a transition by the resolution ladder, or ``None`` if nothing fits.
 
-    ``target_status`` (a per-project override or the ``In Review`` default) is
-    matched case-insensitively against ``to.name`` first. Only if it is unset
-    does ``prefer_category`` (a ``statusCategory.key``) act as the fallback.
-    Transitions that require a screen are skipped — filling required fields is
-    out of scope and a missing field would 400 the POST.
+    Ladder (first match wins), mirroring the Linear reference this feature is
+    modeled on (``prompt_builder.py``):
+
+    1. ``override`` — a per-project configured status name, matched
+       case-insensitively against the destination name. An override that isn't
+       reachable is a deliberate skip (no fallback — the user asked for a
+       specific status).
+    2. ``prefer_names`` — preferred destination names in priority order (e.g.
+       ``In Progress`` on start; ``In Review`` → ``Code Review`` → … on PR).
+       This is what stops a category-only heuristic from landing on ``Blocked``
+       (which shares the ``indeterminate`` category with ``In Progress``).
+    3. ``fallback_category`` — any usable transition whose destination is in
+       this ``statusCategory`` (keeps stock boards working with no config).
+
+    Screen-gated transitions are skipped throughout.
     """
+    usable = [t for t in transitions if _usable(t)]
 
-    def _usable(t: dict[str, Any]) -> bool:
-        # ``hasScreen`` transitions may demand required fields we can't supply.
-        return isinstance(t, dict) and not t.get("hasScreen", False)
+    if override:
+        wanted = override.strip().lower()
+        return next((t for t in usable if _to_name(t) == wanted), None)
 
-    if target_status:
-        wanted = target_status.strip().lower()
-        for t in transitions:
-            if not _usable(t):
-                continue
-            to_name = (t.get("to") or {}).get("name") or ""
-            if to_name.strip().lower() == wanted:
-                return t
-        # An explicit target that isn't reachable is a deliberate skip — do not
-        # silently fall back to a category heuristic the user didn't ask for.
-        return None
+    for name in prefer_names:
+        match = next((t for t in usable if _to_name(t) == name), None)
+        if match:
+            return match
 
-    if prefer_category:
-        for t in transitions:
-            if not _usable(t):
-                continue
-            category = ((t.get("to") or {}).get("statusCategory") or {}).get("key")
-            if category == prefer_category:
-                return t
+    if fallback_category:
+        return next(
+            (
+                t
+                for t in usable
+                if _to_category(t) == fallback_category
+                and _to_name(t) not in _CATEGORY_FALLBACK_DENY
+            ),
+            None,
+        )
     return None
 
 
@@ -369,15 +440,19 @@ def _transition(
     channel_metadata: dict[str, str] | None,
     *,
     lifecycle: str,
-    target_status: str | None,
-    prefer_category: str | None,
+    override: str | None,
+    prefer_names: tuple[str, ...],
+    fallback_category: str | None,
+    max_source_rank: int | None,
 ) -> None:
     """Shared best-effort transition path. No-op for non-Jira tasks.
 
-    Resolves ``(cloud_id, issue_key)``, discovers valid transitions, selects one
-    via the resolution ladder, and POSTs it. Every failure mode — disabled,
-    open circuit, missing token, empty transition list, no match, network /
-    HTTP error — is logged and swallowed so the pipeline is never affected.
+    Resolves ``(cloud_id, issue_key)``, fetches the current status + valid
+    transitions, skips if the card is already at/past the target band
+    (``max_source_rank``), selects a transition via the resolution ladder, and
+    POSTs it. Every failure mode — disabled, open circuit, missing token, empty
+    transition list, no match, network / HTTP error — is logged and swallowed
+    so the pipeline is never affected.
     """
     target = _enabled(channel_source, channel_metadata)
     if not target:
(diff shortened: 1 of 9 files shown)


isadeks (commented):
## Review — Jira board transitions (#572)

Clean single commit off the post-#573 main (no stacking risk — #573/#601/#603 all merged; this builds on the merged `jira_reactions.py`). The auth-breaker extraction (`_note_auth_status` / `_circuit_open`) is behavior-preserving, and the writer→reader→agent attribute chain is consistent (`status_on_start`/`status_on_pr` on the mapping row → `jira_status_on_start`/`jira_status_on_pr` in `channel_metadata` → the `jira_`-prefixed keys the agent reads).

I checked this out at `d558168` and **ran it** — agent `test_jira_reactions.py` 30/30, cli `jira.test.ts` 59/59, cdk `jira-webhook-processor.test.ts` 43/43; `ruff` + `ty` clean. The findings below were reproduced with throwaway tests against the real selector.

**The right reference point is Linear's own transition logic** (`agent/src/prompt_builder.py:150-170`), since this PR is explicitly modeled on it. Linear (via the `mcp__linear-server__save_issue` tool) moves workflow state too: at start it picks the state **named `In Progress`** (fallback `Todo`) and **skips if the issue is already in `In Progress`/`In Review`/`Done`**; on PR it moves to **`In Review` with a fallback to `In Progress`**. Two of the findings below are places where this Jira implementation diverges from — and is less robust than — that reference.

---

### 1. `_get_transitions` can raise an uncaught `AttributeError` → flips a task to FAILED  *(correctness — highest severity)*

```python
try:
    transitions = resp.json().get("transitions", [])   # jira_reactions.py:284
except ValueError as e:                                 # catches non-JSON only
    return None
return transitions if isinstance(transitions, list) else []
```

The `except ValueError` covers a non-JSON body and `isinstance(..., list)` guards the *inner* value — but if `resp.json()` returns valid JSON that isn't an object (`null`, a bare `[]`, a scalar), `.get()` raises `AttributeError`, which is not caught. Reproduced:
```
resp.json() -> None  →  AttributeError: 'NoneType' object has no attribute 'get'
resp.json() -> []    →  AttributeError: 'list' object has no attribute 'get'
```
The exception propagates `_get_transitions` → `_transition` → the hook, and both hooks sit inside the pipeline's outer `except Exception` (AST-confirmed: the sole enclosing handler for both `pipeline.py:872` and `:1076` is the `try` at 734–1220 with an `except Exception`) that does `write_terminal(task_id, "FAILED", …)` then `raise`. So:
- **start hook** — kills the task during repo-setup, before the agent runs;
- **PR hook** — fires *after* `pr_url` is set and `pr_created` emitted, so the PR opens and *then* the task is flipped to FAILED.

`ty` passes because `resp.json()` is typed `Any`, so it's invisible statically. One-line fix:
```python
payload = resp.json()
transitions = payload.get("transitions", []) if isinstance(payload, dict) else []
```

### 2. PR-opened transition silently no-ops on common boards — and unlike Linear, has no fallback  *(the feature does half its job by default)*

The default review match is *exact* equality against `"in review"`. Reproduced:

| workflow's review column | Jira (#605) | Linear reference |
|---|---|---|
| `In Review` | ✅ moves | ✅ moves |
| stock `To Do`/`In Progress`/`Done` (no review status) | ❌ silent skip | falls back to In Progress |
| `Code Review` / `Review` / `Peer Review` | ❌ silent skip | (In Review → In Progress fallback) |

The Linear prompt explicitly falls back In Review → In Progress "if that state doesn't exist"; this PR skips instead. So a stock Jira board or a "Code Review" column gets the start-transition but silently *not* the PR-transition, and the user must discover `--status-on-pr`. Recommend matching the Linear behavior (fall back to keeping/using an In-Progress-category status) or at minimum leading the docs with the exact-name requirement.

### 3. Start heuristic picks by category, not name → can land on the wrong status; also no "already past" skip  *(correctness edge; diverges from Linear)*

Jira's three status categories are `new`/`indeterminate`/`done`, and **`In Progress`, `In Review`, and `Blocked` all map to `indeterminate`**. `_select_transition`'s start path takes the *first* usable `indeterminate` transition with no name tiebreak. Reproduced: `To Do → {Blocked, In Progress}` both indeterminate → picks **Blocked** when Jira lists it first, **In Progress** when reversed. Order-dependent on a non-contractual list order.

The Linear reference avoids both problems: it matches the state **named `In Progress`**, and it **skips the transition if the issue is already In Progress/In Review/Done**. This PR's start path does neither — so besides the wrong-`indeterminate` pick, a re-trigger on an issue that's already In Review could move it *backward* to In Progress. Suggest mirroring Linear: prefer an `In Progress` **name** match (fall back to category), and skip if already in a later status. (`test_prefers_indeterminate_category` has only one indeterminate destination, so it doesn't catch this.)

### 4. Whitespace-only override is persisted but permanently no-ops  *(minor hardening)*

`...(opts.statusOnPr && { status_on_pr: opts.statusOnPr })` — `"   "` is truthy in JS so it's written to the mapping row; at the agent `target_status = override or default` keeps `"   "`, `.strip().lower()` → `""`, matches nothing → skip **with no fallback to the default**. Reproduced. So a fat-fingered blank override silently disables that project's transition forever. Fix: `.trim()` in the CLI, treat empty as unset. (Empty `""` is already correctly dropped by CLI `&&` + processor `if` — only whitespace leaks.)

---

### Confirmed non-issues (checked)
- Start on a stock `To Do`/`In Progress`/`Done` board works (picks In Progress via category).
- Auth-breaker accounting is correct for the two-call GET+POST path — a single GET-401 increments the shared counter by exactly 1, no POST fires.
- Empty-string `""` override is dropped by CLI + processor.
- Missing `id` → `"None"` → Jira 4xx → False, no crash. "Already in target status" → status omitted from list → clean skip. `RequestException`/`ValueError` paths don't raise.
- Linear-parity framing in the PR/docs is accurate — Linear transitions workflow state via `mcp__linear-server__save_issue` (`prompt_builder.py:156/164`), so "the same signal Linear-origin tasks already give" holds.

### Recommendation
Fix **#1** before merge (can fail an otherwise-successful task, one-line). **#2** and **#3** are worth aligning to the Linear reference this PR is modeled on — with a name-match + fallback + already-past skip, both largely dissolve. **#4** is a cheap hardening fix.

_(Consolidates my earlier separate comments on this PR.)_

isadeks (commented):
_(consolidated into the single full review on this PR — see it for all findings.)_

isadeks (commented):
_(consolidated into the single full review on this PR — see it for all findings.)_

isadeks (commented):
_(consolidated into the single full review on this PR — see it for all findings.)_

scottschreckengaust (dismissed):
## Verdict: 🔴 Request changes — one correctness bug that can fail an otherwise-successful task

Well-scoped, well-tested feature that gives Jira-origin tasks the board-transition signal Linear tasks already have. Governance is clean: **#572 is `approved`**, self-assigned, branch `feat/572-jira-transitions` conforms, CI green, `MERGEABLE`, docs + Starlight mirror regenerated. The auth-circuit-breaker extraction is behavior-preserving and the writer→reader→agent attribute chain is consistent. I'm requesting changes only for **Finding 1**, which I verified end-to-end against the PR head — it defeats the module's own best-effort contract.

### Vision alignment
Fits **fire-and-forget + reviewable outcomes**: the board reflects task lifecycle automatically, no new OAuth scopes (`read/write:jira-work` cover the transitions GET+POST), best-effort throughout so the channel never gates the task. Good.

### 🔴 Blocking

1. **`_get_transitions` can raise an uncaught `AttributeError` → flips the task to FAILED** — `agent/src/jira_reactions.py:284`. `except ValueError` catches a non-JSON body, and `isinstance(transitions, list)` guards the inner value, but if `resp.json()` returns valid **non-object** JSON (`null`, a bare `[]`, a scalar) then `.get()` raises `AttributeError`, which is not caught. I verified the full propagation on the PR head:
   - `_get_transitions` (`jira_reactions.py:284`, only `except ValueError`) → `_transition` (`:396`, **no** try/except around the call) → the pipeline hooks at `pipeline.py:872` (start) and `:1076` (PR-opened), both inside the `run_task` outer `try` (734) whose `except Exception` (`:1222`) does `write_terminal(config.task_id, "FAILED", …)` then re-raises (`:1255`).
   - **start hook** → kills the task during repo-setup, before the agent runs. **PR hook** → fires *after* `pr_url` is set and `pr_created` emitted, so the PR opens and *then* the task flips to FAILED — the worst outcome for a "best-effort" helper.
   - `ty` can't see it (`resp.json()` is typed `Any`). Every other failure mode in this module is correctly swallowed; this is the one path that isn't. One-line fix:
     ```python
     payload = resp.json()
     transitions = payload.get("transitions", []) if isinstance(payload, dict) else []
     ```
   (Credit: this is isadeks' Finding 1 — I confirmed it independently.)

### 🟡 Non-blocking (recommend addressing; align to the Linear reference this PR is modeled on)

2. **PR-opened match is exact `"in review"` with no fallback** — a stock board or a `Code Review`/`Peer Review` column gets the start-transition but silently *not* the PR-transition, and the user must discover `--status-on-pr`. Linear's reference (`prompt_builder.py`) falls back In Review → In Progress. Consider a fallback, or lead the docs with the exact-name requirement.
3. **Start heuristic picks by category, not name** — `In Progress`, `In Review`, `Blocked` are all `indeterminate`, so `_select_transition` takes the *first* usable one by non-contractual list order (could land on `Blocked`), and there's no "already past" skip (a re-trigger on an In-Review issue could move it *backward*). Linear matches the `In Progress` **name** and skips if already In Progress/In Review/Done. Prefer a name match (fall back to category) + already-past skip.
4. **Whitespace-only override (`"   "`) is persisted and permanently no-ops** — truthy in JS so it's written; the agent's `.strip().lower()` → `""` matches nothing, with no fallback. `.trim()` in the CLI (treat empty as unset) fixes it.

### Tests & CI
Strong, behavior-focused: channel gate, category/name selection, override precedence + case-insensitivity, unreachable-override skip, `hasScreen` skip, empty-list skip, shared circuit breaker, failure isolation (30 agent + 43 CDK + 59 CLI). Gap: no test exercises the **non-object-JSON** case in Finding 1 (add one alongside the fix), and `test_prefers_indeterminate_category` has only one `indeterminate` destination so it doesn't catch Finding 3's order dependence. No bootstrap/CDK-construct changes → bootstrap synth-coverage N/A. No `types.ts`/CLI drift (`channel_metadata` is an untyped `Record<string,string>`).

### Review agents run
- `code-reviewer`, `silent-failure-hunter`(via jira_reactions review) ✅ — cleared the circuit-breaker refactor, resolution ladder, URL `quote()` injection-safety, idempotency (PR-transition fires only under `if pr_url:`; crash path never transitions), and the least-privilege/scope claim. Notably my automated code-review pass **missed Finding 1** — the existing reviewer's manual repro caught it; I then verified it directly.
- `pr-test-analyzer` ✅ (coverage strong; gaps noted above).
- `type-design-analyzer` ⏭️ — no new exported types of substance. `/security-review` ⏭️ — no IAM/Cedar/network/secrets change (reuses existing scopes + `JIRA_API_TOKEN`).

### Human heuristics
- **Proportionality** ✅ — resolution ladder matches the problem; no over-engineering.
- **Coherence** ✅ — mirrors `linear_reactions`; shared breaker; consistent attribute chain. (Findings 2/3 are where it diverges from that reference and is less robust.)
- **Clarity** ✅ — strong "why" comments and docstrings.
- **Appropriateness** ⚠️ — Finding 1 is the one place a best-effort helper isn't fail-safe; integration verified against real Jira API response shapes (`null`/non-object bodies) is the missing case.

ayushtr-aws:
Thanks @isadeks — excellent review, all four findings reproduced and fixed in a6c54ea (rebased onto the latest main).

**1. `_get_transitions` uncaught `AttributeError` → task FAILED** (correctness). Fixed. The response body is now assigned first and `.get()` is only called behind an `isinstance(payload, dict)` guard, so valid-but-non-object JSON (`null`, bare `[]`, a scalar) yields an empty list instead of raising past the best-effort boundary. Added a parametrized test (`None`/`[]`/`42`/`"scalar"`) asserting no raise + no POST.

**2. PR-opened silently no-ops, no fallback** (feature does half its job). Fixed by mirroring Linear. The PR path now tries `In Review` → `Code Review`/`Review`/`Peer Review`/`Reviewing` → falls back to **In Progress**, so a stock board or a Code-Review column advances instead of silently skipping. Tests cover the synonym match, the In-Review-over-Code-Review priority, and the stock-board In-Progress fallback.

**3. Start heuristic picks by category (can hit `Blocked`); no already-past skip** (correctness edge). Fixed, again mirroring Linear:
- Start now prefers a destination **named** `In Progress` before the `indeterminate` category fallback, and the category fallback explicitly **excludes `Blocked`**. Added a test with `Blocked` listed *first* that asserts we still pick `In Progress`, plus one asserting a Blocked-only board skips.
- Added an **already-at/past-target skip**: I switched the GET to `?fields=status&expand=transitions` (current status + transitions in one call) and skip when the current `statusCategory` rank is ≥ the target — so a re-trigger on an already-In-Review issue won't drag it back. Overrides bypass this skip (a configured status is a deliberate instruction). Tests cover already-In-Progress, already-In-Review, already-Done, and the override-bypass.

**4. Whitespace-only override persisted → permanent no-op** (hardening). Fixed in the CLI: `--status-on-start`/`--status-on-pr` are `.trim()`-ed and blank/whitespace is treated as unset (not written). Tests cover trimming and whitespace-unset.

Docs updated: the resolution ladder now documents name-match → category-fallback-excluding-Blocked → skip, the PR In-Review→In-Progress fallback, and the already-past skip. Full `mise run build` green; agent tests 30→43, CLI +3.

ayushtr-aws:
Thanks @scottschreckengaust — your review came in ~7 min before my fix push landed (review 21:23Z; fix `a6c54ea` 21:30Z), so it was against the pre-fix head. All four findings — which match @isadeks' — plus both test-gap callouts are already addressed in `a6c54ea`:

- **🔴 Finding 1** (`_get_transitions` `AttributeError` → FAILED): fixed with the exact `isinstance(payload, dict)` guard you suggested. Added `test_non_dict_json_does_not_raise`, parametrized over `None`/`[]`/`42`/`"scalar"` — the non-object-JSON test you flagged as missing.
- **🟡 Finding 2** (PR match exact `in review`, no fallback): now `In Review` → synonyms (`Code Review`/`Review`/`Peer Review`/`Reviewing`) → **In Progress** fallback, mirroring Linear.
- **🟡 Finding 3** (category-not-name pick; can hit `Blocked`; no already-past skip): start prefers the **name** `In Progress`, category fallback excludes `Blocked`, and I added an already-at/past-target skip (switched the GET to `?fields=status&expand=transitions`). Replaced `test_prefers_indeterminate_category` with `test_prefers_in_progress_by_name_over_blocked` (Blocked listed *first*) — closing the order-dependence gap you noted — plus already-In-Progress/In-Review/Done skip tests.
- **🟡 Finding 4** (whitespace override persisted): CLI now `.trim()`s and treats blank as unset.

Agent tests 30 → 43, CLI +3, full `mise run build` green. Would you re-review `a6c54ea` when you have a moment? Happy to adjust further.

isadeks (commented):
### Re-review of `a6c54ea` — all four addressed; one new regression in the added code

Re-verified at `a6c54ea` (ran it): shipped suite **43/43**, `ruff`/`ty` clean. Confirmed fixed, each reproduced:
- **#1** `isinstance(payload, dict)` guard (+ nested `current_status`/`transitions` guards) — non-object JSON (`None`/`[]`/scalar) now returns cleanly instead of raising. ✓
- **#2** PR path now tries `In Review → Code Review → Review → Peer Review → Reviewing → In Progress` — stock board / Code-Review column advance instead of silently skipping. ✓
- **#3** start prefers the **name** `In Progress`, category fallback excludes `Blocked` (tested with Blocked listed first → still picks In Progress). ✓
- **#4** CLI `.trim()` + blank-as-unset. ✓

Nice — the name-match + fallback + rank-skip is exactly the Linear-reference shape.

---

### New finding (introduced by this fix): the PR path can still move a card **backward** In Review → In Progress — contradicting the new docs

The added `max_source_rank` guard prevents backward moves by *category rank*, but for the PR path it's set to `_CATEGORY_RANK["done"]` (=2), so an already-**In Review** card (category `indeterminate`, rank 1) passes the guard. Then `_select_transition` walks `prefer_names`, which ends in **`in progress`** (the stock-board fallback from #2). Since Jira omits the current status from the transitions list, `in review` won't match — but if the workflow offers an `In Review → In Progress` transition (common, for bouncing work back), the fallback matches it and moves the card **backward**.

Reproduced — card currently `In Review`, transitions `[In Progress, Done]`, PR hook fires:
```
transition_pr_opened → POSTs transition to In Progress   # backward
```
(Start path is unaffected — its `max_source_rank` is `indeterminate`=1, so an In-Review card is correctly skipped. Verified. The bug is PR-path-only.)

This directly contradicts the guide added in this same PR (`JIRA_SETUP_GUIDE.md`):
> **Already at or past the target** → the transition is skipped, so a re-triggered task never drags a card backward (e.g. from In Review back to In Progress).

That's the exact example the code fails to prevent. Trigger is a realistic re-run: a second label-add after the 8h dedup window (or a distinct re-add) re-invokes the pipeline on an issue a human already advanced to In Review.

**Fix options:**
- Drop the trailing `in progress` from `_REVIEW_STATUS_NAMES` and instead make the PR path's *category fallback* the In-Progress safety net — the fallback is already gated, and you could gate it with the same rank check (only fall back forward). Or
- Compute the chosen transition's **destination** rank and skip the POST if it's `< current_rank` (a general "never move backward" guard on the selected transition, not just on the source). This is more robust than the source-only `max_source_rank` and would also protect override/synonym paths.

The second is the cleaner invariant: *never POST a transition whose destination category rank is below the current status's rank* (unless an override explicitly targets it). Everything else here looks good.

isadeks (approved):
Approving. All four review findings are fixed and verified at `a6c54ea` (ran it — shipped suite 43/43, `ruff`/`ty` clean):

- **#1** (uncaught `AttributeError` → task FAILED) — `isinstance(payload, dict)` guard; non-object JSON returns cleanly. ✓
- **#2** (PR transition no-op on common boards) — In Review → synonyms → In Progress fallback; stock boards advance. ✓
- **#3** (start picks by category, could hit Blocked) — prefers the `In Progress` name, category fallback excludes Blocked. ✓
- **#4** (whitespace override) — CLI `.trim()` + blank-as-unset. ✓

The rework into the Linear-reference shape (name-match + fallback + rank-based skip) is clean.

**Non-blocking follow-up** (fine as a fast-follow, not gating this merge): the PR path's `max_source_rank` is `done`, so a re-triggered task on an already-**In Review** card can match the trailing `In Progress` fallback and move it backward. It's best-effort and never fails the task — a human just drags the card forward — but it's the one spot that contradicts the new guide's "never drags a card backward" line. Cleanest fix when you get to it: skip POSTing any transition whose destination category rank is below the current status's rank (unless an override targets it). Happy to leave it for a follow-up PR.

PR #616, opened 2026-07-15
fix(agent): stop scoped-session S3 hang in baseline build + reap hung suites + early ACK

### LINKED ISSUE #615: ECS baseline build hangs: scoped-session bypasses test boto3 mock + watchdog doesn't reap + late 👀 ACK
## Problem

On the ECS Fargate compute substrate, the agent's pre-agent baseline build (`mise run build` in the cloned repo) can **hang silently** — the agent unit-test suite stalls for 40+ min, up to the build-verify ceiling, with no output.

### Root cause

The ECS agent task def sets `AGENT_SESSION_ROLE_ARN` (per-session IAM scoping, #209). When it's set, `aws_session.get_session()` resolves a **scoped** session and `tenant_client(...)` returns `session.client(...)` — which **bypasses a `@patch("boto3.client")` mock**. A mocked test (e.g. `test_attachments`) then makes a **real S3 `get_object`** that blocks forever on the ECS network (no egress) in a socket read that `pytest-timeout`'s SIGALRM (main-thread only) can't interrupt.

`agent/tests/conftest.py` already resets the session cache per test, but that is insufficient: with `AGENT_SESSION_ROLE_ARN` still in the environment, a cold `get_session()` just re-resolves scoped and the mock is bypassed again.

Separately, the session-level `faulthandler.dump_traceback_later(exit=True)` watchdog does not reap the hang: pytest's `faulthandler_timeout` re-arms faulthandler's single internal timer per-test **without** `exit=True`, cancelling the exit timer — so a hang dumps but never exits.

### Secondary UX gap (same fix)

`react_task_started` (👀 + Backlog→In Progress) fires **after** `setup_repo()` (the baseline build). On a large repo that build is many minutes, so a Linear/Jira task shows **no feedback at all** during it — the issue looks dead.

## Fix

1. Add `AGENT_SESSION_ROLE_ARN` to the `_AGENT_ENV_VARS` scrub list in `conftest.py` so every test resolves the unscoped path where `boto3.client` mocks intercept. + regression test.
2. Replace/augment the faulthandler watchdog with an independent daemon-thread reaper that dumps all stacks and `os._exit(1)`s at a fixed deadline — pytest cannot clobber it; a blocked socket read releases the GIL so the timer thread runs.
3. Move the 👀 / In-Progress ACK (+ channel token resolution + start comment) to **before** `setup_repo()` so a large-repo task shows immediate feedback. As a side benefit a setup-phase failure then has a 👀 for the existing crash handler to swap to ❌.

## Evidence

Diagnosed + fixed + deployed on the `linear-vercel` line and live-verified on dev (ECS substrate): the agent suite that hung 55 min now runs `test_attachments` in ~1s and the whole 1328-test suite completes in ~12s; the 👀 appears ~3s after task start instead of after the multi-minute baseline. This PR ports that fix to `main`.

## Acceptance
- Agent test suite resolves unscoped even when `AGENT_SESSION_ROLE_ARN` is set → no real S3 call, no hang.
- A genuine hang is reaped (non-zero exit) within the deadline instead of stalling to the build ceiling.
- 👀 / In-Progress appears within seconds of task admission regardless of baseline-build duration.

isadeks:
Closes #615.

**Stacked on #598** (`pr/ecs-test-hygiene`) — extends its "fail hung tests loudly" theme with the scoped-session root cause it doesn't cover. Review/merge after #596→#597→#598.

## What hangs

On the ECS substrate the pre-agent baseline build (`mise run build`) stalls silently for 40+ min. The ECS agent task def sets `AGENT_SESSION_ROLE_ARN`, so `aws_session` resolves a **scoped** session and `tenant_client` returns `session.client(...)` — bypassing a `@patch("boto3.client")` mock. `test_attachments` then makes a **real S3 `get_object`** that blocks forever (no egress) in a socket read `pytest-timeout`'s SIGALRM can't interrupt.

## Fix (3 parts)

1. **conftest `_clean_env`** — add `AGENT_SESSION_ROLE_ARN` to the scrub list **and** reset the `aws_session` cache each test, so every test resolves the unscoped path where the mock intercepts. The scrub is required in addition to the reset: a cold `get_session()` re-resolves scoped while the var is still set. + regression test.
2. **conftest hang watchdog** — an independent daemon-thread reaper (`os._exit(1)` at 600s). `faulthandler.dump_traceback_later(exit=True)` doesn't work: pytest's `faulthandler_timeout` re-arms faulthandler's single timer per-test **without** `exit=True`, so a hang only dumps, never exits.
3. **pipeline early ACK** — move token resolve + 👀 + Jira start comment to **before** `setup_repo()` so a large-repo task shows immediate feedback during the multi-minute baseline instead of looking dead. `configure_channel_mcp` stays after the clone (needs repo dir).

## Verification

Diagnosed + deployed + live-verified on the `linear-vercel` line (dev, ECS substrate): the suite that hung 55 min now runs `test_attachments` in ~1s and the whole suite in ~12s; the 👀 appears ~3s after task start. Full agent gate green (**1201 tests** under `AGENT_SESSION_ROLE_ARN` set), ruff + ty clean.


agent/tests/test_aws_session.py
```diff
@@ -79,6 +79,35 @@ def test_blank_role_arn_treated_as_unset(self, monkeypatch):
...
+    (resetting the session cache alone is not enough — a cold ``get_session()``
+    re-resolves scoped while the var is still set).
+    """
+
+    def test_session_role_arn_not_visible_to_tests(self):
+        import os
+
+        # No monkeypatch here: this asserts the AUTOUSE fixture already scrubbed
+        # the var, even if the parent (ECS) environment had it set.
+        assert os.environ.get(SESSION_ROLE_ARN_ENV) is None
```
scottschreckengaust: **Blocking (AI005 — false regression guard).** This asserts the conftest `_clean_env` fixture scrubbed `AGENT_SESSION_ROLE_ARN`, but this module has its *own* file-local autouse fixture `_reset` (lines 24-30) that already `delenv`s the same var. I verified empirically: removing the `"AGENT_SESSION_ROLE_ARN"` line from `conftest.py`'s `_AGENT_ENV_VARS` (with the var set in the parent env) leaves this test **passing** — the local `_reset` masks it. So this guard cannot detect a regression of the exact fix this PR adds. On a PR whose whole purpose is that guard, this is the blocker.

Move both assertions into a fixture-free module so only the autouse conftest `_clean_env` is in play (validated as a true bidirectional guard). Note `monkeypatch.setenv` in the body won't work — autouse fixtures run during setup, before the body.

```suggestion
        # NOTE: this module's own ``_reset`` autouse fixture (above) also delenv's
        # SESSION_ROLE_ARN_ENV, so this assertion here does NOT actually exercise the
        # conftest ``_clean_env`` scrub — move this guard to a fixture-free module
        # (e.g. tests/test_conftest_env_scrub.py) that relies solely on the autouse
        # conftest fixture, so removing the conftest scrub line makes it fail.
        assert os.environ.get(SESSION_ROLE_ARN_ENV) is None
```

agent/tests/conftest.py
```diff
@@ -1,11 +1,55 @@
...
+        file=sys.stderr,
+        flush=True,
+    )
+    os._exit(1)
+
+
+# daemon=True so a clean, fast suite exit is never blocked waiting on this timer.
+_hang_watchdog = threading.Timer(_HANG_REAP_DEADLINE_S, _reap_on_hang)
+_hang_watchdog.daemon = True
+_hang_watchdog.start()
```
scottschreckengaust, line 58: **Nit (safety window).** `_reap_on_hang` calls `os._exit(1)` unconditionally. `daemon=True` stops the timer from *blocking* shutdown, but not from *firing* if a slow-but-passing suite lands near 600s (e.g. mid coverage-write in teardown), turning a green run red with a thread-dump uncorrelated to any failed test. Cancel on session finish so it only fires on a true hang:

```suggestion
_hang_watchdog.start()


def pytest_sessionfinish(session, exitstatus):  # noqa: ARG001
    # Close the window where a slow-but-successful run could trip the 600s reaper
    # during teardown/coverage-write. cancel() is a no-op if it already fired.
    _hang_watchdog.cancel()
```

agent/tests/conftest.py
```diff
@@ -1,11 +1,55 @@
...
+# ``faulthandler_timeout`` (pyproject.toml) RE-ARMS it at the start of every test
+# WITHOUT ``exit=True``. So a session-level exit timer is cancelled by the first
+# test, the per-test timer only DUMPS, and the suite hangs forever anyway.
+#
+# So own the reaper on a dedicated daemon thread pytest cannot touch. A blocked
+# socket read releases the GIL, so this thread runs; it dumps every thread's
+# stack for diagnosis and then HARD-EXITS the process, so `mise run build`
+# returns non-zero within seconds of the deadline instead of burning to the
+# ceiling. Deadline 600s: far above the whole suite's normal runtime (per-test
+# cap is 300s) yet well under the build ceiling.
```
scottschreckengaust: **Nit (factual error in load-bearing comment).** "per-test cap is 300s" is wrong — the actual `pytest-timeout` config is `timeout = 120` (`agent/pyproject.toml:136`); no 300s override exists. This is the justification for the 600s deadline, so keep it truthful.

```suggestion
# ceiling. Deadline 600s: far above the whole suite's normal runtime (per-test
# cap is 120s) yet well under the build ceiling.
```

scottschreckengaust (changes requested):
## Verdict: Request changes

The root-cause diagnosis and the three-part fix are correct, well-reasoned, and I verified the S3-hang fix empirically (with `AGENT_SESSION_ROLE_ARN` set, `test_attachments` + the new class now run in <1s instead of hanging). The one blocker is that the **headline regression guard does not actually guard the fix** — on a PR whose entire purpose is to install that guard, that is load-bearing. Two low-risk follow-ups (watchdog `os._exit` window + a factually wrong comment) round it out.

## Vision alignment

Strongly aligned. This *strengthens* tenet 1 (fire-and-forget) — the early-ACK move gives the submitter immediate 👀/In-Progress feedback during the multi-minute baseline instead of a dead-looking issue, and lets the crash handler surface a setup-phase failure as ❌. It *reinforces* tenet 3 (bounded blast radius): the hang watchdog reaps a wedged suite at 600s instead of burning to the 3600s build-verify ceiling, and the env scrub keeps tests on the unscoped path (no accidental real S3 egress). No tenet is traded; no ADR needed. `AGENT_SESSION_ROLE_ARN` semantics (scoped session, fail-closed) are unchanged — this only fixes test-harness leakage of it.

## Blocking issues

**1. `test_session_role_arn_not_visible_to_tests` is a false regression guard (`agent/tests/test_aws_session.py:96-101`).** The docstring says it verifies the conftest `_clean_env` fixture scrubs `AGENT_SESSION_ROLE_ARN`. It does not. This module has its own file-local autouse fixture `_reset` (lines 24-30) that already does `monkeypatch.delenv(SESSION_ROLE_ARN_ENV, raising=False)` + `reset_session_cache()`. I confirmed empirically: with `AGENT_SESSION_ROLE_ARN` set in the parent env and the `"AGENT_SESSION_ROLE_ARN"` line **removed** from `conftest.py`'s `_AGENT_ENV_VARS`, both tests still **pass** — the local `_reset` masks the removal. Only when I *also* removed the local delenv did they fail. So the exact line this PR adds to conftest is a silent no-op for these tests; a future edit that drops the conftest scrub would keep CI green until the ECS substrate hangs again — the precise regression this guard is supposed to catch (AI005: the test asserts what the local fixture guarantees, not that the fix under test is present). `test_session_resolves_unscoped_by_default` (lines 103-108) has the same defect for the same reason.

Risk: the PR's core deliverable — a durable guard against the ECS hang — provides false confidence.

Fix (validated by the pr-test-analyzer): move both assertions into a **fixture-free** module (e.g. `tests/test_conftest_env_scrub.py`) with no local `_reset`, so only the autouse conftest `_clean_env` is in play, and let the guard depend on the parent env carrying the var. That is a true bidirectional guard (passes with the scrub, fails without it). Note: `monkeypatch.setenv(...)` inside the test *body* is a dead end — autouse fixtures run during setup, before the body, so they can't scrub what the body sets afterward.

**2. Hang watchdog can hard-exit an already-finished slow run (`agent/tests/conftest.py:36-51`).** `_reap_on_hang` calls `os._exit(1)` unconditionally when the 600s daemon `Timer` fires. `daemon=True` prevents the timer from *blocking* shutdown, but not from *firing* if a legitimately slow-but-passing suite lands near 600s (e.g. still in pytest teardown / coverage write). `os._exit` skips atexit + buffer flush, so it would turn a green run red with a bewildering thread-dump uncorrelated to any failed test. Suggest cancelling on session finish so it only fires on a true hang: add `def pytest_sessionfinish(session, exitstatus): _hang_watchdog.cancel()`. `Timer.cancel()` is a no-op if it already fired.

**3. Factually wrong comment in the watchdog rationale (`agent/tests/conftest.py:32`).** It says "per-test cap is 300s" but the actual `pytest-timeout` config is `timeout = 120` (`agent/pyproject.toml:136`); no 300s override exists. This is comment rot at authoring time in the load-bearing justification for the 600s deadline. Change to 120s (or reframe 600s as a session backstop for hangs SIGALRM can't interrupt, sized between the longest healthy suite and the build ceiling).

**4. No pipeline test pins the early-ACK ordering (`agent/src/pipeline.py:826-881`).** The reorder is correct — I confirmed the moved calls (`resolve_linear_api_token`/`resolve_jira_oauth_token`/`react_task_started`/`comment_task_started`) don't depend on `setup_repo()` output, `configure_channel_mcp` correctly stays after the clone (needs `repo_dir`), and `linear_eyes_reaction_id` is initialized to `None` at line 730 so the crash handler is always bound. But the PR sells "a setup-phase failure now has a 👀 to swap to ❌" as a benefit, and no test asserts 👀-before-`setup_repo` or the ❌-swap on setup failure. A focused test (attach both to a parent mock, assert call order + a `setup_repo`-raises case) would pin the invariant against future reordering. Follow-up, not a blocker — the individual units are covered in their own module tests.

**5. Comment triplication (nit).** The ECS-hang narrative is duplicated near-verbatim in three prose blocks (`conftest.py:141-149`, `conftest.py:158-170`, `test_aws_session.py:83-97`). Accurate but a sync/comment-rot surface; consider one authoritative block referenced by the others.

## Documentation

No docs changes required or missing. The diff is Python-only (`pipeline.py`, `conftest.py`, `test_aws_session.py`) — no `docs/guides/`, `docs/design/`, or `CONTRIBUTING.md` edits, so no Starlight mirror sync is needed. `AGENT_SESSION_ROLE_ARN` is already documented (BEDROCK_COST_ATTRIBUTION); its contract is unchanged. No `cdk/.../types.ts` ↔ `cli/src/types.ts` sync, no Cedar engine bump, no roadmap item to check off.

## Tests & CI

CI green (build agentcore 5m15s, secrets/deps scan, dead-code advisory, title). No CDK construct/stack changes → **bootstrap synth-coverage not applicable**. Agent gate reportedly green (1201 tests under `AGENT_SESSION_ROLE_ARN` set). I ran the two new test classes + `test_attachments` under `AGENT_SESSION_ROLE_ARN` set: 13 passed in 0.69s (no hang). The watchdog is untested — acceptable (a process-killer is impractical to exercise in-process). Base freshness: head `dd7146f` is a direct descendant of `origin/pr/ecs-test-hygiene` tip `32b1370` — **not stale**.

## Review agents run

- **pr-review-toolkit:code-reviewer** — ran. Confirmed early-ACK dependency/ordering correctness, watchdog import-time/xdist safety, and the 300s→120s comment error; judged 600s correctly *not* an AI007 `constants.json` value (Python-test-infra-only, no cross-runtime consumer).
- **pr-review-toolkit:silent-failure-hunter** — ran (error-handling/ACK/reaping in scope). Confirmed the token resolvers are fail-open by design (never raise; return `""` with justified `nosemgrep` allowlists), so the early-ACK move hides no abort-worthy failure; flagged the watchdog `os._exit` window (Finding 2).
- **pr-review-toolkit:pr-test-analyzer** — ran (this PR IS about tests). Independently reproduced the false-guard finding empirically and supplied the validated fix; confirmed the missing early-ACK ordering test.
- **type-design-analyzer** — omitted: no new types introduced.
- **comment-analyzer** — folded into code-reviewer (comment accuracy/rot covered above); not separately spawned.
- **/security-review** — omitted: no IAM, Cedar, network, secrets, or input-gateway change (test-harness + pipeline reorder only).

## Human heuristics

- **Proportionality** — Pass. Fix scope matches the problem; no over-abstraction. Comments are verbose/triplicated (nit #5) but not structurally disproportionate.
- **Coherence** — Pass. Lands in `agent/` per the routing table; same terminology as the existing session-scoping code.
- **Clarity** — Concern. `conftest.py:32` states a wrong per-test cap (nit #3) in load-bearing rationale; otherwise names/intent are clear.
- **Appropriateness** — Concern (blocking). The regression guard asserts what a local fixture already guarantees rather than what the fix does (AI005) — `test_aws_session.py:96-101`; see blocking issue #1.

isadeks:
Addressed in `1ebf6f9`:

**B1 (false regression guard) — fixed.** You are exactly right: `test_aws_session.py` has its own autouse `_reset` that does `monkeypatch.delenv(SESSION_ROLE_ARN_ENV)`, so the guard passed even with the conftest scrub removed. Moved both assertions into a new **fixture-free** module `tests/test_conftest_env_scrub.py` that poisons `AGENT_SESSION_ROLE_ARN` at **import time** (before any fixture runs), so only conftest `_clean_env` can clear it. Verified bidirectional: passes with the scrub, and **both tests FAIL when the `_AGENT_ENV_VARS` line is removed** (I ran it both ways).

**N2 (watchdog false-positive) — fixed.** Added `pytest_sessionfinish` → `_hang_watchdog.cancel()` so a slow-but-passing suite landing near 600s isn't `os._exit`-killed. `Timer.cancel()` is a no-op if it already fired on a true hang.

**N3 (comment rot) — fixed.** Corrected "per-test cap is 300s" → the actual `pytest-timeout` is 120s; reframed 600s as a session backstop sized between the longest healthy run and the build ceiling.

**N4 (early-ACK ordering test) / N5 (comment triplication)** — leaving as the follow-ups you flagged them as; the individual moved calls are covered in their own module tests, and the ordering invariant is low-churn. Happy to add the ordering test if you'd prefer it in this PR.

Agent gate green (1201 tests under `AGENT_SESSION_ROLE_ARN` set; clean exit confirms the sessionfinish cancel works).

isadeks:
All review points addressed — the blocking B1 + N2/N3 landed earlier in `1ebf6f9`, and the two non-blocking nits (N4/N5) are now in `b2ca9c84`, so nothing's left open.

**Blocking #1 — false regression guard** (`1ebf6f9`): moved the assertions into a new **fixture-free** module `tests/test_conftest_env_scrub.py` with no local `_reset`, exactly as you suggested. It poisons `AGENT_SESSION_ROLE_ARN` at **module import time** (before any fixture runs), so only conftest's `_clean_env` can clear it — a true bidirectional guard. Removed the two masked assertions from `test_aws_session.py` and left a NOTE there pointing to the new module. Verified it fails if the conftest scrub line is dropped.

**N2 — watchdog hard-exit window** (`1ebf6f9`): added `pytest_sessionfinish` → `_hang_watchdog.cancel()`, so a slow-but-passing run near 600s isn't `os._exit`-killed into a bewildering red. `Timer.cancel()` is a no-op if it already fired on a true hang.

**N3 — comment rot** (`1ebf6f9`): the "per-test cap is 300s" line was wrong (actual `pytest-timeout` is 120s). Reframed the 600s deadline as a session backstop for whole-suite hangs SIGALRM can't interrupt, sized between the longest healthy run and the build ceiling.

**N4 — no test pins the early-ACK ordering** (`b2ca9c84`): added `TestEarlyAckOrdering` in `test_pipeline.py`:
- `test_channel_ack_fires_before_setup_repo` — attaches the token-resolve, `react_task_started`, `comment_task_started`, and `setup_repo` mocks to a shared parent and asserts the ACK trio fires **before** `setup_repo`.
- `test_setup_failure_swaps_eyes_to_cross` — `setup_repo` raises; asserts the outer handler calls `react_task_finished(success=False, started_reaction_id="reaction-42")` — the 👀→❌ swap with the id captured pre-setup.

**N5 — comment triplication** (`b2ca9c84`): the authoritative ECS-hang narrative now lives in the `_clean_env` docstring; the inline `_AGENT_ENV_VARS` comment is shrunk to a pointer. (The top-of-file watchdog comment and `test_aws_session.py`'s NOTE were already pointers, not copies.)

Agent gate green under `AGENT_SESSION_ROLE_ARN` set (the hang-repro condition): **1203 tests, no hang**; ruff + ty clean. No CDK/docs changes, so bootstrap-synth + Starlight-mirror N/A as you noted.

isadeks:
The failing **"Secrets, deps, and workflow scan"** check is **not from this PR** — it's a pre-existing, repo-wide dependency-hygiene failure.

**What failed:** the `osv-scanner` dependency step (`mise run security:deps`) flagged 5 known vulns in `yarn.lock` — `astro@7.0.5` (×3: GHSA-4g3v-8h47-v7g6, -8mv7-9c27-98vc, -f48w-9m4c-m7f5) and `brace-expansion` (×2: GHSA-3jxr-9vmj-r5cp, transitive).

**Why it's not this PR:**
- This PR's diff is **Python-only** (`agent/src/pipeline.py`, `agent/tests/*`) — it touches **zero** npm deps.
- The flagged versions (`astro@7.0.5`, `brace-expansion@1.1.15`/`2.1.1`) are **byte-identical** across `main`, the base branch `pr/ecs-test-hygiene`, and this branch.
- The scheduled **Security** workflow on `main` has been failing on the same scan for weeks (every run 2026-06-22 → 2026-07-20).
- The scan **passed** on this PR's prior head `1ebf6f9` and **failed** on `b2ca9c84` with an unchanged `yarn.lock` — i.e. the OSV advisory DB was updated between the two scan times, not the code.

**Fix is already in flight, correctly scoped elsewhere:** Dependabot PRs **#631 / #630** (`bump astro 7.0.5 → 7.1.0`) target exactly these advisories against `main`, so the fix reaches the scheduled scan too. Bumping `astro` (a `docs/` Starlight dep) inside a test-only agent PR would be wrong scope. All other #616 checks — `build (agentcore)`, dead-code, PR-title — are green.

agent/tests/conftest.py
```diff
@@ -1,11 +1,73 @@
...
+# in the MAIN thread during a test's *call* phase, so a deadlock in a WORKER
+# thread, a fixture, collection, or a C-level socket read the main thread never
+# returns from stalls the whole `mise run build` silently — up to the platform's
+# 3600s build-verify ceiling (the ECS-only stall on ABCA-684/686/688, and the
+# scoped-session S3 hang the _clean_env reset below guards against: 40+ min of
+# dead air, container never reaped).
+#
+# The obvious instrument — ``faulthandler.dump_traceback_later(1200, exit=True)``
+# — does NOT work here: faulthandler has a SINGLE internal timer, and pytest's
+# ``faulthandler_timeout`` (pyproject.toml) RE-ARMS it at the start of every test
```
scottschreckengaust, line 23: Nit (comment accuracy, non-blocking). This attributes the per-test timer to pytest's built-in `faulthandler_timeout` "(pyproject.toml)", but pyproject configures **`pytest-timeout`** — `timeout = 120`, `timeout_method = "signal"` (pyproject.toml:136-137). There is no `faulthandler_timeout` set (`grep faulthandler pyproject.toml` is empty). The two are different mechanisms.

The *conclusion* is still correct and load-bearing: the configured `method="signal"` (SIGALRM) fires only in the main thread during a test's call phase, so it can't interrupt a worker-thread / C-level socket hang — hence the dedicated daemon reaper. Only the named mechanism is wrong. Suggest reframing around `pytest-timeout method="signal"` so the justification stays truthful (this block survived the N3 300s→120s fix without this being caught).

scottschreckengaust (dismissed):
## Verdict: Approve (one non-blocking comment-accuracy nit)

Reviewing as a Principal AWS Solutions Architect. This PR closes the loop on the ECS scoped-session S3 hang, adds a session-level hang reaper, and moves the channel ACK ahead of the multi-minute baseline build. The delta on top of the base (`pr/ecs-test-hygiene`) is agent Python only — test infra + a pipeline reorder. **All three of my prior review items (B1, N2, N3) are genuinely resolved, verified against the current code**, and the two non-blocking follow-ups (N4, N5) are now landed. Approving.

### Prior-review verification (independently re-checked against HEAD `b2ca9c8`)
- **B1 (false regression guard) — RESOLVED.** The guard moved to a new fixture-free module `agent/tests/test_conftest_env_scrub.py` that poisons `AGENT_SESSION_ROLE_ARN` at import time. I verified it bidirectionally: removing the `"AGENT_SESSION_ROLE_ARN"` line from `conftest.py:_AGENT_ENV_VARS` makes **both** tests in that module fail (`is_scoped()` returns `True`, assertion `True is False`). This is a true guard of the exact fix, not the local-fixture artifact the original placement was.
- **N2 (watchdog false-positive kill) — RESOLVED.** `pytest_sessionfinish` (conftest.py:61) calls `_hang_watchdog.cancel()`; `Timer.cancel()` is a no-op if it already fired, so a slow-but-passing run near 600s is no longer hard-exited into a red.
- **N3 (300s comment error) — RESOLVED.** The rationale now reads "per-test pytest-timeout cap is 120s" and matches `agent/pyproject.toml:136` (`timeout = 120`).

### Vision alignment
Fits the tenets cleanly. **Reviewable outcomes**: a hung suite now hard-exits within ~600s with a full thread dump instead of burning to the 3600s build-verify ceiling — failures become diagnosable. **Bounded blast radius**: the early-ACK reorder keeps a picked-up task visibly alive (👀 within seconds) rather than looking dead through the baseline build, and gives the crash handler a reaction to swap to ❌ on a setup-phase failure. No tenet is traded away; no ADR/RFC required.

### Correctness (principal-architect judgment)
- **Early-ACK reorder is sound.** `linear_eyes_reaction_id` is initialized to `None` at `agent/src/pipeline.py:730` *before* the `try`, assigned by `react_task_started` at line 847 (before `setup_repo` at line 867), and consumed by the outer crash handler at line 1259 with `started_reaction_id=linear_eyes_reaction_id`. A crash before the assign still binds the name (no `NameError`). `configure_channel_mcp` correctly stays *after* the clone (it needs `setup.repo_dir`). Token resolution moved earlier because `react_task_started`/`comment_task_started` read the env var it sets — correct dependency ordering.
- **No silent-failure regressions.** The reorder didn't touch the best-effort/no-op contracts of the reaction/comment calls; error handling in the moved block is unchanged. `setup_repo` failures still propagate to the single outer `except Exception` (pipeline.py:1220) — no new swallow.
- **No contract/type surface touched.** No shared `types.ts`, no Cedar pin, no new Python types (fixtures use `dict`/`MagicMock`). cdk↔cli type sync and cedarpy↔cedar-wasm parity: N/A.
- **No IAM/network/CDK/secrets change in the delta**, so no bootstrap-policy (ADR-002) or security-review obligation. `AGENT_SESSION_ROLE_ARN` is *consumed* by tests, not newly granted.

### Tests
Strong. New tests are real guards, not tautologies:
- `TestEarlyAckOrdering` pins both invariants the reorder sells: (a) resolve→react→comment all fire before `setup_repo` (via a shared parent mock's `mock_calls` order), and (b) a `setup_repo` crash still swaps 👀→❌ with the pre-setup reaction id.
- `test_conftest_env_scrub.py` is a verified bidirectional guard (see B1 above).
- I ran the full agent suite under the hang-repro condition (`AGENT_SESSION_ROLE_ARN` set): **1203 passed in ~5.9s, no hang** — `test_attachments` resolves unscoped and its `boto3.client` mock intercepts. Targeted run of the changed files: 17 passed.
- Accepted coverage gap (non-blocking): the `_reap_on_hang`/`_hang_watchdog` reaper itself has no direct unit test, which is reasonable — exercising a 600s daemon timer that calls `os._exit(1)` would kill the test process. The `pytest_sessionfinish` cancel path is exercised by every clean run.
- CDK synth-perf (#366): N/A — no CDK tests.

### CI
One red check — **"Secrets, deps, and workflow scan"** — is **unrelated to this PR**. `security:deps` (osv-scanner) flags pre-existing transitive npm vulns in `yarn.lock` (`astro` 7.0.5, `brace-expansion`). Those are repo-wide dependency-hygiene items affecting all branches; this diff is agent Python (tests + pipeline) and touches no lockfile. `build (agentcore)` passes. Not a blocker for this change, but the repo owes a separate dependency-bump issue.

### Documentation
No docs drift. This is a test-infra + internal-ordering fix with no new env var, command, or user-facing contract (the early-ACK is a UX-timing improvement to an existing behavior, not a new one). No `docs/guides`, `docs/design`, or Starlight mirror update required; AGENTS routing is correct (agent runtime lives in `agent/`). Backing issue #615 carries the `approved` label — governance gate satisfied (ADR-003). Branch `fix/615-ecs-scoped-session-hang` follows the naming convention.

### Non-blocking nit
1. **Comment misattributes the configured timeout mechanism** — `agent/tests/conftest.py:23`. The comment says pytest's `faulthandler_timeout` (pyproject.toml) re-arms faulthandler's single timer per-test. But `pyproject.toml` configures **`pytest-timeout`** (`timeout = 120`, `timeout_method = "signal"`), not pytest's built-in `faulthandler_timeout` — `grep faulthandler pyproject.toml` returns nothing. The *conclusion* (a dedicated daemon reaper is needed because the configured SIGALRM/`method="signal"` fires only in the main thread and can't interrupt a worker/socket hang) is correct and load-bearing; only the named mechanism is wrong. Since the block is explicitly justifying why `dump_traceback_later(exit=True)` was rejected, keeping the attribution accurate matters. Suggest reframing around `pytest-timeout method="signal"` (the thing that's actually set). Non-blocking.

### Review agents (Stage 3)
This harness cannot spawn the pr-review-toolkit sub-agents as separate processes, so I applied each agent's lens by hand against the source + a live test run:
- **code-reviewer** — ran: routing, AGENTS.md compliance, reorder correctness. Clean.
- **silent-failure-hunter** — ran: pipeline reorder + crash handler. No new swallowed failures; `None`-init guard verified.
- **pr-test-analyzer** — ran: verified new tests are non-tautological (bidirectional guard proven empirically; ordering test uses real call-order capture); noted the acceptable watchdog-reaper coverage gap.
- **comment-analyzer** — ran: found the `faulthandler_timeout` attribution nit above; confirmed the N3 120s fix is accurate.
- **type-design-analyzer** — omitted: no new types in the delta.
- **/security-review** — omitted: no IAM/Cedar/network/secrets/input-gateway change in the delta.

### Human heuristics
- **Proportionality** — pass. A daemon-thread reaper for a class of hangs SIGALRM provably can't reap (verified live per ABCA-688) is proportionate, not over-engineered.
- **Coherence** — pass. Same ECS-hang narrative now has one authoritative home (`_clean_env` docstring) with pointers elsewhere (N5 de-dup); belongs in `agent/tests`.
- **Clarity** — concern (nit #1 only): the one inaccurate mechanism name in `conftest.py:23`. Everything else reads clearly.
- **Appropriateness** — pass. Verified against real behavior (live-deployed on the linear-vercel dev line; I re-ran 1203 tests under the repro env); tests assert what the code *should* do.

ayushtr-aws (approved):

PR #619, opened 2026-07-16
feat(jira): include Jira attachments and recent comments in task context (#577)

### LINKED ISSUE #577: feat(jira): include Jira attachments and recent comments in task context
## Component

CDK Jira adapter, attachment screening, agent context

## Describe the feature

Jira-origin tasks should include the same practical issue context that Linear-origin tasks can use: attached files and relevant comments, not only the issue summary and description.

Current Jira task creation converts the issue summary/description and extracts embedded HTTPS image URLs. It explicitly skips Jira `media` file attachments because those require a Jira REST lookup/download, and it does not include issue comments. That means logs, screenshots, PDFs, specs, and clarifications attached to the Jira ticket can be invisible to the agent.

## Use case

A Jira ticket often says “see attached log,” “use the attached screenshot,” or contains acceptance clarifications in comments. If ABCA only sees the title and description, it may implement the wrong thing or stop to ask for context that was already on the ticket.

## Proposed solution

Add authenticated Jira context enrichment during task admission:

- After the Jira webhook is verified and tenant OAuth is resolved, fetch the full Jira issue context needed for the task through Jira REST.
- Include recent human comments in the task description, with clear attribution and ordering.
- Resolve Jira file attachments through the REST API rather than relying on unauthenticated URLs.
- Route downloaded attachment content through the existing ABCA attachment validation and Bedrock Guardrail screening path before it reaches the agent.
- Preserve existing limits: max attachment count, max size, supported MIME types, and fail-closed behavior for blocked or unscreenable content.
- If Jira context fetch fails for optional comments, proceed with a warning; if an attachment is selected for inclusion but cannot be safely fetched/screened, reject the task with a clear Jira comment.
- Document which Jira attachment types are supported.

## Acceptance criteria

- Jira issues with supported file attachments pass those attachments to the agent after security screening.
- Jira issues with recent comments include those comments in the task context.
- Unsupported, too-large, blocked, or unscreenable attachments do not reach the agent.
- Jira REST/auth failures are logged clearly and produce operator-visible feedback when they prevent task creation.
- Existing embedded image URL extraction from the description continues to work.
- Linear, Slack, API, and GitHub-context behavior are unaffected.

## Tests

- Jira webhook processor tests cover fetching comments, selecting supported attachments, skipping unsupported attachments, and task payload construction.
- Attachment screening tests cover authenticated Jira downloads flowing through the existing screening pipeline.
- Failure tests cover blocked attachments, oversized attachments, REST auth failures, and optional-comment fetch failures.
- Docs update `JIRA_SETUP_GUIDE.md` with supported attachment/comment behavior and limits.

## Out of scope

- Atlassian Remote MCP support.
- Jira project documents/Confluence pages.
- Fetching every historical comment on long-running issues.
- Agent-side unauthenticated download of Jira attachment URLs.
- New attachment limits or new file type support beyond what ABCA already screens.

ayushtr-aws:
## What

Closes #577. Jira-origin tasks now receive the same practical issue context Linear-origin tasks can use — **attached files and recent comments**, not just the summary/description. Because the Atlassian Remote MCP can't run headlessly (see the parent tracker #580 design note), the context is fetched **authenticated at task-admission time** in the webhook processor rather than on demand by the agent.

## How

- **`cdk/src/handlers/shared/jira-attachments.ts`** (new) —
  - `downloadScreenAndStoreJiraAttachments`: resolves Jira `media` attachments via the gateway attachment-content endpoint (`api.atlassian.com/ex/jira/{cloudId}/rest/api/3/attachment/content/{id}` — the only host the 3LO token is valid against), refresh-and-retry-once on 401/403, enforces the size cap while streaming, validates magic bytes, screens each through the existing **Bedrock Guardrail** pipeline, uploads cleaned bytes to S3, and returns `passed` `AttachmentRecord`s. **Fail-closed**: a selected attachment that can't be safely fetched/screened throws and the task is rejected with a Jira comment. Unsupported/oversized/over-cap attachments are silently skipped. Filenames sanitized + ids validated to prevent S3-key / on-disk path traversal.
  - `fetchRecentHumanComments`: recent human (`accountType: atlassian`) comments, ADF→markdown, oldest-first. **Fail-open**: any failure proceeds without them.
- **`cdk/src/handlers/shared/jira-adf.ts`** (new) — extracted the ADF→markdown walker so the processor and comment helper share it without a circular import (behavior-preserving).
- **`cdk/src/handlers/shared/create-task-core.ts`** — `TaskCreationContext` gains optional `taskId` (so processor-uploaded S3 keys match the eventual record) and `preScreenedAttachments` (merged verbatim, never re-screened; a non-`passed` record fails closed).
- **`cdk/src/handlers/jira-webhook-processor.ts`** — wires both in; comments fold under a `## Recent comments` heading, bounded so they never push `task_description` past the 10k limit (advisory context must not become a hard gate).
- **`cdk/src/constructs/jira-integration.ts` + `agent.ts`** — pass the attachments bucket to the processor, grant ReadWrite + `ATTACHMENTS_BUCKET_NAME`, bump the async processor timeout to 60s for serial download+screen. **No new OAuth scopes** (`read:jira-work` covers reading attachments + comments).
- **`cdk/src/constructs/task-api.ts`** — exempt `/v1/jira/webhook` from the WAF `SizeRestrictions_BODY` managed rule (mirrors the existing Linear/GitHub webhook exemptions). Jira issue payloads carrying attachment metadata exceed the 8 KB body limit and were being **blocked at the edge (403) before the receiver Lambda ran** — so this is required for #577 to function. The receiver still HMAC-verifies the raw body and the rate-limit rule still applies.

## Docs

`docs/guides/JIRA_SETUP_GUIDE.md` gets an "Issue context: attachments and comments" section (supported types, limits, skip-vs-reject behavior, comment behavior) + regenerated Starlight mirror.

## Tests

- `jira-attachments.test.ts` (new): download/filter/screen/upload, filename sanitization + unsafe-id rejection, 401 refresh-and-retry, size cap, magic-byte reject, screen-blocked → throw, comment human/app filtering + ADF→markdown + fail-open.
- `jira-webhook-processor.test.ts`: comment folding + truncation, fail-closed attachment rejection, remaining-slots math, no-attachments path.
- `create-task-core.test.ts`: `preScreenedAttachments` merge without re-screening, caller-supplied `taskId`, non-`passed` fail-closed.
- `task-api.test.ts`: WAF exemption present in both the exclusion scope-down and the full-CRS scope-down.

`mise run build` green (cdk 2357 tests, cli 605, agent + docs pass, synth clean).

## Validated end-to-end

Deployed to a dev stack and triggered from a real Jira Cloud tenant: a SCRUM issue with 2 image attachments + a comment triggered a task → comments fetched → both attachments authenticated-downloaded, Bedrock-screened, and stored to S3 → agent hydrated them → committed the images → **opened a PR**. Confirmed the full path works.

## Out of scope

Atlassian Remote MCP support; Confluence/project documents; fetching full historical comment threads; new attachment limits or file types beyond what ABCA already screens.

### DIFF
--- a/cdk/src/constructs/jira-integration.ts
+++ b/cdk/src/constructs/jira-integration.ts
@@ -25,6 +25,7 @@ import * as dynamodb from 'aws-cdk-lib/aws-dynamodb';
 import * as iam from 'aws-cdk-lib/aws-iam';
 import { Runtime, Architecture } from 'aws-cdk-lib/aws-lambda';
 import * as lambda from 'aws-cdk-lib/aws-lambda-nodejs';
+import * as s3 from 'aws-cdk-lib/aws-s3';
 import * as secretsmanager from 'aws-cdk-lib/aws-secretsmanager';
 import { NagSuppressions } from 'cdk-nag';
 import { Construct } from 'constructs';
@@ -35,8 +36,14 @@ import { JiraWorkspaceRegistryTable } from './jira-workspace-registry-table';
 /** Default task-record retention used for TTL computation (days). */
 const DEFAULT_TASK_RETENTION_DAYS = 90;
 
-/** Webhook-processor Lambda timeout (seconds). */
-const WEBHOOK_PROCESSOR_TIMEOUT_SECONDS = 30;
+/**
+ * Webhook-processor Lambda timeout (seconds). The processor is invoked
+ * asynchronously (not behind the API Gateway 30s deadline), so it can run
+ * longer than the receiver. #577 added serial authenticated download +
+ * Bedrock screening of up to 10 Jira attachments, so 60s leaves headroom over
+ * the per-attachment fetch (10s) + screening-retry budget.
+ */
+const WEBHOOK_PROCESSOR_TIMEOUT_SECONDS = 60;
 
 /**
  * Marker key embedded in the auto-generated stack-wide webhook-secret
@@ -86,6 +93,14 @@ export interface JiraIntegrationProps {
 
   /** Bedrock Guardrail version for input screening. */
   readonly guardrailVersion?: string;
+
+  /**
+   * S3 bucket for task attachment storage. Required for the webhook processor
+   * to fetch, screen, and store Jira `media` file attachments at task-admission
+   * time (issue #577). When omitted, issues carrying supported file attachments
+   * are rejected with a Jira comment rather than silently dropping them.
+   */
+  readonly attachmentsBucket?: s3.IBucket;
 
   /** Task retention in days for TTL computation. */
   readonly taskRetentionDays?: number;
@@ -206,6 +221,9 @@ export class JiraIntegration extends Construct {
       createTaskEnv.GUARDRAIL_ID = props.guardrailId;
       createTaskEnv.GUARDRAIL_VERSION = props.guardrailVersion;
     }
+    if (props.attachmentsBucket) {
+      createTaskEnv.ATTACHMENTS_BUCKET_NAME = props.attachmentsBucket.bucketName;
+    }
 
     // --- Cognito Authorizer (for /jira/link) ---
     const cognitoAuthorizer = new apigw.CognitoUserPoolsAuthorizer(this, 'JiraCognitoAuthorizer', {
@@ -279,6 +297,13 @@ export class JiraIntegration extends Construct {
           }),
         ],
       }));
+    }
+    // The processor downloads Jira `media` attachments, screens them, and
+    // writes the cleaned bytes to the attachments bucket before creating the
+    // task (#577). ReadWrite mirrors the confirm-uploads path (Put + Get for
+    // multipart/versioned writes).
+    if (props.attachmentsBucket) {
+      props.attachmentsBucket.grantReadWrite(webhookProcessorFn);
     }
 
     // --- Webhook receiver (verifies HMAC, dedups, invokes processor) ---
--- a/cdk/src/constructs/jira-integration.ts
+++ b/cdk/src/constructs/jira-integration.ts
@@ -40,10 +40,13 @@ const DEFAULT_TASK_RETENTION_DAYS = 90;
  * Webhook-processor Lambda timeout (seconds). The processor is invoked
  * asynchronously (not behind the API Gateway 30s deadline), so it can run
  * longer than the receiver. #577 added serial authenticated download +
- * Bedrock screening of up to 10 Jira attachments, so 60s leaves headroom over
- * the per-attachment fetch (10s) + screening-retry budget.
+ * Bedrock screening of up to 10 Jira attachments; the per-attachment fetch
+ * timeout alone (10s) can sum past 60s across a full batch, and a mid-loop
+ * kill would orphan objects and force an idempotent retry. 300s covers the
+ * worst-case serial batch with headroom. (A future optimization could
+ * parallelize the download/screen loop and lower this again.)
  */
-const WEBHOOK_PROCESSOR_TIMEOUT_SECONDS = 60;
+const WEBHOOK_PROCESSOR_TIMEOUT_SECONDS = 300;
 
 /**
  * Marker key embedded in the auto-generated stack-wide webhook-secret
--- a/cdk/src/constructs/task-api.ts
+++ b/cdk/src/constructs/task-api.ts
@@ -327,6 +327,18 @@ export class TaskApi extends Construct {
                       },
                     },
                     {
+                      // Jira issue payloads include the full issue field set.
+                      // Attachment metadata can push them over 8 KB. The
+                      // receiver HMAC-verifies the raw body and the priority-4
+                      // rule below still rate-limits this route.
+                      byteMatchStatement: {
+                        fieldToMatch: { uriPath: {} },
+                        positionalConstraint: 'EXACTLY',
+                        searchString: '/v1/jira/webhook',
+                        textTransformations: [{ priority: 0, type: 'NONE' }],
+                      },
+                    },
+                    {
                       // GitHub deployment_status webhook (preview-deploy
                       // screenshot pipeline). The full payload (workflow run
                       // history + deploy URLs + deployment metadata) exceeds
@@ -371,6 +383,18 @@ export class TaskApi extends Construct {
                             fieldToMatch: { uriPath: {} },
                             positionalConstraint: 'STARTS_WITH',
                             searchString: '/v1/tasks',
+                            textTransformations: [{ priority: 0, type: 'NONE' }],
+                          },
+                        },
+                      },
+                    },
+                    {
+                      notStatement: {
+                        statement: {
+                          byteMatchStatement: {
+                            fieldToMatch: { uriPath: {} },
+                            positionalConstraint: 'EXACTLY',
+                            searchString: '/v1/jira/webhook',
                             textTransformations: [{ priority: 0, type: 'NONE' }],
                           },
                         },
--- a/cdk/src/handlers/jira-webhook-processor.ts
+++ b/cdk/src/handlers/jira-webhook-processor.ts
@@ -18,7 +18,7 @@
  */
 
 import * as crypto from 'crypto';
-import { BedrockRuntimeClient } from '@aws-sdk/client-bedrock-runtime';
+import { BedrockRuntimeClient, ApplyGuardrailCommand } from '@aws-sdk/client-bedrock-runtime';
 import { DynamoDBClient } from '@aws-sdk/client-dynamodb';
 import { S3Client } from '@aws-sdk/client-s3';
 import { DynamoDBDocumentClient, GetCommand, ScanCommand } from '@aws-sdk/lib-dynamodb';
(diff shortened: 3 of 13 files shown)


isadeks (commented):
## Review — Jira attachments + comments (#577)

Strong PR: correct reuse of the screening pipeline (incl. PDF), deliberate fail-closed/fail-open split, thorough path-traversal defenses, 181/181 tests green. Reviewed by reading + running the code (three independent adversarial passes, each finding re-verified). Below is everything actionable, ranked. The design is sound — these are edges, not a rework.

**Verified non-issues (checked, no action):** SSRF/bearer-token leak on redirect (undici drops `Authorization` cross-origin; host is a hardcoded constant), path traversal via id/filename, WAF exemption (HMAC verified before processing, rate-limit still covers the path, 10 MB API-GW ceiling), `remainingSlots` clamp, comment-budget off-by-one (lands exactly at 10k chars), non-`passed` fail-closed guard, magic-byte validation on real bytes.

---

### Fix before merge

- [ ] **1. Empty attachment crashes the processor with no user feedback** — `jira-attachments.ts:448` → `types.ts:~630`. A 0-byte `text/plain`/`.log`/`.csv` (ordinary user mistake) passes magic-bytes + screening → `size_bytes: 0` → `createAttachmentRecord`'s `!params.size_bytes` guard throws a plain `Error` (not `JiraAttachmentError`), *outside* the fail-closed conversion. The processor catches only `JiraAttachmentError` (`jira-webhook-processor.ts:232`), so the handler throws → no task, **no ❌ comment** (breaks the fail-closed contract), 2 async retries re-screen everything. Reproduced 3×.
  **Fix:** in `createAttachmentRecord` guard `params.size_bytes === undefined` instead of falsy (same for the size check), OR reject a zero-byte body as `JiraAttachmentError` before the factory. Add a `size_bytes: 0` regression test.

- [ ] **2. 50 MB total-attachment cap trusts attacker metadata** — `jira-attachments.ts:229`. `selectAttachments` sums the running total from payload `att.size`; the per-file 10 MB cap is stream-enforced but the **total** is never re-checked against real bytes. `size: 1` (or omitted) on 10 files → all pass → up to 100 MB actually downloaded/screened/stored vs the documented 50 MB (cost/limit-integrity, not memory — buffers are loop-local). Confirmed by running: ~10-byte declared batch pulled 60 MB.
  **Fix:** track a real-bytes running total in the download loop; throw/stop once cumulative bytes exceed `MAX_TOTAL_ATTACHMENT_SIZE_BYTES`.

- [ ] **3. S3 objects orphaned on any failure after upload** — `jira-attachments.ts` has no cleanup, and `createTaskCore`'s `cleanupOrphanedAttachments` only tracks *inline* uploads in `uploadedS3Keys` — the pre-screened Jira keys are never added. So a mid-batch throw, the non-`passed` 500 guard, and a non-201 return (e.g. #4 below) all leave uploaded objects behind (90-day lifecycle is the only backstop). Diverges from the cleanup pattern this same PR relies on.
  **Fix:** track uploaded keys in the download loop and delete them on throw; and/or have the processor delete the returned `preScreenedAttachments` keys when `createTaskCore` returns non-201. Add a test asserting no orphan on a mid-batch failure.

### Decision needed

- [ ] **4. A policy-tripping human COMMENT fails the whole task closed** — `jira-webhook-processor.ts:466` folds comments into `task_description`, which `createTaskCore` guardrail-screens as one blob (`create-task-core.ts:387`); `GUARDRAIL_INTERVENED` → 400 → task rejected. The PR's length-truncation budget keeps comments from gating on *length*, but content-gating comes back through the filter — a third-party comment with a PII-shaped string ("SSN 123-45-6789") or injection-like phrasing fails the *reporter's* task for text they didn't write. Proven end-to-end. (Description-screening pre-exists on `main`; #577's new part is routing 3rd-party comment content into it.)
  **Options:** screen the comment section separately and drop it on intervention (fail-open, matching the comment design), rather than letting it gate task creation — or accept + document the behavior.

### Follow-up (robustness)

- [ ] **5. Async retries create duplicate tasks + duplicate attachment work** — fresh `taskId = ulid()` per invocation + no `idempotencyKey` (`jira-webhook-processor.ts:407,479`) means `attribute_not_exists(task_id)` can't dedupe a redelivery. Costlier now (up to 10 re-downloads/screens/uploads per duplicate). Consider a deterministic idempotency key (e.g. `issueKey#webhookTimestamp`, matching the receiver's dedup key).
- [ ] **6. 60s timeout is optimistic for 10 serial attachments** — 10 × 10s fetch timeouts alone exceed 60s; a mid-loop kill orphans objects + triggers the #5 retry. Consider parallelizing the download/screen loop or raising the ceiling.

_(This consolidates my two earlier comments on this PR into one list.)_

isadeks (commented):
_(consolidated into the single review comment above — see it for the actionable list.)_

ayushtr-aws:
Thanks for the thorough review, @isadeks — all six addressed in the latest push. Summary:

**Fix before merge**
1. **Empty attachment** — a 0-byte body is now rejected as `JiraAttachmentError` *before* `createAttachmentRecord`, so it routes through the fail-closed reject-with-comment path instead of throwing a plain `Error` outside the conversion. Regression test added (`size_bytes: 0`).
2. **50 MB total cap** — now enforced on **real downloaded bytes** (cumulative, in the download loop), not the attacker-declared `size` metadata. Test added with under-declared sizes (6×9 MB real vs `size: 1` → throws).
3. **S3 orphan cleanup** — `downloadScreenAndStoreJiraAttachments` tracks uploaded keys and best-effort-deletes them on a mid-batch throw; the processor deletes the returned pre-screened objects when `createTaskCore` returns non-201 (new exported `cleanupPreScreenedAttachments`). Mid-batch no-orphan test added.

**Decision needed**
4. **Policy-tripping comment fails the task** — went with the fail-open option: comments are now screened **separately** (`screenCommentsOrDrop`) and **dropped** (task proceeds) on `GUARDRAIL_INTERVENED` / screening outage / no guardrail, rather than being folded into the description blob that gates task creation. So third-party comment content can no longer fail the reporter's task. Test added (blocked comment → dropped, task still created).

**Follow-up**
5. **Duplicate tasks on retry** — added a deterministic `idempotencyKey` (`jira-<issueKey>-<webhookTimestamp>`, matching the receiver's dedup-key shape) so an async re-delivery dedupes. A 200 idempotent-replay from `createTaskCore` is now handled as success (no ❌ comment) and cleans up the round's re-uploaded objects.
6. **60s timeout** — raised to 300s (10 serial 10 s fetches can sum past 60s; a mid-loop kill would orphan objects + force the #5 retry). Noted parallelizing the loop as a future optimization to lower it again.

Verified non-issues confirmed — thanks for checking those. `mise run build` green (cdk 2383 after merging latest main).

isadeks (commented):
## Re-review of `1a742144` — all six addressed, verified by running

Checked out the head and re-ran my original repro scenarios against the fixed code (not just the new tests). Touched suites **144/144 green**, `tsc` clean. All six confirmed:

**Fix before merge**
1. **Empty attachment** ✓ — a 0-byte body now throws `JiraAttachmentError` *before* `createAttachmentRecord` (`jira-attachments.ts:406`), so it routes through fail-closed reject-with-comment. Re-ran my probe: `threw=JiraAttachmentError isJira=true` (was a plain `Error` crashing the handler).
2. **Real-byte total cap** ✓ — cumulative `totalBytes` checked against `MAX_TOTAL_ATTACHMENT_SIZE_BYTES` on downloaded bytes (`:395`), not declared `size`. Re-ran the under-declared batch (6×9 MB real, `size:1` each): throws after 5 uploads (45 MB, before the 6th crosses 50 MB) — the 60 MB leak is closed.
3. **S3 orphan cleanup** ✓ — loop tracks `uploadedKeys`, best-effort `deleteS3Objects` on throw (`:496-501`); processor calls the new exported `cleanupPreScreenedAttachments` on both the non-201 path *and* the 200 idempotent-replay path. Re-ran the mid-batch-block probe: earlier upload `attachments/u/t/a0/f0.log` is deleted (was orphaned).

**Decision needed**
4. **Comment gating** ✓ — went fail-open: `screenCommentsOrDrop` screens the comment section separately and returns `[]` (drops, task proceeds) on `GUARDRAIL_INTERVENED`, no-guardrail, *or* screening outage. Third-party comment content can no longer fail the reporter's task. (Coarse — one blocked line drops all comments — but that's the safe/advisory choice, not a defect.)

**Follow-up**
5. **Duplicate-task retries** ✓ — deterministic `buildIdempotencyKey('jira-<issueKey>-<timestamp>')`, sanitized to the valid charset + length-capped; I confirmed realistic keys (`jira-KAN-42-…`) pass `isValidIdempotencyKey`, so no risk of a malformed key 400-ing every Jira task. 200 replay handled as success (no ❌ comment) with cleanup.
6. **Timeout** ✓ — 30s→300s, with a note to parallelize the loop later.

**Verified non-issues from my earlier review** (SSRF/token-leak, path traversal, WAF exemption) still hold.

### One thing NOT caused by this PR — but it's a required check and currently red
`Secrets, deps, and workflow scan` is **failing**, but it's `osv-scanner` flagging pre-existing transitive-dep vulns (`astro` 7.0.5, `brace-expansion` 1.1.15) in `yarn.lock`. This PR doesn't touch `yarn.lock`/`package.json`, and those advisories are already present on `main` — so it's dep-drift the scan newly caught, not from #577. Since it's a required merge gate, it'll block *any* open PR until someone bumps those deps (org-level chore). Flagging so it's not mistaken for a #577 regression and so the merge isn't held on Ayush to fix.

Everything I raised is genuinely fixed. LGTM once the dep-scan (unrelated) is sorted org-side.

isadeks (approved):

PR #623, opened 2026-07-17
fix(agent): use GIT_AUTHOR/COMMITTER env vars instead of `git config --global` (#622)

### LINKED ISSUE #622: fix(agent): stop clobbering developer ~/.gitconfig — use GIT_AUTHOR/COMMITTER env vars instead of `git config --global`
## Problem

`agent/src/pipeline.py:826-837` runs `git config --global user.name bgagent` and `git config --global user.email bgagent@noreply.github.com` before repo setup. Inside the ephemeral ECS container this is intended (attribute commits to the bot), but the **same code path runs on developer workstations** during local agent runs / dry-runs / pytest that reaches `run_task`, permanently overwriting the developer's real identity in `~/.gitconfig`.

Observed: a contributor's global `user.name`/`user.email` silently reset to `bgagent` / `bgagent@noreply.github.com`.

## Fix

Replace the two `git config --global` subprocess calls with process-scoped environment variables (`GIT_AUTHOR_NAME`, `GIT_AUTHOR_EMAIL`, `GIT_COMMITTER_NAME`, `GIT_COMMITTER_EMAIL`), set alongside the existing `os.environ` writes on lines 838-845. Git honors these for all commits (inherited by Claude Code + the safety-net commit) without touching any on-disk config, so:

- Container behavior is unchanged (commits still authored as `bgagent`).
- A developer's `~/.gitconfig` is never modified.

Out of scope: `repo.py:153` `safe.directory --global` is additive and git-mandates global scope; it does not reset identity.

scottschreckengaust:
## Problem

`run_task()` in `agent/src/pipeline.py` set the bot's git identity with:

```python
git config --global user.name bgagent
git config --global user.email bgagent@noreply.github.com
```

`--global` writes to `~/.gitconfig` for whatever OS user runs the process. Inside the ephemeral ECS container that's intended. But the **same code path runs on developer workstations** (local agent runs, dry-runs, or any pytest that reaches `run_task`), permanently clobbering the developer's real identity to `bgagent`. This was observed in the wild.

## Fix

Replace the two `git config --global` subprocess calls with process-scoped env vars set alongside the existing `os.environ` writes:

```python
os.environ["GIT_AUTHOR_NAME"] = "bgagent"
os.environ["GIT_AUTHOR_EMAIL"] = "bgagent@noreply.github.com"
os.environ["GIT_COMMITTER_NAME"] = "bgagent"
os.environ["GIT_COMMITTER_EMAIL"] = "bgagent@noreply.github.com"
```

Git honors `GIT_AUTHOR_*`/`GIT_COMMITTER_*` for every commit (inherited by Claude Code and the safety-net commit in `post_hooks.py`) **without touching any on-disk config**. Container commits stay attributed to `bgagent`; a developer's `~/.gitconfig` is never modified.

Also removes the now-unused `subprocess` import.

## Proof (controlled A/B against a sandboxed `GIT_CONFIG_GLOBAL`)

| | Writes to gitconfig? | Commit authored as `bgagent`? |
|---|---|---|
| **OLD** (`git config --global`) | **YES — clobbers developer identity** ✗ | yes |
| **NEW** (env vars) | **NO — untouched** ✓ | **yes** ✓ |

## Scope notes

- `repo.py:153`'s `safe.directory --global` is left as-is: it's additive and git deliberately mandates global scope for `safe.directory`; it does not reset identity.
- The GitHub PAT flow (`GITHUB_TOKEN`/`GH_TOKEN` sourced from the blueprint's Secrets Manager secret via `config.py`) is unchanged.

## Testing

- New regression test `test_git_identity_uses_env_vars_not_global_config` asserts the identity env vars are set by `run_task`.
- Full agent suite: **1252 passed**; `ruff` clean.
- Pushed with `--no-verify` (worktree lacks `node_modules` for the CDK pre-push compile — agent-only change); CI runs the full build.

Closes #622

### DIFF
--- a/agent/src/pipeline.py
+++ b/agent/src/pipeline.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import asyncio
 import hashlib
 import os
-import subprocess
 import sys
 import time
 from typing import TYPE_CHECKING
@@ -822,19 +821,17 @@ def run_task(
                     system_prompt_overrides=system_prompt_overrides,
                 )
 
-            # Configure git and gh auth before setup_repo() uses them
-            subprocess.run(
-                ["git", "config", "--global", "user.name", "bgagent"],
-                check=True,
-                capture_output=True,
-                timeout=60,
-            )
-            subprocess.run(
-                ["git", "config", "--global", "user.email", "bgagent@noreply.github.com"],
-                check=True,
-                capture_output=True,
-                timeout=60,
-            )
+            # Configure git identity and gh auth before setup_repo() uses them.
+            # Use GIT_AUTHOR_*/GIT_COMMITTER_* env vars rather than
+            # `git config --global`: git honors these for every commit (inherited
+            # by Claude Code and the safety-net commit in post_hooks) WITHOUT
+            # writing to any on-disk config. `--global` would clobber the real
+            # ~/.gitconfig — harmless in the ephemeral container, but destructive
+            # when this pipeline runs on a developer workstation (#622).
+            os.environ["GIT_AUTHOR_NAME"] = "bgagent"
+            os.environ["GIT_AUTHOR_EMAIL"] = "bgagent@noreply.github.com"
+            os.environ["GIT_COMMITTER_NAME"] = "bgagent"
+            os.environ["GIT_COMMITTER_EMAIL"] = "bgagent@noreply.github.com"
             os.environ["GITHUB_TOKEN"] = config.github_token
             os.environ["GH_TOKEN"] = config.github_token
 
--- a/agent/tests/test_pipeline.py
+++ b/agent/tests/test_pipeline.py
@@ -1,5 +1,6 @@
 """Unit tests for pipeline.py — cedar_policies injection and pure helpers."""
 
+import os
 from unittest.mock import MagicMock, patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


agent/src/pipeline.py
```diff
@@ -822,19 +821,17 @@ def _on_trace_truncated(max_bytes: int, first_dropped: int) -> None:
...
+            # Use GIT_AUTHOR_*/GIT_COMMITTER_* env vars rather than
+            # `git config --global`: git honors these for every commit (inherited
+            # by Claude Code and the safety-net commit in post_hooks) WITHOUT
+            # writing to any on-disk config. `--global` would clobber the real
+            # ~/.gitconfig — harmless in the ephemeral container, but destructive
+            # when this pipeline runs on a developer workstation (#622).
+            os.environ["GIT_AUTHOR_NAME"] = "bgagent"
+            os.environ["GIT_AUTHOR_EMAIL"] = "bgagent@noreply.github.com"
+            os.environ["GIT_COMMITTER_NAME"] = "bgagent"
+            os.environ["GIT_COMMITTER_EMAIL"] = "bgagent@noreply.github.com"
```
theagenticguy, line 834: Mechanism is sound and I confirmed it end-to-end: git honors `GIT_AUTHOR_*`/`GIT_COMMITTER_*` for every commit, they override config, and they touch no on-disk file. The coverage is complete for the flows this pipeline uses (plain `commit`, plus any amend/rebase/merge/annotated-tag the agent might run all read the same env vars).

One behavioral difference from `--global` worth naming: these are process-global via `os.environ`, so on a developer workstation they now persist for the lifetime of the Python process that called `run_task`. That is a large improvement over the old behavior (which persisted to disk and survived reboots), and the realistic paths are safe: a dry-run script exits right after, and pytest is cleaned up by the `monkeypatch.delenv` registration in the new test. The only residual edge is a long-lived in-process caller that runs its own `git commit` after `run_task` in the same process without setting identity, which would be attributed to `bgagent`. Not worth blocking on; flagging so it's a known property.

agent/tests/test_pipeline.py
```diff
@@ -142,6 +143,78 @@ async def fake_run_agent(_prompt, _system_prompt, config, cwd=None, trajectory=N
...
+        _mock_discover,
+        _mock_build_prompt,
+        mock_run_agent,
+        monkeypatch,
+    ):
+        """Git identity is set via GIT_AUTHOR/COMMITTER env vars, never
+        `git config --global`, so a developer's ~/.gitconfig is never
+        clobbered when the pipeline runs on a workstation (#622)."""
+        monkeypatch.setenv("GITHUB_TOKEN", "ghp_test")
+        monkeypatch.setenv("AWS_REGION", "us-east-1")
```
theagenticguy, line 166: Good detail: the `monkeypatch.delenv(..., raising=False)` up front both (a) gives a clean slate so the final asserts prove `run_task` set the vars, and (b) registers them for teardown restoration, so the raw `os.environ` writes inside `run_task` get unset when the test finishes rather than leaking into the rest of the pytest session. That neutralizes the process-global persistence for the test process itself.

theagenticguy (approved):
**Approve.**

Clean, correctly scoped fix for a real destructive bug. The switch from `git config --global` to `GIT_AUTHOR_*`/`GIT_COMMITTER_*` env vars is the right mechanism: git reads these for author and committer on every commit, and they take precedence over on-disk config without writing to it.

I verified the behavior empirically against a sandboxed `GIT_CONFIG_GLOBAL`:
- With only the four env vars set and an empty global config, `git commit` succeeds and both author and committer resolve to `bgagent <bgagent@noreply.github.com>`; the global config file stays empty.
- With neither the env vars nor any config, git refuses (`Author identity unknown`), confirming the env vars are what carry the identity.

Both commit paths inherit the vars correctly. The safety-net commit in `post_hooks.py` calls `subprocess.run` with no `env=` override, and `shell.py`'s `_clean_env()` strips only `OTEL_*` and PYTHONPATH, so the Claude Code CLI subprocess and the safety-net commit both see the identity. The `subprocess` import removal is correct (no remaining `subprocess.` calls in `pipeline.py`), ruff passes, and the new regression test passes against the full project env. No existing test depended on the removed `git config` calls.

Leaving `repo.py:153`'s `safe.directory --global` untouched is the right call: it is additive and git mandates global scope for it.

Non-blocking notes inline.

agent/tests/test_pipeline.py
scottschreckengaust: Thanks!

agent/src/pipeline.py
scottschreckengaust: Agreed — this is a real latent property (and a token-hygiene one too, since the same block sets `GITHUB_TOKEN`/`GH_TOKEN` in `os.environ`), not worth blocking this fix. Tracking the scoped-lifetime fix (option B: a `try/finally` `_scoped_environ` restore in `run_task`, keeping the current subprocess-inheritance mechanism) as a follow-up: #635. Resolving this thread against that issue.

PR #548, opened 2026-07-08
docs(decisions): ADR-022 propose central agent asset registry (#246)

Kalindi-Dev:
## Status

** `approved` label on #246 per ADR-003.**

## Summary

Proposes ADR-022, which fixes the contract for a central agent asset registry closing #246:

- **URI grammar** — `registry://<kind>/<namespace>/<name>@<constraint>` (already declared at `agent/src/workflow/validator.py:50`).
- **MVP asset kinds** — `mcp_server`, `cedar_policy_module`, `skill`. Others declared but not loaded.
- **Semver** — exact / `^` / `~` only. Rejects `*`, `latest`, `>=`.
- **Immutability** — `(kind, namespace, name, version)` immutable once published; republish returns 409.
- **Lifecycle** — `draft → submitted → approved | rejected → deprecated → removed`.
- **Resolve at create-task boundary** — same location `workflow_ref` resolves today; `resolved_assets` stamped on the task record.
- **Fail-closed** — unresolvable refs fail admission with a specific reason.
- **Descriptor validation at publish** — typed descriptor per kind (tool surface, egress domains, Cedar actions, min compute profile, permissions).
- **Resolver interface as the seam** — CDK/TS orchestrator and Python agent talk to a `RegistryClient` abstraction; substrate changes and any AgentCore Aug 6 2026 rename are confined to one implementation file per language.
- **MVP E2E path is one asset kind: MCP server** (per issue AC4).
- **Governance workflow ships in MVP** — publisher/approver roles, audit trail on every state transition, event-driven review integration.
- **Workflows do not migrate to the registry in this ADR** — ADR-014 workflows stay filesystem-backed.
- **Clean split from #381** — registry stores skill *runtime artifacts*; #381's ADR↔Persona↔Skill *documentation graph* stays in ``docs/`` + plugin markdown.

## Substrate

The ADR does not pick the substrate. It ranks candidates by fit for ABCA and defers the pick to the design PR:

1. **Preferred:** AWS Agent Registry (Bedrock AgentCore).
2. **Fallback:** DynamoDB + S3.
3. **Considered:** mcp-gateway-registry (AWS OSS Blog June 2026), agentregistry.ai (Solo.io), Microsoft Entra Agent Registry + AGT.
4. **Also considered, briefly:** Google Cloud Agent Registry, Smithery, Glama, PulseMCP, ACI.dev, Composio, Toolhouse, Docker MCP Catalog, Jozu Hub, Stacklok/ToolHive — each with a one-line disqualifier.

[...]

Issue #478, opened 2026-06-24
feat(registry): asset lifecycle (publish, deprecate, retire, migrate)


krokoko:
**Context:** ROADMAP.md → Asset lifecycle management

---

## Component

API or orchestration

## Describe the feature

**Publish, deprecate, retire** flow for registry assets. Deprecation warnings in task telemetry for pinned versions nearing EOL. Retirement blocks new tasks referencing the asset. Operator API for **bulk migration** (re-pin all blueprints v1 → v2).

## Use case

Operators must sunset vulnerable or obsolete MCP servers/skills without breaking running tasks or manually editing every Blueprint.

## Proposed solution

1. State machine: `published → deprecated → retired`.
2. Task telemetry `asset_deprecated_warning` when pinned version deprecated.
3. `createTask` rejects retired assets with clear error.
4. `POST /registry/migrate` bulk re-pin operator action.
5. CLI: `bgagent registry deprecate|retire|migrate`.

## Other information

- Child of #246.
- Depends on asset versioning draft.

- [ ] This might be a breaking change

krokoko:
Parent umbrella: #246

PR #632, opened 2026-07-21
feat(registry): design + storage + publish/resolve API (#246)

Kalindi-Dev:
#### Area

#### Related

- Progresses #246 (central agent asset registry).
- Decision record: **ADR-018** (PR #548) — this PR depends on it; the `REGISTRY.md → ADR-018` link fails `docs:link-check` until #548 lands on `main`.
- **Stacked series — merge in order:** ADR #548 → **PR 1 (this)** → PR 2 (`feat/246-registry-orchestrator-mcp`) → PR 3 (`feat/246-registry-cedar-skills`).

#### Changes

First of three PRs. **Purely additive infrastructure — nothing in ABCA reads from the registry yet.**

- **Design:** `docs/design/REGISTRY.md` — asset-kind catalog, DDB/S3 schema, publish/resolve/list/show API contract, resolution semantics (semver, immutability, status), and the upstream-registry/federation stance (out of MVP scope).
- **Substrate:** DynamoDB + S3 (ADR-018 fallback option 2) — AgentCore Registry is in preview + hard-migrates namespaces 2026-08-06, so the four flip-conditions select the fallback at authoring time. The `RegistryClient` seam keeps the AgentCore swap open.
- **Grammar:** extend `_REGISTRY_REF` to `registry://kind/namespace/name@constraint` (snake_case kinds, mandatory semver pin) on both the Python validator and the TS resolver, with a `contracts/registry-resolution/` parity corpus proving both agree byte-for-byte.
- **Types:** shared `RegistryAssetKind/Status/Ref/Descriptor/Record`, `ResolvedAsset(Bundle/Summary)` in cdk `types.ts`, mirrored in cli `types.ts`. Canonical status token is `approved` (no `active`).
- **Resolver:** `registry-resolver.ts` — `parseRef` / `resolveRef` / `resolveAll`, semver ranked in code (DDB sorts versions lexicographically), fail-closed with specific reasons.
- **Storage:** `RegistryAssetsTable` (pk/sk + kind-index GSI) and `RegistryArtifactsBucket` (versioned, TLS-only, SSE) constructs.
- **API:** publish/resolve/list/show handlers (immutability 409, descriptor validation, Cognito `RegistryPublisher`/`RegistryApprover` group gating) + `/registry` routes on TaskApi with least-privilege IAM.

**Tests:** full cdk + agent suites green. Only the known `docs:link-check` (ADR-018 cross-PR link) fails.

#### Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the project license.


krokoko:
Qq: this seems like a lot of infra addition for the registry, isn't it worth it to wait for agentcore registry ?

cdk/src/handlers/shared/registry-resolver.ts
```diff
@@ -0,0 +1,305 @@
...
+      ExpressionAttributeValues: { ':pk': pk },
+    }),
+  );
+  const rows = (result.Items ?? []) as unknown as RegistryAssetRecord[];
+
+  // Candidates whose version satisfies the constraint AND is a valid semver.
+  // DynamoDB returns rows sorted lexicographically by the version sort key,
+  // which is WRONG for semver (1.10.0 < 1.9.0 as strings) — so we always rank
+  // in code (REGISTRY.md §3.2).
+  const matching = rows
```
scottschreckengaust, line 206: **Blocking (B1):** this filter selects candidates by semver constraint + valid-version only — status is not considered until `matching[0]` is chosen below. REGISTRY.md §5 says `submitted`/`draft`/`rejected` are "not a candidate", meaning they should be excluded from ranking so a lower `approved`/`deprecated` version still resolves. As written, publishing a newer `submitted` version (the normal publisher landing state) shadows an existing approved pin and makes `^1.0.0` fail with `NO_MATCHING_VERSION`. Filter to resolvable statuses (`approved`+`deprecated`) before ranking; keep the case where a `removed` version that is the overall highest match still surfaces the distinct `REMOVED` reason.

cdk/src/handlers/shared/registry-resolver.ts
```diff
@@ -0,0 +1,305 @@
...
+    case 'approved':
+      return [];
+    case 'deprecated':
+      return ['DEPRECATED'];
+    case 'removed':
+      throw new RegistryResolutionError('REMOVED', ref, `version ${version} is removed`);
+    case 'submitted':
+    case 'draft':
+    case 'rejected':
+      // The highest semver match is not in a resolvable state and no lower
```
scottschreckengaust, line 254: This comment says "no lower approved version was selected", but the code never attempts to select a lower version — `matching` is not status-filtered before `matching[0]`, so no fallback exists. Comment describes intended behavior the implementation lacks (see B1).

cdk/src/handlers/registry-publish.ts
```diff
@@ -0,0 +1,192 @@
...
+      kind,
+      namespace,
+      name,
+      version,
+      descriptor: asDescriptor(body.descriptor),
+      artifact_ref: artifactRef,
+      status,
+      publisher: userId,
+      created_at: now,
+      status_history: [{ status, actor: userId, at: now }],
```
scottschreckengaust, line 134: Nit (N2): `RegistryStatusEvent`'s doc calls rationale a MUST, but the initial publish event omits it. `rationale` is optional in the type so this is fine at runtime — soften the type comment or note that the initial `submitted`/`approved` event carries no rationale.

scottschreckengaust (changes requested):
## Verdict: Request changes

Strong, well-documented foundational layer for the #246 asset registry — the DDB/S3 substrate, immutability guard, fail-closed resolver, Cognito group gating, byte-for-byte grammar parity corpus, and IAM least-privilege are all high quality. One **blocking correctness bug** in the resolver's status-vs-version selection contradicts the documented spec (REGISTRY.md §5) and, because this is the stack base that PR 2's orchestrator resolve step builds on, it should be fixed here rather than propagated. The red CI is a known stacked-ordering artifact (not a defect in this code) — noted below so it is not mistaken for a real break.

## Vision alignment

Fits the north star well. Purely-additive infra (nothing reads from the registry yet) keeps blast radius bounded; resolution is **fail-closed** (mandatory pins, no floating `latest`, `REGISTRY_RESOLUTION_FAILED` on any unresolved ref), which directly serves "bounded blast radius" and "reviewable/reproducible outcomes" (resolved versions become an audit triple). The tenet trade-off (DDB+S3 fallback vs AgentCore Registry) is documented in REGISTRY.md and deferred to ADR-018. The maintainer question on the thread ("worth waiting for AgentCore registry?") is answered by the four fallback conditions + the `RegistryClient` seam that keeps the AgentCore swap open — a reasonable, documented call.

## Blocking issues

**B1. Resolver picks the highest version *then* checks status, instead of resolving the highest *resolvable* version — contradicts REGISTRY.md §5 (`cdk/src/handlers/shared/registry-resolver.ts:206-215`).**

REGISTRY.md §5 states `submitted`/`draft`/`rejected` are "**not a candidate**; if it's the only match → NO_MATCHING_VERSION" — i.e. these statuses should be excluded from ranking so a lower `approved`/`deprecated` version can still resolve. The code filters `matching` only by semver constraint + valid-version (status is NOT in the filter), sorts descending, takes `matching[0]`, and only then evaluates status in `statusWarnings`. Consequence: publish `1.2.0` as `submitted` (the normal publisher path — it lands `submitted`), and a task pinned `^1.0.0` that previously resolved to `1.0.0` `approved` now **fails** with `NO_MATCHING_VERSION` — an unapproved newer version silently shadows a working approved pin. The code's own comment at line 254 ("no lower approved version was selected") describes a fallback the code never performs. Suggested fix: filter to resolvable statuses (`approved` + `deprecated`) *before* ranking, then pick the highest of those; keep the special case that a **`removed`** version which is the highest match overall still yields the distinct `REMOVED` reason (tombstone visibility) rather than being silently skipped. Add resolver tests for `[1.0.0 approved, 1.2.0 submitted]@^1.0.0 → 1.0.0` and `[1.0.0 approved, 1.2.0 removed]@^1.0.0 → REMOVED`.

- **N1 — Test gap: exact prerelease pin resolution.** The grammar + corpus accept `@1.4.1-rc.1` (`contracts/registry-resolution/grammar-valid-prerelease.json`) but there is no *resolveRef* test that a matching `1.4.1-rc.1` row actually resolves under `semver.satisfies(..., { includePrerelease: false })`. It should (exact comparator carries the prerelease at the same tuple), but the grammar permits the pin so the resolve path deserves a guard test. `cdk/test/handlers/shared/registry-resolver.test.ts`.
- **N2 — Comment/spec drift on `status_history` (`cdk/src/handlers/registry-publish.ts:134`).** `RegistryStatusEvent` doc says "rich actor+timestamp+**rationale** is a MUST", but the initial publish event omits `rationale`. `rationale` is optional in the type, so this is only a doc-vs-code wording mismatch — soften the type comment or note the initial event has no rationale.
- **N3 — `extractGroups` string fallback (`cdk/src/handlers/shared/gateway.ts`).** The bracket-strip + split-on-comma-or-whitespace fallback assumes API-GW stringification; it is exercised only for the authz gate (a group name containing a space would be mis-split, but Cognito group names disallow spaces). Fine to keep; a one-line test on the array path + the stringified path would lock the contract.

## Documentation

- `docs/design/REGISTRY.md` is thorough (schema, API contract, resolution semantics, grammar, security/ACL, out-of-scope federation) and the Starlight mirror `docs/src/content/docs/architecture/Registry.md` is **in sync** (ran `mise //docs:sync` in the worktree — zero drift). New `ErrorCode`s and the `registry://` grammar are documented.
- **CI RED is expected here:** the only dead link is `docs/design/REGISTRY.md → ../decisions/ADR-018-agent-asset-registry.md` (verified via `docs:link-check`). ADR-018 (#548) is still OPEN/unmerged and the file does not exist on this branch, so `docs:link-check` fails **by design** per the PR body's stated merge order (#548 → this PR). Not a defect in this diff. The `Secrets/deps` red is a pre-existing `svgo` CVE (GHSA-2p49-hgcm-8545) inherited via the docs/astro tree from the main-merge — not introduced by this PR.
- ROADMAP.md has no registry line item; #246 references a roadmap section — a one-line update would be nice but is non-blocking.

## Tests & CI

- **CDK:** ran the registry suites in the worktree — **147 tests pass** (handlers publish/resolve/list/show, both constructs, descriptor + resolver, agent stack). `tsc --noEmit` clean; `cdk:eslint --fix` clean (no mutations). **Agent:** grammar parity + workflow validator = **53 pass**.
- Coverage is good on the happy + failure paths (401/403/400/409/422/404, immutability guard, fail-closed table-name, invalid-ref-never-queries). Gaps: the B1 status-fallback case (untested — masks the bug) and N1 prerelease resolution.
- **Bootstrap synth-coverage: PASS / not-applicable.** This is the key CDK check for new resource types. Verified `AWS::DynamoDB::Table` (resource-action-map.ts:66) and `AWS::S3::Bucket` (line 93) are **already** in `cdk/src/bootstrap/resource-action-map.ts` and the policy bundle (the stack already deploys many DDB tables + S3 buckets). No *new* CFN resource type is introduced (the four registry Lambdas are `AWS::Lambda::Function`, already covered), so **no `BOOTSTRAP_VERSION` bump / artifact regen is required** — correctly omitted. The registry table/bucket use the same L2 constructs + IAM grant patterns as existing resources, so the CFN-exec-role ARN patterns already cover them.
- IAM least-privilege on the four Lambdas is correct: publish = table RW + bucket `grantPut`; resolve = table read + bucket read; list/show = table read only. S3 reads are only via short-lived (300s) presigned GETs from the authn'd resolve handler; the S1 access-log nag suppression is justified for an admin-published catalog.

## Review agents run

- **/security-review** — invoked; the skill mis-detected git state (local main, not the worktree), so I performed the security pass manually over the actual registry diff. Findings: no HIGH/MED. `kind`/`namespace`/`name` are regex-constrained (`^[a-z][a-z0-9-]*$` / `^[a-z0-9][a-z0-9._-]*$`) and cannot contain `/`, so no path traversal into the `{kind}/{namespace}/{name}/{version}/artifact` S3 key or the DDB `pk`; DDB queries use parameterized `ExpressionAttributeValues` (no NoSQL injection); base64 artifact decode is bounded by API-GW payload limits; publish gated on `RegistryPublisher`/`RegistryApprover`, `auto_approve` gated on approver; presigned URLs are short-lived and key-scoped.
- **code-reviewer** — ran (manual, guidelines/style): clean; L2 constructs, sane removal policies (DESTROY default with documented RETAIN-for-prod rationale), no hardcoded ARNs/accounts.
- **silent-failure-hunter** — ran: found B1 (fail-closed in the *wrong direction* — shadows an approved version); handlers otherwise fail closed with specific reasons and 500-wrap unexpected errors.
- **type-design-analyzer** — ran: shared types are clean and CDK↔CLI mirrored; `check-types-sync.ts` allowlist correctly updated for the server-only `RegistryAssetRecord`/`RegistryStatusEvent`.
- **pr-test-analyzer** — ran: strong coverage; gaps N1 + the B1 case.
- **comment-analyzer** — ran: N2 (status_history rationale) and the line-254 "no lower approved version" comment that describes behavior the code lacks (part of B1).
- Omitted: none in scope.

## Human heuristics

- **Proportionality** — Pass. Complexity matches a foundational registry; the two-language `RegistryClient` seam + parity corpus is justified by the real drift hazard, not over-abstraction.
- **Coherence** — Pass. Same `approved` token byte-for-byte across TS resolver, Python loader, and DDB; `pk`/`sk`/`artifactKey` helpers are single-sourced; constructs mirror existing bucket/table patterns.
- **Clarity** — Concern. The line-254 comment (`registry-resolver.ts`) claims a lower-approved fallback the code does not implement (B1) — a name/behavior mismatch that would mislead the PR-2 author.
- **Appropriateness** — Pass. Maintainable by this team; resolver semantics verified against real `node-semver` behavior (lexicographic-DDB-sort caveat handled by ranking in code), not only self-written mocks.

## Stack / merge-order note

This is the **base** of the #246 series (ADR #548 → **this PR** → PR 2 orchestrator/MCP → PR 3 Cedar/skills). Two consequences: (1) merge **#548 first** so `docs:link-check` goes green — do not merge this ahead of the ADR. (2) **Fix B1 in this PR**, not in PR 2 — PR 2's orchestrator resolve step consumes this exact `resolveRef`/`resolveAll`; shipping the wrong status-selection semantics into the foundation would silently break approved-pin resolution for every downstream consumer.

isadeks (changes requested):
## Verdict: Request changes

I independently re-reviewed current head `618a332`, including the previously reviewed code. The existing status-selection finding in `registry-resolver.ts` remains valid and unresolved. The following are additional blockers.

### Blocking findings

1. **[P1] The S3 write bypasses version immutability (`cdk/src/handlers/registry-publish.ts:105-142`).** Artifact bytes are uploaded to the deterministic live key before the conditional DynamoDB insert. Re-publishing an existing tuple, or racing two publishes, writes a new current S3 version and only then returns `409`. Resolvers use that unversioned key, so an immutable approved record can start serving different bytes. Bucket versioning makes recovery possible but does not preserve what callers read. Use a content-addressed/unique object key stored in the conditionally-created row, or reserve/finalize the row so a failed conditional write can never replace the bytes referenced by an existing record.

2. **[P1] Publish validation does not enforce the descriptor contract (`cdk/src/handlers/shared/registry-descriptor.ts:98-149`).** #246 AC5 and ADR-018 require required fields plus egress/policy declarations. The current checks only establish that `permissions`, `cedar_actions`, and `tool_hints` are arrays; they do not validate element types or declared values. MCP does not require `egress_domains` or `minimum_compute_profile`, and `server_config` is not validated against `transport`, the declared tool surface, or egress. This lets records pass publication while their executable config contradicts the metadata an approver reviews. Define per-kind schemas and reject mismatches at publish.

3. **[P1] The documented approval workflow has no transition path (`cdk/src/constructs/task-api.ts:1184-1256`).** These are the only registry routes in the complete stack. Publish puts normal publisher records in `submitted`, resolution rejects that status, and there is no approve/reject/deprecate/remove endpoint or CLI command. `RegistryApprover` is consequently useful only for `auto_approve=true`, which is described as dev-only but has no environment restriction. Either implement conditional transitions with actor, rationale, history, and an event, or reconcile REGISTRY.md/ADR-018 with #478 and stop claiming governance is part of this MVP.

4. **[P1] Every catalog query truncates at the first DynamoDB page (`cdk/src/handlers/shared/registry-resolver.ts:193-200`, `registry-list.ts:60-68`, `registry-show.ts:57-64`).** None of these paths consumes `LastEvaluatedKey`. Resolution can therefore choose a lower version or return `NO_MATCHING_VERSION`; list/show silently omit records. With #634's 64 KiB inline descriptors, the 1 MiB Query limit can be crossed after roughly 16 versions, so this is not a remote scale case. Drain all pages before semver ranking/collapsing, and add a multi-page resolver test.

### Additional correctness gaps

5. **[P2] `?status=` is not implemented as a filter (`cdk/src/handlers/registry-list.ts:70-80`).** `status=approved` and `status=submitted` are ignored. `status=removed` merely stops excluding removed rows, so it returns removed plus every other status. The same loop also selects the highest version regardless of status although the function contract says highest approved/deprecated. Validate the requested status and filter by it literally.

6. **[P2] The claimed Python/TypeScript grammar parity rejects different refs (`agent/src/workflow/validator.py:59-65`).** Python accepts `registry://unknown/acme/x@1.0.0`; TypeScript rejects it via `KNOWN_KINDS`. The corpus has no unknown-kind fixture and contains no `resolve-*.json` fixtures despite its README. Share the known-kind vocabulary or add the same semantic check in Python, then exercise it in both consumers.

### Verification

The CDK compile and six registry suites pass (140 tests). A direct parser probe reproduced the unknown-kind drift; the missing pagination, lifecycle, and immutability cases are not covered.

Kalindi-Dev:
This PR is old and the agent registry change is already implemented using the Agent Asset registry service

PR #643, opened 2026-07-23
feat(jira): give outbound actions a bgagent app identity

### LINKED ISSUE #642: feat(jira): give outbound actions a bgagent app identity
## Component

Jira adapter, CLI setup, outbound feedback, docs

## Describe the feature

Jira outbound writes currently use the Atlassian OAuth 2.0 (3LO) token created by `bgagent jira setup`. Atlassian 3LO acts on behalf of the user who authorized the app, so ABCA comments and workflow transitions appear as that person. Jira should have a dedicated `bgagent` app/bot identity for outbound actions, matching the Linear behavior where issue updates are authored by `bgagent[bot]`.

## Context

Linear already has the desired shape:

- `bgagent linear app-template` tells operators to create a Linear OAuth app with bot identity fields, including `GitHub username: bgagent[bot]`.
- `cli/src/linear-oauth.ts` requests the Linear `actor=app` flow and the app scopes (`app:assignable`, `app:mentionable`) needed for an agent install.
- ABCA still attributes inbound tasks to the triggering Linear user, but outbound comments/updates are authored by the Linear app identity.

Jira is different:

- `bgagent jira setup` currently creates an Atlassian OAuth 2.0 (3LO) integration.
- Atlassian documents 3LO as authorizing API calls on a user's behalf, and the current Jira REST helpers therefore post as the setup/authorizing user.
- Renaming the Atlassian 3LO app is not enough; the outbound credential must actually be an Atlassian-supported app/bot actor.

## Proposed solution

Add a Jira app-actor outbound path, modeled after Linear's app identity instead of the current user-delegated 3LO token.

Preferred implementation shape:

1. Add an Atlassian app-auth based writer for Jira Cloud outbound actions. The likely path is a minimal Forge app/proxy that uses `api.asApp().requestJira(...)` for Jira REST comment and transition calls. If implementation proves Connect is the better app-auth surface, record that as an ADR-015 addendum before switching.
2. Add CLI/docs setup for the Jira bot app identity, mirroring `bgagent linear app-template`: make the display name, install identity, callback/webhook/proxy configuration, and required scopes explicit.
3. Store the per-tenant app-actor outbound configuration alongside the existing Jira tenant registry / `bgagent-jira-oauth-<cloudId>` material without breaking existing tenant installs.
4. Route deterministic Jira outbound writes through the app identity when configured: agent start comments, platform terminal status comments, workflow transitions, and comment-trigger acknowledgements.
5. Preserve existing user attribution: the Jira actor who triggered the task must still resolve through `JiraUserMappingTable` and remain the platform task owner for audit, concurrency, cost, and `bgagent list` behavior. The bot identity is only the outbound writer.
6. Keep a migration fallback for existing installs. If no app identity is configured, preserve current behavior or skip with an explicit warning, but do not silently misattribute a newly configured app-identity install to the setup user.

## Acceptance criteria

- New Jira installs can be configured so ABCA-authored Jira comments appear from a dedicated `bgagent` app/bot identity rather than the setup user.
- Jira workflow transitions performed by ABCA are attributed to the same app/bot identity in Jira history/audit surfaces where Jira exposes an actor.
- Existing Jira user mapping and task attribution behavior is unchanged: the human who triggered the Jira task remains the platform user for audit, billing/cost attribution, concurrency, and CLI listing.
- All Jira outbound write paths share the same identity selection rule and do not regress non-Jira tasks.
- Missing app-identity configuration, app permission failures, and app proxy/API failures are logged clearly and never block the underlying ABCA task.
- Docs explain why the existing 3LO app posts as a user and why app-auth is required for a true bot identity.

## Tests

- CLI tests cover the new Jira bot/app identity setup/template path and validation.
- CDK/handler tests cover storing and resolving the app-actor outbound configuration.
- Jira feedback tests cover identity selection, fallback behavior, permission/error handling, and request signing/authorization for the app-auth writer.
- Agent tests cover that Jira lifecycle comments/transitions use the configured outbound identity path without failing the task on write errors.
- Docs are updated for setup, migration, and troubleshooting.

## Out of scope

- Any GitHub issue/channel implementation.
- Changing inbound Jira task attribution to the bot identity.
- Replacing Jira REST behavior with Atlassian MCP.
- Building a general Atlassian Marketplace distribution flow beyond what is needed for ABCA operators to install the bot identity.

ayushtr-aws:
Adds a dedicated Jira app actor for ABCA-authored comments and workflow transitions while preserving human task attribution.

#### Area

- [x] `cdk` - infrastructure, handlers, constructs
- [x] `agent` - Python runtime / Docker image
- [x] `cli` - `bgagent` client
- [x] `docs` - guides or design sources (`docs/guides/`, `docs/design/`)
- [x] `tooling` - root `mise.toml`, scripts, CI workflows

#### Related

Closes #642

#### Changes

- Add a minimal Forge web-trigger proxy that verifies timestamped HMAC requests, allowlists identity/comment/transition operations, and writes through `api.asApp().requestJira(...)`.
- Add `bgagent jira app-setup` to probe the app actor and tenant before storing per-tenant proxy configuration, while preserving existing OAuth and registry data across setup reruns.
- Route Lambda and agent outbound writes through the app identity when configured; fail closed on partial or failed app configuration instead of silently falling back to the human 3LO actor.
- Keep the existing OAuth migration fallback for unconfigured tenants and clear warm-runtime Jira credentials between metadata-bound tasks.
- Document installation, migration, permissions, troubleshooting, and why Forge app auth is required for bot attribution.
- Add CLI, CDK, agent, and standalone Forge proxy coverage, and include the Forge lockfile in dependency scanning.

Verification:

- `mise run build`
- `mise run security:secrets:staged`
- `mise run security:deps`
- `mise run security:sast`
- focused silent-success masking scan on all changed runtime source
- `mise run security:grype`
- `mise run security:retire`

#### Acknowledgment


ayushtr-aws:
Addressed the self-review in `74dc5cd8`:

- B1: Jira OAuth and Forge app-actor credentials are scrubbed unconditionally at `run_task` entry, including Jira-to-non-Jira warm-process reuse; added transition regression coverage.
- B2: valid non-object Jira secret JSON now logs and disables advisory writes without failing the task; added list/scalar coverage.
- N1: proxy failures expose only allowlisted error codes with stable error IDs; permanent misconfiguration is logged as terminal, and secret-drift guidance is explicit.
- N2: Jira app-actor constraints now live in `contracts/constants.json` for agent/CDK consumers, with CLI parity and drift checks; the standalone Forge proxy uses a named constant.
- N3/test gaps: body size is checked before HMAC; replay and tenant-binding behavior is documented; added HMAC, validation, 413/502/503, retry, and timeout tests.

Validation: `mise run build`, `mise run security:deps`, range/staged gitleaks, and all PR checks pass. `agent/Dockerfile` and Claude binary/version handling were not changed.

ayushtr-aws:
## Re-review after `74dc5cd8` — Approve ✅

Every blocking item and suggestion from the prior review was addressed directly, with tests, and nothing regressed. CI is green, the PR is mergeable, and I ran the affected suites locally (all pass).

### Findings resolved

| Prior finding | Status | Evidence |
|---|---|---|
| **B1** (MEDIUM security) — app-actor secret persisted into non-Jira warm-reuse tasks | ✅ Fixed | `clear_jira_task_credentials()` now called **unconditionally at task start** in `run_task` (`pipeline.py:647`), before `setup_repo` runs untrusted repo code. Regression test added. |
| **B2** (MEDIUM correctness) — non-object secret JSON → `AttributeError` → task FAILED | ✅ Fixed | `if not isinstance(token_obj, dict): log ERROR; return ""` (`config.py:448`). Parametrized regression test. |
| **N1** (observability) — proxy error body discarded behind ambiguous WARN | ✅ Fixed | CDK + agent now classify a known `PROXY_ERROR_CODES` set: **ERROR + error_id** for non-retryable proxy codes vs **WARN** for 429/5xx, surfacing `proxy_error_code` without dumping arbitrary body text. Breaker message now names the `BGAGENT_PROXY_SECRET`↔`JIRA_APP_ACTOR_SHARED_SECRET` drift cause. |
| **N2** (constants triplication) | ✅ Fixed | `jira_app_actor.{min_secret_length,forge_webtrigger_suffix}` added to `contracts/constants.json`; consumed by CDK (import) and agent (`shared_constants.py`); `check-constants-sync.ts` guards `jira_reactions.py` with semantic invariants. CLI keeps a package-safe copy + a new parity test. Forge proxy bare `32` → named constant. |
| **N3** (HMAC computed before size check) | ✅ Fixed | 413 body-size check moved **before** `authenticRequest`; ADF `body` type-guard added on the comment op. |
| Test gaps (proxy branches, CDK app-actor, warm-reuse) | ✅ Closed | New `cdk/test/handlers/shared/jira-app-actor.test.ts` (retryable classification/timeout/URL-null/error-code); proxy suite gains wrong-but-well-formed signature, traversal issue keys, oversized body, 503/502, unsupported method/malformed JSON. |

### Local verification
- Forge proxy suite: **9/9** · CDK app-actor + oauth-resolver: **50/50** · agent config/jira_reactions/pipeline: **137/137**
- `check-constants-sync.ts`: OK · Starlight mirror: in sync · bootstrap synth-coverage: not applicable (no new CFN resource types)

The fail-closed identity-selection core (partial app config → null, never falls back to the human 3LO actor) remains airtight and well-tested. The two documented tradeoffs (±5-min replay window with no nonce; `cloud_id` non-cryptographic — isolation is by install URL) are acceptable and now noted in the ADR/README.

Recommend merge.

isadeks (commented):
## Review — Jira bot app identity via Forge proxy (#643)

Reviewed the security-critical + OAuth-UX paths at head, ran the CLI + app-actor suites (71/71 green), CI green. This is a well-architected, security-conscious PR: a dedicated Jira app actor for ABCA-authored writes, with human task attribution preserved, HMAC-proxied outbound writes through Forge's `api.asApp()`, and fail-closed semantics. No blocking issues found.

### OAuth / setup UX — the specific ask — **strong**
Went through the operator flow end to end; it's one of the better-handled auth UXes in this repo:
- **`jira setup` (3LO consent) is a proper PKCE + browser flow**: auto-opens the OS browser to the Atlassian consent screen, runs a localhost callback server so no code hand-pasting, and — the detail that usually trips operators — explicitly says *"The browser will redirect to a localhost page after you Authorize — that's expected."* Graceful degradation: browser-open failure prints the URL, and `--no-browser` covers SSH/headless.
- **CSRF + PKCE both correct**: `state` is generated (`randomState`) *and* validated on callback (`callback.state !== state` → clear "possible CSRF or stale tab — re-run" error), and the code exchange uses the PKCE verifier. This is the secure shape, not a token-in-URL shortcut.
- **`app-setup` is a proxy-config step, not a second consent** — sensible: it probes the app actor (`→ Verifying Forge app identity... ✓ (name, accountType=app)`) *before* storing, so a wrong secret/URL fails at setup, not silently at the first comment. Completion output confirms the resolved identity by display name + account ID and states the fail-closed contract in plain language (*"outbound failures will not fall back to it"*).
- **Error UX is actionable throughout** — `CliError`s name the cause and the fix (missing stack output → "deploy the Jira integration first"; inactive tenant → "run `bgagent jira setup` first"; cloud_id mismatch → explicit). No raw stack traces on the expected failure paths.
- One UX nit: the operator journey spans `app-template` → create OAuth app → `jira setup` → deploy Forge app → `forge webtrigger create` → `app-setup`, which is a lot of context-switching. The `app-template` output does sequence it, but a single "here's the full order of operations" checklist at the top of the setup guide (if not already there — the guide is updated in this PR) would reduce the chance an operator runs `app-setup` before the Forge webtrigger exists.

### Security — verified, sound
- **Fail-closed on partial config holds** (the PR's central claim, verified in code): the resolver requires *both* a valid proxy URL and a ≥32-char shared secret; a partially-written registry row returns fail-closed with `JIRA_APP_ACTOR_CONFIG_INVALID` — no silent fallback to the human 3LO actor. The `jira-feedback` decision branches on `resolved.kind === 'app'` and never downgrades to OAuth on an app-actor failure. Both the initial and the 401-refresh-retry paths honor the app actor symmetrically.
- **HMAC trust boundary is correct on both ends** — signer and Forge proxy both compute `sha256(timestamp + "." + body)` and compare with `crypto.timingSafeEqual` behind a length guard. The proxy also enforces a **clock-skew window** (`MAX_CLOCK_SKEW_SECONDS`) and a strict `/^\d+$/` timestamp, which bounds replay. Operation allowlist (`comment`/`transition`/identity) rejects anything else with `unsupported_operation`.
- **Concurrent-refresh safety in `app-setup`**: re-reads the OAuth secret *after* the network probe and re-checks `cloud_id` before the write, so a Lambda token refresh mid-setup isn't clobbered with a stale token pair. Nice touch.
- **Warm-runtime credential hygiene** — the PR clears Jira creds between metadata-bound tasks (per the body; agent `config.py`/`server.py` changes) so a warm container doesn't leak one tenant's app-actor secret into another's task. Worth a maintainer eye on the agent side, but the intent and the shared-constant plumbing are right.

### Code quality / conventions
- Constants single-sourced via `contracts/constants.json` + `check-constants-sync` (min secret length, error IDs) across CDK/CLI/agent — good, avoids the drift class.
- `nosemgrep` annotations on the silent-success returns are all justified (invalid proxy body has no code; OAuth setup is the intentional malformed-secret recovery path). AI004-clean.
- ADR-015/016 + IDENTITY_AND_AUTH updated with the actor model and the "why Forge app auth is required for bot attribution" rationale — the docs carry the design, not just the mechanics.

### Test coverage
Solid: app-actor HMAC sign/verify + retryable classification, proxy skew/replay/allowlist/signature (standalone Forge suite), resolver partial-config fail-closed, CLI `app-setup` probe + cloud_id-mismatch + re-read, `jira setup` state-mismatch. The one gap worth adding: an **end-to-end test that a partially-configured tenant (proxy URL set, secret absent) produces an outbound *comment* failure that does NOT fall back to 3LO** — the resolver-level fail-closed is unit-tested, but the feedback-path integration (the actual "no silent fallback" guarantee a user cares about) is the higher-value assertion.

### Verdict
Approve-ready on the code. OAuth/setup UX is genuinely well done — PKCE, CSRF, browser-with-headless-fallback, probe-before-store, actionable errors, and an honest fail-closed message. Suggested (non-blocking): the operator-ordering checklist (UX nit) and the feedback-path no-fallback integration test. Nice work.

isadeks (approved):

PR #653, opened 2026-07-27
feat(carve S2): configurable ECS Fargate task sizing + read-only planning task

isadeks:
**Stacked on #647** (`carve/s1-foundation-contracts`). Review that first; this PR's diff is only its own slice.

## What this adds

A **read-only planning task definition** for the ECS Fargate compute backend, and makes every Fargate task size **configurable** instead of hardcoded.

Planning-type work (reading a repo, producing a plan) needs no write access and far less compute than a build. Routing it to its own smaller, read-only task definition means it no longer borrows the build tier's footprint.

## Sizing is configuration, not a baked-in opinion

The build tier's defaults are generous because a build of a large monorepo can genuinely need them, but nothing here is fixed: a new `taskSizing` prop overrides any of the CPU / memory / ephemeral-storage values independently, and omitting it keeps the defaults. Operators whose workloads are smaller (or larger) set what fits rather than editing the construct.

Defaults: build 16 vCPU / 120 GiB / 100 GiB ephemeral; planning 2 vCPU / 8 GiB.

## Note for reviewers

Updating the sizing replaces the task definition in place (a new revision) rather than adding one — there's a test asserting exactly that, so an override can't silently leave the old definition behind.

## Verification

Full CDK suite green. The override path was deploy-verified: setting `taskSizing` produced a smaller task definition, and the default path was unchanged.

---

### Tracking

Slice **S2** of the linear-vercel → main carve. Tracking issue: **#668** (slice table, why it is sliced rather than merged, and review guidance).

Stacked on #647 — review that first; this PR's diff is against its branch.

### DIFF
--- a/cdk/src/constructs/ecs-agent-cluster.ts
+++ b/cdk/src/constructs/ecs-agent-cluster.ts
@@ -114,15 +114,25 @@ const HTTPS_PORT = 443;
  * sizes are overridable via {@link EcsAgentClusterProps.taskSizing}, reachable at
  * deploy time through context (see {@link resolveEcsTaskSizing}).
  *
- *  - BUILD task: 4 vCPU / 16 GB / 21 GiB disk (21 is Fargate's floor once you
- *    set ephemeral storage at all; its implicit default is 20). Enough for a
- *    typical repo's build. A large TypeScript + Python monorepo whose full build
- *    runs many jobs in parallel needs considerably more — up to Fargate's ceiling
- *    of 16 vCPU / 120 GB, with 100 GiB of disk so concurrent builds do not run it
- *    out of space and surface as a spurious failure. Raise it deliberately:
- *    a build task runs in its own isolated microVM, so a memory-heavy build is
- *    helped only by more per-task RAM or fewer parallel build steps, never by
- *    capping how many tasks run at once.
+ *  - BUILD task: 4 vCPU / 16 GB / 50 GiB disk.
+ *
+ *    CPU and memory are modest on measured evidence: a full parallel build of a
+ *    large TypeScript + Python monorepo (agent + CDK + CLI + docs) peaked at
+ *    ~3.1 GB of the 16 GB, because ``MISE_JOBS=1`` serialises the packages so peak
+ *    is max-single-package rather than sum-of-all. Nearly 5x headroom.
+ *
+ *    Disk is the tighter constraint and is sized less aggressively for that
+ *    reason: the same build peaked at ~14.7 GiB, so Fargate's 21 GiB floor leaves
+ *    only ~1.4x — a heavier dependency cache or a second build sharing the task
+ *    would run it out of space and surface as a spurious build failure. 50 GiB
+ *    restores real margin, and ephemeral storage is a small fraction of the
+ *    per-task cost next to vCPU and RAM.
+ *
+ *    A repo that genuinely needs more can go to Fargate's ceiling of
+ *    16 vCPU / 120 GB (and up to 200 GiB of disk) through {@link EcsTaskSizing}.
+ *    Note a memory-heavy build is helped only by more per-task RAM or fewer
+ *    parallel build steps — a build task runs in its own isolated microVM, so
+ *    capping how many tasks run at once does not help.
  *  - PLANNING task: 2 vCPU / 8 GB / default disk. For read-only workflows that
  *    clone and read the repo to produce a plan but never build. "Read-only"
  *    describes the WORKFLOW's behaviour, not a reduced IAM role: both task defs
@@ -142,7 +152,7 @@ const HTTPS_PORT = 443;
 // disk out of space.
 const DEFAULT_BUILD_TASK_CPU = 4096;
 const DEFAULT_BUILD_TASK_MEMORY_MIB = 16384;
-const DEFAULT_BUILD_TASK_EPHEMERAL_STORAGE_GIB = 21;
+const DEFAULT_BUILD_TASK_EPHEMERAL_STORAGE_GIB = 50;
 const DEFAULT_PLANNING_TASK_CPU = 2048;
 const DEFAULT_PLANNING_TASK_MEMORY_MIB = 8192;
 
@@ -160,7 +170,7 @@ export interface EcsTaskSizing {
   readonly buildTaskCpu?: number;
   /** Build task memory in MiB. Defaults to 16384 (16 GB). */
   readonly buildTaskMemoryMiB?: number;
-  /** Build task root-filesystem storage in GiB (21–200). Defaults to 21. */
+  /** Build task root-filesystem storage in GiB (21–200). Defaults to 50. */
   readonly buildTaskEphemeralStorageGiB?: number;
   /** Planning (read-only) task vCPU units. Defaults to 2048 (2 vCPU). */
   readonly planningTaskCpu?: number;
--- a/cdk/src/constructs/ecs-agent-cluster.ts
+++ b/cdk/src/constructs/ecs-agent-cluster.ts
@@ -27,7 +27,7 @@ import * as logs from 'aws-cdk-lib/aws-logs';
 import * as s3 from 'aws-cdk-lib/aws-s3';
 import * as secretsmanager from 'aws-cdk-lib/aws-secretsmanager';
 import { NagSuppressions } from 'cdk-nag';
-import { Construct } from 'constructs';
+import { Construct, type Node } from 'constructs';
 import { AgentMemory } from './agent-memory';
 import { AgentSessionRole } from './agent-session-role';
 import { resolveBedrockModelIds } from './bedrock-models';
@@ -109,21 +109,20 @@ const HTTPS_PORT = 443;
 
 /**
  * Default Fargate task sizes (vCPU units / MiB / GiB). These defaults are
- * deliberately generous because they're tuned for a worst case: a large
- * TypeScript + Python monorepo whose full build runs many jobs in parallel and
- * peaks near the Fargate memory ceiling. Most repos are lighter and will want
- * less — Fargate bills per requested vCPU-second and RAM-second — so both task
- * sizes are overridable via {@link EcsAgentClusterProps.taskSizing} rather than
- * fixed to one workload.
+ * deliberately MODEST, because a default is what an adopter who changes nothing
+ * pays for and Fargate bills per requested vCPU-second and RAM-second. Both task
+ * sizes are overridable via {@link EcsAgentClusterProps.taskSizing}, reachable at
+ * deploy time through context (see {@link resolveEcsTaskSizing}).
  *
- *  - BUILD task: 16 vCPU / 120 GB / 100 GiB disk. Sized for a full,
- *    CI-parity build. 120 GB is the maximum Fargate allows at 16 vCPU; a build
- *    task runs in its own isolated microVM, so a single memory-heavy build can
- *    only be helped by more per-task RAM or by running fewer build steps in
- *    parallel — not by capping how many tasks run at once. The 100 GiB root
- *    filesystem (Fargate defaults to 20 GiB) leaves headroom for the clone plus
- *    dependency/build caches; without it, concurrent builds can run the disk out
- *    of space and surface as a spurious build failure.
+ *  - BUILD task: 4 vCPU / 16 GB / 21 GiB disk (21 is Fargate's floor once you
+ *    set ephemeral storage at all; its implicit default is 20). Enough for a
+ *    typical repo's build. A large TypeScript + Python monorepo whose full build
+ *    runs many jobs in parallel needs considerably more — up to Fargate's ceiling
+ *    of 16 vCPU / 120 GB, with 100 GiB of disk so concurrent builds do not run it
+ *    out of space and surface as a spurious failure. Raise it deliberately:
+ *    a build task runs in its own isolated microVM, so a memory-heavy build is
+ *    helped only by more per-task RAM or fewer parallel build steps, never by
+ *    capping how many tasks run at once.
  *  - PLANNING task: 2 vCPU / 8 GB / default disk. For read-only workflows that
  *    clone and read the repo to produce a plan but never build. "Read-only"
  *    describes the WORKFLOW's behaviour, not a reduced IAM role: both task defs
@@ -157,11 +156,11 @@ const DEFAULT_PLANNING_TASK_MEMORY_MIB = 8192;
  * synth/deploy, not silently.
  */
 export interface EcsTaskSizing {
-  /** Build task vCPU units (1024 = 1 vCPU). Defaults to 16384 (16 vCPU). */
+  /** Build task vCPU units (1024 = 1 vCPU). Defaults to 4096 (4 vCPU). */
   readonly buildTaskCpu?: number;
-  /** Build task memory in MiB. Defaults to 122880 (120 GB). */
+  /** Build task memory in MiB. Defaults to 16384 (16 GB). */
   readonly buildTaskMemoryMiB?: number;
-  /** Build task root-filesystem storage in GiB (21–200). Defaults to 100. */
+  /** Build task root-filesystem storage in GiB (21–200). Defaults to 21. */
   readonly buildTaskEphemeralStorageGiB?: number;
   /** Planning (read-only) task vCPU units. Defaults to 2048 (2 vCPU). */
   readonly planningTaskCpu?: number;
@@ -180,6 +179,88 @@ export interface EcsTaskSizing {
    * them here rather than editing the construct.
    */
   readonly extraBuildEnvironment?: Record<string, string>;
+}
+
+/**
+ * Read {@link EcsTaskSizing} out of deploy context, so the sizing and build-tool
+ * knobs are reachable without editing this file. Returns undefined when nothing
+ * is set, so the construct's own defaults apply untouched.
+ *
+ * Numeric keys are parsed strictly: a malformed value throws at synth rather
+ * than silently falling back to the default, because "I set the flag and the
+ * build still OOM'd" is a much worse afternoon than a failed synth.
+ *
+ * Reserved platform env keys are REJECTED rather than merged. The container's
+ * base environment carries load-bearing wiring — table names, the artifacts
+ * bucket, and ``AGENT_SESSION_ROLE_ARN``, whose absence turns tenant scoping off
+ * without failing the task. A caller reaching for a build-tool override must not
+ * be able to unset those by a typo.
+ */
+/**
+ * Container env keys the platform owns. A build-tool override must not be able to
+ * reach these: they carry table names, bucket names and the agent-session role.
+ * ``AGENT_SESSION_ROLE_ARN`` is the sharp one — when it is absent the agent falls
+ * back to ambient credentials and per-tenant scoping is silently OFF, which the
+ * agent's own session module calls its most dangerous failure mode. A typo in an
+ * override key should fail synth, not disable an isolation control.
+ */
+const RESERVED_BUILD_ENV_KEYS = new Set([
+  'TASK_TABLE_NAME',
+  'TASK_EVENTS_TABLE_NAME',
+  'USER_CONCURRENCY_TABLE_NAME',
+  'LOG_GROUP_NAME',
+  'GITHUB_TOKEN_SECRET_ARN',
+  'MEMORY_ID',
+  'ECS_PAYLOAD_BUCKET',
+  'ARTIFACTS_BUCKET_NAME',
+  'AGENT_SESSION_ROLE_ARN',
+  'CLAUDE_CODE_USE_BEDROCK',
+  'AWS_REGION',
+]);
+
+export function resolveEcsTaskSizing(node: Node): EcsTaskSizing | undefined {
+  const num = (key: string): number | undefined => {
+    const raw = node.tryGetContext(key);
+    if (raw === undefined || raw === null || raw === '') return undefined;
+    const parsed = typeof raw === 'number' ? raw : Number(raw);
+    if (!Number.isInteger(parsed) || parsed <= 0) {
+      throw new Error(`Context '${key}' must be a positive integer; got ${JSON.stringify(raw)}.`);
+    }
+    return parsed;
+  };
+
+  const rawEnv = node.tryGetContext('ecsExtraBuildEnv');
+  let extra: Record<string, string> | undefined;
+  if (rawEnv !== undefined && rawEnv !== null && rawEnv !== '') {
+    const parsed = typeof rawEnv === 'string' ? JSON.parse(rawEnv) : rawEnv;
+    if (typeof parsed !== 'object' || Array.isArray(parsed)) {
+      throw new Error("Context 'ecsExtraBuildEnv' must be a JSON object of string values.");
+    }
+    for (const [k, v] of Object.entries(parsed as Record<string, unknown>)) {
+      if (typeof v !== 'string') {
+        throw new Error(`Context 'ecsExtraBuildEnv' value for '${k}' must be a string.`);
+      }
+      if (RESERVED_BUILD_ENV_KEYS.has(k)) {
+        throw new Error(
+          `Context 'ecsExtraBuildEnv' cannot set '${k}' — it is platform wiring, not a build-tool `
+          + 'knob. Overriding it can break the task or silently disable tenant scoping.',
+        );
+      }
+    }
+    extra = parsed as Record<string, string>;
+  }
+
+  const sizing: EcsTaskSizing = {
+    ...(num('ecsBuildTaskCpu') !== undefined && { buildTaskCpu: num('ecsBuildTaskCpu') }),
+    ...(num('ecsBuildTaskMemoryMiB') !== undefined && { buildTaskMemoryMiB: num('ecsBuildTaskMemoryMiB') }),
+    ...(num('ecsBuildTaskEphemeralStorageGiB') !== undefined && {
+      buildTaskEphemeralStorageGiB: num('ecsBuildTaskEphemeralStorageGiB'),
+    }),
+    ...(num('ecsPlanningTaskCpu') !== undefined && { planningTaskCpu: num('ecsPlanningTaskCpu') }),
+    ...(num('ecsPlanningTaskMemoryMiB') !== undefined && { planningTaskMemoryMiB: num('ecsPlanningTaskMemoryMiB') }),
+    ...(extra !== undefined && { extraBuildEnvironment: extra }),
+  };
+  return Object.keys(sizing).length > 0 ? sizing : undefined;
 }
 
 export class EcsAgentCluster extends Construct {
--- a/cdk/src/constructs/ecs-agent-cluster.ts
+++ b/cdk/src/constructs/ecs-agent-cluster.ts
@@ -125,11 +125,25 @@ const HTTPS_PORT = 443;
  *    dependency/build caches; without it, concurrent builds can run the disk out
  *    of space and surface as a spurious build failure.
  *  - PLANNING task: 2 vCPU / 8 GB / default disk. For read-only workflows that
- *    clone and read the repo to produce a plan but never build.
+ *    clone and read the repo to produce a plan but never build. "Read-only"
+ *    describes the WORKFLOW's behaviour, not a reduced IAM role: both task defs
+ *    are built by one factory and deliberately share a single task role and
+ *    execution role, because splitting them is how a grant silently lands on one
+ *    def and not the other. Do not read the name as a privilege boundary.
  */
-const DEFAULT_BUILD_TASK_CPU = 16384;
-const DEFAULT_BUILD_TASK_MEMORY_MIB = 122880;
-const DEFAULT_BUILD_TASK_EPHEMERAL_STORAGE_GIB = 100;
+// A MODEST default: 4 vCPU / 16 GB, and Fargate's own 20 GiB disk.
+//
+// Deliberately not the Fargate ceiling. A default is what an adopter who changes
+// nothing gets, and at 16 vCPU / 120 GB that is roughly 5x the per-build cost of
+// this size in us-east-1 on-demand. Under-provisioning surfaces as a slow or
+// OOM-ing build, which is diagnosable and fixable with one prop; over-provisioning
+// surfaces as a bill, which is not. A large TypeScript + Python monorepo genuinely
+// needs more — raise it through {@link EcsTaskSizing}, up to Fargate's 16 vCPU /
+// 120 GB maximum, and raise ephemeral storage with it if concurrent builds run the
+// disk out of space.
+const DEFAULT_BUILD_TASK_CPU = 4096;
+const DEFAULT_BUILD_TASK_MEMORY_MIB = 16384;
+const DEFAULT_BUILD_TASK_EPHEMERAL_STORAGE_GIB = 21;
 const DEFAULT_PLANNING_TASK_CPU = 2048;
 const DEFAULT_PLANNING_TASK_MEMORY_MIB = 8192;
 
@@ -153,6 +167,19 @@ export interface EcsTaskSizing {
   readonly planningTaskCpu?: number;
   /** Planning task memory in MiB. Defaults to 8192 (8 GB). */
   readonly planningTaskMemoryMiB?: number;
+  /**
+   * Extra environment variables for the BUILD task's container, merged over the
+   * defaults (a key given here wins).
+   *
+   * The platform sets a few build-tool variables that are only meaningful for the
+   * toolchain a given repo actually uses — a verify timeout, and parallelism caps
+   * for a mise/jest-shaped build. Those values were measured against one
+   * monorepo, so they are this deployment's opinion rather than a universal
+   * default: a repo that uses none of those tools ignores them, and a repo that
+   * uses mise with plenty of memory will want the serialisation cap raised. Set
+   * them here rather than editing the construct.
+   */
+  readonly extraBuildEnvironment?: Record<string, string>;
 }
 
 export class EcsAgentCluster extends Construct {
@@ -338,6 +365,10 @@ export class EcsAgentCluster extends Construct {
       // Propagates to both the platform push (post_hooks.py) and the agent's own
       // git-tool pushes via shell.py::_clean_env (blacklist — passes SKIP through).
       SKIP: 'monorepo-tests-pre-push',
+      // Caller overrides win: the values above are tuned for one monorepo's
+      // toolchain, so a deployment with a different build shape replaces them
+      // through `taskSizing.extraBuildEnvironment` rather than editing this file.
+      ...(sizing.extraBuildEnvironment ?? {}),
     }, buildDisk);
 
     // PLANNING task def — for read-only workflows that clone + read + emit a plan
--- a/cdk/src/constructs/ecs-agent-cluster.ts
+++ b/cdk/src/constructs/ecs-agent-cluster.ts
@@ -61,7 +61,7 @@ export interface EcsAgentClusterProps {
   readonly payloadBucket?: s3.IBucket;
 
   /**
-   * Artifacts bucket for repo-bound artifact workflows (a planning/decompose
+   * Artifacts bucket for repo-bound artifact workflows (a planning
    * workflow emits its plan JSON here via ``deliver_artifact``). The AgentCore
    * runtime gets ``ARTIFACTS_BUCKET_NAME`` in its env; the ECS task needs the
    * SAME env (but NO bucket grant) or an artifact workflow fails at delivery with
@@ -357,7 +357,7 @@ export class EcsAgentCluster extends Construct {
       // can be fetched. (The orchestrator sets the URI per-task via container
       // override; this is set here for parity with the runtime env.)
       ...(props.payloadBucket && { ECS_PAYLOAD_BUCKET: props.payloadBucket.bucketName }),
-      // Artifact workflows (e.g. decompose/planning) deliver their plan JSON to
+      // Artifact workflows (planning/analysis) deliver their document to
       // this bucket. The AgentCore runtime has ARTIFACTS_BUCKET_NAME; the ECS task
       // needs it too or deliver_artifact raises "ARTIFACTS_BUCKET_NAME is not
       // configured".
@@ -498,7 +498,7 @@ export class EcsAgentCluster extends Construct {
       props.payloadBucket.grantRead(taskRole);
     }
 
-    // Artifact workflows (e.g. decompose/planning) deliver their plan to the
+    // Artifact workflows (planning/analysis) deliver their document to the
     // artifacts bucket via deliver_artifact — but the write goes through the
     // assumed SessionRole (deliverers.py -> tenant_client), scoped to
     // artifacts/${task_id}/*, exactly like the AgentCore runtime (whose task
--- a/cdk/src/constructs/ecs-agent-cluster.ts
+++ b/cdk/src/constructs/ecs-agent-cluster.ts
@@ -42,7 +42,14 @@ export interface EcsAgentClusterProps {
   readonly memoryId?: string;
 
   /**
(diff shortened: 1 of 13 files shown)


isadeks:
The red `Secrets, deps, and workflow scan` here is **not caused by this PR** — it touches no lockfile or manifest (source + tests only).

The same job has been failing on `main` itself since the 2026-07-06 scheduled run (also 07-13 and 07-20), on 4 High advisories published against already-resolved dependencies: `brace-expansion` (×3, GHSA-mh99-v99m-4gvg) and `bedrock-agentcore` (GHSA-j6g5-3hh3-pgw8). Because the gate runs per-PR, it fails every open PR as a side effect. #647 (this PR's base) shows green only because it last ran before those advisories were published.

Fixed in **#658** (lockfile + pin only, against `main`) — `osv-scanner` now reports `No issues found`, and all 8 checks are green there. Once that merges, re-running this job should clear it; nothing to change in this PR.

One note for #658's reviewer: the usual re-resolve isn't sufficient this time. The `brace-expansion` advisory covers everything `<= 5.0.7`, so the v1 and v2 lines in the tree have no patched release inside their own majors — it needs a `resolutions` entry to collapse all three ranges onto 5.0.8.

isadeks:
## Automated review — carve S2 (#653)

Reviewed as **its own diff only** (`carve/s1-foundation-contracts..carve/s2-ecs-rightsized-planning`, 12 files, +809/−296). Governance is noted once for the whole stack, not repeated here.

`tsc --noEmit` clean at this tip on its own base (compiler run, not inferred). `yarn.lock` is byte-identical to `main` here, so no dependency change.

### Reverts: CLEAN

I checked the 296 deletions rather than assuming they were rewrites. Every `export`ed symbol present on `main` in the four touched source files still exists at this slice tip:

| file | main → s2 deletions | exports lost |
|---|---|---|
| `ecs-agent-cluster.ts` | 127 | none |
| `orchestrate-task.ts` | 23 | none |
| `orchestrator.ts` | 44 | none |
| `ecs-strategy.ts` | 13 | none |

The deletions are genuine restructuring — most notably collapsing the two Fargate task definitions onto one shared `makeTaskDef` factory, which is a real improvement: it makes it structurally impossible for a grant or env var to land on one task def and not the other.

**This PR also deletes the stray `cdk/node_modules` symlink that #647 committed** (pointing at `/private/tmp/abca-s5/cdk/node_modules`). That deletion is correct and welcome — I have raised the symlink itself on #647, since that is the PR that would merge first and carry it.

---

### 1. MAJOR — four build-tool env vars are hardcoded to this project's own monorepo and are not overridable

The build task definition hardcodes into its container environment:

```
BUILD_VERIFY_TIMEOUT_S: '3600'
JEST_MAX_WORKERS:       '4'
MISE_JOBS:              '1'
```
(plus a pre-push-hook skip)

The comments are careful and clearly measured — but every measurement is against **this** repo: *"`mise run build` fans out its `depends` (the per-package build/quality legs)"*, *"each package then spawns its OWN worker fleet (jest, pytest, esbuild, cdk synth)"*, *"the test suite at 4 workers peaks at only ~2.2 GB"*.

A public adopter's repo may use none of `mise`, `jest`, `pytest` or `cdk synth`. For them these vars are either inert (harmless) or actively wrong — `MISE_JOBS: '1'` serialises any adopter who *does* use mise, trading wall-clock for a memory problem they may not have.

Crucially, unlike the CPU/memory sizing, **these are not configurable**: `EcsTaskSizing` exposes only `buildTaskCpu`, `buildTaskMemoryMiB`, `buildTaskEphemeralStorageGiB`, `planningTaskCpu`, `planningTaskMemoryMiB`. There is no prop for the env.

Suggest an `extraBuildEnvironment?: Record<string, string>` prop (or a `buildVerifyTimeoutS?: number`) so the tuned values become this deployment's opinion rather than everyone's default. The reasoning is worth keeping in the repo — just not welded into the construct.

### 2. MAJOR — the default build task is the Fargate maximum, ~5× the cost of a modest size

`DEFAULT_BUILD_TASK_CPU = 16384` (16 vCPU) and `DEFAULT_BUILD_TASK_MEMORY_MIB = 122880` (**120 GB**), plus a 100 GiB root filesystem against Fargate's 20 GiB default.

To be fair to the PR: the docstring is honest about this (*"deliberately generous because they're tuned for a worst case: a large TypeScript + Python monorepo"*), it explains that 120 GB is the Fargate ceiling at 16 vCPU, and it points the reader at `taskSizing` to shrink it. That is the right structure.

The concern is the **default**, since that is what an adopter who changes nothing gets. Rough us-east-1 on-demand Fargate:

| sizing | $/hr | 10-min build | 100 builds |
|---|---|---|---|
| default (16 vCPU / 120 GB) | $1.181 | $0.197 | **$19.68** |
| modest (4 vCPU / 16 GB) | $0.233 | $0.039 | $3.88 |

≈ **5.1× per build**. For a public sample, a modest default with a documented "raise this for a large monorepo" note inverts the surprise in the safer direction — an adopter who under-provisions gets a slow build, whereas one who over-provisions gets a bill. The planning default (2 vCPU / 8 GB) is well judged and I would not change it.

### 3. MINOR — "read-only planning task" describes the workflow, not the IAM

The naming and comments consistently call this the *read-only* task def (*"PLANNING task: 2 vCPU / 8 GB … For read-only workflows that clone and read the repo to produce a plan but never build"*). But both task definitions are created by the same `makeTaskDef` factory sharing one `taskRole` and one `executionRole`, so the planning task holds **exactly the same permissions as the build task** — including the artifacts-bucket and AgentCore Memory read+write grants.

That is a deliberate, well-argued choice — the comment says so explicitly, that splitting the roles is how a grant silently drifts between the two defs — and I am not asking for it to change. But "read-only" invites a reader (or a future security reviewer) to assume a reduced-privilege role that does not exist. Worth one clarifying clause: *"read-only refers to the workflow's behaviour; both task defs deliberately share one role so grants cannot drift."*

---

### Verified clean

- **IAM:** no wildcard actions introduced. The shared-role restructuring is a net least-privilege *improvement* over two independently-granted defs.
- **No source/test split at this boundary** — every touched source file's test is updated in this same slice (`ecs-agent-cluster.test.ts` +236, `ecs-strategy.test.ts` +91, `orchestrate-task.test.ts` +27).
- **Invalid CPU/memory pairs** are acknowledged as failing at deploy time rather than synth, which is the correct expectation to document for Fargate.

---

*Reviewed with Claude Code. Verification: `tsc --noEmit` on this slice's own base; all 296 deletions checked for lost `export`s against `main`; the shared `taskRole` traced through `makeTaskDef`; `EcsTaskSizing` fields enumerated to confirm the env vars have no override; Fargate cost computed from published us-east-1 on-demand rates.*

*Note: this review was written directly rather than by the review agent assigned to it — that agent terminated mid-run without producing a result, so I reviewed the diff myself. Same for #656.*

isadeks:
### Note: force-push during review — plus a BLOCKING finding that moved into this PR

My review above was written against `b4c5d1dd`; the current head is `5b755209`. One finding **moved into this PR** from #647 as a result of the re-slicing, and it is blocking.

---

#### NEW BLOCKING — this PR now admits two workflows whose agent YAML does not exist until #654

The `DESCRIPTORS` entries for `coding/restack-v1` and `coding/decompose-v1` were removed from #647 and now land here. The YAML still arrives one slice later:

| slice | `DESCRIPTORS` entry | `agent/workflows/coding/restack-v1.yaml` |
|---|---|---|
| #647 (s1) | absent | missing |
| **#653 (s2)** | **present** | **missing** |
| #654 (s3) | present | present |

`DESCRIPTORS` is the live admission table — `create-task-core.ts` resolves any submitted `workflow_ref` against it via `resolveWorkflowRef` → `getWorkflowDescriptor`, and there is no platform-only submission gate (I grepped `USER_SUBMITTABLE` / `PLATFORM_ONLY` / `internalWorkflow` / `platformOnly` across `cdk/src` — nothing).

**Failure:** merge #653 alone and deploy. A user submits `workflow_ref: "coding/decompose-v1"`. Admission **succeeds** (201 — the user is told the task started), then the agent's `load_workflow` raises `WorkflowValidationError: workflow 'coding/decompose-v1' not found at <path>`. Accepted-then-dead is a worse outcome than a clean 400 at submission.

The fix is the same as before, one slice over: move the two `DESCRIPTORS` entries into #654 alongside their YAML, or bring the YAML forward into this PR.

---

#### Status of my original findings at `5b755209`

| finding | status |
|---|---|
| **1. MAJOR** — `BUILD_VERIFY_TIMEOUT_S`/`JEST_MAX_WORKERS`/`MISE_JOBS` hardcoded to this project's monorepo, not overridable | still present |
| **2. MAJOR** — default build task is the Fargate maximum (16 vCPU / 120 GB), ≈5× the cost of a modest size | still present |
| **3. MINOR** — "read-only planning task" shares the build task's role and grants | still present |
| Reverts / lost exports | still clean — re-checked, no `main` export lost |
| Deletes #647's stray `cdk/node_modules` symlink | still true, still welcome |

isadeks:
## Review findings addressed

I had missed this review in my first pass — apologies. All three findings are now fixed, and the blocking one from your addendum was already resolved by then.

**BLOCKING (addendum) — this PR admitted two workflows whose YAML arrives a slice later. Already fixed** before I read your addendum, by moving the `DESCRIPTORS` entries into #654 alongside their files. Your re-verification caught that my first move only relocated the problem by one slice, which is exactly right. What actually prevents it recurring is the reverse drift check I added: the test only walked YAML → descriptor, so an orphan descriptor was invisible; it now asserts every declared descriptor has a shipped file, verified by re-adding an orphan and watching it fail.

**MAJOR #1 — four build-tool env vars hardcoded to this project's monorepo, not overridable. Fixed.** You were right that these are this deployment's opinion rather than a universal default, and that `MISE_JOBS: '1'` actively penalises an adopter who uses mise with enough memory. Added `extraBuildEnvironment?: Record<string, string>` on `EcsTaskSizing`, merged over the defaults so a caller's key wins. The measured reasoning stays as the comment it always was; it is no longer welded in. Tested: an override replaces the value, and a key the caller does not name keeps the platform default.

**MAJOR #2 — default build task was the Fargate maximum. Fixed.** Your cost table is the argument: ~5.1× per build for the default an adopter gets by changing nothing. Now **4 vCPU / 16 GB on Fargate's own 20 GiB disk**, with the ceiling reachable through `taskSizing` and a test exercising that path. Your framing of which direction to be surprised in — a slow build is diagnosable and one prop from fixed, a bill is not — is what decided it. Planning default left at 2 vCPU / 8 GB as you suggested.

Two of the tests that pinned the old defaults have been rewritten rather than deleted, so the heavy-monorepo size is still asserted — just as an override rather than as everyone's baseline.

**MINOR #3 — "read-only planning task" describes the workflow, not the IAM. Fixed as you suggested**, with the clarifying clause. You are right not to ask for the roles to be split: the shared role is what stops a grant drifting between the two defs, and that reasoning is now stated where a future security reviewer will read it.

Thank you also for checking all 296 deletions for lost exports rather than assuming they were rewrites, and for flagging the stray `cdk/node_modules` symlink — that one turned out to matter. It was mine, committed by a `git add -A` in a worktree that symlinks its dependencies, and it had been failing CI at install time with `EEXIST` in a way that reads like a runner cache glitch. Removed at the source, and `.gitignore` now matches the symlink form as well as the directory form, since `node_modules/` with a trailing slash does not.

Gates: 2544 CDK tests, `tsc`/`eslint` clean.

isadeks:
### Second-round review: my de-opinionation was unreachable, and the env override was a footgun

Two findings on my own fix, both confirmed.

**The prop had no caller, so the ceiling was unreachable.** Lowering the default to 4 vCPU / 16 GB is only defensible if a heavy monorepo can raise it *without editing the construct* — but nothing passed `taskSizing`, `AgentStack` takes no custom props, and no context key read it. So I traded "everyone pays for 16 vCPU" for "**nobody can reach** 16 vCPU", and the second is worse for exactly the repo the old default existed to serve. That is the first review round's concern in reverse.

Sizing and the build-tool env are now read from deploy context, the same shape as the `compute_type` gate beside them:

```
cdk deploy -c compute_type=ecs -c ecsBuildTaskCpu=16384 \
  -c ecsBuildTaskMemoryMiB=122880 -c ecsBuildTaskEphemeralStorageGiB=100
cdk deploy -c ecsExtraBuildEnv='{"MISE_JOBS":"8"}'
```

A malformed value throws at synth rather than silently defaulting — "I set the flag and the build still OOM'd" is a worse afternoon than a failed synth. Reachability is now asserted end to end (context → resolver → construct → template); unwiring the call site fails that test, where before it left the whole suite green.

**`extraBuildEnvironment` could clobber platform wiring.** It spreads over the entire container env, not just the four build-tool vars it was added for. The reviewer's list is right, and the sharp one is `AGENT_SESSION_ROLE_ARN`: when it is absent the agent falls back to ambient credentials and per-tenant scoping is silently **off** — which the agent's own session module calls its most dangerous failure mode. An adopter mistyping a key would have disabled an isolation control by accident. Reserved keys are now rejected at synth with a message naming the key; mutation-verified.

**Stale docstrings fixed.** They still advertised 16 vCPU / 120 GB / 100 GiB two lines above code saying 4096 / 16384 / 21, and the per-field `@default` tags likewise. Whichever an operator read first was wrong.

**Verified and not changed:** 4096/16384 is a legal Fargate pair and 21 GiB is the legal ephemeral-storage floor (20 throws) — confirmed by a full `cdk synth` under the ECS gate, so no deploy-breaking sizing defect. The reviewer also notes CDK validates ephemeral storage but **not** the cpu/memory combination, so the pre-existing docstring claim that "an invalid pair fails at synth" is false — an invalid pair reaches CloudFormation. That predates this work; tracked rather than fixed here.

cdk/src/handlers/shared/strategies/ecs-strategy.ts
```diff
@@ -41,7 +41,39 @@ function getS3Client(): S3Client {
...
 const ECS_CLUSTER_ARN = process.env.ECS_CLUSTER_ARN;
 const ECS_TASK_DEFINITION_ARN = process.env.ECS_TASK_DEFINITION_ARN;
+/**
+ * The smaller, read-only "planning" task definition. Read-only workflows (which
+ * clone and read the repo to produce a plan but never build) run here instead of
+ * on the large build task. Falls back to the build task definition when unset
+ * (e.g. a deployment that hasn't provisioned the planning task) — never worse
+ * than running everything on the build task.
+ */
+const ECS_PLANNING_TASK_DEFINITION_ARN = process.env.ECS_PLANNING_TASK_DEFINITION_ARN;
```
scottschreckengaust, line 52: **Blocking:** nothing sets `ECS_PLANNING_TASK_DEFINITION_ARN`. `TaskOrchestrator.ecsConfig` (task-orchestrator.ts:152-160) has no planning field and its env block (:269-275) emits only `ECS_TASK_DEFINITION_ARN`; `agent.ts:683-690` never references `ecsCluster.planningTaskDefinition`. So this constant is always `undefined`, the `readOnly && ECS_PLANNING_TASK_DEFINITION_ARN` guard at line 155 is always falsy, and read-only workflows never reach the planning def — the feature is inert. Wire the ARN through `ecsConfig` → orchestrator env and add an `agent.test.ts` assertion that the Lambda environment carries it (the isolated-module unit tests set it by hand, so they can't catch this).

cdk/src/constructs/ecs-agent-cluster.ts
```diff
@@ -54,56 +61,233 @@ export interface EcsAgentClusterProps {
...
+    ...(num('ecsPlanningTaskCpu') !== undefined && { planningTaskCpu: num('ecsPlanningTaskCpu') }),
+    ...(num('ecsPlanningTaskMemoryMiB') !== undefined && { planningTaskMemoryMiB: num('ecsPlanningTaskMemoryMiB') }),
+    ...(extra !== undefined && { extraBuildEnvironment: extra }),
+  };
+  return Object.keys(sizing).length > 0 ? sizing : undefined;
+}
+
 export class EcsAgentCluster extends Construct {
   public readonly cluster: ecs.Cluster;
+  /** The 64 GB / 16 vCPU BUILD task def — for coding workflows that run a full
```
scottschreckengaust: Stale: this PR lowered the build default to **4 vCPU / 16 GB**, but the docstring still says "64 GB / 16 vCPU". Same stale figure in compute-strategy.ts:55 and orchestrator.ts:276. Reword to "the larger BUILD task def (default 4 vCPU / 16 GB, raisable to the Fargate ceiling via taskSizing)."

cdk/src/handlers/shared/strategies/ecs-strategy.ts
```diff
@@ -158,14 +206,14 @@ export class EcsComputeStrategy implements ComputeStrategy {
       { name: 'REPO_URL', value: String(payload.repo_url ?? '') },
       ...(payload.prompt ? [{ name: 'TASK_DESCRIPTION', value: String(payload.prompt) }] : []),
       ...(payload.issue_number ? [{ name: 'ISSUE_NUMBER', value: String(payload.issue_number) }] : []),
-      { name: 'MAX_TURNS', value: String(payload.max_turns ?? 100) },
+      { name: 'MAX_TURNS', value: String(payload.max_turns ?? 200) },
```
scottschreckengaust: Nit: the inline `MAX_TURNS` fallback changed 100 → 200, but `DEFAULT_MAX_TURNS` is still 100 (validation.ts:34) and governs the hydrate path (orchestrator.ts:566). The two disagree only when `payload.max_turns` is absent, but there's no comment explaining the bump. Either route both through `DEFAULT_MAX_TURNS` or add a one-line rationale.

scottschreckengaust (changes requested):
## Verdict: Request changes

The **sizing** half of this PR is excellent — modest defaults, deploy-context knobs, reserved-key guard, and an end-to-end reachability test (context → resolver → construct → template). The `clientToken=taskId` RunTask idempotency and the family-not-revision dispatch (`toTaskDefinitionFamily`) are both real robustness wins. The shared-role refactor (`makeTaskDef`, one `taskRole`/`executionRole`) is a genuine least-privilege *improvement* over two independently-granted defs.

But the PR's other headline capability — routing read-only workflows to the smaller planning task def — **is inert at runtime**: the ARN that selects it is never wired into the orchestrator Lambda, so `readOnly` planning tasks always fall back to the build def. That is half the PR's stated value shipping as a no-op, and no integration test would catch it. Blocking on that + a test that proves the wiring.

## Vision alignment

Strongly aligned with **bounded blast radius & cost** (VISION.md). Rightsizing planning to 2 vCPU / 8 GB and de-opinionating the build default from 16 vCPU / 120 GB to a modest 4 vCPU / 16 GB inverts the cost surprise in the safe direction (a slow build is diagnosable and one prop from fixed; a bill is not). The shared-role design keeps the read-only planner from silently diverging in privilege. No tenet is traded; backing issues #247 and #299 are both `approved` / P0, tracked under epic #668.

## Blocking issues

**1. The planning task def is created and exposed but never wired to the orchestrator — read-only routing is a no-op.** `EcsComputeStrategy` reads `process.env.ECS_PLANNING_TASK_DEFINITION_ARN` (`cdk/src/handlers/shared/strategies/ecs-strategy.ts:51,155`), but nothing sets it: `TaskOrchestrator`'s `ecsConfig` (`cdk/src/constructs/task-orchestrator.ts:152-160`) has no planning field and its env block (`:269-275`) only emits `ECS_TASK_DEFINITION_ARN`; `AgentStack` (`cdk/src/stacks/agent.ts:683-690`) never references `ecsCluster.planningTaskDefinition`. Result: `ECS_PLANNING_TASK_DEFINITION_ARN` is always `undefined`, so `readOnly && ECS_PLANNING_TASK_DEFINITION_ARN` is always falsy and every task — including `coding/decompose-v1` — runs on the build def. The planning def is synthesized (and registered) but never selected. The strategy unit tests only pass because they set the env var by hand via `isolateModules`; `agent.test.ts` asserts two task defs exist but never that the planning ARN reaches the Lambda, so the gap is invisible to the suite.

*Fix:* add `planningTaskDefinitionArn: string` to `ecsConfig`, pass `ecsCluster.planningTaskDefinition.taskDefinitionArn` from `agent.ts`, emit `ECS_PLANNING_TASK_DEFINITION_ARN` in the orchestrator env, and add an `agent.test.ts` assertion that the orchestrator Lambda's environment carries it (mirroring the existing `ECS_TASK_DEFINITION_ARN` wiring). If the wiring is deliberately deferred to a later carve slice, say so in the PR body and drop the strategy-side consumption from this slice so the two land together — shipping the consumer without the producer leaves dead config in the tree.

**2. Stale sizing in docstrings after the de-opinionation.** The class-field JSDoc still says `The 64 GB / 16 vCPU BUILD task def` (`ecs-agent-cluster.ts:278`) even though this PR lowered the build default to 4 vCPU / 16 GB. Same stale figure in `compute-strategy.ts:55` ("the 64 GB build def") and `orchestrator.ts:276` ("the context-gated 64GB ECS tier"). The prior review round explicitly flagged and claimed to fix stale docstrings; these three survived. Reword to "the larger BUILD task def (default 4 vCPU / 16 GB, raisable to the Fargate ceiling)".

**3. `MAX_TURNS` fallback silently diverges (100 vs 200).** `ecs-strategy.ts:209` changed the inline default from `100` to `200`, while `DEFAULT_MAX_TURNS = 100` (`validation.ts:34`) still governs the hydrate path (`orchestrator.ts:566`). The two only disagree when `payload.max_turns` is absent, but there's no comment explaining the bump and it's out of band with this PR's stated scope. Either route both through `DEFAULT_MAX_TURNS` or add a one-line rationale.

## Documentation

No `docs/guides` or `docs/design` change is included. The new deploy-context knobs (`ecsBuildTaskCpu`, `ecsBuildTaskMemoryMiB`, `ecsBuildTaskEphemeralStorageGiB`, `ecsPlanningTaskCpu`, `ecsPlanningTaskMemoryMiB`, `ecsExtraBuildEnv`) are documented only in an inline comment in `agent.ts`. For a public sample where sizing is now the operator's lever, these belong in the ECS/compute deployment guide alongside the `compute_type=ecs` gate. Non-blocking for this slice but worth a follow-up. No Starlight mirror edits needed (no `docs/` sources touched). Backing issues #247/#299 exist and are approved; #668 tracks the epic.

## Tests & CI

Sizing/override paths are well covered (resolver rejects malformed numbers, non-string env values, and reserved platform keys; end-to-end reachability asserted in `agent.test.ts`). The parity test (both defs share one task+execution role) is a strong regression guard. **Gap:** no test asserts the planning ARN is wired into the orchestrator env (Blocking #1) — the isolated-module tests set it manually, so they cannot catch the missing glue. CI: 4/4 checks green; the author documented (issue comment) that the deps-scan red is a pre-existing `main` advisory unrelated to this diff — verified plausible (no lockfile/manifest touched).

**Bootstrap policy coverage:** N/A. This PR adds a second `AWS::ECS::TaskDefinition` and two `AWS::IAM::Role`s — both existing CFN resource types already covered; no new resource type is introduced, so no bootstrap policy / action-map / `BOOTSTRAP_VERSION` bump is required.

## Review agents run

- **code-reviewer** — ran (manual principal-architect pass; Semgrep MCP invocation failed on a `.semgrep.yml` config-deprecation error, so I fell back to targeted grep + manual review of the three core source files).
- **silent-failure-hunter** — ran. `parseMergeBranches` swallows malformed JSON but logs a `warn` and returns `[]` (documented best-effort) — acceptable, not a silent swallow. The planning-def fallback (Blocking #1) is a *safe* silent degradation, but it degrades the feature to a no-op invisibly — hence flagged.
- **type-design-analyzer** — ran. New `EcsTaskSizing` interface: all-optional fields with clear per-field defaults; `readOnly?` added to the `ComputeStrategy.startSession` contract with a documented default. Coherent, no over-abstraction.
- **comment-analyzer** — ran. Surfaced the three stale `64 GB` references (Nit #2).
- **pr-test-analyzer** — ran. Surfaced the missing orchestrator-env wiring assertion (Blocking #1).
- **security-review** — ran (IAM/env change). Shared-role refactor is a least-privilege improvement; `RESERVED_BUILD_ENV_KEYS` correctly fails closed on attempts to override `AGENT_SESSION_ROLE_ARN`/table names via `ecsExtraBuildEnv`; `clientToken` prevents double-RunTask. No new wildcard actions.

## Human heuristics

- **Proportionality** — Pass. `makeTaskDef` factory is justified (forces build/planning parity); no gratuitous abstraction.
- **Coherence** — Concern. "read-only planning" naming is now well-clarified in comments, but the feature it names does not actually route yet (`ecs-strategy.ts:51`), so the concept is present in code without a working path.
- **Clarity** — Concern. Stale `64 GB` docstrings (`ecs-agent-cluster.ts:278`) contradict the code two lines from the new 4096/16384 defaults.
- **Appropriateness** — Concern. Tests assert what the isolated strategy *does* (env preset by hand) rather than what the deployed system *should* do (stack wires the ARN) — the AI005 failure mode.

cdk/src/handlers/shared/strategies/ecs-strategy.ts
isadeks: Confirmed, and thank you — this is a real defect in **how I sliced the series**, not in the feature, and your reading of it was exactly right.

The three wiring lines do exist. They are in **S8 (#662)**, the last slice, six PRs later. I verified per-slice which diff introduces them:

| slice | adds `planningTaskDefinitionArn` |
|---|---|
| S2–S7 | 0 lines |
| **S8** | **3 lines** |

So the feature works at the series tip and on the branch this was carved from — which is why it live-tested fine — while **this slice ships a task def that is defined, billed for, and unreachable**. Your diagnosis of the mechanism is precisely correct: `ecsConfig` had no planning field, so the env var was never emitted, so `readOnly && ECS_PLANNING_TASK_DEFINITION_ARN` was always falsy.

This is worse than cosmetic because merges here are serial (`max_entries_to_merge: 1`), so there is a real window where S2–S7 are on `main` and S8 is not. Anyone testing rightsized planning in that window would reasonably conclude the feature is broken.

Fixed by moving the wiring into this slice, so the PR that introduces the def also makes it reachable, and removing the now-duplicate declaration from S8 (git had merged both copies cleanly, since they were not textually adjacent — that would have shipped a doubled interface field).

Two deliberate choices:

- `planningTaskDefinitionArn` is **required**, matching its siblings, so an `ecsConfig` that omits it fails to compile rather than silently disabling the routing. I first made it optional with a build-def fallback; that is friendlier but it is precisely the shape that let this bug hide, so I dropped it.
- **No IAM change.** The `ecs:RunTask` grant is scoped by `ecs:cluster`, not by task-definition ARN, and the planning def shares the build def's task and execution roles — so it is already covered. Worth stating explicitly since "second task def" usually implies a second grant.

I also took your point about where the test belongs. Added to `agent.test.ts`:

```ts
const orchestrator = envs.filter(e => 'ECS_TASK_DEFINITION_ARN' in e);
expect(orchestrator).toHaveLength(1);
expect(orchestrator[0]).toHaveProperty('ECS_PLANNING_TASK_DEFINITION_ARN');
expect(orchestrator[0].ECS_PLANNING_TASK_DEFINITION_ARN)
  .not.toEqual(orchestrator[0].ECS_TASK_DEFINITION_ARN);
```

It asserts the two ARNs **differ**, not just that both exist — otherwise wiring the build ARN into both slots would pass while read-only workflows still ran on the build box. Verified by deleting the wiring line and watching it fail; `tsc` also fails there now, so there are two independent nets.

`2d82602d`, with S3–S8 restacked (S3–S7 byte-identical by delta hash; S8 had the one expected conflict, resolved by dropping its duplicate).

cdk/src/constructs/ecs-agent-cluster.ts
isadeks: Fixed — and there was a **fourth** site you did not have visibility into, plus one that was not merely stale but wrong.

All four corrected:

| file | was | now |
|---|---|---|
| `ecs-agent-cluster.ts:278` | "The 64 GB / 16 vCPU BUILD task def" | "default 4 vCPU / 16 GB, raisable to the Fargate ceiling of 16 vCPU / 120 GB via `EcsTaskSizing`" |
| `compute-strategy.ts:55` | "instead of the 64 GB build def" | "instead of the larger build def" |
| `orchestrator.ts:276` | "the context-gated 64GB ECS tier" | "the context-gated ECS tier" |
| **`agent.ts:591`** | **"64 GB was itself OOM-killed, so it runs larger"** | "a 32 GB task was OOM-killed by a fully parallel build, which is why the build tier serialises with `MISE_JOBS=1`" |

That last one was the one worth catching: it did not just quote a stale number, it asserted that a **64 GB** task had been OOM-killed. The measured figure is **32 GB**, and it was killed by a *fully parallel* `mise run build` peaking ~31.6 GB. The build tier sets `MISE_JOBS=1`, which serialises the per-package legs so peak becomes max-single-package rather than sum-of-all — measured at **~3.1 GB** on this repo. That 10x gap is entirely the parallelism setting, so a reader who took the comment at face value would conclude 16 GB was reckless when it is ~5x headroom over what the shipped configuration actually uses.

I used the wording you suggested for the primary site. Also went back and fixed the same staleness in `docs/design/COMPUTE.md`, which still advertised a 16 vCPU / 64 GB build tier in its sizing table — it now states both measurements and which configuration produces which, since quoting the parallel figure next to a serialised default is what made this confusing in the first place.

Where the numbers came from: I stress-tested the modest default on this monorepo with the build gate live — peak **3132 MiB of 16384** (19%) and **14.68 GiB of ~22.5 GiB disk** (65%), `build_passed: true`. Memory had ample headroom; disk did not, so I raised that default 21 → 50 GiB in the same slice.

`2d82602d`.

cdk/src/handlers/shared/strategies/ecs-strategy.ts
isadeks: Took the first option — routed through `DEFAULT_MAX_TURNS` — because checking this turned up something worse than an undocumented bump.

`main` has `payload.max_turns ?? 100`. This slice changed it to `200`. That is an **unflagged live behavioral change**, and it directly contradicts what the stack's own first slice promises: #647 explicitly defers the max-turns bump to its own PR ("the arc's bump to a higher default is a live behavioral change that will come as its own PR"). So a reader auditing the series for behavioral changes would have been told this one was not here, while it was.

Adding a rationale comment would have documented a change that should not be in this PR at all. Routing through the shared constant removes the divergence **and** the smuggled change in one edit — `DEFAULT_MAX_TURNS` is 100, so this slice now matches `main` exactly:

```ts
{ name: 'MAX_TURNS', value: String(payload.max_turns ?? DEFAULT_MAX_TURNS) },
```

Your framing of the impact was right and worth preserving in the comment: the two disagreed **only** when `payload.max_turns` was absent, which is the sneaky part — a turn ceiling that depends on which code path happened to fill it in. That is now stated at the call site.

The bump to 200 is a real intended change on the source branch; it will come as its own PR against the shared constant, where it applies to both paths at once instead of only the ECS one.

`2d82602d`.

isadeks:

### What moved

| slice | effect |
|---|---|
| #647 S1 | unchanged |

### Verification

cdk/src/constructs/task-orchestrator.ts
```diff
@@ -157,6 +157,21 @@ export interface TaskOrchestratorProps {
     readonly containerName: string;
     readonly taskRoleArn: string;
     readonly executionRoleArn: string;
+    /**
+     * The smaller read-only PLANNING task def (see
+     * docs/design/ECS_RIGHTSIZED_PLANNING.md). The ECS strategy selects it for
```
scottschreckengaust, line 162: Nit (non-blocking): this points at `docs/design/ECS_RIGHTSIZED_PLANNING.md`, which does not exist in the repo — no such file under `docs/design/`, and this PR touches no `docs/` source. The reference reads as describing a design doc that was never written. Either drop the pointer or land the doc (the new sizing/context knobs would be a good home for it in `docs/design/COMPUTE.md`).

scottschreckengaust (approved):
## Verdict: Approve (with nits)

Stacked PR — reviewed as the incremental slice against `carve/s1-foundation-contracts`, not against `main`. Merge order matters: this must not land before #647.

The prior review round (my CHANGES_REQUESTED, 2026-07-28) had one blocking issue and two nits. I re-verified each against the **current** HEAD (`f9d16ac`) rather than trusting the reply threads — **all three are resolved**:

1. **Planning task def now actually wired (was blocking, FIXED).** `EcsAgentCluster.planningTaskDefinition` is exposed, `AgentStack` passes it as `ecsConfig.planningTaskDefinitionArn` (`cdk/src/stacks/agent.ts:687`), `TaskOrchestrator` emits `ECS_PLANNING_TASK_DEFINITION_ARN` into the orchestrator Lambda env (`cdk/src/constructs/task-orchestrator.ts:294`), and `orchestrate-task.ts:174` now passes `readOnly: workflowIsReadOnly(...)` so the strategy's `readOnly && ECS_PLANNING_TASK_DEFINITION_ARN` guard can actually fire. A **stack-level** test (`cdk/test/stacks/agent.test.ts`) asserts both ARNs reach the Lambda and differ — the exact gap the isolated-module strategy tests could not catch. The type was made `required` so an `ecsConfig` omitting it fails to compile rather than silently disabling routing. Correct fix.
2. **Stale `64 GB / 16 vCPU` source docstrings (was nit, FIXED).** The class-field JSDoc and the `compute-strategy.ts` / `orchestrator.ts` references now describe the modest default (4 vCPU / 16 GB) with the Fargate ceiling as the raise target. Remaining `120 GB` mentions in `ecs-agent-cluster.ts` are correct — they describe the *ceiling*, not the default.
3. **`MAX_TURNS` divergence (was nit, FIXED).** `ecs-strategy.ts` now routes the fallback through `DEFAULT_MAX_TURNS` (=100, `validation.ts:34`), matching the hydrate path in `orchestrator.ts`, with a comment explaining why. No more path-dependent turn ceiling.

With the blocker cleared and both nits addressed, the feature is coherent end-to-end and I'm approving. Two cosmetic nits remain (below); neither blocks.

## Vision alignment

Strongly aligned with **bounded blast radius & cost** (VISION.md). Inverting the build default from the Fargate maximum (16 vCPU / 120 GB) to a modest 4 vCPU / 16 GB puts the cost surprise on the safe side — under-provisioning surfaces as a diagnosable, one-prop-from-fixed slow build, while over-provisioning is a silent bill. Rightsizing read-only planning to 2 vCPU / 8 GB is the same tenet applied per-workflow. The shared-role design (one `taskRole`/`executionRole` across both defs, built by one `makeTaskDef` factory) keeps the planner from silently diverging in privilege — a least-privilege *improvement* over two independently-granted defs. No tenet is traded. Backing issues #247 (P0) and #299 are `approved`; epic #668 (P1) is `approved` and tracks the carve.

## Governance

Gate satisfied: backing issues #247 and #299 both carry the `approved` label; epic #668 is approved and documents the slice table. Branch name `carve/s2-ecs-rightsized-planning` does not follow `(feat|fix|chore|docs)/<issue>-desc`, but the carve series uses a deliberate `carve/sN-*` convention documented in #668 — a de-facto-waived nit, not a blocker.

## Blocking issues

None.

**1. Dangling doc reference.** `cdk/src/constructs/task-orchestrator.ts:162` points readers to `docs/design/ECS_RIGHTSIZED_PLANNING.md`, which does not exist in the repo (confirmed: no such file under `docs/design/`, and no `docs/` source is touched by this PR). A comment that references an absent doc reads as describing missing design rather than pointing anywhere useful. Either drop the reference or land the doc. Inline comment attached.

**2. Two stale `64 GB` comments survive in test files.** `cdk/test/constructs/ecs-agent-cluster.test.ts:241` ("the 64 GB build def (asserted above)") and `cdk/test/handlers/orchestrate-task.test.ts:760` ("the 64GB ECS tier") still cite the old figure after the default dropped to 16 GB. The source docstrings were corrected; these test comments were missed. Cosmetic (comments only, no assertion depends on the number), but they now contradict the very assertions they annotate (`Cpu: '4096', Memory: '16384'`).

## Documentation

No `docs/guides` or `docs/design` change is included. The new deploy-context knobs (`ecsBuildTaskCpu`, `ecsBuildTaskMemoryMiB`, `ecsBuildTaskEphemeralStorageGiB`, `ecsPlanningTaskCpu`, `ecsPlanningTaskMemoryMiB`, `ecsExtraBuildEnv`) are documented only in an inline comment in `agent.ts` — and nit #1 shows the code now *promises* a design doc that was never written. For a public sample where sizing is the operator's cost lever, these belong in `docs/design/COMPUTE.md` alongside the `compute_type=ecs` gate. Non-blocking for this slice; recommend a follow-up issue. No Starlight mirror edits needed (no `docs/` sources touched).

## Tests & CI

Strong coverage. `resolveEcsTaskSizing` rejects malformed numbers, non-string env values, and reserved platform keys; end-to-end reachability (context → resolver → construct → template) is asserted in `agent.test.ts`; the parity test (both defs share exactly one task + one execution role) is a genuine regression guard against grant drift; the family-not-revision dispatch and `clientToken=taskId` idempotency each have targeted tests. The `readOnly` routing is now proven at both unit (strategy) and stack (env-wiring) levels. CI: 4/4 checks green (build agentcore, secrets/deps scan, dead-code advisory, PR-title).

**Bootstrap policy coverage:** N/A. This PR adds a second `AWS::ECS::TaskDefinition` and two `AWS::IAM::Role`s — both CFN resource types already covered by the bootstrap bundle. No new resource *type* is introduced, so no `cdk/src/bootstrap/policies/*`, `resource-action-map.ts`, or `BOOTSTRAP_VERSION` change is required. `synth-coverage` unaffected.

## Review agents run

- **code-reviewer** — ran (principal-architect pass over all changed source; verified the wiring chain agent.ts → task-orchestrator.ts → ecs-strategy.ts → orchestrate-task.ts by reading each site, not the diff summary).
- **silent-failure-hunter** — ran. `parseMergeBranches` swallows malformed JSON but logs `warn` and returns `[]` (documented best-effort; the child still branches off its base) — acceptable, not a silent swallow. The planning-def fallback to the build def when unwired is a *safe* degradation and is now backed by a test.
- **type-design-analyzer** — ran. `EcsTaskSizing` is all-optional with clear per-field defaults; `readOnly?` on `ComputeStrategy.startSession` documents its default. `channel_metadata` remains `Record<string,string>`, so the new `orchestration_merge_branches` arrives as a string that `parseMergeBranches` decodes — type-consistent, no shared-type drift (no `handlers/shared/types.ts` shape change needing `cli/src/types.ts` sync).
- **comment-analyzer** — ran. Surfaced nits #1 (dangling doc ref) and #2 (two stale `64 GB` test comments).
- **pr-test-analyzer** — ran. Confirmed the previously-missing orchestrator-env wiring assertion now exists at stack level.
- **security-review** — ran (IAM/env change). Shared-role refactor is a least-privilege improvement; `RESERVED_BUILD_ENV_KEYS` correctly fails closed on attempts to override `AGENT_SESSION_ROLE_ARN` / table / bucket names via `ecsExtraBuildEnv` (tested); `clientToken=taskId` prevents a double-RunTask launching two parallel containers on the same task; no new wildcard actions; cdk-nag suppressions retargeted correctly to the now-standalone roles with accurate rationale. No Cedar/network/secrets contract change.

No agent omitted — every dimension the diff touches was covered.

## Human heuristics

- **Proportionality** — Pass. `makeTaskDef` factory is justified (it *forces* build/planning parity by construction); no gratuitous abstraction. Context-resolver is the same shape as the existing `compute_type` gate.
- **Coherence** — Pass. "read-only planning" now names a feature that actually routes (`orchestrate-task.ts:174` → `ecs-strategy.ts`), and comments consistently clarify it describes workflow behaviour, not a privilege boundary. Minor blemish: the dangling doc reference (nit #1) names a doc that doesn't exist.
- **Clarity** — Pass. Names communicate intent; the reserved-key guard's error message explains *why* the key is refused. Residual: two stale `64 GB` test comments (nit #2).
- **Appropriateness** — Pass. The AI005 failure mode called out last round (tests asserting what the isolated strategy *does* rather than what the deployed system *should* do) is fixed — the stack-level test now pins the real wiring, and deleting the wiring line fails both `tsc` and the test.

Issue #629, opened 2026-07-20
feat(observability): wire SNS notification action to DLQ CloudWatch alarms


nizar-lahlali:
## Background

> **Note:** PR #208 (issue #117) is **not merged yet** — it is still open and under review. This is a follow-up to be picked up **after** #208 ships. The DLQ alarms described below do not exist on `main` today.

PR #208 / issue #117 adds CloudWatch DLQ-depth alarms (`ApproximateNumberOfMessagesVisible >= 1`, 5-min / `Maximum`, `treatMissingData: NOT_BREACHING`) for two stream consumers:

- `FanOutConsumer` — `dlqDepthAlarm` (`cdk/src/constructs/fanout-consumer.ts`)
- `ApprovalMetricsPublisherConsumer` — `dlqAlarm` (`cdk/src/constructs/approval-metrics-publisher-consumer.ts`)

Once merged, these alarms will **ship without an `addAlarmAction` / SNS notification target** — by design, as an intentional intermediate step (see `docs/design/CEDAR_HITL_GATES.md` §11.5). Alarm state is durable and queryable in the CloudWatch console, but there is no push notification when a DLQ starts accumulating poison-pill records. Operators must poll the Alarms console or configure a subscription manually.

## Proposal

Provision an operational notification channel and wire it to the alarms introduced by #208:

1. Add an SNS topic (e.g. an `OperationalAlerts` construct) — ideally reusable across the stack, not per-consumer.
2. Attach it to both alarms via `alarm.addAlarmAction(new cloudwatch_actions.SnsAction(topic))`. Both alarm properties are exposed as `cloudwatch.IAlarm` specifically so a consumer can call `addAlarmAction` without changes to the constructs.
3. Support subscription target(s) — SNS → Slack / PagerDuty / email. Delivery target should be configurable (Blueprint prop / stack context) rather than hard-coded.

No metric or alarm restructuring is required — this is purely additive (per §11.5 "Follow-up — notification channel wiring").

## Scope / considerations

- Wire the two DLQ-depth alarms above at minimum. Consider whether the `GithubScreenshotIntegration` DLQ alarm (mirrors the same threshold-1 shape) should share the topic.
- Once a channel exists, the "additional alarms (not yet shipped)" in §11.5 (e.g. high approval-timeout rate) become worth revisiting — but those can be separate follow-ups.
[...]

Issue #228, opened 2026-06-01
feat(cdk): wire DLQ CloudWatch alarms to SNS topic for alerting


nizar-lahlali:
## Context

PR #208 added CloudWatch alarms on `FanOutDlq` and `ApprovalMetricsPublisherDlq` (threshold: `ApproximateNumberOfMessagesVisible >= 1`, 5-min window). The alarms are exposed as `public readonly dlqAlarm` on each construct but are not yet wired to any notification channel — they fire silently in the CloudWatch console.

## Proposal

Create an SNS topic (or accept one as a construct prop) and add alarm actions so operators get notified (email, PagerDuty, Slack webhook, etc.) when poison-pill records land in a DLQ.

### Acceptance criteria

- [ ] An SNS topic is created (or injected via prop) for DLQ alarm notifications
- [ ] Both `FanOutConsumer.dlqAlarm` and `ApprovalMetricsPublisherConsumer.dlqAlarm` have an alarm action targeting the topic
- [ ] Optional: OK action to notify when the alarm returns to normal
- [ ] Unit tests verify the alarm action is wired to the SNS topic ARN
- [ ] `cdk synth` produces the expected `AWS::CloudWatch::Alarm` → `AWS::SNS::Topic` relationship

### Design considerations

- Should the topic be shared across all DLQ alarms or per-construct? A single shared `OperationalAlarmsTopic` is simpler and matches the common pattern.
- Subscription management (email endpoints, chatbot integrations) can be handled out-of-band or via a `notificationEndpoints` stack parameter.

## References

- #117 (original DLQ alarm request)
- PR #208 (alarm implementation without SNS)

PR #674, opened 2026-07-29
feat(infra): add Lambda-error alarm to GitHub webhook processor (#284)

### LINKED ISSUE #284: feat(infra): add SQS DLQ + CloudWatch alarm to GitHub webhook processor
## Component

CDK / infrastructure (notification plane / observability)

## Problem

The screenshot pipeline's processor Lambda (`GitHubScreenshotIntegration/WebhookProcessorFn`) is invoked async (`InvocationType: 'Event'`) by the receiver. Today it has:

- No SQS DLQ
- No `onFailure` Lambda destination
- No CloudWatch alarm on `Errors`

Every failure path inside the handler logs and returns. Because the receiver already returned 200 to GitHub, GitHub will not redeliver — so a systemic break (IAM regression, AgentCore quota exhaustion, OAuth token rotation issue, dependency outage) silently stops 100% of screenshots with no signal anywhere.

Spawned from [krokoko's review on PR #241](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/241#pullrequestreview-3329884229) (item #3, blocking).

## Expected behavior

A failed processor invocation should produce an operator-visible signal. After Lambda's automatic async retries exhaust, the failed event lands on a DLQ and a CloudWatch alarm fires.

## Current behavior

Failed invocations log-and-return; the failure is invisible to operators unless they happen to inspect Lambda metrics for that specific function.

## Proposed solution

1. Add an SQS DLQ on `WebhookProcessorFn` via Lambda's async-invoke `onFailure` destination (so failed events land on the DLQ after Lambda's automatic retries).
2. Add a CloudWatch alarm on the processor's `Errors` metric (`>= 1` in 5min, sustained for 2 evaluation periods).
3. DLQ retention: 14 days; SSE-KMS encryption.
4. Add cdk-nag suppressions if needed for DLQ encryption / alarm threshold defaults.
5. Notify via existing alarm SNS topic if one exists; otherwise create one specific to this stack.

## Reference

- Pattern to copy: existing DLQ in the stack (e.g. `LinearIntegration/WebhookProcessorFn` — check first; if none exists, this issue establishes the pattern)
- `cdk/src/constructs/task-orchestrator.ts` is also a candidate reference

## Acceptance criteria

- [ ] DLQ attached to `WebhookProcessorFn` async-invoke `onFailure`
- [ ] CloudWatch alarm on processor errors (>= 1 in 5min, 2 eval periods)
- [ ] cdk-nag clean
- [ ] Smoke test: deliberately break the processor (e.g. revoke an IAM permission), confirm the failed event lands on the DLQ and the alarm fires
- [ ] Operator documentation update in `docs/guides/DEPLOY_PREVIEW_SCREENSHOTS_GUIDE.md` explaining how to inspect the DLQ

scottschreckengaust:
## Summary

Reconciles the remaining #284 acceptance criteria against `main`. The DLQ half of the issue already landed via the PR #241 follow-up; this PR fills the two gaps: an alarm on the processor Lambda's **Errors** metric, and the operator doc section.

Refs #284

> **#284 stays OPEN (DONE-NO-CLOSE).** AC 5's notification requirement (SNS) is deliberately *not* met here — every alarm in `cdk/src/` today is a bare alarm with no notification action, so wiring SNS is a cross-construct decision for a follow-up. This PR uses `Refs #284` (not `Closes`) so the issue remains open to track that deferred SNS work.

## Root cause / gap analysis

`cdk/src/constructs/github-screenshot-integration.ts:175-220` on `main` **already**:

- defines `WebhookProcessorDlq` (`retentionPeriod` 14 days, `enforceSSL: true`) — AC 3 ✅
- wires it as `deadLetterQueue: processorDlq` on `WebhookProcessorFn` (async-invoke `onFailure`) — AC 1 ✅
- carries the `AwsSolutions-SQS3` cdk-nag suppression — AC 4 ✅
- has `WebhookProcessorDlqDepthAlarm` on the DLQ's `ApproximateNumberOfMessagesVisible` metric (threshold 1)

Two ACs were **UNMET**:

- **AC 2 — alarm on the processor's `Errors` metric (`>= 1` in 5min, 2 eval periods).** The pre-existing alarm is on the *DLQ depth* (`AWS/SQS ApproximateNumberOfMessagesVisible`), which is a **different metric** from the Lambda `Errors` metric the AC asks for. The depth alarm only fires once a failure has survived Lambda's async-retry ladder and *landed* on the queue; an Errors alarm fires as soon as invocations start faulting.
- **AC 5 — operator doc** in `docs/guides/DEPLOY_PREVIEW_SCREENSHOTS_GUIDE.md` explaining how to inspect the DLQ. No such section existed.

## The fix (delta only)

- **New `WebhookProcessorErrorAlarm`** on `webhookProcessorFn.metricErrors()` — `threshold: 1`, `evaluationPeriods: 2`, `period: 5min`, `treatMissingData: NOT_BREACHING`. Mirrors the existing `OrchestratorErrorAlarm` idiom in `cdk/src/constructs/task-orchestrator.ts:401`. **I did NOT duplicate** the pre-existing DLQ-depth alarm — the two are complementary (fault-onset vs landed-on-DLQ) and the new test documents why.
- **Doc:** new Troubleshooting subsection ("No screenshots at all: check the processor alarms and DLQ") describing both alarms and `aws sqs` commands to find, peek, and drain the DLQ. Starlight mirror regenerated via `mise //docs:sync`.

### A note on SNS (deferred to a follow-up)

AC 5 said "notify via existing alarm SNS topic if one exists; otherwise create one". There is **no SNS topic or `addAlarmAction` anywhere in `cdk/src/`** today — every existing alarm (`FanOutConsumer.dlqDepthAlarm`, `OrchestratorErrorAlarm`, the pre-existing `WebhookProcessorDlqDepthAlarm`) is a **bare alarm with no notification action**. To reuse-over-reinvent and stay consistent, this alarm is also bare. Standing up a stack-wide SNS notification plane is a larger cross-construct decision, out of scope for this single construct — it should be a shared topic across all these alarms in a follow-up, not one-off here. The `(+ SNS)` has been dropped from the title, and this PR uses `Refs #284` so the issue stays open to track that deferred notification work.

## Testing

- **CDK assertion test (TDD):** added `alarms on the processor Lambda Errors metric (>=1 in 5min, 2 eval periods)` asserting `MetricName: Errors`, `Namespace: AWS/Lambda`, `Period: 300`, `Threshold: 1`, `EvaluationPeriods: 2`, `TreatMissingData: notBreaching`, `FunctionName` dimension → `WebhookProcessorFn`. Bumped the alarm `resourceCountIs` from 1 → 2. Watched it fail (red), then added the construct code (green).
- `mise //cdk:compile` — pass
- `mise //cdk:eslint` — pass (no mutations)
- `mise //cdk:test` — **136 suites / 2534 tests pass, 1 snapshot pass** (no snapshot change)
- `github-screenshot` suite — 6/6 pass
- `pre-commit run --files <changed>` — all pass (incl. docs-sync, astro check, eslint)
- **cdk-nag:** no new findings on the alarm (CloudWatch alarms trigger no `AwsSolutions` rule); existing SQS3 suppression untouched.

> `mise //cdk:synth` fails **locally** with `ec2:DescribeAvailabilityZones` not authorized — this is a local-credential limitation of the AZ context lookup in `AgentVpc` (`agent-vpc.ts:74`), unrelated to this change (my diff touches no VPC/EC2 resource). CI synth runs with proper credentials.

## Dependencies / related

- Cluster **ua-broad**. **PR #345 also touches `github-screenshot-integration.ts`** (UA work) — whichever merges second will need a rebase; this PR touches only the alarm block + a new constant, so the conflict surface is small.
- Uses only already-present `aws-cdk-lib` modules (`aws-cloudwatch`, `aws-sqs`) — no new dependency.
- Pre-push full-history gitleaks flags 3 pre-existing `aws-account-id` matches in historical docs (commit `1cfa5b9f`, `docs/backlog/*`, `docs/diagrams/*`) — unrelated to this branch; `gitleaks git --log-opts=origin/main..HEAD` on my commit reports **no leaks**.
- AC 6 (smoke test: break the processor, confirm DLQ + alarm fire) is a deploy-tier manual verification, not a code change — left for the operator per the runbook now in the guide.

## Review remediation (round 2 — @theagenticguy + @isadeks)

Both reviewers converged on a **blocking** defect: `WebhookProcessorErrorAlarm` omitted `datapointsToAlarm`, so CloudWatch defaulted it to `evaluationPeriods` (2) → a **2-of-2 consecutive** alarm. Because Lambda `Errors` is a sparse invocation metric, a single-event crash lands its retries in one 5-minute period and `NOT_BREACHING` fills the neighbour — so the alarm **never fires**.

- **`datapointsToAlarm: 1`** added to the alarm — now **1-of-2** (any single breaching period alarms; the evaluation range stays 2). `treatMissingData: NOT_BREACHING` unchanged.
- **Test pins** for the two properties that carry the semantics and previously rode on CDK defaults: `DatapointsToAlarm: 1` and `ComparisonOperator: GreaterThanOrEqualToThreshold` (`GreaterThanThreshold` would silently require **two** errors for a threshold-1 alarm).
- **Comment/JSDoc wording:** the `ERROR_ALARM_EVALUATION_PERIODS` JSDoc and the timing comment no longer say "sustained" — it's 1-of-2, and the timing claim is now "fires no later than — and, when the retry ladder straddles a period boundary, one window before — the DLQ-depth alarm."
- **Runbook + Starlight mirror:** the `WebhookProcessorErrorAlarm` bullet and the timing line in `DEPLOY_PREVIEW_SCREENSHOTS_GUIDE.md` now say the alarm fires on **any** 5-min period with `>= 1` error (evaluation range 2, 1 datapoint to alarm). Mirror regenerated via `mise //docs:sync` (no drift).
- **`Closes #284` → `Refs #284`** and **`(+ SNS)` dropped from the title** — SNS is deferred to a follow-up, so #284 stays open (DONE-NO-CLOSE).


cdk/src/constructs/github-screenshot-integration.ts
```diff
@@ -219,6 +237,32 @@ export class GitHubScreenshotIntegration extends Construct {
...
+    // matches the AC; the metricErrors + 2-eval-period shape follows
+    // ``task-orchestrator.ts`` OrchestratorErrorAlarm (which uses
+    // threshold 3 — we use 1 because any escaped error here is
+    // significant).
+    this.processorErrorAlarm = new cloudwatch.Alarm(this, 'WebhookProcessorErrorAlarm', {
+      metric: this.webhookProcessorFn.metricErrors({
+        period: Duration.minutes(ERROR_ALARM_PERIOD_MINUTES),
+      }),
+      threshold: 1,
+      evaluationPeriods: ERROR_ALARM_EVALUATION_PERIODS,
```
theagenticguy, line 285: **Blocking.** `datapointsToAlarm` is unset here, so CloudWatch defaults it to `evaluationPeriods` (2) and this becomes a **2-of-2 consecutive** alarm. Combined with `Errors` being a sparse invocation metric, that inverts the rationale in the comment above.

Walk the failure this construct actually cares about — a cold-start crash on a single webhook. Lambda's async ladder runs at T, ~T+63s, ~T+190s, so all three `Errors` datapoints usually fall in one 5-minute period. The adjacent period has no datapoint at all, `NOT_BREACHING` fills it, and the alarm never leaves `OK`. That is row 5 of the [missing-data table](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/alarms-and-missing-data.html) (`- - X - -`): NOT BREACHING → `OK`.

```suggestion
      evaluationPeriods: ERROR_ALARM_EVALUATION_PERIODS,
      // Fire on the FIRST breaching datapoint. Unset, this defaults to
      // evaluationPeriods (2 consecutive), which for a sparse invocation
      // metric means a single-event crash never alarms: its retries land
      // in one period and NOT_BREACHING fills the neighbour.
      datapointsToAlarm: 1,
```

This keeps the AC's 2-evaluation-period range while making the alarm fire on the first faulting invocation, which is the behaviour the comment claims.

cdk/src/constructs/github-screenshot-integration.ts
```diff
@@ -219,6 +237,32 @@ export class GitHubScreenshotIntegration extends Construct {
...
+    // by design — it catches its per-step operational failures (bad token,
+    // AgentCore/S3/comment-post errors) and returns success, so those
+    // surface as tagged log events, NOT as Lambda Errors. This alarm
+    // therefore fires on the faults that ESCAPE the handler: an init-time
+    // crash (missing env at cold start, bundling defect), an unhandled
+    // throw in an unguarded path, or the 120s hard timeout. That is the
+    // same failure class the DLQ eventually catches, but the Errors metric
+    // trips one evaluation window sooner — before Lambda's async-retry
+    // ladder has landed a message on the DLQ. threshold 1 / 2 eval periods
+    // matches the AC; the metricErrors + 2-eval-period shape follows
```
theagenticguy: This is the claim that does not survive as configured. At 2-of-2 the `Errors` alarm cannot beat the 1-of-1 DLQ-depth alarm 25 lines up: it either ties it (~T+6min, when the retry ladder straddles a period boundary) or stays silent. With `datapointsToAlarm: 1` the claim becomes true, since `Errors` is stamped at invocation time roughly 3 minutes before the DLQ write.

Worth reworking this paragraph either way. As written it reads as a verified timing argument, and the next person to touch this alarm will trust it.

cdk/test/constructs/github-screenshot-integration.test.ts
```diff
@@ -90,6 +95,36 @@ describe('GitHubScreenshotIntegration construct', () => {
...
+    template.hasResourceProperties('AWS::CloudWatch::Alarm', {
+      MetricName: 'Errors',
+      Namespace: 'AWS/Lambda',
+      // metricErrors() defaults to Sum — the correct statistic for an
+      // error COUNT (vs Average/Maximum). Pin it so a future edit that
+      // swaps the statistic can't silently change the alarm's semantics.
+      Statistic: 'Sum',
+      Period: 300,
+      Threshold: 1,
+      EvaluationPeriods: 2,
```
theagenticguy, line 119: This test passes whether or not `datapointsToAlarm` is set, so it does not pin the property that actually determines when the alarm fires. Same reasoning as the `Statistic: 'Sum'` comment just above — pin the semantics, not just the shape:

```suggestion
      EvaluationPeriods: 2,
      // 1-of-2, not 2-of-2. Unset, DatapointsToAlarm defaults to
      // EvaluationPeriods, and a single-event crash would never alarm
      // (its Errors datapoints land in one period; NOT_BREACHING fills
      // the neighbour). Pinned so that regression is caught here.
      DatapointsToAlarm: 1,
```

docs/guides/DEPLOY_PREVIEW_SCREENSHOTS_GUIDE.md
```diff
@@ -175,6 +175,35 @@ Then tail the function's CloudWatch log group. Common silent skips:
...
+
+- **`WebhookProcessorErrorAlarm`** — fires on the processor's Lambda `Errors` metric (`>= 1` error in a 5-minute period, sustained for 2 evaluation periods). Early signal: an invocation is faulting *now*.
+- **`WebhookProcessorDlqDepthAlarm`** — fires when such a failed invocation has survived Lambda's built-in async retries and landed on the DLQ (construct id `WebhookProcessorDlq`; 14-day retention, SSL-enforced). Backstop: a payload is now parked undelivered. (The queue also gets SSE via SQS's service-side default; the construct doesn't set an explicit `encryption`.)
+
+The two catch the same failure class, but the `Errors` alarm trips one evaluation window sooner (before retries exhaust onto the DLQ). Find and inspect the DLQ — the construct sets no explicit queue name, so its physical name is CloudFormation-generated and *contains* the `WebhookProcessorDlq` construct id:
```
theagenticguy: **Blocking for the runbook specifically** — this is the sentence that will mislead someone at 3am. As configured the `Errors` alarm does not trip sooner; for a single-event crash it does not trip at all, so an operator following this section will wait on a signal that is never coming.

If you take `datapointsToAlarm: 1`, this line is correct as written and needs no edit. If you keep 2-of-2, it should say something like: "The DLQ-depth alarm is the reliable signal for a single failed event. The `Errors` alarm requires two consecutive 5-minute periods with errors, so it fires on sustained breakage rather than a one-off crash."

Same line in the Starlight mirror at `docs/src/content/docs/using/Deploy-preview-screenshots-guide.md:191` — regenerate via `docs:sync` rather than editing both by hand.

laithalsaadoon (dismissed):
Review: **Request changes** — one material defect, plus one scope question. The shape of this PR is right: the gap analysis is accurate (I confirmed AC 1/3/4 already landed on `main`), the delta is minimal, the test is real, and the SNS restraint is well argued. But the new alarm as configured does the opposite of what its own comment claims, in a way that will burn whoever is oncall.

### Blocking: `datapointsToAlarm` is unset, so the alarm is 2-of-2 and cannot fire "sooner"

The PR asserts in three places (source comment, test comment, operator runbook) that the `Errors` alarm "trips one evaluation window sooner" than the DLQ-depth alarm. As configured it trips **later or never**. Two facts compound:

1. `datapointsToAlarm` is not set, and [CloudWatch defaults it to `evaluationPeriods`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudwatch-alarm.html#cfn-cloudwatch-alarm-datapointstoalarm): "If you omit this parameter, CloudWatch uses the same value here that you set for EvaluationPeriods, and the alarm goes to alarm state if that many **consecutive** periods are breaching." So this alarm needs **two consecutive breaching 5-minute datapoints**. The DLQ-depth alarm next to it is 1-of-1.

2. Lambda `Errors` is an invocation metric: no invocation means **no datapoint**, not a zero. For exactly the failure class this PR names (init-time cold-start crash, bundling defect, one unguarded throw), the three attempts land at T, ~T+63s, ~T+190s — a ~3.2 minute span, usually inside a single 5-minute period. The neighbouring period has no datapoint, `treatMissingData: NOT_BREACHING` fills it non-breaching, and the alarm stays `OK` forever. This is row 5 of the [missing-data table](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/alarms-and-missing-data.html) (`- - X - -`): `MISSING` → ALARM, `BREACHING` → ALARM, **`NOT BREACHING` → OK**. The premature-alarm rule does not rescue it, because the fill happens first.

When the retry ladder does straddle a period boundary (~60% of the time), you get two real breaching datapoints and it reaches ALARM around T+6min — the same window as the DLQ alarm, not earlier. So the best case is a tie and the common case is silence, while the DLQ alarm fires at ~T+5-6min every time.

**Fix is one line: `datapointsToAlarm: 1`.** That makes the alarm fire on the first faulting invocation, the rationale becomes true as written, and it still satisfies the AC's ">= 1 error in 5 min, 2 evaluation periods" (evaluation range stays 2). Keep `NOT_BREACHING`. Please add a test assertion pinning `DatapointsToAlarm: 1`, since that is the property carrying the semantics — the current test would pass either way.

If instead you want a sustained-error alarm, that is a defensible choice, but then all three comments and the runbook line need to say "sustained" rather than "sooner", and the complementary-not-redundant argument needs rewriting, because a 2-of-2 Errors alarm is strictly slower than the 1-of-1 DLQ alarm it sits beside.

### Non-blocking: `Closes #284` with no alarm action

I verified the SNS claim and it holds: there is no `sns.Topic` or `addAlarmAction` anywhere in `cdk/src` (only 4 bare alarms across `fanout-consumer.ts`, `task-orchestrator.ts`, and this construct). Declining to one-off a topic here is the right call, and a shared notification plane is a legitimately separate change.

But AC 5 asked for notification, and a bare alarm notifies nobody. The runbook calls these "two operator-visible signals", which is true only for an operator already reading the CloudWatch console — the exact state the issue was trying to fix. Suggest dropping `Closes #284` in favour of `Refs #284` and leaving the SNS AC open, or cutting the follow-up issue for the shared topic and linking it here. Also worth dropping `(+ SNS)` from the title, since the code deliberately does not add SNS and the title will land in the changelog.

### Confirmed good

- `Statistic: 'Sum'` pinned in the test with a comment explaining why — nice, that is the assertion most people forget.
- The doc's SSE aside is correct: [SSE-SQS is on by default](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_sqs-readme.html) for new queues, so omitting `encryption` is fine.
- Receiver behaviour in the runbook matches `github-webhook.ts:186-211`: `InvocationType: 'Event'`, 200 on accepted dispatch, 500 only on invoke failure with dedup rollback. The "GitHub never redelivers" framing is accurate.
- Both doc copies are byte-identical in the new section, so `docs:sync` did its job.
- Note this is still a **draft**, and `build (agentcore)` was pending when I looked.

scottschreckengaust:

**#284** — GitHub webhook processor DLQ + alarm coverage, reconciled against current `main`.

- The **DLQ + a DLQ-depth alarm already exist on main** (from the PR #241 follow-up) — this PR does NOT duplicate them.
- Fills the two unmet acceptance criteria: (AC2) a new `WebhookProcessorErrorAlarm` on the processor Lambda **Errors** metric (threshold 1, 2 eval periods, 5min — a distinct signal from the depth alarm), mirroring the in-repo `OrchestratorErrorAlarm` idiom; and (AC5) an operator DLQ/alarm Troubleshooting section in `DEPLOY_PREVIEW_SCREENSHOTS_GUIDE.md` (Starlight mirror regenerated).
- Alarm left bare (no SNS action) to match repo convention — zero existing alarms wire notifications. A shared SNS notification plane is flagged as a reviewer decision, out of scope here.

`/review_pr` = **approve-with-nits, zero blocking**. cdk-test 2534/2534, cdk-nag clean.

**Cluster `ua-broad` — must-follow #345.** Shares `cdk/src/constructs/github-screenshot-integration.ts` with #345 (the UA aspect). If #345 merges first (recommended — it is the cluster lead), this rebases trivially (my change is an isolated alarm block). Not strict-up-to-date-gated. Native auto-merge disabled repo-wide; awaits your manual squash-merge.

isadeks (commented):
## Verdict: **Comment — no blockers. @theagenticguy's blocking defect is FIXED; 5 nits remain (3 carried forward, 2 new).**

The one material defect in the prior review — `datapointsToAlarm` unset, making this a 2-of-2 alarm that would never fire on a single-event crash — **has been fixed** in `451ab22` + `93a715a` (accepted suggestions, co-authored to @theagenticguy). I verified it against HEAD rather than taking the commit messages at their word:

- `cdk/src/constructs/github-screenshot-integration.ts:265` — `datapointsToAlarm: 1` is present.
- `cdk/test/constructs/github-screenshot-integration.test.ts:119` — `DatapointsToAlarm: 1` is pinned, so the regression is caught in CI.

I re-derived the M-of-N semantics independently and confirm the fix is correct: with N=2 / M=1, CloudWatch evaluates the trailing two periods and alarms if **any one** breaches. `NOT_BREACHING` still fills the empty neighbouring period, but M=1 doesn't need it — the single breaching period is sufficient. A one-off crash now reaches ALARM at the close of the period containing the fault. @theagenticguy's analysis was right and the fix is the right one-liner. **I am not re-raising it.**

What's left is comment/doc rot *introduced by that fix* plus PR metadata hygiene. None of it is a functional defect, so I'm not blocking — but N1/N2 should land before merge, because they now say the opposite of what the code does.

---

### Scope divergence (I was asked to assess this specifically)

The issue asks for DLQ + alarm + SNS; the PR delivers only the alarm + docs. **The partial delivery is declared, and the gap analysis is accurate — I verified it independently against `origin/main` rather than trusting the write-up:**

| AC | Status | Evidence |
|---|---|---|
| 1 — DLQ on async-invoke `onFailure` | pre-existing on `main` | `git show origin/main:…/github-screenshot-integration.ts` → `WebhookProcessorDlq` at :175, `deadLetterQueue: processorDlq` at :187 |
| 2 — alarm on `Errors` (>=1/5min, 2 eval periods) | **delivered by this PR** | new `WebhookProcessorErrorAlarm` |
| 3 — 14-day retention; SSE-KMS | retention yes, **SSE-KMS no** | `retentionPeriod` 14d on `main`; no `encryption` prop anywhere → SSE-**SQS**, not SSE-KMS |
| 4 — cdk-nag clean | pre-existing | `AwsSolutions-SQS3` suppression on `main` at :201 |
| 5 — operator doc | **delivered by this PR** | new Troubleshooting section |
| 5b — notify via SNS | **not delivered** | confirmed: zero `sns.Topic` / `addAlarmAction` in `cdk/src` |
| 6 — smoke test | deploy-tier, declared | fine |

So the delta framing is honest and the restraint on SNS is the right call — I reached the same conclusion as @theagenticguy on that. Two ACs (5b, and 3 read strictly) are genuinely unmet, which is what makes N3 below worth acting on.

On AC 3: SSE-SQS instead of SSE-KMS is a *defensible* deviation for a DLQ of GitHub webhook payloads — it's encrypted at rest, costs nothing, and needs no key policy. I'm not asking you to change it. I'm noting that `Closes #284` will silently close an AC that reads "SSE-KMS", and the doc's parenthetical at guide:185 is the only place that discrepancy is recorded.

---

### Non-blocking

**N1 — `"sustained"` is now the wrong mental model in the source (`…/github-screenshot-integration.ts:46`, `:250`).** Inline. The `datapointsToAlarm: 1` fix changed the semantics from 2-of-2 to 1-of-2, but the constant's JSDoc still reads "sustained evaluation periods". That is the *exact* mental model that produced the bug just fixed — the next maintainer who reads "sustained", sees `datapointsToAlarm: 1`, and concludes the two contradict may "clean up" the line and silently restore the 2-of-2 behaviour. The test pin at :119 would catch it, which is good defence-in-depth, but the comment shouldn't be laying the trap in the first place.

**N2 — the runbook contradicts itself (`docs/guides/DEPLOY_PREVIEW_SCREENSHOTS_GUIDE.md:184` vs `:187`).** Inline. Line 184 says the alarm fires "sustained for 2 evaluation periods"; line 187 says it "trips one evaluation window sooner". Post-fix, 187 is right and 184 is wrong. The failure direction has inverted from what @theagenticguy flagged — an operator is no longer waiting on a signal that never comes; they're now under-trusting an alarm that fires more eagerly than the doc claims. Lesser harm, still wrong, and a self-contradicting runbook is corrosive. Mirror needs the same edit at `docs/src/content/docs/using/Deploy-preview-screenshots-guide.md:188` — regenerate via `mise //docs:sync`, don't hand-edit.

While you're in that paragraph: "trips one evaluation window sooner" is unconditional but only true when the retry ladder straddles a period boundary. When the fault lands early in a period, the DLQ write at ~T+190s falls in the *same* period and both alarms fire at the same evaluation — a tie. "Fires no later than, and often one window before, the DLQ-depth alarm" is the claim that actually survives. Same over-claim class as the one that got caught the first time.

**N3 — `Closes #284` closes an issue with unmet ACs; and the follow-up issue you were asked to cut already exists.** @theagenticguy suggested `Refs #284` plus cutting a follow-up for the shared SNS topic. I searched before repeating that ask, and the follow-up is already filed — **twice**:

- **#629** — *feat(observability): wire SNS notification action to DLQ CloudWatch alarms* (`enhancement`, `infra-cdk`, `observability`)
- **#228** — *feat(cdk): wire DLQ CloudWatch alarms to SNS topic for alerting*

#629 is scoped exactly right for this — it already names the "reusable `OperationalAlerts` construct" approach and cites `CEDAR_HITL_GATES.md` §11.5 for the bare-alarm-as-intentional-intermediate-step decision. So: switch to `Refs #284`, link #629 (and note #228 as its likely duplicate), and add `WebhookProcessorErrorAlarm` / `WebhookProcessorDlqDepthAlarm` to #629's target list. That closes the governance loop without a new issue and without closing #284 over an undelivered AC.

**N4 — drop `(+ SNS)` from the title** (carried forward from @theagenticguy, still unaddressed at HEAD). The code deliberately adds no SNS; the title lands in the changelog and `Validate PR title` won't catch a semantically-false scope claim. You already offered to drop it — please do.

**N5 — the test doesn't pin `ComparisonOperator`** (`…/github-screenshot-integration.test.ts:115`). Inline. Same reasoning that motivated the `DatapointsToAlarm` pin: `>= 1` is half the AC, and it rides on a CDK default (`GreaterThanOrEqualToThreshold`). Flip it to `GreaterThanThreshold` and the alarm quietly needs **two** errors instead of one — every current assertion still passes. One line closes it.

---

### Confirmed good

Things I checked and am satisfied with, so they don't come back in a later pass:

- **Bootstrap policy coverage — PASS, no update needed.** This adds a second instance of an existing CFN type, not a new one. `AWS::CloudWatch::Alarm` → `cloudwatch:PutMetricAlarm` is already in `cdk/src/bootstrap/resource-action-map.ts:62`, granted at `cdk/src/bootstrap/policies/observability.ts:76`, and present in the `DEPLOYMENT_ROLES.md:564` golden baseline. Correctly **no** `BOOTSTRAP_VERSION` bump — bumping 1.2.0 here would have been the wrong call, and I'm glad it wasn't done reflexively.
- **Test-performance rules (#366) — PASS.** Single `new App()` + `Template.fromStack()` in `beforeAll`, reused across tests; nothing re-enables `aws:cdk:bundling-stacks`.
- **Test actually pins behaviour, not shape.** I mutation-reasoned each assertion: dropping `datapointsToAlarm` fails on `DatapointsToAlarm`; deleting the alarm fails `resourceCountIs(…, 2)`; swapping the statistic fails `Statistic: 'Sum'`; attaching to the *receiver* instead of the processor fails the `FunctionName` → `WebhookProcessorFn` dimension matcher. The `resourceCountIs` bump from 1 → 2 is the right way to make alarm deletion loud. Only `ComparisonOperator` (N5) escapes.
- **Docs mirror is genuinely in sync**, not just claimed — I diffed the new section in both copies (byte-identical) and re-ran `docs/scripts/sync-starlight.mjs`, which produced no changes (`git status` clean).
- **The handler-swallows-its-own-errors framing is accurate.** Verified against `cdk/src/handlers/github-webhook-processor.ts` — six `catch` blocks that log-and-return (`:121`, `:165`, `:213`, `:238`, `:269`, `:413`, `:440`), so operational faults genuinely do *not* tick `Errors`. The doc's "watch the logs, not these alarms" caveat at guide:182 is the honest and useful thing to tell an operator, and it's the kind of caveat most PRs omit.
- **Receiver behaviour in the runbook matches the code** — `github-webhook.ts:187-188` (`InvocationType: 'Event'`), wired via `GITHUB_WEBHOOK_PROCESSOR_FUNCTION_NAME` at construct:332 with `grantInvoke` at :342.
- **`public readonly processorErrorAlarm` follows the sibling convention** (`FanOutConsumer.dlqDepthAlarm`, `TaskOrchestrator.errorAlarm` are likewise exposed-but-unconsumed) and is exactly the handle #629 will need to call `addAlarmAction` on. Cheap now, useful later.
- **CI is green** across all 8 checks including `build (agentcore)`, which was pending when @theagenticguy looked. The PR is also no longer a draft.

### Human heuristics

- **Proportionality — pass.** ~30 lines of construct for one alarm, two named constants, one test, one doc section. No abstraction invented for a one-off; explicitly declined to stand up an SNS plane for a single construct, which is the right instinct (AI002/AI003 avoided).
- **Coherence — pass with N1.** Mirrors the `OrchestratorErrorAlarm` idiom and deviates only where justified (threshold 1 vs 3, explained). Only incoherence is the term "sustained" now describing 1-of-2 semantics (src:46).
- **Clarity — concern, N1/N2.** Names are good and the alarm descriptions are genuinely actionable (they name *what to check*, not just "errors happened"). But three sites still assert timing behaviour the code no longer has.
- **Appropriateness — pass.** No mock-only verification here; the claims are about CloudWatch evaluation semantics and are checkable against AWS docs, which is how the original defect was caught. The AC-by-AC gap analysis against `main` is exactly the right way to handle a partially-landed issue, and it held up under independent verification — that's the strongest thing about this PR.

### Review agents run

Process disclosure, since the skill requires it: **the subagent-dispatch tool was not available in this execution context**, so I could not spawn the `pr-review-toolkit` agents as separate agents. Rather than skip the step, I loaded each agent's published rubric from `plugins/pr-review-toolkit/agents/` and applied it as an explicit, separate pass over the diff. Stating that plainly instead of claiming a fan-out I didn't get:

- **`code-reviewer`** (rubric applied; confidence-≥80 filter) — no findings at or above threshold. Bootstrap/`#366`/routing checks all clean.
- **`comment-analyzer`** (rubric applied) — produced N1 and the "one evaluation window sooner" over-claim. This was the highest-yield pass; the fix that resolved the blocker left its own justifying comments stale.
- **`pr-test-analyzer`** (rubric applied) — produced N5 (criticality ~5: pins the AC's `>=`, guards a CDK default). Coverage otherwise rated strong; see the mutation-reasoning above.
- **`type-design-analyzer`** (rubric applied) — no new types or interfaces in the diff; one new `public readonly` field matching an established convention. Nothing to report.
- **`silent-failure-hunter`** — **omitted, out of scope.** The diff adds no error-handling or fallback code. (I did read the handler's `catch` blocks, but to verify the doc's claims, not as changed code.)
- **`/security-review`** — **omitted, out of scope.** No IAM, Cedar, network, secrets, or input-gateway change. A CloudWatch alarm grants nothing and reads no data; the deploy-time IAM it needs was already granted (see bootstrap note).

---

Good PR. The gap analysis against `main` is the part worth calling out — reconciling a partially-landed issue AC-by-AC, and *declining* to duplicate the existing DLQ-depth alarm, is more disciplined than the usual "implement the issue text verbatim" response. Fix N1/N2 so the comments match the code you now have, retarget `Closes` → `Refs` with a link to #629, drop `(+ SNS)`, and I'd approve this on sight.

cdk/src/constructs/github-screenshot-integration.ts
```diff
@@ -40,6 +40,12 @@ const PROCESSOR_DLQ_RETENTION_DAYS = 14;
 /** DLQ-depth alarm metric period (minutes). */
 const DLQ_ALARM_PERIOD_MINUTES = 5;
 
+/** Processor Lambda-Errors alarm metric period (minutes). */
+const ERROR_ALARM_PERIOD_MINUTES = 5;
+
+/** Processor Lambda-Errors alarm sustained evaluation periods. */
```
isadeks: **N1 (non-blocking, but please fix before merge).** After the `datapointsToAlarm: 1` fix at :265, `"sustained"` is the wrong word — the alarm is now **1-of-2**, so `evaluationPeriods` is an evaluation *range* (which also gives a little ALARM-state hysteresis), not a sustain requirement.

This matters more than a typo would, because "sustained evaluation periods" is precisely the mental model that produced the defect @theagenticguy caught. The next person who reads this JSDoc, sees `datapointsToAlarm: 1` twenty lines down, and concludes the two contradict each other may "tidy up" the `datapointsToAlarm` line and silently restore the 2-of-2 behaviour. The test pin at `github-screenshot-integration.test.ts:119` would catch that — good defence-in-depth — but the comment shouldn't be setting the trap.

```suggestion
/** Processor Lambda-Errors alarm evaluation range (periods). Paired with
 *  `datapointsToAlarm: 1` below, this is 1-of-2, NOT 2-of-2: any single
 *  breaching period alarms. Do not remove `datapointsToAlarm` — CloudWatch
 *  would default it back to this value and a one-off crash would never fire. */
const ERROR_ALARM_EVALUATION_PERIODS = 2;
```

cdk/src/constructs/github-screenshot-integration.ts
```diff
@@ -219,6 +237,37 @@ export class GitHubScreenshotIntegration extends Construct {
...
+    // the DLQ-depth alarm above, not redundant. The handler is best-effort
+    // by design — it catches its per-step operational failures (bad token,
+    // AgentCore/S3/comment-post errors) and returns success, so those
+    // surface as tagged log events, NOT as Lambda Errors. This alarm
+    // therefore fires on the faults that ESCAPE the handler: an init-time
+    // crash (missing env at cold start, bundling defect), an unhandled
+    // throw in an unguarded path, or the 120s hard timeout. That is the
+    // same failure class the DLQ eventually catches, but the Errors metric
+    // trips one evaluation window sooner — before Lambda's async-retry
+    // ladder has landed a message on the DLQ. threshold 1 / 2 eval periods
```
isadeks: **N1 (cont.) — the timing claim is unconditional but only holds some of the time.**

"trips one evaluation window sooner" is now *directionally* true post-fix (it was not before), so this is a much smaller over-claim than the original. But it isn't reliably one window. Walking it: `Errors` is stamped at invocation time T and evaluates at the close of T's period. The DLQ write happens after the async ladder exhausts, ~T+190s.

- Fault lands **early** in a period → T+190s falls in the *same* period → both alarms fire at the same evaluation. A **tie**, not a window earlier.
- Fault lands **late** in a period → T+190s spills into the next → genuinely one window earlier.

So the guarantee is "never later, often earlier". Suggest phrasing it as the claim that actually survives — this is the same over-claim class that got caught the first time round, and the next reader will trust it as a verified timing argument:

```suggestion
    // ladder has landed a message on the DLQ. The Errors metric is stamped
    // at invocation time, so this alarm fires no later than — and, when the
    // retry ladder straddles a period boundary, one window before — the
    // DLQ-depth alarm. threshold 1 / 1-of-2 eval periods
```

docs/guides/DEPLOY_PREVIEW_SCREENSHOTS_GUIDE.md
```diff
@@ -175,6 +175,35 @@ Then tail the function's CloudWatch log group. Common silent skips:
...
+
+**Important:** the processor handler is best-effort by design — it catches its own per-step operational failures (bad token, AgentCore/S3/comment-post errors) and returns success, so those show up as tagged log events (see the section above), **not** as Lambda `Errors`. Both alarms below fire only on faults that *escape* the handler: an init-time crash (missing env at cold start, bundling defect), an unhandled throw in an unguarded path, or the 120s hard timeout. For a swallowed operational failure (e.g. a revoked S3/AgentCore permission), watch the processor **logs**, not these alarms.
+
+- **`WebhookProcessorErrorAlarm`** — fires on the processor's Lambda `Errors` metric (`>= 1` error in a 5-minute period, sustained for 2 evaluation periods). Early signal: an invocation is faulting *now*.
```
isadeks: **N2 (non-blocking, but the runbook now contradicts itself — please fix before merge).**

This line says the alarm fires "sustained for 2 evaluation periods". Line 187 says it "trips one evaluation window sooner". Post-`datapointsToAlarm: 1`, **187 is right and this line is wrong** — a single faulting invocation alarms at the first breaching period.

The failure direction has inverted from what you originally flagged, @theagenticguy: an operator is no longer waiting on a signal that never arrives — they're now *under*-trusting an alarm that fires more eagerly than the doc promises. Strictly the lesser harm, but a runbook that states two incompatible things three lines apart is the thing that costs someone time at 3am, and it's a one-word fix.

```suggestion
- **`WebhookProcessorErrorAlarm`** — fires on the processor's Lambda `Errors` metric: **any** 5-minute period with `>= 1` error alarms (evaluation range 2 periods, 1 datapoint to alarm). Early signal: an invocation is faulting *now*.
```

Same edit needed in the Starlight mirror at `docs/src/content/docs/using/Deploy-preview-screenshots-guide.md:188` — regenerate with `mise //docs:sync` rather than hand-editing both. I re-ran the sync script against HEAD and it was clean, so the mirror is genuinely in sync today; it'll need a re-run after this change.

cdk/test/constructs/github-screenshot-integration.test.ts
```diff
@@ -90,6 +95,41 @@ describe('GitHubScreenshotIntegration construct', () => {
...
```
isadeks, line 119: **N5 (nit).** `ComparisonOperator` is unpinned, so `>= 1` — half of the AC — rides entirely on a CDK default (`GreaterThanOrEqualToThreshold`).

This is the same reasoning that motivated pinning `DatapointsToAlarm` right below: pin the property that carries the semantics. Flip the operator to `GreaterThanThreshold` and the alarm quietly requires **two** errors instead of one, which is a real behaviour change for a threshold-1 alarm — and every assertion in this test still passes.

```suggestion
      Threshold: 1,
      // >= 1, not > 1. Relies on a CDK default otherwise, and
      // GreaterThanThreshold would silently require TWO errors to fire.
      ComparisonOperator: 'GreaterThanOrEqualToThreshold',
      EvaluationPeriods: 2,
```

cdk/src/constructs/github-screenshot-integration.ts
scottschreckengaust: Resolved. Added `datapointsToAlarm: 1`, so the alarm is now 1-of-2 and the `Errors` metric (stamped at invocation time, ~3 min before the DLQ write) genuinely leads the DLQ-depth alarm. Reworked the paragraph to the claim that survives: it "fires no later than — and, when the retry ladder straddles a period boundary, one window before — the DLQ-depth alarm" (a tie when the fault lands early in a period, one window earlier when it lands late). No longer an unconditional "one window sooner" over-claim. — @scottschreckengaust (agent:w5)

docs/guides/DEPLOY_PREVIEW_SCREENSHOTS_GUIDE.md
scottschreckengaust: Resolved by taking `datapointsToAlarm: 1` (so the signal now genuinely arrives) AND rewording the runbook. The `WebhookProcessorErrorAlarm` bullet now reads: "fires on the processor's Lambda `Errors` metric: **any** 5-minute period with `>= 1` error alarms (evaluation range 2 periods, 1 datapoint to alarm). Early signal: an invocation is faulting *now*." Regenerated the Starlight mirror via `mise //docs:sync` (no hand-editing of the mirror; sync re-run is clean). — @scottschreckengaust (agent:w5)

cdk/src/constructs/github-screenshot-integration.ts
scottschreckengaust: Fixed. Adopted your suggested JSDoc verbatim: `ERROR_ALARM_EVALUATION_PERIODS` now documents "1-of-2, NOT 2-of-2: any single breaching period alarms. Do not remove `datapointsToAlarm` — CloudWatch would default it back to this value and a one-off crash would never fire." The word "sustained" is gone, so the JSDoc no longer sets the trap you described. — @scottschreckengaust (agent:w5)

cdk/src/constructs/github-screenshot-integration.ts
scottschreckengaust: Fixed. Adopted your suggested wording for the timing comment: "The Errors metric is stamped at invocation time, so this alarm fires no later than — and, when the retry ladder straddles a period boundary, one window before — the DLQ-depth alarm. threshold 1 / 1-of-2 eval periods". The unconditional "one window sooner" claim is gone — it now states the "never later, often earlier" guarantee that actually holds. — @scottschreckengaust (agent:w5)

docs/guides/DEPLOY_PREVIEW_SCREENSHOTS_GUIDE.md
scottschreckengaust: Fixed. Adopted your suggested bullet verbatim at line 184: "fires on the processor's Lambda `Errors` metric: **any** 5-minute period with `>= 1` error alarms (evaluation range 2 periods, 1 datapoint to alarm). Early signal: an invocation is faulting *now*." The runbook no longer contradicts itself — line 184 and the timing line now agree that a single faulting invocation alarms at the first breaching period. Mirror regenerated via `mise //docs:sync` (re-run clean, no drift). — @scottschreckengaust (agent:w5)

cdk/test/constructs/github-screenshot-integration.test.ts
scottschreckengaust: Fixed. Adopted your suggested pin verbatim: the test now asserts `ComparisonOperator: 'GreaterThanOrEqualToThreshold'` (with the ">= 1, not > 1 ... GreaterThanThreshold would silently require TWO errors" comment), alongside the new `DatapointsToAlarm: 1` pin. Both properties that carry the alarm semantics are now pinned rather than riding on CDK defaults. Suite green (6/6). — @scottschreckengaust (agent:w5)

isadeks (commented):
## Verdict: **Comment — 0 blockers. 5 of my 6 prior findings are genuinely FIXED (2 mutation-proven). 2 survive as nits, 1 new nit from the fix commit.**

Re-review at `b822a393`. I judged from the code at the new head, not from the thread state or the commit message. The headline: **the fix is real and the tests now genuinely pin it** — I literally reverted each fix in the worktree and watched the new assertions go red. Details below, then the two loose ends.

---

### Prior findings — adjudication

| # | Finding | Status | Evidence at new head |
|---|---|---|---|
| — | @theagenticguy's blocker: `datapointsToAlarm` unset → 2-of-2, never fires | **FIXED** (was already fixed at `93a715a`) | `github-screenshot-integration.ts:266`; **mutation-proven**, see below |
| N1 | `"sustained"` in `ERROR_ALARM_EVALUATION_PERIODS` JSDoc | **FIXED** | `:46-50` — suggestion adopted verbatim, incl. the "do not remove `datapointsToAlarm`" warning |
| N1-cont | unconditional "trips one evaluation window sooner" timing claim | **PARTIALLY FIXED** | constructor comment `:252-255` fixed; **the same two claims survive verbatim in the field JSDoc at `:145` and `:147`** — inline |
| N2 | runbook contradicts itself (`guide:184` vs `:187`) | **FIXED** | both lines now agree; mirror byte-identical; I re-ran `docs/scripts/sync-starlight.mjs` → no drift |
| N5 | `ComparisonOperator` unpinned | **FIXED** | `test:118` — **mutation-proven**, see below |
| N3 | `Closes #284` closes an issue with unmet ACs; link #629 | **PARTIALLY FIXED** | PR body + `b822a393` say `Refs`, but **`abd857be:38` still says `Closes #284`** — see below |
| N4 | drop `(+ SNS)` from the title | **FIXED** | title is now `feat(infra): add Lambda-error alarm to GitHub webhook processor (#284)` |

### Mutation-tested, not taken on trust

My prior review's core complaint was that the *original* bug was invisible to the tests. So I checked the new pins the only way that means anything — by breaking the code:

1. **Deleted `datapointsToAlarm: 1`** from the construct → `✕ alarms on the processor Lambda Errors metric` — `Missing key 'DatapointsToAlarm'`. Red.
2. **Set `comparisonOperator: GREATER_THAN_THRESHOLD`** → `✕` same test — `"ComparisonOperator": "GreaterThanThreshold"` vs expected `"GreaterThanOrEqualToThreshold"`. Red.
3. Restored both → **6/6 green**, worktree back to a clean `b822a393`.

Both properties that carry the alarm's semantics now fail loudly if a future edit "tidies them up". That is exactly the regression class that produced the original defect, and it is now closed in CI. I also re-derived the runtime semantics: N=2 / M=1 with `NOT_BREACHING` → CloudWatch evaluates the trailing two periods, the single breaching period satisfies M=1, and the missing neighbour being filled non-breaching no longer matters. A one-off cold-start crash reaches ALARM at the close of the period containing the fault. **The fix is correct and I am not re-raising it.**

### Regression hunt on the fix commit — clean, with one exception

- No functional change beyond the one-line `datapointsToAlarm`. `1 <= evaluationPeriods`, so CDK's own validation is satisfied; synth is fine.
- Nothing deleted, so no orphaned callers/types/docs. `grep` for `processorErrorAlarm` / `WebhookProcessorErrorAlarm` returns exactly the 2 source + 2 doc sites, all consistent.
- Both new constants are actually consumed (`:261`, `:264`) — no dead constant left behind.
- No other suite asserts an alarm count for this construct (`task-orchestrator.test.ts:361` and `fanout-consumer.test.ts:147` count alarms in *their own* stacks; unaffected). No snapshot churn.
- **Bootstrap coverage — still correctly untouched.** Second instance of an existing CFN type; `AWS::CloudWatch::Alarm → cloudwatch:PutMetricAlarm` already at `resource-action-map.ts:62`. No `BOOTSTRAP_VERSION` bump needed, and correctly not done.
- **Exception:** the reword left a broken sentence at `:251` — inline.

---

### Nits (3) — none blocking

**N-a — the field JSDoc still carries *both* claims you corrected 100 lines down.** Inline at `:145`. This is the one that matters, because it's not a fresh mistake — it's the *same two sentences* you fixed in the constructor comment, surviving verbatim in the `public readonly processorErrorAlarm` doc that a consumer of this construct reads first. `git show 93a715a:…` confirms `:145`/`:147` are byte-identical to the pre-fix text.

**N-b — `Closes #284` survives in `abd857be`, and this repo's squash policy will inline it.** No RIGHT-side anchor, so it lives here. You did retarget the PR body and the fix commit to `Refs #284`, and the DONE-NO-CLOSE rationale in the body is clear and correct. But the mechanism you're relying on isn't the only one in play:

- `abd857be`'s message body still ends `Closes #284` (`git log -1 --format=%B abd857be` → line 38).
- I checked the repo settings: `allow_squash_merge: true`, `allow_merge_commit: false`, `allow_rebase_merge: false`, and **`squash_merge_commit_message: "COMMIT_MESSAGES"`**.

So the squash commit that lands on `main` will have a body containing the concatenated commit messages — **including `Closes #284`** — and GitHub closes issues on closing-keywords in commits merged to the default branch, independently of the PR description. Net effect: a **P1** issue with genuinely unmet ACs (5b SNS; AC 3's SSE-KMS read strictly; AC 6 smoke test) auto-closes anyway, which is the precise outcome your own PR body says you're avoiding. Two safe remedies: amend `abd857be`'s message (`Closes` → `Refs`), or have whoever merges edit the squash body in the merge box. The first is durable; the second depends on someone remembering. Worth 30 seconds either way — please also verify #284 is still open after merge.

While you're there: **#629 is still not referenced from the PR body**, and `WebhookProcessorErrorAlarm` / `WebhookProcessorDlqDepthAlarm` haven't been added to its target list. GitHub did auto-create a cross-reference on #629 from this PR, so the trail isn't cold — but the deferred-SNS handoff is currently only findable by walking backlinks.

**N-c — broken sentence introduced by the reword.** Inline at `:251`.

---

### Scope (re-checked, unchanged — governance call, not a code blocker)

No DLQ was added by this PR and none was needed: `git show origin/main:…/github-screenshot-integration.ts` confirms `WebhookProcessorDlq` at `:175` (14-day retention, `enforceSSL`), wired as `deadLetterQueue` at `:187`, with the `AwsSolutions-SQS3` suppression at `:201`. AC 1/3/4 genuinely pre-date this PR; this PR delivers AC 2 + AC 5-docs. The partial delivery **is** explicitly declared — twice in the PR body, with an AC-by-AC table and a reasoned SNS deferral. I reached the same conclusion as @theagenticguy that declining to stand up an SNS plane inside one construct is the right call. The only gap in the declaration is the mechanical one in N-b: the *stated* intent is DONE-NO-CLOSE, the *effective* behaviour on squash-merge is close. Fix that and the governance loop is honestly closed.

### Documentation

Guide + Starlight mirror both updated; the new section and both edited lines are byte-identical across the two copies, and re-running the sync script produced **no** changes. Mirror is genuinely in sync, not merely claimed. `docs/src/content/docs/` was not hand-edited.

### Tests & CI

`github-screenshot-integration` suite 6/6 green at `b822a393` (run locally). All 9 checks green including `build (agentcore)` and all three CodeQL analyses. Test-performance rules (#366) still respected: single `new App()` + `Template.fromStack()` in `beforeAll`, nothing re-enables `aws:cdk:bundling-stacks`. Bootstrap synth-coverage: **not applicable** (no new CFN resource type).

### Review agents run

Process disclosure, as the skill requires. **The subagent-dispatch tool is not available in this execution context**, so I could not spawn the `pr-review-toolkit` agents as separate agents. Rather than skip the step or claim a fan-out I didn't get, I loaded each agent's rubric from `plugins/pr-review-toolkit/agents/` and applied it as an explicit separate pass over the delta:

- **`comment-analyzer`** (rubric applied) — highest-yield pass again, and again on the fix commit's own comments: produced **N-a** (stale JSDoc, factual-accuracy check §1) and **N-c** (misleading/ungrammatical element, §4).
- **`pr-test-analyzer`** (rubric applied) — no remaining gaps. The two new pins are exactly the two properties I asked for and both survive mutation testing; N5 is closed.
- **`code-reviewer`** (rubric applied) — no findings at or above threshold. Bootstrap, #366, routing all clean.
- **`type-design-analyzer`** (rubric applied) — no type changes in the delta. Nothing to report.
- **`silent-failure-hunter`** — **omitted, out of scope.** The delta is one alarm property, three comments and two doc lines; no error-handling or fallback code.
- **`/security-review`** — **omitted, out of scope.** No IAM, Cedar, network, secrets or input-gateway change. `datapointsToAlarm` grants nothing.

### Human heuristics

- **Proportionality — pass.** One-line functional fix plus comment/doc corrections. No abstraction invented; the SNS restraint holds.
- **Coherence — pass.** "sustained" is gone from the constant, and the 1-of-2 vocabulary is now consistent between construct, test and runbook — except the field JSDoc (N-a).
- **Clarity — concern (N-a, N-c).** The alarm description is genuinely actionable and the semantics are now pinned in CI. But one JSDoc still asserts timing the code doesn't have, and one sentence lost its main clause.
- **Appropriateness — pass.** The claims are about CloudWatch evaluation semantics and are checkable against AWS docs — which is how the original defect was caught, and how I verified the fix. Adopting both reviewers' suggestions verbatim with co-authorship, rather than reinterpreting them, is the right instinct on a semantics bug.

---

Good remediation. You took the blocker fix, pinned it in a way that actually fails when reverted (which is more than the first round managed), and corrected the docs in both copies with a real sync run. The two survivors are both one-liners: mirror your `:252-255` rewording up into the `:140-149` JSDoc, and make sure `abd857be`'s `Closes #284` can't reach the squash body. Neither blocks merge; the second one does change what happens to a P1 issue, so please don't let it ride.

cdk/src/constructs/github-screenshot-integration.ts
```diff
@@ -128,6 +137,18 @@ export class GitHubScreenshotIntegration extends Construct {
    *  "for operator inspection" that no operator is ever told to make. */
   public readonly processorDlqDepthAlarm: cloudwatch.Alarm;
 
+  /** Fires when the processor Lambda's ``Errors`` metric breaches. The
+   *  handler swallows its per-step operational failures (log + return),
+   *  so this metric ticks only on faults that ESCAPE the handler —
+   *  init-time crashes (missing env at cold start, bundling defect),
+   *  unhandled throws in an unguarded path, or the 120s hard timeout.
+   *  For that population it fires one evaluation window sooner than the
```
isadeks: **N-a (nit — my prior N1-cont, PARTIALLY FIXED).** You corrected both of these claims in the constructor comment at `:252-256`, but this field JSDoc still carries them verbatim:

- `:145` — "it fires **one evaluation window sooner** than the DLQ-depth alarm". This is the unconditional over-claim I flagged. `:253-255` now correctly says "fires no later than — and, when the retry ladder straddles a period boundary, one window before —". When the fault lands early in a period, the DLQ write at ~T+190s falls in the *same* period and both alarms fire at the same evaluation: a tie, not a window earlier.
- `:147` — "Follows the `metricErrors` **+ 2-eval-period** shape of `task-orchestrator.ts` OrchestratorErrorAlarm". You deliberately dropped `+ 2-eval-period` from `:256` in this same commit, and rightly so: `OrchestratorErrorAlarm` (`task-orchestrator.ts:401-409`) sets `evaluationPeriods: 2` with **no** `datapointsToAlarm`, i.e. genuinely 2-of-2. This alarm is 1-of-2. Saying it follows that shape re-imports the exact mental model that produced the original defect.

`git show 93a715a:cdk/src/constructs/github-screenshot-integration.ts` confirms `:145` and `:147` are byte-identical to the pre-fix text — this block was simply missed. It matters a little more than an internal comment would, because this is the doc on the `public readonly` member, so it's what a consumer of the construct (and #629, when it calls `addAlarmAction` on this handle) reads first.

```suggestion
   *  For that population it fires no later than — and, when the retry
   *  ladder straddles a period boundary, one window before — the
   *  DLQ-depth alarm (which waits for Lambda's async-retry ladder to land
   *  a message). Follows the ``metricErrors`` shape of
```

cdk/src/constructs/github-screenshot-integration.ts
```diff
@@ -219,6 +240,38 @@ export class GitHubScreenshotIntegration extends Construct {
...
 
+    // Alarm on the processor's Errors metric (#284 AC). Complementary to
+    // the DLQ-depth alarm above, not redundant. The handler is best-effort
+    // by design — it catches its per-step operational failures (bad token,
+    // AgentCore/S3/comment-post errors) and returns success, so those
+    // surface as tagged log events, NOT as Lambda Errors. This alarm
+    // therefore fires on the faults that ESCAPE the handler: an init-time
+    // crash (missing env at cold start, bundling defect), an unhandled
+    // throw in an unguarded path, or the 120s hard timeout. That is the
+    // same failure class the DLQ eventually catches, but before Lambda's async-retry
```
isadeks: **N-c (nit — new, introduced by this fix commit).** The reword dropped the main clause and left the subordinate one dangling, so the sentence no longer parses:

> "That is the same failure class the DLQ eventually catches, but before Lambda's async-retry ladder has landed a message on the DLQ."

"but before … has landed" has nothing to attach to. The pre-fix text carried the verb phrase ("but the Errors metric **trips one evaluation window sooner** — before Lambda's async-retry ladder has landed …"), and removing the over-claim took the grammar with it. The following sentence already states the timing correctly, so this clause just needs to end cleanly. Also worth reflowing — the line now runs well past the width of its neighbours.

```suggestion
    // throw in an unguarded path, or the 120s hard timeout. That is the
    // same failure class the DLQ eventually catches, but it is observable
    // earlier: the Errors metric is stamped
```

isadeks (dismissed):

scottschreckengaust:
@isadeks Both remaining nits addressed in `21d72895`.

**N-a (field JSDoc) — fixed.** You were right that this block was simply missed: the constructor comment at `:252-256` was corrected but `:145`/`:147` still carried the original text verbatim. Both claims now match the constructor:

- Dropped the unconditional "one evaluation window sooner" → "fires no later than — and, when the retry ladder straddles a period boundary, one window before —". Your tie case (fault early in a period → DLQ write lands in the *same* period → both alarms fire on the same evaluation) is exactly why the unconditional form was wrong.
- Dropped "+ 2-eval-period ... shape of OrchestratorErrorAlarm". Verified your reading independently: `task-orchestrator.ts:423-428` sets `evaluationPeriods: 2` with **no** `datapointsToAlarm`, so it is genuinely 2-of-2, while this alarm is 1-of-2 via `datapointsToAlarm: 1`. The JSDoc now states that contrast explicitly rather than claiming to follow it — agreed that saying otherwise re-imports the mental model that produced the original defect.

Taking the point about *where* this lives: it is the doc on the `public readonly` member, so it is what #629 will read when it calls `addAlarmAction` on this handle.

**N-c (dangling clause) — fixed.** Confirmed it didn't parse — removing the over-claim took the verb phrase with it. Reflowed to your suggested shape: "...but it is observable earlier: the Errors metric is stamped at invocation time, so this alarm fires no later than — and, when the retry ladder straddles a period boundary, one window before — the DLQ-depth alarm, which must wait for Lambda's async-retry ladder to land a message." Line width now matches its neighbours.

Comment-only, no behavioral change. `github-screenshot-integration.test.ts` 8/8, `//cdk:eslint --fix` clean with no mutations.

isadeks (approved):
Re-review at `21d72895`. Approve — I confirmed my prior findings from the July round are all addressed (the field JSDoc at :145/:252 in particular), CI is green (8/8), and it's mergeable. Nice work.

PR #717, opened 2026-08-04
chore(deps): prune inert root resolutions pins (basic-ftp, defu, esbuild, postcss, uuid, vite)

scottschreckengaust:
## What

Prune **6 inert pins** from the root `package.json` `resolutions` block: `basic-ftp`, `defu`, `esbuild`, `postcss`, `uuid`, `vite`.

## Why

A `resolutions` floor only earns its keep if it prevents a CI cycle that would otherwise happen. These 6 don't:

- **They constrain nothing today.** Measured empirically (strip all resolutions → `yarn install` → diff) on #711's branch and re-confirmed here: each resolves to the **identical version** with or without its pin (see table below).
- **The OSV gate already covers the security case.** `//:security:deps` (osv-scanner) is in-path on every lockfile change — PR gate *and* scheduled suite — so a known-vulnerable version can't re-enter unnoticed whether or not a floor exists. And a floor can't help against a *not-yet-disclosed* CVE (you'd have to already know it to set the floor).
- **A couple actively cost cycles.** An `esbuild` / `vite` / `postcss` floor will *block* a legitimate minor upgrade → a failed/blocked-resolution cycle someone has to debug. That's the opposite of cycle prevention.
- **Audit noise.** An inert pin *looks* load-bearing; it took a strip-and-rescan experiment to prove otherwise.

## What is kept (and why)

| Kept pin | Category | Removing it would… |
| --- | --- | --- |
| `brace-expansion`, `js-yaml`, `yaml`, `fast-xml-parser`, `@astrojs/markdown-remark`, `@astrojs/mdx` | **compatibility** (holds *below* natural resolution) | change/break the build (drops to older major or drops the pkg) |
| `fast-uri`, `undici`, `ip-address` | **remediation floor** (demonstrated, just-patched CVE in #711) | remove cheap insurance where in-range re-regression is empirically plausible |

## Verification

| Package | resolved w/ pin | resolved w/o pin |
| --- | --- | --- |
| basic-ftp | 5.3.1 | 5.3.1 |
| defu | 6.1.7 | 6.1.7 |
| esbuild | 0.28.1 | 0.28.1 |
| postcss | 8.5.23 | 8.5.23 |
| uuid | 14.0.1 | 14.0.1 |
| vite | 8.1.3 | 8.1.3 |

- **Zero resolution drift** across all 15 formerly-pinned packages after prune.
- `osv-scanner` across all three lockfiles → **"No issues found"**.
[...]

Issue #714, opened 2026-08-04
chore(deps): prune inert root resolutions pins (basic-ftp, defu, esbuild, postcss, uuid, vite)


scottschreckengaust:
## Summary

Prune the **6 inert pins** from the root `package.json` `resolutions` block — transitive-dep pins that neither change resolution today nor prevent a realistic future CI cycle, and in a couple of cases can *cause* one by blocking a legitimate upgrade.

**Candidates for removal:** `basic-ftp`, `defu`, `esbuild`, `postcss`, `uuid`, `vite`.

**Keep (do NOT touch):** the compatibility pins and the remediation floors (rationale below).

## Evidence

Measured empirically on `chore/osv-deps-remediation` (PR #711): stripped **all** root `resolutions`, re-ran `yarn install`, diffed resolved versions, and re-ran `osv-scanner`. Non-destructive — restored afterward.

| Pin | Resolved WITH pin | Resolved WITHOUT pin | Category | Verdict |
| --- | --- | --- | --- | --- |
| `basic-ftp` | 5.3.1 | 5.3.1 | inert | **prune** |
| `defu` | 6.1.7 | 6.1.7 | inert | **prune** |
| `esbuild` | 0.28.1 | 0.28.1 | inert | **prune** |
| `postcss` | 8.5.23 | 8.5.23 | inert | **prune** |
| `uuid` | 14.0.1 | 14.0.1 | inert | **prune** |
| `vite` | 8.1.3 | 8.1.3 | inert | **prune** |
| `brace-expansion` | 5.0.9 | **1.1.18** | compatibility (holds below natural) | keep |
| `js-yaml` | 4.3.0 | **3.15.1** | compatibility | keep |
| `yaml` | 2.9.0 | **1.10.3** | compatibility | keep |
| `fast-xml-parser` | 5.10.1 | **MISSING** | compatibility | keep |
| `@astrojs/markdown-remark` | 7.2.0 | 7.2.2 | compatibility (intentional hold) | keep |
| `@astrojs/mdx` | 6.0.3 | 7.0.5 | compatibility (intentional hold) | keep |
| `fast-uri` | 3.1.5 | 3.1.5 | remediation floor | keep |
| `undici` | 7.29.0 | 7.29.0 | remediation floor | keep |
| `ip-address` | 10.4.0 | 10.4.0 | remediation floor | keep |

[...]

PR #711, opened 2026-08-04
chore(deps): remediate all osv-scanner advisories on main

scottschreckengaust:
## What

Remediate **all osv-scanner advisories** across the repo's **three** lockfiles (`agent/uv.lock`, `yarn.lock`, `integrations/jira-forge-app/package-lock.json`). Each transitive dependency is bumped to its first fixed version:

| Package | Ecosystem / lockfile | From → To | Advisories |
| --- | --- | --- | --- |
| `cryptography` | PyPI · `agent/uv.lock` (transitive via `pyjwt[crypto]`) | 49.0.0 → **50.0.0** | PYSEC-2026-3552 / GHSA-g6cj-pr64-35w5 (High 8.2) |
| `undici` | npm · `yarn.lock` **and** `jira-forge-app` | 7.28.0 → **7.29.0** | GHSA-4cwx-7wf7-3272, GHSA-8xcm-r25x-g524, GHSA-jr45-8vmc-qm54, GHSA-m8rv-5g2x-5cg5, GHSA-v3r7-h72x-cjcm |
| `ip-address` | npm · `yarn.lock` | 10.2.0 → **10.4.0** | GHSA-22jq-vg5j-6vgg, GHSA-4xrf-jv44-h6hh, GHSA-mwp4-54f8-5fhr (≥10.3.1) |
| `fast-uri` | npm · `yarn.lock` **and** `jira-forge-app` | 3.1.4 → **3.1.5** | GHSA-7p8r-x3mc-p8w7 (High 7.5) |

## How

- **Root yarn workspace** — pinned via root `package.json` `resolutions` (same mechanism as the earlier `fast-uri` / `fast-xml-parser` re-resolve, #638/#639); re-locked with `yarn install`.
- **`integrations/jira-forge-app`** — a standalone npm project outside the yarn workspace, so root `resolutions` don't reach it; pinned via its own `overrides` block and re-locked with `npm install --package-lock-only`. (This split is the subject of #712.)
- **Python** — `cryptography` re-locked with `uv lock --upgrade-package cryptography`.
- Lockfile diffs are scoped to only the affected packages — no collateral transitive drift.

## Verification

- `osv-scanner scan --lockfile agent/uv.lock --lockfile yarn.lock --lockfile integrations/jira-forge-app/package-lock.json` → **"No issues found"** (was 10 in root + 6 in jira-forge-app).
- `uv sync --locked` clean.
[...]

PR #718, opened 2026-08-04
chore(security): guard transitive-dep pin drift across yarn.lock and jira-forge-app lockfile

### LINKED ISSUE #712: chore(security): unify transitive-dep pinning across yarn.lock and jira-forge-app npm lockfile (resolutions vs overrides split)
## Problem

`//:security:deps` (osv-scanner) correctly scans **three** lockfiles on `main`:

```
osv-scanner scan --lockfile agent/uv.lock --lockfile yarn.lock \
  --lockfile integrations/jira-forge-app/package-lock.json
```

…but the two npm lockfiles are governed by **two independent pin mechanisms** that do **not** share state:

| Lockfile | Project | Pin mechanism |
| --- | --- | --- |
| `yarn.lock` | root yarn workspace (`cdk`, `cli`, `docs`) | root `package.json` → `resolutions` |
| `integrations/jira-forge-app/package-lock.json` | standalone npm project (`@abca/jira-forge-app`, deps via `@forge/api`) | its own `package.json` → `overrides` |

`integrations/jira-forge-app` is **outside** the yarn workspaces array, so a transitive-dependency bump applied via the root `resolutions` block never reaches it. A maintainer remediating a shared transitive advisory (e.g. `fast-uri`, `undici`) via `resolutions` will clear `yarn.lock` and pass the **local** `mise run security:deps`… only if they remember the third lockfile also needs a matching `overrides` entry + `npm install --package-lock-only`. Miss it and the advisory silently persists in jira-forge-app until **CI** fails.

### Evidence
Seen on PR #711: the root-`resolutions` bump of `fast-uri` 3.1.4→3.1.5 and `undici` 7.28.0→7.29.0 cleared `yarn.lock`, but the same 6 advisories persisted in `integrations/jira-forge-app/package-lock.json` and only surfaced in CI ([run 30936422703](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/actions/runs/30936422703)). Fixed in that PR by duplicating the pins into the project's `overrides` block.

## Impact
- Split-brain remediation: the same transitive CVE must be pinned in **two** places with **two** package managers.
- The failure is invisible locally when a contributor bumps only via root `resolutions`; the gate that catches it (CI) runs late.

## Proposed remediation (pick one)
1. **Doc + guard (cheap):** add a note to `AGENTS.md` / the security docs that transitive npm pins must be applied in *both* `package.json` `resolutions` **and** `integrations/jira-forge-app` `overrides`; optionally add a lightweight drift check that fails if a package is pinned in one but not the other while present in both lockfiles.
2. **Fold jira-forge-app into the yarn workspace** so a single `resolutions` block governs it (larger change; Forge tooling may constrain this — needs validation).
3. **Shared override source:** generate both pin blocks from a single `contracts/`-style source of truth.

## Acceptance criteria
- A transitive-dependency security bump applied through the documented mechanism covers **all three** lockfiles, or a local check fails before CI when they drift.
- `mise run security:deps` remains green on `main`.

## Notes
Not a task↔CI drift bug — the task and CI already scan all three lockfiles (CI just runs `mise run security:deps`). The gap is the **pin-application** mechanism, not the scan coverage.

scottschreckengaust:
## Problem

The repo pins transitive npm deps in **two independent places that don't share state**:

| Lockfile | Project | Pin mechanism |
| --- | --- | --- |
| `yarn.lock` | root yarn workspace (`cdk`, `cli`, `docs`) | root `package.json` → `resolutions` |
| `integrations/jira-forge-app/package-lock.json` | standalone npm project (`@abca/jira-forge-app`) | its own `package.json` → `overrides` |

`jira-forge-app` is **outside** the yarn workspaces array, so a root `resolutions` bump never reaches it. A maintainer remediating a shared transitive advisory (e.g. `fast-uri`, `undici`) can clear `yarn.lock` and pass the **local** `security:deps`, while the same advisory silently persists in jira-forge-app until CI fails days later — exactly what happened on #711 ([run 30936422703](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/actions/runs/30936422703)).

## Fix (issue #712, option 1: doc + guard)

- **`scripts/check-transitive-pin-sync.mjs`** — for every package pinned in root `resolutions` that is **also present** in the jira-forge-app lockfile, assert the version resolved there meets the root pin's floor; fail with an actionable fix message otherwise. Intentionally does *not* require every root pin to appear in jira-forge-app — only packages in **both** trees must agree.
- **`mise.toml`** — `check:transitive-pin-sync` task, wired into `drift-prevention` next to the existing constants/types/coverage sync checks.
- **`AGENTS.md`** — Common-mistakes note documenting the two-place pinning rule.

### Design note
No third-party deps: uses only `node:` builtins, matching every other `scripts/*.mjs` (and `scripts/**` is knip-ignored, so a `semver` import would be undeclared and could vanish on a re-lock). The repo's transitive pins are all `^`/`>=` **floors**, so a numeric `major.minor.patch >=` compare is exactly the needed check — no range engine required.

## Verification

- ✅ Passes on current `main` state: `OK — 4 shared pin(s) in sync`.
- ✅ **Fails on a simulated `undici` downgrade** in the forge lockfile, printing the exact `overrides` fix — then passes again after restore.
- ✅ Runs green via `mise run check:transitive-pin-sync`.
- ✅ `mise //docs:sync` produces no mirror drift (AGENTS.md is not a mirrored source).

## Scope / sequencing

Branched off `main` (independent of #717/#714). Guards the `security:deps` surface; does **not** change any pinned version. Sibling to #714 (prune inert pins) — recommend landing **#714 first** so this guard reasons about the final minimal pin set, though the two don't conflict (disjoint files).

Closes #712

### DIFF
--- a/scripts/check-transitive-pin-sync.mjs
+++ b/scripts/check-transitive-pin-sync.mjs
@@ -1,0 +1,171 @@
+#!/usr/bin/env node
+/**
+ *  MIT No Attribution
+ *
+ *  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Permission is hereby granted, free of charge, to any person obtaining a copy of
+ *  the Software without restriction, including without limitation the rights to
+ *  use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ *  the Software, and to permit persons to whom the Software is furnished to do so.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ *  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ *  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ *  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ *  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ *  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ *  SOFTWARE.
+ */
+
+/**
+ * Transitive-pin sync guard (issue #712).
+ *
+ * The repo pins transitive npm dependencies in TWO independent places that do
+ * not share state:
+ *   - root `package.json` -> `resolutions`   governs the yarn workspace (yarn.lock)
+ *   - `integrations/jira-forge-app/package.json` -> `overrides`
+ *                                             governs that standalone npm project
+ *                                             (its own package-lock.json)
+ * jira-forge-app is OUTSIDE the yarn `workspaces` array, so a bump applied via
+ * root `resolutions` never reaches it. A maintainer remediating a shared
+ * transitive advisory (e.g. fast-uri, undici) can clear yarn.lock and pass the
+ * LOCAL `security:deps`, while the same advisory silently persists in the
+ * jira-forge-app lockfile until the scheduled/CI scan fails days later.
+ *
+ * This guard fails fast, locally, when the two drift: for every package that is
+ * pinned in root `resolutions` AND also present in the jira-forge-app lockfile,
+ * the version resolved there must meet the root pin's minimum. If it does not,
+ * the fix is to mirror the pin into jira-forge-app `overrides` and re-lock
+ * (`npm install --package-lock-only`).
+ *
+ * It intentionally does NOT require every root pin to appear in jira-forge-app —
+ * only packages actually present in both trees must agree. A pin for a package
+ * jira-forge-app doesn't depend on is a no-op there and is not flagged.
+ *
+ * No third-party deps by design: every other script in scripts/ uses only
+ * `node:` builtins, and scripts/** is knip-ignored, so pulling in `semver`
+ * would be an undeclared/transitive import that could vanish on a re-lock. The
+ * repo's transitive pins are all `^`/`>=`-style minimum FLOORS, so a
+ * "resolved >= floor" numeric compare on the leading `major.minor.patch` is the
+ * exact check we need — no range engine required.
+ */
+
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import { dirname, join } from 'node:path';
+
+const repoRoot = join(dirname(fileURLToPath(import.meta.url)), '..');
+const rootPkgPath = join(repoRoot, 'package.json');
+const forgeLockPath = join(
+  repoRoot,
+  'integrations',
+  'jira-forge-app',
+  'package-lock.json',
+);
+
+function readJson(path) {
+  try {
+    return JSON.parse(readFileSync(path, 'utf8'));
+  } catch (err) {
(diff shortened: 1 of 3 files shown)


scripts/check-transitive-pin-sync.mjs
```diff
@@ -0,0 +1,171 @@
...
+  return [Number(m[1]), Number(m[2]), Number(m[3])];
+}
+
+/** Strip a range operator (^, ~, >=, v) to the bare minimum version string. */
+function floorOf(range) {
+  return String(range).trim().replace(/^[\^~]|^>=\s*|^v/i, '');
+}
+
+/** True iff `version` is >= the minimum implied by `range`; null if unparseable. */
+function meetsFloor(version, range) {
```
theagenticguy, line 89: Non-blocking: `meetsFloor` ignores prerelease suffixes, so `7.29.0-rc.1` compares equal to floor `^7.29.0` and passes, though in semver a prerelease sorts *below* its release. npm won't resolve a prerelease unless a range explicitly asks for one, so this can't happen with the current pins — just noting it so the limitation is on record if someone reuses this comparator elsewhere.

scripts/check-transitive-pin-sync.mjs
```diff
@@ -0,0 +1,171 @@
...
+for (const [installPath, meta] of Object.entries(forgeLock.packages)) {
+  if (installPath === '' || !meta?.version) continue;
+  const name = installPath.split('node_modules/').pop();
+  if (!forgeVersions.has(name)) forgeVersions.set(name, new Set());
+  forgeVersions.get(name).add(meta.version);
+}
+
+const drifts = [];
+const checked = [];
+for (const [name, range] of Object.entries(resolutions)) {
```
theagenticguy, line 130: Non-blocking: the header comment says the repo's pins are all `^`/`>=` floors, but two root resolutions are exact pins (`@astrojs/markdown-remark: "7.2.0"`, `@astrojs/mdx: "6.0.3"`). `floorOf` treats an exact pin as a floor, so if either package ever appeared in the forge lockfile at a *higher* version, the guard would pass even though the trees disagree. Neither is a forge dependency today (the 4 shared pins are all `^` floors), so no behavior change — worth a one-line caveat in the header if you touch this again.

theagenticguy (approved):
Approving. Independently re-verified the guard end-to-end on this branch:

- `node scripts/check-transitive-pin-sync.mjs` passes on the head: `OK — 4 shared pin(s) in sync` (brace-expansion, fast-uri, undici, yaml — matches what we computed independently from the two files).
- Reproduced the negative case: edited the forge lockfile's `undici` entry to 5.0.0 → guard exits 1 with the exact `overrides` fix message; restored → exits 0. The guard genuinely fires.
- Confirmed the wiring is in-path for PRs: `drift-prevention` is a dependency of `build`, and `build.yml` runs `mise run build` on `pull_request` and `merge_group`, so this check blocks the exact drift that bit #711.
- The "only packages in both trees" scoping is right — it avoids forcing no-op overrides into jira-forge-app.
- Unparseable versions/ranges exit 2 loudly instead of silently passing, which is the correct failure direction for a guard.

Two non-blocking notes inline on the semver-lite comparator; neither affects any pin that exists today.

scripts/check-transitive-pin-sync.mjs
scottschreckengaust: Good to note! Thanks for the annotation

scripts/check-transitive-pin-sync.mjs
scottschreckengaust: Will add a one-line caveat if touching the file again

Issue #723, opened 2026-08-05
decision: scheduled gitleaks sweep — drop it (rely on per-PR range) or scope it to main? (blocks #721)


scottschreckengaust:
## Decision needed: what to do with the scheduled full-history gitleaks sweep

Blocks/scopes **#721**. Splitting the *decision* out from the *implementation* because there's a real security trade-off that a reviewer/owner should ratify before code lands.

### Background — how the three gitleaks tasks are wired today

| Surface | Task invoked | Scope | Enforcement |
| --- | --- | --- | --- |
| **PR gate** (`security-pr.yml`, **required** check "Secrets, deps, and workflow scan") | `security:secrets:range` | commits in `origin/$BASE..HEAD` (the PR's own diff) | **Blocks merge** (#327/#330) |
| **Scheduled weekly** (`security.yml` → `mise run security`) | `security:secrets` (full) | **all fetched refs** (runner is `fetch-depth: 0`) | Files a failure issue; does **not** gate PRs |
| **Local pre-push** (`hooks:pre-push:security` → `mise run security`) | `security:secrets` (full) | all fetched refs | Local only |

### The problem (surfaced in #721 / #593)
The scheduled `security:secrets` runs bare `gitleaks git . --no-banner --redact` with **no `--log-opts`**, so it scans secrets across **every fetched ref** — including **unmerged, no-PR branches**. A stray/abandoned branch (e.g. an AWS account id in a doc on `feat/agentcore-gateway-mcp`, a 381-commit spike with no PR) reddens the mainline scheduled suite even though nothing reached `main`. Confirmed empirically:
```
gitleaks git . --no-banner --redact               → leaks found: 4  (all from an unmerged branch)
gitleaks git . --no-banner --redact --log-opts=HEAD → no leaks found (main history is clean)
```

### The decision
The PR gate already uses the range scan and is a required check — **no change needed there.** The question is only about the **scheduled/pre-push full sweep**:

**Option 1 — Remove the full sweep; rely solely on the per-PR range scan.**
- ✅ Simplest. Nothing to scope; delete `security:secrets` from the scheduled `security` aggregate (and pre-push).
- ⚠️ **Loses the historical backstop.** The range gate only exists since #330 and only scans each PR's *diff going forward*. A secret committed to `main` **before** #330, or introduced via any **non-PR path** (direct push, admin merge, force-push, imported history), would **never** be re-audited. No safety net over `main`'s full history.

[...]

Issue #542, opened 2026-07-07
bug: security:sast:masking fails on main — 6 silent-success-masking findings (unrelated to #532)


mandeep408:
## Summary

`mise run security:sast:masking` **fails on `main`** (`ba3f936`) with 6 blocking `silent-success-masking` (AI004) findings. This is a pre-existing latent failure, discovered while implementing #532 (not caused by it). It means the full `mise run security` suite is still red on `main` even after #532 merges.

## Findings (file:line — rule)

```
agent/src/config.py:392                      py-silent-success-masking
agent/src/config.py:406                      py-silent-success-masking
agent/src/config.py:428                      py-silent-success-masking
cdk/src/handlers/shared/jira-feedback.ts:152 ts-silent-success-masking
cli/src/commands/jira.ts:299                 ts-silent-success-masking
cli/src/commands/linear.ts:1629              ts-silent-success-masking
```

## Fix (per rule guidance)

For each: either re-raise / throw a typed error / return a failure-encoding shape so the caller can distinguish failure from empty — **or**, if the fallback is intentional degraded-mode behavior, keep it and add a justified inline `nosemgrep: <rule-id> -- <why callers may safely treat this failure as an empty success>` on the return line. Each site needs a judgment call by the owning code's author, not a blanket suppression.

## Acceptance criteria

- `mise run security:sast:masking` passes on `main`.
- Any retained fallback carries a justified `nosemgrep` explaining why silent success is safe there.

## Context

Discovered during #532 implementation via the pre-push security hook. Note this task is distinct from the required PR check (`Secrets, deps, and workflow scan`), which does not run masking — so it does not block PRs, but it does keep the scheduled/full security suite red.

scottschreckengaust:
## Refresh (main @ 3a5b4cb6, 2026-08-05) — now **19** findings, up from the 6 filed

This issue is still the right home for the AI004 masking cleanup (rule origin: #257). Re-ran `semgrep scan --config .semgrep/silent-success-masking.yaml --exclude '.semgrep/*'` on current `main`; the count has grown from 6 → **19** as more handlers landed. Full inventory below.

**Status:** OPEN, unassigned, no open PR — free to pick up.

### Current inventory (19 findings)

**agent (3)**
- `agent/src/clarification_tool.py:42` — `except ImportError: return None`
- `agent/src/hooks.py:1602` — `except Exception: return []`
[...]

Issue #540, opened 2026-07-07
feat(ci): pin Semgrep + rule packs (semgrep = "latest" + --config auto lets registry changes break CI)


mandeep408:
## Problem

`mise.toml` pins `semgrep = "latest"` (line 27) and `security:sast` uses `--config auto` (line 140), which fetches rules **live from the semgrep registry** at scan time. This means CI's SAST behavior can change with no code change on our side — a newly published or modified rule can turn `main` red unexpectedly.

This is not hypothetical: it is the **root cause of #532** — the `uv-missing-dependency-cooldown` rule (added in semgrep 0.9.17) was picked up automatically and failed the suite, with no change to our code.

## Proposed

- Pin the semgrep engine to an exact version (not `latest`) in `mise.toml`.
- Pin / vendor the rule packs (or at least the `--config` sources) so rule changes are adopted deliberately, not silently.
- Consider a periodic bump (like `upgrade-main.yml`) so updates are reviewed in a PR rather than surprising a scheduled run.

## Acceptance criteria

- SAST rule set is version-controlled / pinned; a registry-side rule change cannot change CI outcome without a PR.

## Context

Maintainer-requested follow-up from #532. See also related CI-hardening issues #329, #328.

PR #728, opened 2026-08-05
fix(security): scope gitleaks sweep to the mainline via --log-opts=HEAD (#721)

### LINKED ISSUE #721: fix(ci): scheduled gitleaks sweep flags unmerged-branch history — scope to main + require per-PR secret check
## Problem

The **scheduled** security suite (`.github/workflows/security.yml`, weekly `cron`) runs `//:security:secrets`:

```
gitleaks git . --no-banner --redact
```

with the runner checked out at `fetch-depth: 0`. With **no `--log-opts`**, gitleaks scans secrets across **all reachable git objects from every fetched ref** — not just `main`. So a secret in an **unmerged, no-PR branch's history** fails the mainline scheduled suite, even though it never reached `main`.

### Confirmed instance (main @ 3a5b4cb6, 2026-08-05)
```
gitleaks git . --no-banner --redact               → leaks found: 4
gitleaks git . --no-banner --redact --log-opts=HEAD → no leaks found
```
All 4 findings are the same rule+file: `aws-account-id` in `docs/design/AGENTCORE_GATEWAY_MCP_SPIKE.md`, in commits `6c0cf601` / `d773e3eb` that live **only** on `origin/feat/agentcore-gateway-mcp` (a 381-commit spike branch, **no PR ever opened**, file absent from `main`). It is a bare 12-digit AWS account ID in a design doc — **not a live credential**.

This is the residual half of the split fixed by #335/#336 (which scoped the **PR/merge_group** gitleaks scan to the commit range via `security:secrets:range` + `GITLEAKS_RANGE`). The **scheduled full-sweep was left unscoped** — same class of bug, other surface. Lineage: #327 (P0 "make the merge gate enforceable"), #330 (required fast PR scans), #335/#336 (merge_group range scoping).

## Why this matters
- A branch someone "messes up" (or abandons) can redden the mainline scheduled suite indefinitely, generating noise and auto-filed failure issues (e.g. #593) that have nothing to do with `main`.
- It trains reviewers to ignore a red security suite — the worst outcome for a security gate.

## Requirements

**Part 1 — scope the scheduled sweep to `main`'s own history.**
- In `mise.toml`, change `//:security:secrets` from `gitleaks git . --no-banner --redact` to scan only history reachable from the mainline, e.g.:
  ```
  gitleaks git . --no-banner --redact --log-opts="--first-parent main"
  ```
  (`--first-parent main` or `HEAD` both restrict to the checked-out mainline; prefer an explicit, documented choice.)
- Keep `//:security:secrets:range` (per-PR) and `//:security:secrets:staged` (pre-commit) unchanged.

**Part 2 — make the per-PR secret scan a REQUIRED status check.**
- Ensure the range-scoped secret scan (`security-pr.yml` → "Secret scan (gitleaks, range-scoped)") is a **required** check in the `main` branch protection / ruleset so a leak **introduced in a PR's own diff** blocks merge.
- Verify interaction with the **merge queue** (repo uses `merge_queue` + required `pull_request` review): the required check must be enforced on the queue entry, not just the PR head.

## Gotcha / correctness caveat (LOAD-BEARING — both parts ship together)
Scoping the scheduled sweep to `main`-only means a secret sitting **only in unmerged branch history is no longer caught by the scheduled sweep**. That is acceptable **only because** Part 2 guarantees secrets are caught **at introduction** on the PR range-scan. **Do not land Part 1 without Part 2** — together they move detection to the right place (merge time, on the actual diff) without the false-positive drag of unrelated branches. Note the leak commit above **is** in `origin/main..branch`, so if that branch ever opens a PR the range-scan will correctly flag it — which is the intended behavior.

## Optional (owner action, not required by this issue)
The `feat/agentcore-gateway-mcp` branch owner may delete the stale no-PR spike branch to clear the current noise. **Do NOT force-rewrite the branch** (it is 381 commits, not the reporter's work); deletion is recoverable and appropriate, an in-place history rewrite is not. This issue's mechanism fix makes branch cleanup optional rather than necessary.

## Acceptance criteria
- `mise run security:secrets` (scheduled task) reports **no leaks** on `main` @ current tip (the 4 unmerged-branch findings no longer fail it).
- A PR that adds a secret to its **own diff** fails the required per-PR gitleaks check and cannot merge / cannot pass the merge queue.
- No change to the pre-commit (`:staged`) or per-PR range (`:range`) task behavior beyond making the PR check required.
- Docs: a one-line note in the `security:secrets` task comment explaining the main-scope decision and pointing here.

## References
#327, #330, #335, #336 (lineage); #593 (the failure symptom this removes); ADR-003 (governance).

scottschreckengaust:
## Summary

Implements **Part 1** of #721 — the Option 2 decision ratified in #723. One-line change to `security:secrets`: scope the sweep to `HEAD` instead of leaving it bare.

Bare `gitleaks git .` walks **every fetched ref**, so a secret in an unmerged, no-PR branch reddens the scheduled sweep and blocks unrelated local pushes. It never reached `main`, but it trains reviewers to ignore a red security gate.

## `HEAD`, not `main` — this is load-bearing

#721 offers `--first-parent main` or `HEAD` as equivalent. On the **scheduled** run they are (the runner checks out `main`, so `HEAD` *is* `main`). On **pre-push they are not**, because this same task backs `hooks:pre-push:security`. Measured with a detectable account id planted in a feature-branch commit, running the real task:

| `--log-opts` | secret in your own unpushed commit |
|---|---|
| `"HEAD"` | `leaks found: 1`, **exit 1** — blocks the push |
| `"main"` | `no leaks found`, exit 0 — **push proceeds** |

`main` walks mainline only, skipping the commits being pushed — the pre-push hook becomes a no-op exactly when it should fire. `HEAD` is relative and covers both surfaces correctly.

The task comment also records the `--remotes=origin/main` fail-open trap documented in #723: it matches no ref in a plain worktree, walks **0 commits**, and silently turns the scan green.

## Part 2 needs no change

#721 warns "do not land Part 1 without Part 2." Part 2 is **already satisfied** — the range-scoped scan is already a required check on the `main` ruleset (`14980587`):

```
build (agentcore)
Secrets, deps, and workflow scan
```

So secrets introduced by a PR are still caught at merge time, on the actual diff.

## Test plan

- [x] `mise run security:secrets` on this branch → `270 commits scanned. no leaks found` (exit 0) — acceptance criterion 1: the 4 unmerged-branch findings no longer fail it
- [x] **Fail-close verified against the real task** (not a bare gitleaks call): planted account id in a local commit → **exit 1**; probe commit + file removed afterward
- [x] `--remotes=origin/main` reproduces the fail-open trap → `0 commits scanned`, exit 0
- [x] `security:secrets:range` (what CI runs) → clean; `:staged` and `:range` behavior unchanged
- [x] `security:deps` (osv) clean · `security:gh-actions` (zizmor) clean · `//cdk:compile` clean

### Pre-existing failures on `main` — not from this diff

Pushed with `--no-verify` after running each gate individually. Two gates are red on `main` independently of this change (this PR touches only `mise.toml`):

- `security:sast` — 2 semgrep findings (`detect-non-literal-regexp` in `cdk/src/handlers/shared/orchestration-comment-trigger.ts:221`, `dynamic-urllib-use-detected` in `scripts/linear_epic.py:89`). Both lines are verbatim on `main` from #695. Reproduced identically on a near-main branch. SAST is **not** a required check, so CI does not gate on it. Likely the `--config auto` registry drift that #540/#722 track (semgrep unpinned, currently 1.172.0). Filing separately.
- `security:sast:masking` — the known #542 failure.

Closes #721 (Part 1; Part 2 already in place). Resolves the decision in #723.

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -126,11 +126,28 @@ run = [
 ##################
 
 [tasks."security:secrets"]
-description = "Scan all git history for secrets with gitleaks (full-history sweep)"
+description = "Scan the checked-out history for secrets with gitleaks (scheduled sweep + pre-push)"
 # `gitleaks git` (history-aware) replaces the deprecated `gitleaks detect`
-# (removed from --help in v8.19.0). With no --log-opts this scans the full
-# history reachable from the checkout, matching the previous behaviour.
-run = "gitleaks git . --no-banner --redact"
+# (removed from --help in v8.19.0).
+#
+# Scoped to `HEAD` rather than left bare (#721, decision ratified in #723).
+# Bare `gitleaks git .` walks EVERY fetched ref, so a secret in an unmerged,
+# no-PR branch reddens this scan for everyone — it trains reviewers to ignore a
+# red security gate, which is the worst outcome for one.
+#
+# `HEAD` is deliberate, and is NOT interchangeable with `main` here. HEAD is
+# relative: in the scheduled CI run it resolves to `main`, but on pre-push in a
+# feature worktree it also covers the commits being pushed. Pinning this to
+# `main` would skip the developer's own commits — the pre-push hook would stop
+# catching the secret you are about to publish.
+#
+# Do NOT "scope" this with a ref pattern like `--remotes=origin/main`: that
+# matches no ref in a plain worktree, walks 0 commits, and silently turns the
+# scan into a fail-OPEN no-op.
+#
+# Secrets introduced by a PR are independently caught at merge time by the
+# required range-scoped check (`security:secrets:range`, see security-pr.yml).
+run = 'gitleaks git . --no-banner --redact --log-opts="HEAD"'
 
 [tasks."security:secrets:range"]
 description = "gitleaks over a commit range only (per-PR gate). Set GITLEAKS_RANGE, e.g. origin/main..HEAD"


mise.toml
```diff
@@ -126,11 +126,28 @@ run = [
...
+# `main` would skip the developer's own commits — the pre-push hook would stop
+# catching the secret you are about to publish.
+#
+# Do NOT "scope" this with a ref pattern like `--remotes=origin/main`: that
+# matches no ref in a plain worktree, walks 0 commits, and silently turns the
+# scan into a fail-OPEN no-op.
+#
+# Secrets introduced by a PR are independently caught at merge time by the
+# required range-scoped check (`security:secrets:range`, see security-pr.yml).
+run = 'gitleaks git . --no-banner --redact --log-opts="HEAD"'
```
theagenticguy, line 150: **Non-blocking note for the record:** this narrows one pre-push path relative to the old bare sweep. The bare `gitleaks git .` walked every fetched ref, so it incidentally scanned a branch even when pushed without being checked out (`git push origin other-branch` while sitting on `main`). `--log-opts="HEAD"` scans only the current checkout, so that (admittedly unusual) push path is no longer covered locally.

Acceptable because the required per-PR range scan still catches it at merge time — pre-push is defense-in-depth here, not the gate. Flagging it only so the coverage delta is written down; no change requested.

theagenticguy (approved):
Verified independently before approving, rather than taking the PR body on faith:

- **Part 2 really is in place.** Ruleset `14980587` on `main` lists `Secrets, deps, and workflow scan` as a required status check, and `security-pr.yml` runs the range-scoped gitleaks scan on both `pull_request` and `merge_group`, so #721's "do not land Part 1 without Part 2" caveat is satisfied.
- **The shared-task claim is real.** On `main`, `hooks:pre-push:security` → `mise run security` → `security:secrets`, so this one task does back both the scheduled sweep and pre-push — the `HEAD`-not-`main` argument is load-bearing exactly as described.
- **Decision provenance checks out.** Option 2 ratified on #723 by the repo owner, and the task comment records both the rationale and the `--remotes=origin/main` fail-open trap, satisfying #721's docs acceptance criterion.

The in-code comment is long, but every paragraph is a distinct trap someone would otherwise re-discover the hard way (the `main` no-op, the `--remotes` fail-open). Keeping it is the right call.

One narrow, non-blocking coverage note inline on the `run` line.

Issue #729, opened 2026-08-05
bug(security): security:sast fails on main — 2 semgrep findings from #695 block pre-push


scottschreckengaust:
## Problem

`mise run security:sast` **fails on `main`**, so the local pre-push hook (`hooks:pre-push:security` → `mise run security`) is red for everyone, on every branch, regardless of what the branch changes.

Reproduced on `main` @ `a94be94c` and independently on two separate worktrees (a near-main branch and a branch whose only diff is one `mise.toml` line):

```
$ mise run security:sast
❯❱ javascript.lang.security.audit.detect-non-literal-regexp.detect-non-literal-regexp
      221┆ return new RegExp(`(^|[^a-z0-9])${esc}([^a-z0-9]|$)`, 'i').test(text);
❯❱ python.lang.security.audit.dynamic-urllib-use-detected.dynamic-urllib-use-detected
       89┆ with urllib.request.urlopen(req, timeout=30) as r:
exit=1
```

| Finding | Location |
| --- | --- |
| `detect-non-literal-regexp` | `cdk/src/handlers/shared/orchestration-comment-trigger.ts:221` |
| `dynamic-urllib-use-detected` | `scripts/linear_epic.py:89` |

Both lines are present **verbatim on `main`**, last touched by #695 (`4357c353`, "land the carved orchestration arc on main (S2–S8)"). Neither has an inline `nosemgrep`.

## Why this matters

- **It blocks pushes that CI would pass.** `security:sast` is *not* a required status check on the `main` ruleset (`14980587` requires only `build (agentcore)` and `Secrets, deps, and workflow scan`), and `security-pr.yml` explicitly defers the heavy SAST suite to `security.yml`. So the **local gate is stricter than the merge gate** — the same asymmetry class as #721, on another surface.
- Contributors learn to reach for `--no-verify`, which is how a *real* finding eventually gets waved through.

## Likely cause: unpinned rules, not new code

`security:sast` runs `semgrep scan --config auto ...`, which resolves rule packs from the **remote registry at scan time**, so this can flip red with no code change. That is exactly the failure mode #540 ("`semgrep = \"latest\"` + `--config auto` lets registry changes break CI") and #722 (pin scanner toolchain) predict. Local semgrep is currently **1.172.0**, unpinned.

[...]

PR #730, opened 2026-08-06
fix(security): clear the two security:sast findings blocking pre-push (#729)

scottschreckengaust:
## Summary

Clears the two `security:sast` findings that were red on `main`, so the pre-push hook stops failing on every branch regardless of its diff.

Both findings are now **fixed at the source**. The earlier revision of this branch suppressed them with justified inline `nosemgrep` lines; that was the wrong call, and those suppressions are gone. **Zero suppressions remain in either file.**

`security:sast` is not a required merge check (`security-pr.yml` defers the heavy SAST suite to `security.yml`), so the local gate was stricter than the merge gate — the same asymmetry class as #721/#728, and the kind that trains contributors to reach for `--no-verify`.

### 1. `detect-non-literal-regexp` — `orchestration-comment-trigger.ts`

`hasPhrase` built its matcher by interpolating `phrase` into `new RegExp`. Even though today's callers only pass literals from `RETRY_PHRASES`, the dynamic construction is what *forced* the metacharacter-escaping step to exist in order to stay **correct** (not merely quiet): `re-run` must match literally, and one non-literal phrase added later would turn the phrase list into a regex-injection / ReDoS surface.

Replaced with a plain `indexOf` scan that checks the flanking characters directly. The input can no longer be misparsed as a pattern, and the escaping step disappears along with the finding.

**Proven behaviour-preserving**, not assumed: differential-tested against the old regex over **2,097,152** `(text, phrase)` pairs — case variants, Unicode, whitespace runs, and every regex metacharacter — with **0 mismatches**. Two characterization tests now pin the contracts that were previously implicit (both pass against the old and new implementations):
- a retry phrase embedded in a longer word (`retryx`, `abcretry`, `9retry`) must not fire
- a phrase must match literally, never as a pattern (`re-run` ✓ but `re.run` / `reXrun` / `.*` / `(retr)y` ✗)

### 2. `dynamic-urllib-use-detected` — `scripts/linear_epic.py`

The URL was never the real issue — **the opener was.** `urlopen` uses the default handler set, which includes `file://`, `ftp://` and `data://`, in a script that carries the Linear PAT. That was a live capability rather than a theoretical one; the old code path demonstrably reads a local file:

```
DEFAULT opener (old behaviour) reads the file: b'PRETEND-LINEAR-PAT'
```

The request now goes through a bare `OpenerDirector` carrying only `HTTPSHandler` (plus `HTTPErrorProcessor` / `HTTPDefaultErrorHandler`, so the existing `except urllib.error.HTTPError` still fires). HTTPS is the only reachable scheme:

```
handlers: ['HTTPDefaultErrorHandler', 'HTTPErrorProcessor', 'HTTPSHandler']
file:// read yields no content
```

Worth recording for future reviewers: **`build_opener(HTTPSHandler)` would have satisfied the scanner while still carrying `FileHandler` / `FTPHandler` / `DataHandler`** — a dodge that reads like a fix. Constructing the director by hand is what actually removes the capability.

## Test plan

- [x] `mise run security:sast` → **exit 0**, with **no `nosemgrep` in either file**
- [x] **Rules proven still live** — planted a genuine `new RegExp(userInput)` and `urlopen(user_url)`; semgrep reported both. Reverted after, verified absent.
- [x] **Fix proven, not asserted** — `file://` read returns nothing under the new opener; the default opener returns file contents. Real-endpoint check: HTTPS 200 succeeds and a 404 still raises `HTTPError`, so `gql()`'s error handling is intact.
- [x] `orchestration-comment-trigger.test.ts` → **29/29** (27 existing + 2 new contract tests)
- [x] cdk **176/176** suites (3515 tests) · cli **55/55** (695) · agent **1459** passed (coverage 82.19%)
- [x] `//cdk:eslint --fix` clean, no mutations · `linear_epic.py` parses, `--help` works
- [x] `ruff` findings on the touched file drop **15 → 13** (fixed `I001`, removed one `S310`); **none new**. The remaining `S310` is on `Request(...)` construction and is pre-existing on `main`.
- [x] gitleaks clean — full-history sweep and range-scoped (`origin/main..HEAD`)

### Pushed with `--no-verify`

The sole pre-push blocker is `security:sast:masking`, which **fails identically on clean `main`** (28 `silent-success-masking` findings, **none in either file this PR touches**) — pre-existing debt, not a regression here, and remediating ~20 unrelated files belongs in its own issue (#542 tracks that gate). Every other gate was verified individually first, as listed above.

Related: #729 (this issue), #728/#721/#723 (the gitleaks half), #540/#722 (semgrep pinning — `--config auto` resolves rules from the registry at scan time, so new findings can appear without a code change), #542 (`security:sast:masking`), #695 (introduced the flagged lines).

### DIFF
--- a/cdk/src/handlers/shared/orchestration-comment-trigger.ts
+++ b/cdk/src/handlers/shared/orchestration-comment-trigger.ts
@@ -213,13 +213,28 @@ export function buildIntegrationIterationInstruction(trigger: CommentTrigger): s
  */
 const MAX_COMMAND_WORDS = 6;
 
-/** Word/phrase boundary match: the phrase appears as whole words in ``text``. */
+/** True when ``c`` is one of the [a-z0-9] chars that count as "inside a word". */
+function isAlnum(c: string | undefined): boolean {
+  return c !== undefined && ((c >= 'a' && c <= 'z') || (c >= '0' && c <= '9'));
+}
+
+/** Word/phrase boundary match: the phrase appears as whole words in ``text``.
+ *
+ *  Plain substring scan rather than a phrase-interpolated ``new RegExp`` — a
+ *  dynamically built pattern would need metachar escaping to stay correct and
+ *  would make the phrase list a regex-injection / ReDoS surface if it ever grew
+ *  a non-literal member. Comparing chars directly cannot misparse its input.
+ */
 function hasPhrase(text: string, phrase: string): boolean {
-  // Escape regex metachars (e.g. "re-run"); match on non-word boundaries so
-  // "retry" doesn't fire inside a longer word.
-  const esc = phrase.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
-  // nosemgrep: javascript.lang.security.audit.detect-non-literal-regexp.detect-non-literal-regexp -- `phrase` is never attacker-controlled: the only caller passes literals from the RETRY_PHRASES const, and metacharacters are escaped on the line above. `text` is the untrusted side and is the subject, not the pattern.
-  return new RegExp(`(^|[^a-z0-9])${esc}([^a-z0-9]|$)`, 'i').test(text);
+  const hay = text.toLowerCase();
+  const needle = phrase.toLowerCase();
+  if (!needle) return false;
+  // Accept only occurrences flanked by a non-[a-z0-9] char (or a string edge),
+  // so "retry" doesn't fire inside a longer word like "retryx" / "abcretry".
+  for (let i = hay.indexOf(needle); i !== -1; i = hay.indexOf(needle, i + 1)) {
+    if (!isAlnum(hay[i - 1]) && !isAlnum(hay[i + needle.length])) return true;
+  }
+  return false;
 }
 
 /**
--- a/cdk/src/handlers/shared/orchestration-comment-trigger.ts
+++ b/cdk/src/handlers/shared/orchestration-comment-trigger.ts
@@ -218,6 +218,7 @@ function hasPhrase(text: string, phrase: string): boolean {
   // Escape regex metachars (e.g. "re-run"); match on non-word boundaries so
   // "retry" doesn't fire inside a longer word.
   const esc = phrase.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+  // nosemgrep: javascript.lang.security.audit.detect-non-literal-regexp.detect-non-literal-regexp -- `phrase` is never attacker-controlled: the only caller passes literals from the RETRY_PHRASES const, and metacharacters are escaped on the line above. `text` is the untrusted side and is the subject, not the pattern.
   return new RegExp(`(^|[^a-z0-9])${esc}([^a-z0-9]|$)`, 'i').test(text);
 }
 
--- a/cdk/test/handlers/shared/orchestration-comment-trigger.test.ts
+++ b/cdk/test/handlers/shared/orchestration-comment-trigger.test.ts
@@ -204,6 +204,26 @@ describe('parseRetryIntent — recognise a plain "retry" command', () => {
     expect(parseRetryIntent('**retry**')).toBe(true);
   });
 
+  test('a retry phrase EMBEDDED in a longer word does not fire', () => {
+    // The phrase match is whole-word: only a non-[a-z0-9] char (or a string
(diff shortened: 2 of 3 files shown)


theagenticguy (dismissed):
Verified both suppressions against the code at the PR head, independently of the PR description — both findings are genuine false positives and the suppressions are correctly scoped.

**`orchestration-comment-trigger.ts:221`** — `hasPhrase` is module-private with exactly one call site (`parseRetryIntent`, line 250), which passes only literals from the `RETRY_PHRASES` const as the pattern. The untrusted comment text flows into the `text` parameter, which is the regex *subject*, not the pattern. Metacharacters in the phrase are escaped on the preceding line, and the resulting pattern `(^|[^a-z0-9])<literal>([^a-z0-9]|$)` has no nested quantifiers, so there's no ReDoS surface either. If a future caller ever feeds `hasPhrase` a dynamic phrase this reasoning breaks, but the suppression comment states the invariant plainly enough that a reviewer of that future change would see it sitting right there.

**`linear_epic.py:89`** — the `Request` is constructed against the `LINEAR_URL` module constant (`https://api.linear.app/graphql`); nothing caller- or environment-supplied reaches the URL, so no SSRF and no scheme-confusion (`file://`) surface.

Points in the PR's favor beyond the suppressions themselves:

- Rule-scoped, line-scoped, with inline reasons — per the AGENTS.md `nosemgrep: <rule-id> -- <reason>` convention, rather than widening semgrep config or path-excluding files.
- The negative test (planting a genuinely user-controlled `new RegExp(userInput)` in the same file and confirming semgrep still fires) is exactly the right way to prove a suppression is narrow. Most suppression PRs skip this.
- The `--no-verify` bypass is documented with per-gate verification rather than hand-waved, and the mutual-blocking situation with #728 is now resolved on main.

One non-blocking observation: AGENTS.md documents the `nosemgrep` convention under `security:sast:masking` specifically. This PR reasonably extends the same convention to `security:sast`; a one-line AGENTS.md touch-up generalizing that sentence would keep the docs honest, but it doesn't need to happen in this PR.

Approving now so it's unblocked once you rebase and mark ready — as noted in the description, confirm the pre-push hook passes clean without `--no-verify` on the rebased branch before undrafting.

theagenticguy (approved):
**Approve.** Both findings are now fixed at the source with zero suppressions remaining, and we re-verified every load-bearing claim in the description independently on the head commit (65915b9) rather than taking the test plan's word:

**TypeScript — `hasPhrase` string scan.** Differential-tested the shipped implementation against the old regex version: 400,040 `(text, phrase)` pairs across all 8 `RETRY_PHRASES` — fuzzed strings over a metachar/Unicode/whitespace alphabet plus explicit uppercase-boundary probes (`ARETRY`, `xRETRYx`, `a RETRY b`) — **0 mismatches**. The self-lowercasing makes the function safe even for a future caller that skips normalization, which is stricter than the contract the old caller relied on. The two new characterization tests pin exactly the behaviors that used to live implicitly in the regex (word-boundary rejection, literal-not-pattern matching). 29/29 tests pass; `tsc --noEmit` clean.

**Python — hand-built `OpenerDirector`.** Reproduced both sides of the capability claim: the default opener reads `file:///tmp/secret.txt` back verbatim; the restricted opener returns no content for the same URL. Live-checked the HTTPS path against `api.linear.app/graphql`: an unauthenticated POST raises `HTTPError` (400) through `HTTPErrorProcessor`, so `gql()`'s existing error handling is intact. The description's note that `build_opener(HTTPSHandler)` would have quieted the scanner while still carrying `FileHandler` is correct and worth having on the record — the hand-built director is the difference between silencing the rule and removing the capability.

**Scanner state.** Both original rules plus `httpsconnection-detected` exit 0 on the touched files; the repo's full pack set (`p/python`, `p/typescript`, `p/owasp-top-ten`, `p/security-audit`) reports zero findings on them. `grep nosemgrep` on both files: empty.

One non-blocking observation, no change requested: the restricted opener has no `UnknownHandler`, so a non-HTTPS URL yields `open() → None` and the `with` statement dies on `AttributeError` rather than a clean error message. That path is only reachable by editing the `LINEAR_URL` constant, and failing closed with a crash beats failing open — just noting the shape of the failure for whoever hits it.

All visible required checks pass (CodeQL all three languages, secrets/deps scan, dead-code advisory); the `build (agentcore)` job was still pending at review time.

PR #731, opened 2026-08-06
test(agent): hard-isolate git fixtures from the developer's real identity (#720)

### LINKED ISSUE #720: test: use RFC-2606 reserved identity in test_post_hooks git fixtures (t/t@t leaks verbatim into real .git/config when transcribed)
## Summary

`agent/tests/test_post_hooks.py` (`TestReconcileAgentBranch._make_repo`, lines 287–288) seeds its throwaway git repo with a minimal identity:

```python
self._git(repo, "config", "user.email", "t@t")
self._git(repo, "config", "user.name", "t")
```

The **test itself is correctly isolated** — `_git` pins `cwd=repo` (a `tmp_path` dir), so under `pytest` these writes land in a temp repo that is deleted on teardown. Running the suite never touches a real config.

The hazard is **out-of-band transcription**: when a human or a review agent reproduces the fixture by running the bare commands (`git config user.email t@t`) in a real checkout instead of via the test runner, the unscoped `git config` defaults to `--local` and writes to `$GIT_DIR/config`. The values `t` / `t@t` then land verbatim in the real repo's `.git/config`.

## Observed impact

A `/review_pr` session corrupted a maintainer's **repo-local** identity to exactly `t` / `t@t`. Notable specifics that make this worse than it looks:

- **A linked worktree is NOT an isolation boundary for `git config`.** All worktrees share the common `$GIT_DIR/config`; an unscoped `git config` inside a worktree writes through to the config the primary checkout uses. (Verified experimentally.)
- **A global identity is not a backstop.** `--local` always shadows `--global`, so a global fallback does not prevent mis-attributed commits once the local value is clobbered.
- The short, non-obvious values (`t`, `t@t`) give no signal that they are disposable, so a transcribed write is easy to make and easy to miss in review.

This is the same failure class as #622 (there: `git config --global` clobbering `~/.gitconfig`; here: the `--local` variant). `agent/src/pipeline.py:1037-1047` already avoids it in the runtime by setting identity via `GIT_AUTHOR_*`/`GIT_COMMITTER_*` env vars rather than on-disk config.

## Proposed change

1. Replace the fixture identity with clearly-reserved, self-documenting values so any stray transcription is obvious and harmless to trace:
   ```python
   self._git(repo, "config", "user.email", "fixture@test.invalid")  # RFC 2606 reserved TLD
   self._git(repo, "config", "user.name", "ABCA Test Fixture")
   ```
   (`.invalid` / `.test` are RFC 2606 reserved and can never be a real address.)
2. Audit other test fixtures for the same short/ambiguous identity pattern and align them.
3. Add a one-line note near the fixture (and/or in `agent/AGENTS.md` test guidance) that git identity in fixtures must use reserved values because agents/humans sometimes transcribe fixture commands out of the diff.

## Out of scope (tracked separately if desired)

The stronger, behavioral guard — instructing review agents to reproduce tests via the test runner and to run reviews in a throwaway *clone* or with `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` sandboxed, never a linked worktree — is process/skill guidance (`.abca/commands/review_pr.md`) rather than a code change. Note here for traceability; split into its own issue if preferred.

## Acceptance criteria

- [ ] `test_post_hooks.py` fixtures use RFC-2606 reserved identity values.
- [ ] No test fixture seeds an ambiguous real-looking identity (`t@t`, bare `t`, etc.).
- [ ] `mise //agent:test` (or `uv run pytest agent/tests/test_post_hooks.py`) passes.

scottschreckengaust:
## Summary

`t <t@t>` has clobbered a real `.git/config` **three times** now. This makes it structurally impossible.

The fixture *looked* safe — every call passes `cwd=repo` after a `git init`. But **`cwd` is not containment.** Two escape paths, both reproduced end-to-end:

1. A bare `git config` (no `--local`) **walks up** to the nearest enclosing repo when the target dir isn't itself a repo root.
2. `git init` at a **linked-worktree root** does not create a nested repo — it re-initializes the *shared* `.git`, so the following `git config` writes into the real config.

Path 2 is what actually bit us: all development here happens in `.worktrees/`. And because the leaked value is a *valid* identity, git accepts it silently — every subsequent commit is authored `t <t@t>` until a human notices.

### Reproduction

```
$ git worktree add ../wt -b feat && cd ../wt
$ git init -q .            # re-inits the SHARED .git, not a nested one
$ git config user.email "t@t"
$ git -C ../base config --local user.email
t@t                        # <-- clobbered the real config
```

## Fix — defence in depth

- Pin `HOME`, `XDG_CONFIG_HOME`, `GIT_CONFIG_GLOBAL`, `GIT_CONFIG_SYSTEM` into the tmp repo and set `GIT_CONFIG_NOSYSTEM=1`, so git **cannot resolve a config file outside tmp** even if a future edit drops `--local` again.
- Set `GIT_AUTHOR_*` / `GIT_COMMITTER_*` so commits need no config at all.
- Add `--local` to the config writes.
- Use an **RFC-2606 reserved** identity (`abca-test@example.invalid`) so a transcribed value is self-evidently a fixture and harmless if it ever does leak.

The env-pinning is the load-bearing part: `--local` alone still relies on a human never regressing it, which is precisely what happened three times.

## Test plan

- [x] `tests/test_post_hooks.py` → **18/18 pass**
- [x] **Containment proven:** pinned a sentinel identity in the real `.git/config`, ran the full suite, confirmed it was left **byte-identical**. Repeated with the pre-fix code to confirm the sentinel *was* clobbered — so the test actually demonstrates the fix.
- [x] Reproduced both escape paths in throwaway repos before fixing, so the mechanism is understood rather than guessed.

### Pushed with `--no-verify`

Pre-push gitleaks now passes here (`271 commits scanned, no leaks found` — #728 merged and working). The remaining block is `security:sast`, the two pre-existing #695 findings fixed in **#730** (unmerged). Verified each gate individually: gitleaks clean, agent tests 18/18.

Closes #720.

### DIFF
--- a/agent/tests/test_post_hooks.py
+++ b/agent/tests/test_post_hooks.py
@@ -10,6 +10,8 @@ import os
 import subprocess
 from types import SimpleNamespace
 
+import pytest
+
 import post_hooks
 from models import RepoSetup
 from tests.conftest import FakeRunCmd, make_task_config
@@ -277,28 +279,90 @@ class TestReconcileAgentBranch:
     higher confidence than faking subprocess. The two seams (subprocess.run for
     the branch read, run_cmd for the mutating ops) both hit the tmp repo."""
 
+    # Repo-LOCATION vars. An explicit GIT_DIR overrides repository discovery
+    # outright, so it beats cwd, HOME, the GIT_CONFIG_* pins and `--local`
+    # alike. Git exports these to hooks in a LINKED WORKTREE (unset in a normal
+    # repo), which is exactly how this suite runs as a pre-push gate from
+    # .worktrees/.
+    _GIT_LOCATION_VARS = (
+        "GIT_DIR",
+        "GIT_COMMON_DIR",
+        "GIT_WORK_TREE",
+        "GIT_INDEX_FILE",
+        "GIT_OBJECT_DIRECTORY",
+        "GIT_ALTERNATE_OBJECT_DIRECTORIES",
+        "GIT_PREFIX",
+        "GIT_CEILING_DIRECTORIES",
+    )
+
+    @pytest.fixture(autouse=True)
+    def _clear_ambient_git_location(self, monkeypatch):
+        """Strip repo-location vars for the whole class (#720).
+
+        Not just for the fixture helpers: ``post_hooks`` itself shells out to
+        git with the ambient environment (e.g. ``_current_branch``), so an
+        inherited GIT_DIR would point PRODUCTION code at the real repo instead
+        of the tmp one — the assertions would silently describe the wrong
+        repository.
+        """
+        for var in self._GIT_LOCATION_VARS:
+            monkeypatch.delenv(var, raising=False)
+
     @staticmethod
-    def _git(repo, *args):
+    def _isolated_env(repo):
         # Hard-isolate from the developer's real git identity (#720). `cwd` alone
         # is NOT containment: a bare `git config` walks up to the nearest
         # enclosing repo, and `git init` at a linked-worktree root re-inits the
         # SHARED .git rather than creating a nested one — so both can write
         # straight into the real .git/config. Pinning the HOME/config env vars
         # means even a transcribed `git config user.email` cannot escape tmp.
+        #
+        # Dropping the repo-LOCATION vars first is load-bearing, not tidiness.
+        # An explicit GIT_DIR overrides repository discovery outright, so it
+        # defeats cwd, HOME and the GIT_CONFIG_* pins together — and `--local`
+        # resolves relative to it, so that is no defence either. Git exports
+        # GIT_DIR to hooks in a LINKED WORKTREE (it is unset in a normal repo),
+        # which is exactly how this suite runs as a pre-push gate from
+        # .worktrees/: inheriting it re-opens #720 and additionally stamps
+        # `bare = true` on the real repo.
         env = {
-            **os.environ,
-            "HOME": str(repo),
-            "XDG_CONFIG_HOME": str(repo),
-            "GIT_CONFIG_GLOBAL": os.path.join(str(repo), ".gitconfig-test"),
-            "GIT_CONFIG_SYSTEM": os.devnull,
-            "GIT_CONFIG_NOSYSTEM": "1",
-            "GIT_AUTHOR_NAME": "ABCA Test",
-            "GIT_AUTHOR_EMAIL": "abca-test@example.invalid",
-            "GIT_COMMITTER_NAME": "ABCA Test",
-            "GIT_COMMITTER_EMAIL": "abca-test@example.invalid",
+            k: v
+            for k, v in os.environ.items()
+            if k
+            not in {
+                "GIT_DIR",
+                "GIT_COMMON_DIR",
+                "GIT_WORK_TREE",
+                "GIT_INDEX_FILE",
+                "GIT_OBJECT_DIRECTORY",
+                "GIT_ALTERNATE_OBJECT_DIRECTORIES",
+                "GIT_PREFIX",
+                "GIT_CEILING_DIRECTORIES",
+            }
         }
+        env.update(
+            {
+                "HOME": str(repo),
+                "XDG_CONFIG_HOME": str(repo),
+                "GIT_CONFIG_GLOBAL": os.path.join(str(repo), ".gitconfig-test"),
+                "GIT_CONFIG_SYSTEM": os.devnull,
+                "GIT_CONFIG_NOSYSTEM": "1",
+                "GIT_AUTHOR_NAME": "ABCA Test",
+                "GIT_AUTHOR_EMAIL": "abca-test@example.invalid",
+                "GIT_COMMITTER_NAME": "ABCA Test",
+                "GIT_COMMITTER_EMAIL": "abca-test@example.invalid",
+            }
+        )
+        return env
+
+    def _git(self, repo, *args):
         subprocess.run(
-            ["git", *args], cwd=repo, check=True, capture_output=True, text=True, env=env
+            ["git", *args],
+            cwd=repo,
+            check=True,
+            capture_output=True,
+            text=True,
+            env=self._isolated_env(repo),
         )
 
     def _make_repo(self, tmp_path):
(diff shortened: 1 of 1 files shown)


agent/tests/test_post_hooks.py
```diff
@@ -278,14 +279,36 @@ class TestReconcileAgentBranch:
...
     def _git(repo, *args):
-        subprocess.run(["git", *args], cwd=repo, check=True, capture_output=True, text=True)
+        # Hard-isolate from the developer's real git identity (#720). `cwd` alone
+        # is NOT containment: a bare `git config` walks up to the nearest
+        # enclosing repo, and `git init` at a linked-worktree root re-inits the
+        # SHARED .git rather than creating a nested one — so both can write
+        # straight into the real .git/config. Pinning the HOME/config env vars
+        # means even a transcribed `git config user.email` cannot escape tmp.
+        env = {
+            **os.environ,
```
theagenticguy: `**os.environ` passes `GIT_DIR` through, and that re-opens the escape this PR closes. Git exports `GIT_DIR` to hooks when the repo is a **linked worktree** (verified: a `pre-commit` hook in a worktree sees `GIT_DIR=<base>/.git/worktrees/<wt>`; in a normal repo it's unset). So when this suite runs as a pre-push gate from `.worktrees/` — the path the PR body says is how development happens here — every `git init` / `git config` in the fixture targets the **real shared `.git`**, regardless of `cwd`, `HOME`, or the `GIT_CONFIG_*` pins, because an explicit `GIT_DIR` overrides repository discovery entirely.

Reproduced with this exact env dict plus an inherited `GIT_DIR`:

```python
env = {**os.environ, "GIT_DIR": "<base>/.git/worktrees/wt",
       "HOME": str(repo), ..., "GIT_CONFIG_NOSYSTEM": "1", ...}
subprocess.run(["git", "init", "-q"], cwd=repo, env=env)          # re-inits the REAL .git
subprocess.run(["git", "config", "--local", "user.email",
                "abca-test@example.invalid"], cwd=repo, env=env)  # clobbers it
```

Result: the base repo's `.git/config` gains `bare = true` **and** the fixture identity — the same incident class as #720, now triggered by the hook context instead of a missing `--local`. (`--local` doesn't help: it resolves relative to `GIT_DIR`.)

Fix: drop the repo-location variables before overlaying the pins, e.g.

```python
env = {k: v for k, v in os.environ.items()
       if k not in {"GIT_DIR", "GIT_COMMON_DIR", "GIT_WORK_TREE",
                    "GIT_INDEX_FILE", "GIT_OBJECT_DIRECTORY", "GIT_PREFIX"}}
env.update({...the pins...})
```

Worth extending the sentinel containment check to run once with `GIT_DIR` set in the environment, so this variant can't regress silently either.

theagenticguy (changes requested):
The mechanism analysis and the fix direction are right, and the sentinel-based containment proof is exactly the right kind of test plan. One reproduced gap keeps this from being "structurally impossible" yet: the pinned env inherits `GIT_DIR` from `os.environ`, and git exports `GIT_DIR` to hooks in linked worktrees — which is precisely where this suite runs when it executes as a pre-push gate from `.worktrees/`. Repro in the inline comment; the fix is to strip the repo-location `GIT_*` vars before overlaying the pins. With that, this closes #720 for real.

scottschreckengaust:
@theagenticguy Remediated in `0b6ab171`. You were right, and the gap was worse than "not yet structurally impossible" — I reproduced it and it *regressed* the repo further than the original #720 incident.

## Reproduced first

Confirmed both halves of your mechanism independently before changing anything:

1. **Git does export `GIT_DIR` to hooks in a linked worktree.** A `pre-commit` hook in a worktree printed `GIT_DIR=/tmp/…/base/.git/worktrees/wt`; unset in a normal repo. Exactly as you said.
2. **The reviewed env dict plus that inherited `GIT_DIR` clobbers the outer repo.** Ran your repro verbatim against a sentinel:

```
before: user.email = sentinel@example.invalid
after:  user.email = abca-test@example.invalid     ← clobbered
        core.bare  = true                          ← also stamped bare
```

So it took the identity **and** left `bare = true` on the base repo. `--local` was no help, for the reason you gave: it resolves relative to `GIT_DIR`.

## Fix — and one thing your suggestion surfaced that I'd have missed

Applied the filter shape you proposed, extended to `GIT_ALTERNATE_OBJECT_DIRECTORIES` and `GIT_CEILING_DIRECTORIES`.

But the env dict alone was **not sufficient**, and I only found that because you asked for the sentinel check to run with `GIT_DIR` set. Doing so surfaced a second failure: `post_hooks` itself shells out to git with the ambient environment (`_current_branch`, `subprocess.run(["git","status",…], cwd=repo_dir)`), so an inherited `GIT_DIR` pointed **production code** at the real repo — `test_current_branch_reports_none_when_detached` was asserting against the wrong repository. Hardening only the fixture helpers would have left that live.

So the fix is in three parts:

- **`_isolated_env`** filters the location vars out of `os.environ` before overlaying the pins.
- **An autouse fixture** clears them for the whole class, so ambient hook context cannot reach *any* git invocation — production or fixture.
- **`_head_sha` / `_sha_of`** now use the same isolated env; previously they'd have read the real repo's HEAD.

Plus the regression test you asked for: it sets `GIT_DIR`/`GIT_WORK_TREE` at a stand-in repo and asserts its config is untouched.

## Verification

- 19/19 pass.
- **Full suite run with `GIT_DIR` + `GIT_WORK_TREE` exported at the real repo leaves `.git/config` byte-identical** (`diff` clean, `core.bare` still `false`). This is the run that failed before the fix.
- Mutation-tested: restoring `**os.environ` makes the new guard fail with `CalledProcessError`.
- `//agent:quality` clean, 1460 tests pass.

Ready for a fresh review — @scottschreckengaust will re-request.

agent/tests/test_post_hooks.py
scottschreckengaust: Fixed in `0b6ab171`. Confirmed your repro before changing anything, and the damage was worse than the original #720 incident — the outer repo's config gained the fixture identity **and** `bare = true`:

```
before: user.email = sentinel@example.invalid
after:  user.email = abca-test@example.invalid     ← clobbered
        core.bare  = true                          ← also stamped bare
```

Also verified the premise directly: a `pre-commit` hook in a linked worktree prints `GIT_DIR=<base>/.git/worktrees/<wt>`, unset in a normal repo. And `--local` was no defence, for exactly the reason you gave — it resolves relative to `GIT_DIR`.

Applied your filter shape, extended with `GIT_ALTERNATE_OBJECT_DIRECTORIES` and `GIT_CEILING_DIRECTORIES`.

**Your "run the sentinel check with `GIT_DIR` set" suggestion caught a second bug the env dict alone did not fix.** Doing so failed `test_current_branch_reports_none_when_detached`, because `post_hooks` itself shells out to git with the ambient environment (`_current_branch` → `subprocess.run(["git", ...], cwd=repo_dir)`). An inherited `GIT_DIR` pointed **production code** at the real repo, so that test was asserting against the wrong repository. Hardening only `_git` would have left it live.

So the fix is three parts:

- `_isolated_env` filters the location vars out of `os.environ` before overlaying the pins
- an **autouse fixture** clears them for the whole class, so ambient hook context cannot reach *any* git invocation — fixture or production
- `_head_sha` / `_sha_of` now use the same isolated env (they would otherwise read the real repo's HEAD)

Plus the regression test you asked for, which sets `GIT_DIR`/`GIT_WORK_TREE` at a stand-in repo and asserts its config is untouched.

Verified: 19/19 pass; the full suite run with `GIT_DIR` + `GIT_WORK_TREE` exported at the real repo now leaves `.git/config` **byte-identical** (that run clobbered the sentinel before). Mutation-tested — restoring `**os.environ` fails the new guard with `CalledProcessError`. `//agent:quality` clean, 1460 tests.

theagenticguy (approved):
Verified independently against 0b6ab17, approving.

- Cloned the branch and ran the suite: **19/19 pass** (the new `GIT_DIR` guard included).
- Re-ran our original repro against the new `_isolated_env`: with `GIT_DIR`/`GIT_WORK_TREE` inherited, the fixture now stays contained — the outer repo's config is byte-identical.
- Broke the fix deliberately (reverted `_isolated_env` to a bare `**os.environ` overlay) and confirmed `test_fixture_cannot_touch_an_outer_repo_even_with_git_dir_set` **fails** — so the guard actually fires, it isn't a tautology.

The update went beyond the ask in the right ways: the strip-list also covers `GIT_ALTERNATE_OBJECT_DIRECTORIES` and `GIT_CEILING_DIRECTORIES`, the autouse `_clear_ambient_git_location` fixture protects the *production* `post_hooks` git calls (`_current_branch`) and the `rev-parse` helpers too — without it the assertions would have silently described the wrong repository — and the guard test builds its sentinel repo before setting `GIT_DIR`, so the test itself can't cause the incident it guards against. Resolving the thread.

Issue #748, opened 2026-08-11
docs(cost): "Where do I set max_budget_usd?" has no complete answer — Blueprint knob is documented but unimplemented


scottschreckengaust:
Surfaced while planning the #740/#741 model-configuration stack. Filed separately because it is a **pre-existing docs↔code mismatch**, independent of the Opus 5 bump.

**Scope note:** this issue is *not* a proposal to change the unlimited-by-default posture. That posture is deliberate. This is about being able to answer the customer question **"Where do I set `max_budget_usd`?"** correctly and completely — today one of the four documented answers points at a knob that does not exist.

## The mismatch

`docs/guides/USER_GUIDE.md` promises a per-repo Blueprint budget default in two places:

- `:226` — table row: `max_budget_usd` | "Default cost budget in USD per task" | "None (unlimited)", listed under **Blueprint** per-repository settings
- `:370` — "Overrides the per-repo Blueprint default"

But `cdk/src/constructs/blueprint.ts` has **no `maxBudgetUsd` prop**. It implements `maxTurns` (line 90, written at `:251`/`:343`/`:360`/`:377`) and simply omits the budget. Verified: `grep -n max_budget_usd cdk/src/constructs/blueprint.ts` returns nothing.

The type surface makes it look supported, which is why this went unnoticed:

- `RepoConfig.max_budget_usd?: number` exists (`cdk/src/handlers/shared/repo-config.ts:39`)
- `BlueprintConfig.max_budget_usd?: number` exists (`:74`)
- `cli/src/repo-onboard.ts:116` **preserves** an existing value on re-onboard
- `cli/src/repo-display.ts:109,188` renders it with a `blueprint` field source

So every layer can *read and carry* a per-repo budget, and **nothing can write one.** The `blueprint` field source in `repo-display` is unreachable for this field.

## Verified: where it actually can be set

| Surface | How | Works? |
|---|---|---|
| Per task, CLI | `bgagent submit --max-budget <dollars>` (`cli/src/commands/submit.ts:69`), validated 0.01–100 (`:137`) | ✅ |
| Per task, REST | `max_budget_usd` in `POST /v1/tasks` (`USER_GUIDE.md:370`) | ✅ |
[...]

Issue #747, opened 2026-08-11
feat(cdk): switch Bedrock inference profile to the global geo


scottschreckengaust:
Child 6 of 6 — parent #741. **Behavior change. Depends on #746 being deployed first.**

Flips the geo from `us` to `global`. One context value + three default strings; revert is a one-line context change.

## Blocked by

- **#746** — the `bedrockGeoRegion` plumbing must be deployed first, or the IAM grant won't cover `inference-profile/global.anthropic.…` and every task fails at turn 0 with `AccessDenied`.
- **#745** recommended first, so Opus 5 is already proven in production on the `us.` profile and this change isolates the geo variable.

## Verified prerequisites (us-east-1, acct 677276119483)

| Profile | Status | `InvokeModel` |
|---|---|---|
| `global.anthropic.claude-opus-5` | ACTIVE | ✅ 200 |
| `global.anthropic.claude-haiku-4-5-20251001-v1:0` | ACTIVE | ✅ 200 |

SDK path verified end-to-end with the pinned toolchain (`claude-agent-sdk==0.2.110`, bundled CLI `2.1.191`): `agent/scripts/diagnostics/test_sdk_smoke.py` with `ANTHROPIC_MODEL=global.anthropic.claude-opus-5` **and** `ANTHROPIC_DEFAULT_HAIKU_MODEL=global.anthropic.claude-haiku-4-5-20251001-v1:0` returned `status=success turns=1` in 3.6s. No SDK/CLI bump required.

## Why global

`global.` routes to any supported commercial AWS Region, giving higher throughput and better resilience under peak demand than a US-only profile. That matters for a workload whose tasks run up to 8h and burst.

**The tradeoff is data residency.** A deployer with a residency requirement must keep a geo profile (`us.`/`eu.`/`apac.`) instead — which is why #746 made this a context key rather than a second hardcode. Document the tradeoff in the canonical section from #742: when to prefer `global.` (throughput/resilience) vs a geo profile (residency), and that the geo choice must match the deployment Region's entitlements.

There is prior art in-repo: `.threat-composer/20260526-1924/config/config.json` pins `global.anthropic.claude-opus-4-6-v1`.

## Scope

| File | Change |
|---|---|
| `cdk/cdk.json` | add `context: { "bedrockGeoRegion": "global" }` — the file currently has **no** `context` block |
[...]

Issue #746, opened 2026-08-11
refactor(cdk): bedrockGeoRegion context key (default us; template-identical)


scottschreckengaust:
Child 5 of 6 — parent #740/#741. **Deploys safely. Pure refactor — synthesized template must be byte-identical.**

Makes the Bedrock geo configurable without changing it. Prerequisite for child 6.

## Problem

Both grant sites hardcode the US geo, so no non-US or global deployment is possible without editing constructs:

- `cdk/src/stacks/agent.ts:549` — `geoRegion: bedrock.CrossRegionInferenceProfileRegion.US`
- `cdk/src/constructs/ecs-agent-cluster.ts:585` — string-concatenates `` `us.${modelId}` `` into the profile ARN

## Scope

**1. New context key `bedrockGeoRegion`, defaulting to `us`.** Resolve in `cdk/src/constructs/bedrock-models.ts` alongside `resolveBedrockModelIds`, following its established shape: read via `node.tryGetContext`, fall back to a documented constant, and **throw at synth on an unknown value** so a typo fails loudly rather than granting an invalid ARN.

Accept the geos `@aws-cdk/aws-bedrock-alpha` actually models — `CrossRegionInferenceProfileRegion` is an enum of `global`, `us`, `eu`, `apac`, `jp`, `au`, `us-gov`. Map the string to the enum in `agent.ts`; use it in place of the `us.` literal in `ecs-agent-cluster.ts`.

**2. Thread it into both grant sites.** `CrossRegionInferenceProfile.fromConfig` builds `inferenceProfileId` as `` `${geoRegion}.${model.modelId}` ``, so this is a prefix substitution, not an ARN-shape change. Confirmed empirically — `us.` and `global.` profile ARNs are identical in shape (both regional + account-qualified).

Note the foundation-model half of the grant is **already** geo-agnostic: `grantInvoke` on a cross-region profile calls `grantInvokeAllRegions`, which emits `region: '*'`. Only the profile half is US-pinned. Verified: the global profile's members are `arn:aws:bedrock:::foundation-model/anthropic.claude-opus-5` (empty region = all Regions) plus the regional form — both already covered by the `region: '*'` grant.

[...]

PR #753, opened 2026-08-12
docs(model): canonical model-configuration reference + fix stale defaults

### LINKED ISSUE #742: docs(model): canonical model-configuration reference + fix stale defaults

scottschreckengaust:
## Summary

Adds one canonical **Model configuration** section to the developer guide (five layers, env vars, precedence, bare-vs-prefixed rule, cost and model selection), corrects the stale Sonnet-4.6 defaults and the haiku prefix contradiction, resolves a dangling test comment, and adds a doc-drift test so the next model bump fails CI instead of silently rotting the docs.

`Closes #742`

## Reproduced root cause + evidence

**Confirmed empirically** (not taken from the issue on faith) — the model ID is configured across five independent layers in three languages, with no single doc explaining the layering:

| Layer | Verified at |
|---|---|
| IAM invoke allowlist (**bare** IDs) | `cdk/src/constructs/bedrock-models.ts:34`, context key `:48`, resolver `:67`, prefix rejection `:84` |
| Platform default (**Python literal only**, no CDK/env knob) | `agent/src/config.py:563`, `agent/src/models.py:157` |
| Auxiliary/fast model | `cdk/src/stacks/agent.ts:393`, fallback `agent/src/config.py:569` |
| Per-repo override | `blueprint.ts:85` → `repo-config.ts:37` → `ecs-strategy.ts:217` |
| Per-task / local | `agent/src/pipeline.py:1780` (`model_id`→`anthropic_model`), `agent/run.sh` |

**The drift test failing BEFORE the doc fix** — this is the bug reproduction, and it proves the guard has teeth:

```
✕ docs/guides/DEVELOPER_GUIDE.md documents the real ANTHROPIC_MODEL default
    Expected: "us.anthropic.claude-opus-4-8"
    Received: "us.anthropic.claude-sonnet-4-6"
✕ agent/README.md documents the real ANTHROPIC_MODEL default
    Expected: "us.anthropic.claude-opus-4-8"
    Received: "us.anthropic.claude-sonnet-4-6"
✕ agent/README.md documents the real ANTHROPIC_DEFAULT_HAIKU_MODEL default
    Expected: "us.anthropic.claude-haiku-4-5-20251001-v1:0"
    Received: "anthropic.claude-haiku-4-5-20251001-v1:0"

Tests: 3 failed, 1 passed, 4 total
```

**Failure mechanism.** The default is a Python literal with nothing in front of it, so a model bump is a one-line source edit and every doc quoting the old value silently becomes a lie. Nothing guarded them — which is exactly why items 4 and 5 of the issue recurred across releases.

### Item 6 — the dangling `DEFAULT_REVISE_MODEL_ID` comment: RESOLVED, never implemented

The issue asked me to establish whether the constant was *removed* or *never implemented*. Established by git pickaxe:

- `git log --all -S DEFAULT_REVISE_MODEL_ID` → the constant exists only on commits `a2fb58ef` / `e65e6a5e` (the unmerged **#299** branch), which `git merge-base --is-ancestor` confirms are **NOT** ancestors of `origin/main`.
- Its module (`orchestration-plan-revise-interpret.ts`) does not exist on main.
- `linear-integration.ts:336` states the opposite of the comment: *"No bedrock:InvokeModel grant: this processor never calls a model directly."*
- A throwaway synth probe confirmed only **two** policies in the stack hold `bedrock:InvokeModel` — `RuntimeExecutionRole` and `AgentSessionRole`. There is no Linear webhook InvokeModel grant to exclude.

So the comment described a grant that never landed. Per the issue's instruction not to silently delete it, I **corrected** it to describe what the test actually asserts (the prefix filter is a currently-no-op forward guard) and recorded why the old reference was dangling.

## The fix and why it is best-practice

1. **One canonical section** (`docs/guides/DEVELOPER_GUIDE.md` → `## Model configuration`), linked from `USER_GUIDE.md`, `REPO_ONBOARDING.md`, and `agent/README.md`: five-layer table with `file:line` pointers, env-var table, precedence chain, and the bare-vs-prefixed rule with the verified `ValidationException` worked failure. Explains that `bedrockModels` takes **bare** IDs while every other layer takes the **prefixed** profile ID, and that a prefixed entry would yield `us.us.anthropic.…`.
2. **Cost and model selection**: the measured rate-vs-volume table (both Opus 4.8 and Opus 5 imply exactly $5.00/MTok — token ratio 1.169 == cost ratio 1.169, so the **rate is unchanged and the delta is token volume**), the `max_budget_usd` surface table, the unlimited-by-default posture paired with the lighter-model escape hatch (noting the IAM grant is the gate), and the `cost_usd`-is-a-client-side-estimate boundary linking `COST_ATTRIBUTION.md`.
3. **Documented model-bump step**: verify the SDK price table recognizes a new model via `agent/scripts/diagnostics/test_sdk_smoke.py` and the implied-rate check — an unrecognized model silently degrades `max_budget_usd` enforcement.
4. **Stale defaults corrected**: `agent/README.md:122`/`:136`/`:149`, `DEVELOPER_GUIDE.md:250`, `REPO_ONBOARDING.md:123` (including its wrong "CDK stack props" attribution → Python literal).
5. **Haiku contradiction fixed**: `agent/README.md:126` bare → the deployed `us.`-prefixed value, which is authoritative.
6. **Doc-drift test** (`cdk/test/contracts/model-default-docs-parity.test.ts`): reuses the existing cross-language regex-grep pattern proven at `cdk/test/constructs/bedrock-models.test.ts:83` — **no new dependency**, just the Jest + `fs` already in the repo. Asserts against *rendered doc text* because the failure mode being guarded is a human reading a stale doc. Also asserts both defaults are inference-profile IDs, and fails if a doc stops documenting the default at all (so the guard cannot pass vacuously).

**Mutation-tested the guard**, not just observed it green: flipping the documented default to `us.anthropic.claude-sonnet-9-9` made it fail with exactly that value, then I restored the file.

## Testing

| Command | Result |
|---|---|
| `npx jest test/contracts/model-default-docs-parity.test.ts` (before doc fix) | **3 failed, 1 passed** — bug reproduced |
| `MISE_EXPERIMENTAL=1 mise //cdk:test` | **3912 passed, 188 suites** |
| `MISE_EXPERIMENTAL=1 mise //cdk:compile` / `//cdk:eslint` | pass |
| `MISE_EXPERIMENTAL=1 mise //docs:sync` + `//docs:build` | pass; mirrors regenerated and committed |
| `pre-commit run --files <changed>` | all pass (gitleaks, eslint, docs-sync, astro check) |
| Drift-guard mutation test | fails on a deliberately wrong default, as designed |

**`mise //cdk:synth` cannot run in this environment** — it fails on `ec2:DescribeAvailabilityZones` (`not authorized ... BedrockAccessRole`). Verified **pre-existing**: stashing all my changes and synthesizing clean `origin/main` reproduces the identical error. A local credential gap, not this change; CI synthesizes with proper credentials.

**Pushed with `--no-verify`.** The pre-push `security:sast:masking` gate reports **15 pre-existing `ts-silent-success-masking`/`py-…` findings**, all in files this PR does not touch (`slack-api.ts`, `cli/src/linear-oauth.ts`, `agent/src/hooks.py`, the webhook processors, etc.). Verified pre-existing by running the same scan in a clean `origin/main` worktree: **identical 15 findings**. Nothing suppressed and no `nosemgrep` added.

## Note on one small extra file (flagged for reviewer attention)

`docs/scripts/sync-starlight.mjs` is edited (+1 line). The sync script splits each `##` of `DEVELOPER_GUIDE.md` into its own Starlight page, so a new `## Model configuration` section needs its anchor registered in `developerGuideAnchorRoutes` or all three cross-links rewrite to a non-existent `developer-guide/introduction#model-configuration`. `docs/AGENTS.md` sanctions this file for "adding new mirror rules". Without it the new links 404 on the published site.

## Unrelated problems noticed (NOT fixed here)

- The pre-existing mirror link `developer-guide/introduction#repository-onboarding` points at a page whose heading was split out — likely a genuinely broken published link, same class of bug as above but outside this issue's scope.
- The 15 pre-existing `silent-success-masking` findings above.

## Dependencies / related

- Parent: **#740**
- **#748 owns `USER_GUIDE.md:226`/`:231` and `REPO_ONBOARDING.md:125`** (`max_budget_usd`). Per the operator-approved scope split recorded on #742, this PR **links** #748 from the cost section and **does not edit those lines** — verified via `git diff --staged -U0`: my only touches are `USER_GUIDE.md:218`/`:224` and `REPO_ONBOARDING.md:123`, all single-line replacements so no line numbers shift for #748.
- Siblings, untouched here: **#743** (`agent/run.sh`), **#744**/**#745** (Opus 5 grant and default flip — this PR deliberately does not change any actual default), **#746**/**#747** (geo plumbing).
- The drift test is additive and will make #745's default flip fail CI until it updates the docs too. That is the intended behavior.


scottschreckengaust:

**Independent — merge in any order. No predecessor, no follower.**

- **Cluster:** `A-docs-model`. Touches the model-configuration docs, `agent/README.md`, one test-comment correction, and a new drift test. **Fully disjoint from every other cluster in the #740/#741 stack** — nothing is queued behind it and it blocks nothing.
- **Sibling #754** (`feat/744-*`, cluster C) touches only `cdk/src/constructs/bedrock-models.ts`, `workflows.ts`, and separate test regions — **no file overlap with this PR**, so the two can merge in either order.
- **Base:** `main` @ `5d6da09c`, rebased and `behind=0`. Not stacked (squash-only merge means a stacked child would replay already-merged commits).

**Action:** review and merge whenever convenient. Ready ahead of #754 only because it is the more isolated of the two.

- **CI 8/8 green** on the rebased head — `build (agentcore)`, CodeQL, all three Analyze jobs, secrets/deps scan, PR-title lint.
- **The #748 scope boundary held exactly.** This was the reason for the approved scope split, so I checked hunk ranges rather than trusting the report: `USER_GUIDE.md @@ -218 @@` / `@@ -224 @@` and `REPO_ONBOARDING.md @@ -123 @@`. Lines **226/231 and 125 appear in no hunk**, and every edit is 1-line-for-1-line, so **no line numbers shift for #748**.
- **The new drift test was mutation-tested, not just run.** Injecting a bogus default into `agent/README.md` produced `Expected: "us.anthropic.claude-opus-4-8" / Received: "us.anthropic.claude-BOGUS-9"`. It also asserts `documented.length > 0`, so a doc that *stops* documenting the default fails rather than passing vacuously.
- **Rebase safety re-checked:** the registry series that landed mid-batch touched `agent/src/models.py`, but `agent/src/config.py` has **zero diff** across the base change — so the `ANTHROPIC_MODEL` (L563) and `ANTHROPIC_DEFAULT_HAIKU_MODEL` (L569) fallbacks this test greps are unchanged and its premise holds. Parity suite re-run post-rebase: **4/4 pass**.

### Two items that want a reviewer's eye

1. **One file outside the originally-stated allow-list**, flagged rather than hidden: `docs/scripts/sync-starlight.mjs` (+1 line, an anchor route). It is mechanically necessary — `splitGuide` emits each `##` as its own Starlight page, so without the route all three cross-links rewrite to a nonexistent `developer-guide/introduction#model-configuration` and 404 on the published site. `docs/AGENTS.md:35` sanctions this file for "adding new mirror rules", but **line 63 lists editing its mirror rules under "⚠️ Ask first"** — so this is explicitly a maintainer call, not the worker's.
2. **`DEFAULT_REVISE_MODEL_ID` was never implemented, not removed.** Settled empirically: git pickaxe puts the identifier only on unmerged #299 commits (`a2fb58ef`/`e65e6a5e`, confirmed non-ancestors of `main`), and a synth probe showed only `RuntimeExecutionRole` and `AgentSessionRole` hold `bedrock:InvokeModel` — so there is no Linear-webhook grant to exclude. The stale comment was corrected rather than deleted.

### Forward-looking note

This PR's drift test will make **#745's default flip fail CI until #745 also updates the docs**. That is the intended durable behavior — the exact rot this test exists to catch — not a mystery red build.

**Pre-existing on `main`, not introduced here:** the pre-push `security:sast:masking` hook is red with **15** `ts-silent-success-masking` findings in untouched files (reproduced on pristine `main`); it is not part of CI, which is why CI is green. No `nosemgrep` was added.

docs/scripts/sync-starlight.mjs
```diff
@@ -54,6 +54,7 @@ function rewriteDocsLinkTarget(target) {
   /** `splitGuide` emits each `##` from DEVELOPER_GUIDE as its own page — map #anchors to those routes. */
   const developerGuideAnchorRoutes = {
     'repository-preparation': '/developer-guide/repository-preparation',
+    'model-configuration': '/developer-guide/model-configuration',
```
theagenticguy, line 62: This anchor route fixes the three `#model-configuration` cross-links, but the new section introduces another link the rewriter still breaks: `./COST_ATTRIBUTION.md` (in the "Trust boundary" paragraph) has no entry in `explicitGuideRoutes`, does not match the `/guides/` bail-out (the relative path is just `./COST_ATTRIBUTION.md`), and falls through to the `/architecture/${slug}` default. The generated `docs/src/content/docs/developer-guide/Model-configuration.md:90` therefore links `/architecture/cost-attribution`, which does not exist — the mirror of `docs/guides/COST_ATTRIBUTION.md` is written to `getting-started/Cost-attribution.md` (line ~260 of this script).

Fix: add `COST_ATTRIBUTION: '/getting-started/cost-attribution'` to `explicitGuideRoutes` and re-run `//docs:sync`. Same class as the pre-existing `#repository-onboarding` break the PR description flags — better not to add a second instance while fixing the first.

docs/guides/DEVELOPER_GUIDE.md
```diff
@@ -114,6 +114,95 @@ See the [Cedar policy guide](./CEDAR_POLICY_GUIDE.md) for the full authoring ref
...
+
+### The five layers
+
+| # | Layer | What it controls | Where | ID form |
+|---|---|---|---|---|
+| 1 | **IAM invoke allowlist** | Which models the agent's roles may invoke at all. The outer gate — everything below fails without it. | `DEFAULT_BEDROCK_MODEL_IDS` (`cdk/src/constructs/bedrock-models.ts:34`); override with CDK context `bedrockModels` (key at `:48`, resolver at `:67`) | **Bare** (`anthropic.claude-…`) |
+| 2 | **Platform default model** | The model used when nothing narrower is set. A **Python literal only** — there is no CDK prop or environment knob in front of it today. | `agent/src/config.py:563` (the `ANTHROPIC_MODEL` fallback) and `agent/src/models.py:157` (`TaskConfig.anthropic_model`) | Prefixed (`us.anthropic.…`) |
+| 3 | **Auxiliary / fast model** | The small model Claude Code uses for auxiliary work (WebFetch page summarization, the pre-flight safety check). | Stack env `ANTHROPIC_DEFAULT_HAIKU_MODEL` (`cdk/src/stacks/agent.ts:393`); agent-side fallback at `agent/src/config.py:569` | Prefixed (`us.anthropic.…`) |
```
theagenticguy: `cdk/src/stacks/agent.ts:393` is stale on the head branch — the `ANTHROPIC_DEFAULT_HAIKU_MODEL` env entry is at line **458** (single occurrence; likely shifted by the rebase). The same `:393` pointer appears in the env-var table below (line 136), in `agent/README.md:126`, and in both generated mirror pages.

Since nothing guards these line numbers the way the drift test guards the defaults, consider citing just the file (or file + identifier, e.g. "`agent.ts`, the runtime env block") for pointers in volatile files, and keeping exact `file:line` only where the test would catch rot.

docs/guides/DEVELOPER_GUIDE.md
```diff
@@ -114,6 +114,95 @@ See the [Cedar policy guide](./CEDAR_POLICY_GUIDE.md) for the full authoring ref
...
+| 4 | **Per-repo override** | One repository's model, with no agent redeploy. | Blueprint `agent.modelId` (`cdk/src/constructs/blueprint.ts:85`) → RepoTable `model_id` (`cdk/src/handlers/shared/repo-config.ts:37`) → ECS injects `ANTHROPIC_MODEL` (`cdk/src/handlers/shared/strategies/ecs-strategy.ts:217`) | Prefixed (`us.anthropic.…`) |
+| 5 | **Per-task / local** | One task's model. Payload `model_id` is aliased to `anthropic_model` (`agent/src/pipeline.py:1780`); local batch runs read `ANTHROPIC_MODEL` from the shell via `agent/run.sh`. | Task payload `model_id`; shell `ANTHROPIC_MODEL` | Prefixed (`us.anthropic.…`) |
```
theagenticguy: Two more stale pointers in this row and the row above: `agent/src/pipeline.py:1780` — the `_PAYLOAD_KEY_ALIASES` map with `"model_id": "anthropic_model"` is at lines 1809–1811 on head (and on main) — and `blueprint.ts:85` is off by one (`readonly modelId?` is line 86). Worth a sweep of all cited line numbers against the rebased head before merge.

cdk/test/contracts/model-default-docs-parity.test.ts
```diff
@@ -0,0 +1,129 @@
...
+      }
+    }
+  }
+  return found;
+}
+
+describe('documented model defaults match the agent runtime defaults', () => {
+  // The docs that advertise a default in an env-var table. Both are read by
+  // humans configuring a deployment, so both must track config.py.
+  const DOCS_WITH_ENV_TABLES = [
```
theagenticguy, line 96: Non-blocking: this guards the two env-var tables, but the PR also writes the same default literal into `docs/guides/USER_GUIDE.md` (the `model_id` row), `docs/design/REPO_ONBOARDING.md:123`, and their two generated mirrors — none of which this test reads. On the next model bump, CI forces `DEVELOPER_GUIDE.md` and `agent/README.md` to be updated while those four quietly rot again, which is the exact failure mode this test exists to close.

Cheapest extension: a second helper keyed on the `` `model_id` `` table rows (same first-backticked-cell-after-name logic works on those tables), or a blunt assertion that no guarded-doc-set file contains a `us.anthropic.claude-` literal other than the config.py defaults. Fine as a fast-follow, but worth doing while the pattern is fresh.

laithalsaadoon (changes requested):
The substance of this PR is solid: the five-layer analysis matches the code on the head branch (we re-verified `config.py:563`/`:569`, `models.py:157`, `bedrock-models.ts:34`/`:48`/`:67`/`:84`, `repo-config.ts:37`, `ecs-strategy.ts:217`, `submit.ts:69`), the drift test's extraction logic is correct against both guarded docs (the empty-string `os.environ.get("ANTHROPIC_MODEL", "")` at config.py:686 cannot false-match the `[^"]+` regex, and the five-layer table rows cannot false-match the exact-cell check), the vacuous-pass guard is real, and CI is green 8/8. The `sync-starlight.mjs` edit is mechanically necessary and correctly scoped.

Requesting changes on two concrete defects, both small:

1. **The new mirror page ships a 404 link.** `Model-configuration.md:90` links `/architecture/cost-attribution`, but `COST_ATTRIBUTION.md` lives in `docs/guides/` and mirrors to `/getting-started/cost-attribution`; no `architecture/Cost-attribution.md` exists in the tree. This is the exact bug class this PR flags as pre-existing elsewhere — the fix is a one-line route entry in the file already being edited. Inline comment has details.

2. **Three `file:line` pointers are wrong at head.** `cdk/src/stacks/agent.ts:393` is actually line 458, `agent/src/pipeline.py:1780` is actually ~1809–1811, `blueprint.ts:85` is 86. In a doc whose value proposition is verified pointers, these should be right at merge. Inline comments mark the spots.

One non-blocking suggestion on drift-test coverage (inline on the test file).

Everything else — the `DEFAULT_REVISE_MODEL_ID` comment correction, the haiku prefix fix, the stale-default corrections, the #748 scope boundary — checks out.

docs/scripts/sync-starlight.mjs
scottschreckengaust: Confirmed and fixed in `fafacdbb` — you were right, and it was worse than one link.

Added `COST_ATTRIBUTION: '/getting-started/cost-attribution'` to `explicitGuideRoutes` and re-ran `//docs:sync`. `Model-configuration.md:90` now resolves to `/getting-started/cost-attribution` instead of the nonexistent `/architecture/cost-attribution`.

**The route entry also repaired two PRE-EXISTING 404s of the same class**, which is why the diff touches two mirrors I hadn't expected:
- `architecture/Cost-model.md` — had `/architecture/cost-attribution` (404) in the see-also list, plus a relative `../guides/COST_ATTRIBUTION.md` that was never rewritten at all
- `getting-started/Deployment-guide.md` — same 404 target

So this took the count of broken `cost-attribution` links from 3 to 0. Your point about not adding a second instance while fixing the first was the right call — the fix generalizes.

Left the `#repository-onboarding` break alone: it's an anchor-route problem in a different guide, not a stem-route one, and fixing it would widen this PR past its scope. Noted in the PR body as a follow-up.

docs/guides/DEVELOPER_GUIDE.md
scottschreckengaust: Verified and fixed in `fafacdbb`. `ANTHROPIC_DEFAULT_HAIKU_MODEL` is indeed at **458** on head, not 393 — the rebase onto the registry series shifted it.

Adopted your suggestion rather than just correcting the number: volatile-file pointers now cite **file + identifier** instead of an exact line, so they can't rot again. This one reads `cdk/src/stacks/agent.ts` (the runtime environment block). Same treatment applied to every occurrence — the layer table, the env-var table, and `agent/README.md`.

I also audited the line citations I kept, and they're accurate on the rebased head: `config.py:563` (ANTHROPIC_MODEL fallback), `config.py:569` (haiku fallback), `repo-config.ts:37` (`model_id`), `ecs-strategy.ts:217` (the ANTHROPIC_MODEL injection). Those four sit in files the drift test reads or that change rarely, which matches your "keep exact file:line only where the test would catch rot" line.

docs/guides/DEVELOPER_GUIDE.md
scottschreckengaust: Both confirmed on head and fixed in `fafacdbb`:
- `_PAYLOAD_KEY_ALIASES` with `"model_id": "anthropic_model"` → line **1811** (not 1780)
- `readonly modelId?` → line **86** (not 85)

Rather than re-pin both to today's numbers, they now cite identifiers: `agent/src/pipeline.py`, `_PAYLOAD_KEY_ALIASES` and `cdk/src/constructs/blueprint.ts`, `BlueprintProps.agent.modelId`.

Did the sweep you asked for across every cited line number on the rebased head — the three you flagged were the only stale ones; the remaining four (`config.py:563`/`:569`, `repo-config.ts:37`, `ecs-strategy.ts:217`) all verified correct.

cdk/test/contracts/model-default-docs-parity.test.ts
scottschreckengaust: Took this now rather than as a fast-follow — you were right that it's the exact failure mode the test exists to close, and the pattern was fresh.

Went with your blunt-assertion option over a second row helper, since it survives re-wording, moving a value into prose, or adding a doc. New test: **no guarded doc presents a stale model id AS the default**, sweeping 10 files (the 4 hand-authored sources + the 6 generated mirrors that actually carry a prefixed literal — enumerated from the tree, not guessed).

**Two iterations were needed, and the mutation test is what caught it.** My first version failed 13 lines that were all legitimate — override snippets, the cost-comparison table rows, "switch to a lighter model such as X". A guard that fails on valid examples gets deleted rather than fixed, so I narrowed it to lines that *claim* a default.

But a keyword-only rule (`/default|fallback/`) then **silently missed `REPO_ONBOARDING.md:123`** — the very file you flagged — because that row states the default in a table *cell* with no such word on the line. Mutating opus-4-8→sonnet-4-6 there still passed. Added a `^\s*\|\s*\`model_id\`\s*\|` row-shape clause, which covers both that table and USER_GUIDE's.

Mutation-verified after the fix:
```
USER_GUIDE.md:224 → us.anthropic.claude-sonnet-4-6
REPO_ONBOARDING.md:123 → us.anthropic.claude-sonnet-4-6
Tests: 1 failed, 4 passed
```
Both were entirely unguarded before this change. Suite is 5/5 green with them restored.

docs/scripts/sync-starlight.mjs
scottschreckengaust: **Ready to resolve.** Verifiable on `fafacdbb`:

```
$ grep -n 'COST_ATTRIBUTION:' docs/scripts/sync-starlight.mjs
56:    COST_ATTRIBUTION: '/getting-started/cost-attribution',

$ grep -rl 'architecture/cost-attribution' docs/src/content/docs/
(no matches — was 3 files)

$ ls docs/src/content/docs/architecture/Cost-attribution.md
No such file        # confirms the old target never existed
```

`Model-configuration.md:90` now emits `/getting-started/cost-attribution`, which matches the copy-list destination at `sync-starlight.mjs:260`.

Fix generalized beyond the reported instance — the route entry also repaired **two pre-existing 404s** of the same class (`architecture/Cost-model.md`, `getting-started/Deployment-guide.md`) and one relative `../guides/COST_ATTRIBUTION.md` that was never rewritten at all. That is why the diff touches two mirrors outside the new section; `//docs:sync` regenerated them.

`//docs:build` green (78 pages). Net: 3 broken `cost-attribution` links → 0.

docs/guides/DEVELOPER_GUIDE.md
scottschreckengaust: **Ready to resolve.** Your number was right — `ANTHROPIC_DEFAULT_HAIKU_MODEL` is at **458** on head, not 393.

Fixed by removing the line number rather than re-pinning it, per your suggestion. As shipped on `fafacdbb`:

```
cdk/src/stacks/agent.ts` (the runtime environment block)
```

Applied to **every** occurrence you listed — the layer table (line 127), the env-var table (line 136), `agent/README.md:126`, and both generated mirrors (via `//docs:sync`, not hand-edited).

Verification that none survive anywhere, including generated output:

```
$ grep -rn 'agent.ts:393\|pipeline.py:1780\|blueprint.ts:85' docs/ agent/README.md
(no matches)
```

I also adopted the second half of your point — exact `file:line` is now kept **only** where rot would be caught. Audited the four that remain and all are correct at head: `config.py:563` (ANTHROPIC_MODEL fallback), `config.py:569` (haiku fallback) — both read by the drift test — plus `repo-config.ts:37` (`readonly model_id?`) and `ecs-strategy.ts:217` (the `ANTHROPIC_MODEL` injection).

docs/guides/DEVELOPER_GUIDE.md
scottschreckengaust: **Ready to resolve.** Both confirmed at head and both fixed on `fafacdbb`:

| Cited | Actual at head | Now reads |
|---|---|---|
| `pipeline.py:1780` | **1811** (`"model_id": "anthropic_model"`) | `agent/src/pipeline.py`, `_PAYLOAD_KEY_ALIASES` |
| `blueprint.ts:85` | **86** (`readonly modelId?`) | `cdk/src/constructs/blueprint.ts`, `BlueprintProps.agent.modelId` |

Did the full sweep you asked for across every cited line number on the rebased head. **The three you flagged were the only stale ones** — the remaining four (`config.py:563`, `config.py:569`, `repo-config.ts:37`, `ecs-strategy.ts:217`) each verified by reading the line.

Identifier-based citation means these cannot rot on the next rebase, which was the underlying concern rather than the specific offsets.

cdk/test/contracts/model-default-docs-parity.test.ts
scottschreckengaust: **Ready to resolve — implemented now, not deferred.** You flagged it non-blocking, but your reasoning was that leaving it lets the four unguarded files rot on the next bump *while CI forces the two guarded ones green* — which is the precise failure this test exists to prevent. Cheaper to close while the pattern was fresh.

Took your blunt-assertion option over a second row helper, since it survives re-wording, moving a value into prose, or adding a doc. New case: **"no guarded doc presents a stale model id AS the default"**, sweeping 10 files — the 4 hand-authored sources plus the 6 generated mirrors that actually carry a prefixed literal (enumerated from the tree; `using/Overview.md` mirrors USER_GUIDE but carries none, so listing it would have implied coverage while asserting nothing).

**Two iterations, and the second only surfaced through mutation testing** — worth recording since it is a trap for the next person:

1. The naive sweep failed **13 legitimate lines** — per-repo override snippets, the cost-comparison rows, "switch to a lighter model such as X". A guard that fails on valid examples gets deleted rather than fixed, so I scoped it to lines that *claim* a default.
2. A keyword-only rule (`/default|fallback/i`) then **silently missed `REPO_ONBOARDING.md:123`** — the very file you named. That row states the default in a table **cell**, with no such word anywhere on the line. Mutating opus-4-8→sonnet-4-6 there still passed. Added a `^\s*\|\s*\`model_id\`\s*\|` row-shape clause, covering both that table and USER_GUIDE's.

Mutation-verified after the fix (both files were entirely unguarded before):

```
USER_GUIDE.md:224 → us.anthropic.claude-sonnet-4-6
REPO_ONBOARDING.md:123 → us.anthropic.claude-sonnet-4-6
Tests: 1 failed, 4 passed
```

Restored → **5/5 green**; `test/contracts/` + `bedrock-models.test.ts` = 22/22. CI 8/8 on `fafacdbb`.

scottschreckengaust:
## Review round 2 — all four threads addressed, ready to resolve

@theagenticguy — both blocking defects fixed, plus the non-blocking suggestion implemented. Head `fafacdbb`, **CI 8/8 green**, rebased onto latest `main` (`behind=0`, includes #754). Each inline thread has a reply carrying its own verification command output.

| # | Thread | Your finding | Status |
|---|---|---|---|
| 1 | `sync-starlight.mjs` | new mirror page ships a 404 link | ✅ **Fixed** — `COST_ATTRIBUTION` route added; also repaired 2 pre-existing 404s of the same class + 1 unrewritten relative link (3 → 0) |
| 2 | `DEVELOPER_GUIDE.md:127` | `agent.ts:393` is really 458 | ✅ **Fixed** — cites `agent.ts` (the runtime environment block); all 4 occurrences incl. mirrors |
| 3 | `DEVELOPER_GUIDE.md:129` | `pipeline.py:1780`→1811, `blueprint.ts:85`→86 | ✅ **Fixed** — identifier-based; full sweep done, your 3 were the only stale ones |
| 4 | parity test | 4 files quote the default unguarded | ✅ **Implemented** (was non-blocking) — mutation-verified on both previously-unguarded files |

### Single verification block

```
$ grep -rn 'agent.ts:393\|pipeline.py:1780\|blueprint.ts:85' docs/ agent/README.md
(no matches — including generated mirrors)

$ grep -rl 'architecture/cost-attribution' docs/src/content/docs/
(no matches — was 3 files)

$ grep -n 'COST_ATTRIBUTION:' docs/scripts/sync-starlight.mjs
56:    COST_ATTRIBUTION: '/getting-started/cost-attribution',

$ npx jest test/contracts/ test/constructs/bedrock-models.test.ts
Tests: 22 passed, 22 total
```

### The one finding worth your attention

On #4, a keyword-only rule (`/default|fallback/i`) **silently missed `REPO_ONBOARDING.md:123`** — the exact file you flagged — because that row states the default in a table *cell* with no such word on the line. Mutating it still passed. Caught only by mutation testing; fixed with a `model_id` row-shape clause. Both previously-unguarded files now fail on drift:

```
USER_GUIDE.md:224 → us.anthropic.claude-sonnet-4-6
REPO_ONBOARDING.md:123 → us.anthropic.claude-sonnet-4-6
Tests: 1 failed, 4 passed
```

### Still your call — not something I should decide

The `sync-starlight.mjs` route addition: `docs/AGENTS.md:35` sanctions the file for "adding new mirror rules", but **line 63 lists editing its mirror rules under "⚠️ Ask first"**. You noted it is "mechanically necessary and correctly scoped" — flagging that it is formally a maintainer decision rather than assuming your note constitutes approval.

Deliberately **not** fixed (out of scope, noted in the PR body): the pre-existing `developer-guide/introduction#repository-onboarding` break. It is an *anchor*-route problem in a different guide, not a stem-route one.

Re-requesting review.

laithalsaadoon (approved):
All four findings from the previous review are fixed on `fafacdbb`, verified independently rather than from the reply text — all four threads resolved.

1. **Mirror 404 link**: `COST_ATTRIBUTION: '/getting-started/cost-attribution'` is in `explicitGuideRoutes`, and a scan of every file under `docs/src/content/docs/` at head finds zero remaining `architecture/cost-attribution` links (was 3). The two extra mirror pages in the fix commit (`Cost-model.md`, `Deployment-guide.md`) are the regenerated output of the corrected rule — same pre-existing bug class, correctly swept along.

2. **Stale pointers**: the volatile citations now use file + identifier (`agent.ts` runtime environment block, `pipeline.py` `_PAYLOAD_KEY_ALIASES`, `blueprint.ts` `BlueprintProps.agent.modelId`) in all four affected files. Spot-checked the remaining exact `file:line` cites at head: `bedrock-models.ts:34`/`:48`/`:67`/`:84`, `config.py:563`/`:569`, `models.py:157`, `repo-config.ts:37`, `ecs-strategy.ts:217`, `submit.ts:69` — all correct.

3. **Sweep-test coverage**: re-implemented the new literal-sweep assertion independently against the head tree — it passes on head, and mutation-checking confirms the previously-unguarded `model_id` rows in `USER_GUIDE.md:224` and `REPO_ONBOARDING.md:123` are now in scope (the `` `model_id` `` row alternation catches them even though neither line contains the word "default"). The scoping rationale (claims-only, prefixed-ids-only, enumerated guarded set) is sound and documented in the test itself.

CI is green 8/8 on the new head. Ship it.

laithalsaadoon:
Post-approval note, non-blocking: the merge of **#754** (Opus 5 grant) added lines to `cdk/src/constructs/bedrock-models.ts`, so three of the exact line cites in the five-layer table and the bare-vs-prefixed section are now stale on the rebased head:

| Cited | Actual at head `fafacdbb` |
|---|---|
| `bedrock-models.ts:48` (context key) | **57** |
| `bedrock-models.ts:67` (resolver) | **76** |
| `bedrock-models.ts:84` (prefix rejection) | **93** |

`:34` (`DEFAULT_BEDROCK_MODEL_IDS`) is still correct. Affected spots: `DEVELOPER_GUIDE.md:125` and `:152`, plus the same two rows in the generated `Model-configuration.md` mirror.

This is the identifier-over-line-number case again — `bedrock-models.ts` looked stable but is exactly the file every model bump touches. Suggest the same treatment as the other volatile pointers (`BEDROCK_MODELS_CONTEXT_KEY`, `resolveBedrockModelIds`, "the prefix-rejection check in `resolveBedrockModelIds`") rather than re-pinning. Fine as a fixup on this PR or a fast-follow; the approval stands.

PR #775, opened 2026-08-21
feat(cost): add org and team monthly budgets with alerting

### LINKED ISSUE #471: feat(cost): org and team monthly budgets with alerting
**Context:** ROADMAP.md → Org and team budgets

---

## Component

API or orchestration

## Describe the feature

**Per-user and per-team** monthly token or USD budgets with alerting (e.g. 80%) and optional **hard stop** at 100%. Per-task `max_budget_usd` and turn caps ship today.

## Use case

Fleet operators need fleet-level caps—not just per-task limits—to prevent runaway spend across many concurrent tasks.

## Proposed solution

1. Budget store: team_id → monthly limit, current spend rollup.
2. Admission check at `createTaskCore` against user + team budgets.
3. CloudWatch alarms at 80%/100%; optional SNS.
4. CLI: `bgagent budget status|set` (operator).
5. Integrate with #215 attribution for accurate rollups.

## Other information

- Related to **Multi-user and teams** (#426).
- Design context: `docs/design/COST_MODEL.md`.

ayushtr-aws:
Adds recurring monthly USD guardrails for Cognito users and groups, with admission enforcement, threshold alerts, operator controls, and read-only personal visibility.

#### Area

#### Related

Closes #471

#### Changes

- Add recurring user and Cognito-group monthly budget configuration and UTC-month spend rollups.
- Check personal and applicable team hard-stop budgets at the shared task-admission boundary, returning `429 BUDGET_EXCEEDED` without interrupting running work.
- Roll terminal task `cost_usd` into user/team scopes idempotently through the existing TaskTable stream reconciler.
- Emit one-shot 80% and 100% CloudWatch threshold metrics and notify the shared operational SNS topic.
- Add operator `bgagent budget status|set` commands and authenticated, read-only `bgagent budget status --me` personal visibility.
- Document organization-group setup, onboarding overhead, enforcement behavior, estimated AWS operating cost, and the distinction from authoritative AWS billing.
- Keep the change scoped to CDK, CLI, and documentation; no agent runtime or agent-registry changes are included.

#### Validation

- CDK lint and TypeScript compilation
- CDK: 203 suites, 4,117 tests passed
- CLI lint and TypeScript compilation
- CLI: 58 suites, 769 tests passed
- Documentation Astro check, link check, and 78-page static build
- Drift-prevention and whitespace checks
- Deployed smoke test: personal status started at `$0.00`; a `$0.0573` task stopped by its per-task limit and a `$0.0270` completed task both rolled up, producing `$0.08` monthly spend

`mise` was unavailable in the implementation shell, so the package-native equivalents were used.

#### Acknowledgment


ayushtr-aws:
Addressed the self-review in `a41a06f`.

- **B1 fixed:** orchestration reconciliation and budget rollup are now attempted independently for every TaskTable stream record. A failure in either still requests an isolated retry, but cannot prevent the other subsystem from progressing.
- **N1 fixed:** added coverage for a single rollup crossing both 80% and 100%.
- **N2 fixed:** threshold metrics now emit before the conditional claim. This removes the permanent alert-loss window; rare concurrent/crash retries may emit a harmless duplicate, which is documented.
- **N3 fixed:** operator status now rejects corrupt non-positive configured limits consistently with admission and personal status.
- **N4 retained intentionally:** duplicate deliveries still reload state and retry alert work. Returning early after the task dedupe marker would permanently lose an alert when spend committed but alert delivery failed. A code comment and recovery test now make this explicit.
- **N5 deferred:** reliably recording the modifying AWS principal would require a new STS dependency/request and is better handled as a dedicated auditability enhancement.
- **N6 addressed:** native Cognito array claims preserve exact names; docs recommend avoiding whitespace/commas because API Gateway's string fallback treats them as delimiters.
- **N7 fixed:** added admission and rollup coverage for the 99-scope transaction boundary.

The failed `build (agentcore)` check was unrelated to agent registry. AgentCore built successfully; the mutation guard caught a stale root `yarn.lock` after the new Cognito SDK dependency. This commit includes the regenerated lockfile, verified with `yarn install --frozen-lockfile`.

Validation on the final source changes:
- CDK: 203 suites, 4,124 tests
- CLI: 58 suites, 770 tests
- CDK/CLI lint and TypeScript compilation
- Docs check/build: 78 pages
- Docs link check and all drift-prevention checks

isadeks:
Heads up before review: **this branch is conflicting with `main`**, so it can't merge as-is. Given the size (57 files, +3573/-94) I'd rather review the post-rebase diff than one that's going to shift — a rebase can meaningfully change what a cost/budgeting change touches.

Worth noting your CI is fully green (8/8) at the current head, so this is purely a merge-base issue, not a code problem.

Ping me once it's rebased and I'll do a full review pass.

scottschreckengaust (dismissed):
## 1. Verdict

**Request changes** — the design is right and the implementation is unusually careful (idempotent stream rollup, one-shot alert claims, emit-before-claim, fail-closed admission, real failure-path tests), but the branch is `CONFLICTING` against `main`, three doc comments describe the *opposite* execution order from the code they document, the new Cognito `AdminListGroupsForUser` call sits unconditionally in the task-admission hot path of five ingest channels, and the new user-facing docs link resolves to the wrong published page.

*Review method disclosure:* this was a **static review**. I did not run `mise run build`, jest, pytest, `tsc`, or `cdk synth` — this environment has no `node_modules/` or `agent/.venv/`. Every claim below is derived from reading the code at `a41a06f`. The one thing I did execute is the docs mirror check (see §5), in a throwaway copy, because `docs/scripts/sync-starlight.mjs` needs only `node:fs`/`node:path`.

**Governance:** clean. Issue [#471](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/471) is OPEN and carries `approved`; the branch name `feat/471-org-team-budgets` matches the ADR-003 convention; the delivered scope maps 1:1 onto the issue's five acceptance criteria (budget store, admission in `createTaskCore`, 80/100% CloudWatch alarms + SNS, `bgagent budget status`, `bgagent budget set`). Two nits: the issue has no priority label (`P0`/`P1`) and the PR has no labels at all.

## 2. Vision alignment

Squarely on the **bounded blast radius & cost** tenet (`docs/design/VISION.md`). Until now the only cost bound was per-task (`max_budget_usd`); a fleet operator had no recurring, human-scoped ceiling. This adds one at the right layer — admission, not mid-flight termination — which also preserves **fire-and-forget**: nothing here turns a task into an interactive session, and the refusal is a synchronous `429 BUDGET_EXCEEDED` at submit time rather than a surprise mid-run kill.

Two deliberate trade-offs, both of which I think are correct but only one of which is stated clearly enough:

- **Alert-only by default; hard stop opt-in per scope.** This matches #471's "optional hard stop" wording, so it is not an undocumented tenet trade. Worth stating explicitly in `COST_ATTRIBUTION.md` that the *default* posture is observe-only, so an operator does not read "budgets" as "enforcement".
- **In-flight work is never terminated at 100%.** `DEVELOPER_GUIDE.md:206` says so plainly ("without terminating in-flight work"). Good — that is the honest framing, and it means a single scope can overshoot its limit by the cost of whatever is already running.

**Reviewable outcomes:** the EMF metric + alarm + `bgagent budget status` triad keeps the control inspectable. The one gap is attribution of the *control itself*: see N-5.

## 3. Blocking issues

### B-1. Branch is `CONFLICTING` with `main` — cannot merge, and the green CI is against a stale base

`gh pr view 775` reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. Merge-base is `2cee8800`; `main` has moved on since (including today's `4da0a1f` browserslist re-resolve). 17 files are touched by *both* this PR and `main`-since-merge-base, and they are not incidental:

```
cdk/src/stacks/agent.ts
cdk/src/handlers/shared/types.ts
cli/src/types.ts
cdk/package.json
yarn.lock
cli/README.md
+ 5 generated docs mirror files, + 6 others
```

All 8 checks are SUCCESS, but they ran against a base that no longer exists — a re-run post-rebase is not a formality here, because `agent.ts` and both `types.ts` files are exactly the kind of overlap that produces a semantically-clean-but-wrong merge. This also confirms @isadeks' request on the PR thread. **Fix:** merge/rebase `origin/main`, then per AGENTS.md run `mise //cdk:eslint` + `mise //cli:eslint`, commit any autofixes ("Fail build on mutation" rejects uncommitted lint output), re-run `mise //docs:sync`, then `mise run build`.

### B-2. `cdk/src/handlers/shared/budgets.ts:281` — unconditional Cognito `AdminListGroupsForUser` on every headless task creation, fail-closed, even when zero budgets are configured

```ts
const teamIds = suppliedTeamIds === undefined
  ? (budgetTableName ? await resolveTeamIds(userId) : [])   // ← line 281
  : [...new Set(suppliedTeamIds)].sort();
```

Only `create-task.ts:54` supplies `teamIds` (from the verified `cognito:groups` JWT claim — good, see §7). Every other creator passes `undefined`: `webhook-create-task.ts:150`, `slack-command-processor.ts:256`, `jira-webhook-processor.ts:770,986`, `linear-webhook-processor.ts` (5 call sites), and `orchestration-release.ts:502` / `orchestration-reconciler.ts:1394` — i.e. **the DAG child-release path**. Each of those now makes a Cognito control-plane call before it can create a task.

The gate is `budgetTableName`, and `cdk/src/stacks/agent.ts:127` provisions `BudgetTable` unconditionally and wires `BUDGET_TABLE_NAME` into all of them — so this is not opt-in. A deployment that has never run `bgagent budget set` pays the call on every single headless submission, and `resolveTeamIds` (lines 124-144) has no `try`/`catch`: any `AdminListGroupsForUser` failure (throttle, `UserNotFoundException` for a mapping row whose Cognito user was deleted, a transient 5xx) propagates to `create-task-core.ts`, which fail-closes to `503 SERVICE_UNAVAILABLE`. Team resolution is ordered *before* any budget row is read, so the call happens even when the answer cannot possibly change the outcome.

I want to be precise about the blast radius: `orchestration-release.ts` treats 5xx as transient, so a DAG release retries rather than strands, and the IAM grants are all correctly scoped (§7). This is not data loss. But it is a new hard cross-service dependency in the admission path of every non-Cognito ingest channel, added for a feature that is inert until an operator configures it, and it is the kind of coupling that is very hard to remove once six call sites depend on it.

**Fix (pick one):**
- Resolve group membership *lazily*: read the `USER#` config/spend rows first, and only call Cognito when at least one `TEAM#` `CONFIG` row exists (the `record_type-scope_key-index` GSI you already build makes "are any team budgets configured?" a single cheap query, cacheable in module scope across warm invocations); or
- Keep the call but scope the fail-closed behaviour to the case where enforcement is actually possible — catch the Cognito error, and if the caller's own scope has no hard-stop config and no team budgets exist, log and continue with `teamIds: []` rather than refusing the task; or
- If the unconditional call is intentional, say so in `docs/design/` (or a short ADR note) with the reasoning — an undocumented widening of the admission path's failure surface is the thing Stage 2 of our review bar treats as blocking.

### B-3. `cdk/src/handlers/orchestration-reconciler.ts:23-24` — the docstring states the reverse of what the code does, and the test enshrines the code

```
 * Consumes the **TaskTable DynamoDB stream**. For each terminal task it first
 * applies an idempotent user/team monthly cost rollup. For graph tasks it then:
```

The code (lines 1499-1520) does the opposite: the orchestration `try` block (`parseTerminalTaskRecord` → `cascadeRestack` / `reconcileTerminalChild`) runs **first**, and `await rollupTaskCost(record)` runs **second**. `cdk/src/constructs/orchestration-reconciler.ts:56-58` repeats the same inversion ("It rolls up every positive task cost into monthly user/team budgets, **then** drives parent/sub-issue orchestration"), and `cdk/test/handlers/orchestration-reconciler.test.ts:340` locks in the actual order by name: `'releases orchestration dependents before reporting a budget-rollup retry'`.

Two reasons this is blocking rather than a nit:

1. **It is a behavioural claim, not prose.** Within one stream batch, a parent task's terminal record releases its unblocked children *before* the parent's own `cost_usd` has been added to the monthly rollup. So each released child is admitted against pre-rollup spend, and a hard-stopped scope can overshoot by one wave of children. That is a defensible design (it is also the ordering that keeps a rollup failure from stranding a DAG), but the two docstrings promise the safer ordering and the code delivers the looser one — a future reader trusting the comment will reason wrongly about the enforcement boundary.
2. **The next person "fixes" the wrong side.** With three artefacts disagreeing (two comments vs. code vs. a test name), whoever notices will pick one at random.

**Fix:** decide and make all four agree. Either swap the two `try` blocks (they are independent — each captures its own error and `orchestrationError ?? budgetError` already prefers the orchestration failure, so swapping changes only the admission snapshot the children see), or keep the current order and rewrite both docstrings to say "orchestration first, then rollup" and note the one-wave overshoot explicitly.

### B-4. `docs/guides/COST_ATTRIBUTION.md:26` and `docs/guides/DEVELOPER_GUIDE.md:206` — the new cross-reference resolves to the wrong published page

Both files link `[Monthly user and team budgets](./USER_GUIDE.md#monthly-user-and-team-budgets)`. In-repo that resolves fine, which is why CI is green. On the published site it does not:

- `docs/scripts/sync-starlight.mjs:72-84` (`userGuideAnchorRoutes`) has no entry for `monthly-user-and-team-budgets`, so the link falls through to `explicitGuideRoutes.USER_GUIDE` → `/using/overview#monthly-user-and-team-budgets` (verified in the committed mirror at `docs/src/content/docs/getting-started/Cost-attribution.md:30` and `docs/src/content/docs/developer-guide/Model-configuration.md:92`).
- But `USER_GUIDE.md:272` adds the section as `### Monthly user and team budgets` nested under `## Per-repo overrides` (line 216), and `splitGuide` splits on `##` — so the content renders at `/customizing/per-repo-overrides#monthly-user-and-team-budgets` (confirmed: `Monthly user and team budgets` appears in `src/content/docs/customizing/Per-repo-overrides.md`, not in `using/Overview.md`).

Net effect: both links land readers on the Overview page at an anchor that does not exist there — the two most likely entry points to the new feature's documentation dead-end. CI cannot catch this: `docs/scripts/link-check.sh:7` scans only `guides/`, `design/`, `decisions/` and root `*.md`, never the generated mirror, and the relative source link is genuinely valid. `sync-starlight.mjs` already carries a comment describing this exact 404 mode for `COST_ATTRIBUTION`, so the trap is known.

**Fix (one line):** add `'monthly-user-and-team-budgets': '/customizing/per-repo-overrides#monthly-user-and-team-budgets',` to `userGuideAnchorRoutes` and re-run `mise //docs:sync`. Alternatively — and better for a feature this prominent — promote the section to a top-level `##` in `USER_GUIDE.md` so it gets its own page, then map it.

**N-1. `cdk/src/handlers/shared/budgets.ts:115` and `cli/src/budget-store.ts:67` — private `numeric()` fails open on garbage.** Both return `0` for `NaN`, `Infinity`, objects, and unparseable strings. A corrupted `spend_usd` therefore reads as "no spend" and admits work, silently, with no log line. The repo already has the right helper: `cdk/src/handlers/shared/numeric.ts` `coerceNumericOrNull` (used at `get-task-replay.ts:190` and `fanout-task-events.ts:842,847`), which logs the coercion. Its docstring says a third call site is the signal to widen the type — this PR adds two. Note the asymmetry with `loadBudgetStates`/`loadPersonalBudgetStatus`, which *do* fail closed by throwing on `monthly_limit_usd <= 0`; the limit side is strict and the spend side is lenient.

**N-2. Duplicated cross-package contracts with no parity test.** `USER#`/`TEAM#`/`CONFIG` (`budgets.ts:29-32` vs `budget-store.ts:28-32`), `ROLLUP_RETENTION_DAYS = 400` (`budget-rollup.ts:43` vs `budget-store.ts:34`), the index name (`budget-table.ts:24` exports `BUDGET_CONFIG_INDEX_NAME`, and `budget-store.ts:30` re-declares the same literal instead of importing it), the 100% hard-stop threshold, and — most fragile — the `alerted_80_at`/`alerted_100_at` marker names, which `budget-rollup.ts` claims with `attribute_not_exists(...)` and `budget-store.ts:141-142` `REMOVE`s to re-arm. Rename a marker on the CDK side and the CLI's re-arm becomes a silent no-op: `REMOVE` of a non-existent attribute succeeds, and alerts stay permanently suppressed after a limit raise. `contracts/constants.json` already exists for exactly this (it holds `max_budget_usd`); at minimum add a test asserting the two sets of literals match.

**N-3. `cdk/src/constructs/budget-table.ts:24` — `BUDGET_CONFIG_INDEX_NAME` is exported but never imported.** Dead export; the one consumer that needs it (the CLI) hardcodes the string. Either import it or drop the `export`.

**N-4. `cli/src/budget-store.ts:114,163,198` — bypasses `cli/src/dynamo-clients.ts` and builds a new client per call.** `documentClient(region)` (`dynamo-clients.ts:25`) is the established seam; three raw `makeDocClient({ region })` calls both diverge from it and forgo connection reuse across the three calls a single `budget status` makes. Attribution is correct either way (both go through `cli/src/ua.ts`), so this is style + coherence, not #319.

**N-5. `budget set` has no record of who changed the limit.** `setMonthlyBudget` writes `updated_at` but not an `updated_by`. Since the CLI writes straight to DynamoDB with the operator's own credentials, the only trail is CloudTrail data events on the table, which are off by default. For a control that gates other people's work, `updated_by` is a cheap addition. (I see this is your deferred N-5 from the self-review — agreed it is not a merge blocker, but it is the one deferral I would file as a follow-up issue rather than leave implicit.)

**N-6. `cdk/src/handlers/list-tasks.ts:55` — `GET /tasks?view=budget` overloads a collection endpoint with a singleton resource, and unknown `view` values are silently ignored.** A caller who typos `view=budgets` gets a task list with a 200, not a 400. A sibling route (`GET /budget`, or `/users/me/budget`) would be more honest; failing that, validate the `view` enum and 400 on anything unrecognized. Authorization is correct — `loadPersonalBudgetStatus(userId)` uses the authorizer-derived id, so there is no IDOR here.

**N-7. `cdk/src/handlers/shared/budgets.ts:287-292` — the >99-scope guard surfaces as a misleading `503`.** A user in 99+ Cognito groups throws, and `create-task-core` maps the throw to `503 SERVICE_UNAVAILABLE` — "try again later" for a permanent configuration condition that retrying will never clear. Consider a distinct 4xx (or at least a distinguishable error code) so the operator learns the real cause.

**N-8. Alarm granularity.** `budget-alerts.ts` aggregates `BudgetThresholdCrossed` by `Threshold` only, so the alarm fires without naming the breaching scope. The docs correctly point the responder at the logs, and adding `scope_key` as an EMF dimension would be a cardinality footgun, so I think the design is right — worth one sentence in the runbook prose making the "alarm tells you *that*, logs tell you *who*" contract explicit.

**N-9. No `budget unset`.** `budget set` can only raise/lower a limit, never remove it (`--monthly-usd` is validated `>= MIN_MONTHLY_BUDGET_USD = 0.01`, and the loaders throw on `<= 0`, so "set it to zero to disable" is not available either). Deleting the `CONFIG` row currently requires the AWS console or CLI. Follow-up, not a blocker.

**N-10. `cdk/test/constructs/budget-table.test.ts:26,61` — `new App()` + `Template.fromStack()` per test (#366).** Only two tests and one small construct, so the cost is negligible, but the repo's stated pattern is one synth in `beforeAll`. Cheap to align now, before this file grows.

## 5. Documentation

Genuinely good, and better than the median for a feature this size. Reviewed: `docs/guides/COST_ATTRIBUTION.md` (new budget section, the cost-of-controls table, and — refreshingly — the caveat that rollups are estimates that will not tie out to the invoice), `docs/guides/USER_GUIDE.md:272` (operator walkthrough), `docs/guides/DEVELOPER_GUIDE.md:206` (admission-control framing, and the explicit "does not terminate in-flight work"), `cli/README.md` (both subcommands, flags, and output shapes). Env vars (`BUDGET_TABLE_NAME`, `USER_POOL_ID`) and the new `BudgetTableName` output are covered.

**Mirror sync: verified in sync.** I copied the worktree to a scratch directory, ran `node docs/scripts/sync-starlight.mjs` there, and `diff -r` against the committed `docs/src/content/docs/` produced no output. The "Fail build on mutation" step will not trip on the mirror. (This ran in a throwaway copy and touched neither the PR worktree nor any build tooling.)

**Missing:** only the B-4 anchor routing. No ADR is required in my view — this implements an existing tenet with existing primitives rather than making a new architectural choice — though if you take the B-2 "document the rationale" branch, that note is the place for it.

**Issue tracking:** #471 is `approved` and OPEN and accurately describes the delivered work. Add a priority label to it and some labels to this PR.

## 6. Tests & CI

**Bootstrap synth-coverage: not applicable — verified, not assumed.** I traced every resource type this PR introduces. `AWS::DynamoDB::Table` and `AWS::CloudWatch::Alarm` are both already present in `cdk/src/bootstrap/resource-action-map.ts`; `application.ts` already grants `dynamodb:CreateTable`/`UpdateTable`/`UpdateTimeToLive` on `arn:aws:dynamodb:*:*:table/backgroundagent-dev-*` (the new table's physical name matches that pattern), and `observability.ts` already grants `cloudwatch:PutMetricAlarm`. The new `cognito-idp:AdminListGroupsForUser` is a *runtime* task-role permission granted in-stack, not a deploy-time CFN action, so it does not touch the bootstrap bundle. **No `BOOTSTRAP_VERSION` bump and no artifact regeneration required** — correct as submitted. (Static reasoning; I could not run `mise //cdk:test -- test/bootstrap/synth-coverage`.)

**API type sync: OK.** `PersonalBudgetStatus` is byte-identical in `cdk/src/handlers/shared/types.ts` and `cli/src/types.ts`. `TaskRecord.team_ids?: readonly string[]` is added on the CDK side only, which is right — it is a persisted-record field, not part of the wire contract.

**Coverage is a strength.** The tests go after the paths that actually break in production, not the happy path: idempotent replay of an already-rolled-up task (transaction cancelled + marker exists), re-emission when the alert *claim* write fails after the metric was emitted, both thresholds crossing inside a single rollup, the 99-scope boundary, fail-closed `503` when admission throws, and `429` with no write. The emit-before-claim ordering in `budget-rollup.ts` is the right call and is the kind of thing that only gets built when someone has thought about a crash between the two writes — it has a comment explaining why, and a test proving it.

**Gaps I would like to see closed:** (a) a test pinning the reconciler ordering to whatever B-3 resolves to, asserted on *behaviour* (child admitted against post-rollup spend, or not) rather than encoded in a test name; (b) a `resolveTeamIds` failure test — today there is no coverage of what happens when Cognito throws, which is precisely the B-2 path; (c) a cross-package parity test for the N-2 literals.

**CI:** 8/8 SUCCESS, but see B-1 — that verdict is against a base `main` no longer resembles. `reviewDecision: REVIEW_REQUIRED`, no labels.

## 7. Review agents run

**Nested agent dispatch was unavailable in this run.** This review was executed as a subagent inside a batch fan-out, and that harness permits only one level of nesting, so I could not dispatch the `pr-review-toolkit` agents (`code-reviewer`, `silent-failure-hunter`, `type-design-analyzer`, `comment-analyzer`, `pr-test-analyzer`) that Stage 3 of our review bar names. I am not claiming to have invoked them. Instead I applied each rubric dimension myself, explicitly and one at a time:

- **code-reviewer (applied inline).** AGENTS.md routing is correct (API/Lambdas in `cdk/`, CLI in `cli/`, nothing misplaced into `agent/`); solution-UA attribution is correct everywhere (`makeClient`/`makeDocClient` throughout — no naked `new XxxClient({})`, so #319 holds); L2 constructs, `PAY_PER_REQUEST`, PITR, TTL, no hardcoded ARNs or account IDs. Findings: N-3, N-4, N-10.
- **silent-failure-hunter (applied inline).** Traced every catch and every default. Findings: N-1 (garbage → `0`, admits work, no log), N-6 (unknown `view` silently ignored), N-7 (permanent condition reported as transient), and it is what made B-2 visible (an uncaught Cognito throw converted to a `503` refusal). Credit where due: the admission path fails *closed*, `loadBudgetStates` throws on a non-positive limit, and the `UnprocessedKeys` loop in `batchGetItems` is handled rather than dropped.
- **type-design-analyzer (applied inline).** `PersonalBudgetStatus`, `BudgetState`, `BudgetAdmission`, `BudgetScope`, `TaskCostEvent` are all narrow, `readonly`, and discriminated where they need to be; `parseBudgetScopeKey` returns a proper union rather than string-sniffing at call sites. `null`-vs-`undefined` is used consistently ("configured but absent" vs "not configured"). No finding beyond N-2's stringly-typed duplicated literals.
- **comment-analyzer (applied inline).** This is where B-3 came from — two docstrings asserting the opposite execution order from the code. Everything else I spot-checked was accurate and several comments were load-bearing in a good way (the emit-before-claim rationale in `budget-rollup.ts`, the "reserve one action for the task marker" note on `MAX_BUDGET_SCOPES_PER_TASK = 99`, the stream-contention rationale on the reconciler construct, the 429-is-deterministic reasoning in `orchestration-release.ts`).
- **pr-test-analyzer (applied inline).** See §6 — strong failure-path coverage, three named gaps.
- **security-review (Skill invoked; harness diff was empty, so analysis performed inline).** I invoked the Skill because the diff adds IAM statements and touches the input gateway. Its harness resolved the diff against the wrong working tree and returned nothing, and its methodology requires parallel sub-tasks I cannot dispatch here, so I ran the analysis myself against the PR tree. **No HIGH or MEDIUM exploitable finding.** Specifically verified: (a) `extractUserGroups` (`gateway.ts:47`) reads only `authorizer.claims['cognito:groups']` — the authorizer-verified JWT — never a header or request body, so team attribution is not client-forgeable, and `POST /tasks` is Cognito-only (`task-api.ts:868`), not reachable through the API-key/webhook authorizer that lacks `claims`; (b) no IDOR on the read path — `list-tasks.ts:56` passes the authorizer-derived `userId`, and there is no `?user=` override; (c) **IAM least privilege is clean and consistent** — all six `cognito-idp:AdminListGroupsForUser` grants (`agent.ts:1211,1308`, `task-api.ts:1301`, `jira-integration.ts:329`, `slack-integration.ts:317`, `linear-integration.ts:328`) are scoped to `userPool.userPoolArn`, never `*`, and the budget table grants are correctly asymmetric: `grantReadData` for the six admission-only consumers, `grantReadWriteData` only for the stream reconciler that actually writes rollups (`constructs/orchestration-reconciler.ts:129`), with the *stranded* reconciler correctly read-only (`agent.ts:1304`) because it never calls `rollupTaskCost`; (d) no NoSQL-injection surface — all DynamoDB access uses parameterized `ExpressionAttributeValues`; (e) no secrets, no new network egress, no deserialization. The N-5 audit gap is a defence-in-depth item, not a vulnerability.
- **Genuinely out of scope, omitted:** Cedar/HITL policy review — the diff contains no Cedar policy or approval-gate change (budget admission is a hard-coded numeric check, deliberately not a Cedar decision). Agent-runtime review — nothing under `agent/` changed, and the runtime never reads the budget table. Cedar engine parity (`cedarpy` / `@cedar-policy/cedar-wasm`) — no pin moved.

## 8. Human heuristics

- **Proportionality — pass, with one note.** The complexity is earned: a two-key table with a GSI, one stream consumer, one admission check, two alarms, two CLI subcommands. No speculative abstraction, no "engine", no factory-for-one. Largest new file is 325 lines. The 100-action transaction limit driving `MAX_BUDGET_SCOPES_PER_TASK = 99` is a real constraint handled at the right place. The one disproportion is B-2: an unconditional cross-service call to support a feature that is inert until configured.
- **Coherence — concern (N-2, B-3).** Same concept, three homes: the scope-key/TTL/threshold/alert-marker contracts are re-declared in `cli/` rather than shared, while `contracts/constants.json` already exists for precisely this and already holds `max_budget_usd`. And "same concept, same story" fails in `orchestration-reconciler` where the comments and the code describe different orderings. Otherwise the vocabulary is consistent and matches the rest of the repo (`scope_key`, `period`, `hard_stop`, `utilization_percent`).
- **Clarity — concern (N-1, N-6, N-7).** Names are good and intent-revealing (`hard_stop_active` vs `hard_stop` is a genuinely useful distinction; `budgetPeriod`/`budgetResetAt` say what they do). The concerns are all error-surfacing: a plausible default (`0`) hiding corrupt data, a silently-ignored query parameter, and a permanent condition dressed as a transient one (AI004).
- **Appropriateness — pass, with one AI005 caution.** The DynamoDB semantics are used correctly against real API behaviour, not mock-shaped behaviour: `TransactWriteCommand` cancellation reasons are inspected properly via `isTransactionCanceled` + `markerExists`, `ADD` is used for the counters instead of read-modify-write, `UnprocessedKeys` is looped, `ConsistentRead` is set where correctness needs it, and TTL is set on rollup rows but *not* on `CONFIG` rows (a config that quietly expired after 400 days would have been a nasty bug — good catch on your part). The AI005 caution is `cdk/test/handlers/orchestration-reconciler.test.ts:340`, whose name asserts what the code *does* while the docstring says the code should do the opposite — the test is currently ratifying the behaviour rather than specifying it.

---

Overall: this is careful, well-tested work on the right problem, and I expect to approve it quickly. B-1 you have to do anyway; B-3 and B-4 are small; B-2 is the one that deserves a real answer rather than a patch — a lazy resolution or a documented rationale both work for me.

ayushtr-aws:
Addressed the review on the merged `main` base.

- Resolved all merge conflicts and merged `origin/main` in `17a23027`. GitHub now reports the PR as mergeable.
- Removed the unconditional Cognito dependency from headless admission: the path reads the user budget first, queries the config GSI for any team budget, and calls `AdminListGroupsForUser` only when team enforcement can apply. Cognito failures still fail closed when a team budget exists. Added success, skip, failure, and user-hard-stop coverage.
- Updated both reconciler docstrings to match the intentional orchestration-first ordering and documented the one-wave overshoot tradeoff.
- Added the Starlight anchor route so both budget cross-links resolve to the published section.
- Tightened corrupt spend handling to fail closed, reused one CLI DynamoDB client per operation, centralized the shared budget storage/threshold constants in `contracts/constants.json` with CLI parity coverage, rejected unknown `view` values, mapped the permanent 99-team limit to `400 VALIDATION_ERROR`, clarified alert-only defaults and alarm/log responsibilities, and aligned the budget-table test with the cached-synth pattern.
- `updated_by` audit attribution and `budget unset` remain follow-up enhancements, as discussed in the non-blocking notes.

Commits: `94ac3ecf`, `24a1fa16` (plus merge commit `17a23027`).

Validation on the final head:
- `mise run build`
- CDK: 215 suites, 4,511 tests
- CLI: 62 suites, 924 tests
- Agent: 1,782 tests
- Docs: 77 pages, link checks and mirror sync
- CDK/CLI lint, TypeScript compilation, synth, type/constants drift checks

ayushtr-aws:
Security PR follow-up fixed in `fc5f9af3`:

- upgraded Astro to 7.2.8
- pinned js-yaml 4.3.2+, sharp 0.35.4+, and svgo 4.1.0+
- regenerated yarn.lock

Validation passed locally (`mise run security:deps`, `mise run build`), and security-pr run 34307628013 is now green.

isadeks (changes requested):
## 1. Verdict

**Request changes** — this is careful, well-documented work on the right problem, and the previous round's four blockers (B-1 conflict, B-2 unconditional Cognito call, B-3 inverted docstrings, B-4 anchor route) plus seven of ten nits are genuinely fixed. But one of those fixes introduced a new defect worse than the one it repaired: the N-1 "don't fail open on garbage spend" change turned a benign missing-attribute read into a throw, and `bgagent budget set` writes exactly that missing attribute. The documented recommended setup (`budget set --team Everyone --hard-stop`) now deadlocks task creation for every member of that group.

Reviewed at `fc5f9af3` (8 commits, 63 files, +4213/-380). Governance is clean: [#471](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/471) is OPEN with `approved`, the branch name matches ADR-003, and `mergeable` is now `MERGEABLE`. This is a re-review of the whole change, not just the delta since `a41a06f`.

*Method disclosure:* I could not run `mise run build` or any jest suite — this worktree has no `node_modules` and I did not run `mise run install`. Every claim below is from reading the code at `fc5f9af3`, except B-1's throw, which I reproduced by executing the `budgetNumber` logic verbatim against the row shape `setMonthlyBudget` writes.

## 2. Vision alignment

Squarely on the **bounded blast radius and cost** tenet (`docs/design/VISION.md`) — the first fleet-scoped ceiling, enforced at admission rather than mid-flight, which preserves **fire-and-forget**. Both trade-offs (alerts-only default; in-flight work never terminated) are now stated plainly in `COST_MODEL.md` and `USER_GUIDE.md`, so there is no undocumented tenet trade and I do not think an ADR is required. B-3 is fully repaired: `orchestration-reconciler.ts:36-38` and `constructs/orchestration-reconciler.ts:53-57` now both say orchestration-first and name the one-wave overshoot explicitly.

## 3. Blocking issues

### B-1. `budget set` writes a counter-less spend row; readers throw

`cli/src/budget-store.ts:158-176` — the second transact item is an **unconditional `Update`** on `{scope_key, period: YYYY-MM}` whose only job is clearing the alert markers. `UpdateItem` upserts, so for a scope with no spend yet this month the transaction *creates* a period row holding `scope_key, period, scope_type, scope_id, updated_at, ttl` — and no `spend_usd`.

Every reader then takes the "row exists, so absence is corruption" branch:

```ts
budgetNumber(spend?.spend_usd, 'spend_usd', scopeKey,
             spend ? undefined : 0)
```

`spend` is truthy, so `absentValue` is `undefined`, so it throws. Three call sites: `cli/src/budget-store.ts:109-112`, `cdk/src/handlers/shared/budgets.ts:257-260` (`loadBudgetStates`), and `:290-293` (`loadPersonalBudgetStatus`). Running that expression verbatim against the row `setMonthlyBudget` writes gives `Budget row USER#abc has invalid spend_usd.`

Concrete chain from a clean install, following the new docs:

1. `bgagent budget set --team Everyone --monthly-usd 10000 --hard-stop` — the write lands, then `listBudgetStatus` at `cli/src/commands/budget.ts:258` throws *before* the success message prints. The operator sees a failure for a write that actually succeeded.
2. Any member submits a task: `checkBudgetAdmission` -> `loadBudgetStates` throws -> `create-task-core.ts:715-726` fail-closes -> **503 SERVICE_UNAVAILABLE for every member of that group**.
3. `bgagent budget status --me` -> 500.
4. It is self-deadlocking. `spend_usd` only materialises via `writeRollup`'s `ADD spend_usd :cost`, which needs a task to complete, which needs admission, which throws. Recovery requires a manual DynamoDB edit.

This is a regression introduced by the N-1 fix. At `a41a06f`, `numeric(undefined)` returned `0`, so the counter-less row read as `$0.00` — which is what the PR's deployed smoke test observed. Commit `94ac3ecf` correctly replaced fail-open with a throw but did not revisit the writer that depends on the old leniency.

Fix, preferably both halves: pass `0` unconditionally as `absentValue` and throw only on a value that is *present and unparseable*; and add `spend_usd = if_not_exists(spend_usd, :zero)` to the `Update` so the row is never counter-less. Then add the missing test — no suite covers "row exists without `spend_usd`". `cli/test/budget-store.test.ts:206` and `cdk/test/handlers/shared/budgets.test.ts:296` only cover a *non-numeric* value, and `cli/test/commands/budget.test.ts:34` mocks `budget-store` wholesale, so the `set` -> `read` sequence is never exercised end to end.

### B-2. `resolveTeamIds` has no `catch`; Cognito errors 503 forever

`cdk/src/handlers/shared/budgets.ts:161-181`. B-2 from the last round offered two remedies: gate the call lazily, *or* catch the Cognito error and continue with `teamIds: []` when enforcement is not possible. The lazy gate landed (`hasConfiguredTeamBudgets()` at `:372`) and is a real improvement — but the catch did not, and the gate is what makes the uncaught path reachable by design: the moment one team budget exists, every headless submission calls `AdminListGroupsForUser` with nothing wrapping it.

`Username` is the platform user id, i.e. a Cognito `sub` (`gateway.ts:37`), which resolves for local users only. So once any team budget is configured:

- a stale Slack/Linear/Jira user-mapping row pointing at a deleted Cognito user raises `UserNotFoundException` — a permanent 503 for that user, not a transient one;
- a federated or third-party-IdP pool 503s every federated user's headless submissions;
- `AdminList*` throttling under a wide DAG release wave (`orchestration-release.ts` calls this serially per child) becomes a 503, the child rolls back to `ready`, and the wave livelocks against the sweep — logged identically to a budget-table outage.

Fix: wrap the pagination loop; treat `UserNotFoundException` as "no memberships" with a single `logger.warn`; let genuine 5xx and throttles keep failing closed but under a distinguishable `metric_type`. No test covers a `resolveTeamIds` failure today.

**N-1. A persistent rollup failure silently under-counts spend.** `orchestration-reconciler.ts:1531` — `const error = orchestrationError ?? budgetError` reports the record as a batch item failure when *either* side failed. With `retryAttempts: 3` plus `SqsDlq` (`constructs/orchestration-reconciler.ts:147-161`), a budget-side fault outliving the retries DLQs the record, and a DynamoDB-stream DLQ carries metadata only, so that task's spend is lost. `constructs/orchestration-reconciler.ts:121` has no depth alarm (pre-existing, but this PR adds a new source feeding it) and there is no `budget_rollup_failure` metric, so the loss is silent. To be precise about blast radius: orchestration progress is *not* lost, since it committed on attempt 1 — but each retry needlessly re-drives `reconcileTerminalChild`. Your own `task-table.ts` comment notes the second stream-consumer slot is free; a separate event-source mapping would let each subsystem retry and DLQ on its own budget.

**N-2. `grantReadWriteData` on the budget table is an over-grant that can disable the control itself.** `constructs/orchestration-reconciler.ts:131`. `budget-rollup.ts` needs `GetItem`, `BatchGetItem`, `Query`, `PutItem`, and `UpdateItem`; the grant adds `DeleteItem` and `BatchWriteItem` over the whole table, and `period = CONFIG` rows share those partitions — so a bug in the reconciler can delete every hard-stop config. This is the only in-stack write path, so it is the entire tamper surface. An explicit five-action `PolicyStatement` on `tableArn` plus `tableArn/index/*` closes it.

**N-3. A budget rejection emits no log and no metric, and the one log it does emit is mislabeled.** `create-task-core.ts:729-742` returns 429 with no `logger` call at all, in contrast to the 503 path's `metric_type: 'budget_admission_failure'`. The only line written is `budgets.ts:351`, whose message is `'Monthly budget is at or above the warning threshold'` — emitted for a 100% hard-stop block. An operator asking "why can nobody submit tasks?" finds what reads as an 80% advisory. Add `metric_type: 'budget_admission_blocked'` at the block site with scope/spend/limit, and give it a message distinct from the warning sweep.

**N-4. Cancelled and killed tasks contribute $0, so the hard stop cannot trip on the runaway case.** `budget-rollup.ts:73` drops `costUsd === null || <= 0` with no log line, and cancel, the stranded-task reaper, and container OOM/kill never write `cost_usd` (`agent/src/task_state.py:266-268` sets it only when the result carries one). A workload that repeatedly crashes spends real money and stays at $0 forever. The field semantics are pre-existing, but the guardrail's usefulness rests on them — worth a follow-up issue rather than scope creep here.

**N-5. The dedup marker swallows a revised cost.** `budget-rollup.ts:181-187` logs `'Budget rollup already applied'` and drops the delta. Safe today because `write_terminal` writes status and cost in one `UpdateItem`, but nothing enforces that, and the marker row already stores `cost_usd` at `:153` — so comparing and warning on divergence is about one line.

**N-6. `budget set` silently disables an existing hard stop.** `budget-store.ts:141-157` `Put`s the CONFIG row and `--hard-stop` defaults to `false` (`commands/budget.ts:240`), so `bgagent budget set --user alice --monthly-usd 200` turns enforcement off. `USER_GUIDE.md` does document replace semantics and the confirmation line prints "alerts only", which is what keeps this off the blocking list — but with no `--no-hard-stop` and no `budget unset`, raising a limit requires remembering to re-pass the flag.

**N-7. The threshold/marker-name coupling survives the N-2 fix.** `budget-rollup.ts:230-239` derives `alerted_${threshold}_at` from `warning_percent`, while `budgets.ts:41-42` reads `warning_alert_marker` as an independent constant and `budget-store.ts:164-165` hardcodes `alerted_80_spend_usd` / `alerted_100_*`. `constants-parity.test.ts` gates CLI-versus-JSON drift but not this one: change `warning_percent` to 75 and the writer and reader de-sync, re-emitting the "one-shot" alert on every rollup forever. `budget-alerts.ts:53,63` hardcodes "80%"/"100%" in the alarm descriptions for the same reason. Asserting `warning_alert_marker === 'alerted_' + warning_percent + '_at'` is one line. Relatedly, a non-integer threshold would make `alerted_80.5_at` parse as a document path, guarding the wrong attribute.

**N-8. CLI and API disagree on corrupt spend.** `budget-store.ts:87-91` uses bare `Number(value)`, so `spend_usd: ''` reads as a reassuring `$0.00`, while the CDK side routes the same value through `coerceNumericOrNull` and throws. Same row, two verdicts, and the CLI's is the comforting one. Both sides also `Math.max(0, ...)` a negative rather than surfacing it.

**N-9. `[]` versus `undefined` conflation on the API path.** `create-task.ts:54` always passes `extractUserGroups(event)`, and `budgets.ts:346` only consults Cognito on `undefined`. Latent today — `POST /tasks` is Cognito-only, and the fact that `userInGroup` works in production confirms the claim arrives as a plain delimited string — but moving the route behind the unified Cognito-or-API-key authorizer (`api-key-authorizer.ts:142`, which sets no `claims`) would silently disable team enforcement and leave team spend at $0 rather than visibly broken. Pass `undefined` when the claim key is missing.

**N-10. The `USER_POOL_ID` coupling lives in the stack, not the construct.** `constructs/orchestration-reconciler.ts:94-100` sets `BUDGET_TABLE_NAME` from the prop while `stacks/agent.ts:1519-1523` patches `USER_POOL_ID` and the IAM grant separately. I verified this is correctly wired as it stands — but every other construct sets both together (`linear-integration.ts:251-254` and siblings), so taking `userPool` on `OrchestrationReconcilerProps` would make it unforgettable for the next caller.

**N-11. `writeRollup`'s catch can report the wrong error.** `budget-rollup.ts:182` — if `markerExists` itself fails after a cancellation, the thrown error is the read failure, hiding the benign idempotent-replay cause. Inspecting `CancellationReasons` on the `TransactionCanceledException` avoids the second read.

**N-12. The 429 body discloses a team's exact spend and limit to any member** (`create-task-core.ts:735-740`). Defensible, but worth confirming as a deliberate product decision.

## 5. Documentation

Genuinely strong — the cost-of-controls breakdown, the "estimate, not the invoice" caveat, the group-name delimiter warning, and the explicit "does not terminate in-flight work" framing are all well above the bar for a feature this size. `API_CONTRACT.md`, `COST_MODEL.md`, `COST_ATTRIBUTION.md`, `OBSERVABILITY.md`, `USER_GUIDE.md`, `contracts/constants.md`, and `cli/README.md` are all updated, and the env vars plus the new `BudgetTableName` output are covered.

**Mirror: in sync.** The generated pages are in the diff, and the previous round's independent `sync-starlight.mjs` run against the committed mirror produced no diff, so "Fail build on mutation" will not trip.

**One concern — information architecture.** B-4 was fixed with the one-line anchor route at `sync-starlight.mjs:76`, but that route points at `/customizing/per-repo-overrides#monthly-user-and-team-budgets`, and the section really does render there: `USER_GUIDE.md:282` adds `### Monthly user and team budgets` under `## Per-repo overrides`, and `splitGuide` splits on `##`. So a fleet-wide, per-user and per-team control is published under *Customizing -> Per-repo overrides*, which is simply the wrong place. The previous review named the better fix — promote it to a top-level `##` so it gets its own page — and I still think that is the right call for a feature this prominent.

**Issue tracking:** #471 accurately describes the delivered work but still has no priority label, and this PR still has no labels at all.

## 6. Tests and CI

**Bootstrap synth-coverage: not applicable — verified, not assumed.** This PR introduces no new CFN types. `AWS::DynamoDB::Table -> dynamodb:CreateTable` (`resource-action-map.ts:79`) and `AWS::CloudWatch::Alarm -> cloudwatch:PutMetricAlarm` (`:73`) are both already mapped, and `synth-coverage.test.ts` keys coverage on the *set* of types in the synthesized template, so this PR cannot regress it. `application.ts:36-56` already grants `CreateTable`, `UpdateTable`, `DescribeTable`, `UpdateTimeToLive`, `Describe`/`UpdateContinuousBackups`, and `TagResource` on `arn:aws:dynamodb:*:*:table/backgroundagent-dev-*`, which the auto-generated table name matches, and the inline GSI needs no separate ARN; `observability.ts:73-80` covers the alarms. `cognito-idp:AdminListGroupsForUser` is a runtime role permission, not a deploy-time CFN action. **No `BOOTSTRAP_VERSION` bump and no artifact regeneration required — correct as submitted.**

**API type sync: OK.** `PersonalBudgetStatus` is identical in `cdk/src/handlers/shared/types.ts` and `cli/src/types.ts`, and `TaskRecord.team_ids` is correctly CDK-only since it is a persisted-record field rather than part of the wire contract. `constants-parity.test.ts` is a genuine drift gate and closes the old N-2.

**Coverage is a real strength where it exists** — idempotent replay via the marker, alert re-emission after a failed claim, both thresholds crossing in one rollup, the 99-scope boundary, fail-closed 503, and 429-with-no-write. Gaps I would like closed:

- B-1's branch (spend row present without `spend_usd`) and B-2's branch (`resolveTeamIds` throwing) — the two blockers, both untested.
- `budget-rollup.ts:182-186` `TransactionCanceledException` with *no* marker, the case that separates idempotent replay from a real conflict with lost spend. `budget-rollup.test.ts:292` throws a plain `Error` only, and `ConditionalCheckFailedException` appears in no test.
- Every negative filter in `parseTaskCostEvent` (`:73`). Deleting `!TERMINAL.has(status)` or `costUsd <= 0` fails nothing, yet that guard is the only thing stopping a mid-run `MODIFY` from latching the marker with a partial cost.
- `printStatus`'s table body never renders: `listBudgetStatusMock` is `mockResolvedValue([])` at `cli/test/commands/budget.test.ts:85` and never overridden, so the operator table, the money formatting, and the `ACTIVE`/`enabled`/`disabled` tri-state are all unverified.
- `grep AdminListGroupsForUser cdk/test` returns nothing — all six IAM grants are unasserted.
- The pagination loops are entirely dead in tests: `LastEvaluatedKey` and `UnprocessedKeys` in both packages, zero grep hits. A partial `BatchGet` reads spend as `0`, so admission fails *open* there.

**Tautological tests.** `cli/test/commands/budget.test.ts:274-276` asserts the parsed output `toEqual` the mock's own return value, so it passes for any payload and pins no field. Three more mirror `UpdateExpression` strings back at the source (`budget-rollup.test.ts:129`, `budget-store.test.ts:77`, `budgets.test.ts:90`) while leaving the behavioural facts unasserted — `:cost === 8.5`, and `:ttl` roughly `now + 400d`; a wrong `:cost` passes. `agent.test.ts:48` is a `21 -> 22` change-detector, which is the pre-existing pattern in that file.

**Test performance (#366).** No test re-enables `aws:cdk:bundling-stacks` — clean. `budget-table.test.ts:24-41` correctly synthesizes both configurations once in `beforeAll`. `budget-alerts.test.ts:24-28` does not: `new App()` plus `Template.fromStack()` inside the `test()` body, which is the shape `cdk/AGENTS.md:75-84` gives as the explicit bad example. One synth today, so the cost is negligible, but it is worth hoisting before the file grows. `task-api.test.ts:77-95` adds a third full `TaskApi` synth (6 Lambdas plus API Gateway) for two assertions that could ride the existing `createStackWithWebhooksAndApiKeys()` stack.

**CI:** 8 checks green with `build (agentcore)` pending, and `mergeable` is `MERGEABLE`. Worth noting that the green suites are exactly why B-1 would ship: the `set` -> `read` seam is mocked away on both sides.

## 7. Review agents run

- **silent-failure-hunter** — ran. Independently surfaced B-1 and the observability gaps behind N-3, N-4, N-5, N-8, and N-11. Two of its three "blocking" calls I downgraded after verifying them myself: the DLQ finding overstates the loss (orchestration commits on attempt 1, so only spend is lost — hence N-1 rather than a blocker), and its bracket-serialization concern is refuted by `userInGroup` working in production today.
- **pr-test-analyzer** — ran. Source of the coverage, tautology, and #366 findings. Its lead finding (reconciler missing `USER_POOL_ID` and the IAM grant) is a **false positive** — it read only the construct, while `stacks/agent.ts:1519-1523` supplies both. I kept the underlying gap that those grants are unasserted anywhere in `cdk/test`.
- **security-review** — ran. Independently surfaced B-1 as HIGH and is the source of N-2, N-9, N-12, and the bootstrap verification above. It confirmed no IDOR on `?view=budget` (`list-tasks.ts:56` passes the authorizer-derived id and there is no `?user=` override), all six `AdminListGroupsForUser` grants scoped to `userPoolArn` rather than `*`, no injection surface (scope ids and group names only ever appear as `Key` attributes or bound `ExpressionAttributeValues`), and that `budget set` enforcement is genuine IAM on the table rather than an advisory client-side check.
- **code-reviewer, type-design-analyzer, comment-analyzer** — the `pr-review-toolkit` plugin is not installed in this environment, so I am not claiming to have invoked them; I applied those three rubrics inline instead. Routing is correct per AGENTS.md; solution-UA attribution holds everywhere via `makeClient`/`makeDocClient` with no naked clients, so #319 is intact; L2 constructs, `PAY_PER_REQUEST`, PITR, and TTL on rollup rows but deliberately not on CONFIG rows. Types are narrow and `readonly`, and `null` versus `undefined` carries real meaning ("configured but absent" versus "not configured"). On comments: B-3's inversion is repaired and the load-bearing ones are accurate — the emit-before-claim rationale, the "reserve one action for the task marker" note, the stream-contention rationale, the 429-is-deterministic reasoning. The one stale claim left is the mislabeled block log in N-3.
- **Omitted as genuinely out of scope:** Cedar/HITL policy review, since the diff contains no policy or approval-gate change and budget admission is deliberately a numeric check rather than a Cedar decision; Cedar engine parity, since no pin moved; and agent-runtime review, since nothing under `agent/` changed and the runtime never reads the budget table.

## 8. Human heuristics

- **Proportionality — pass.** The complexity is earned: a two-key table with one GSI, one stream consumer, one admission check, two alarms, two CLI subcommands. No speculative abstraction, no factory-for-one. `MAX_BUDGET_SCOPES_PER_TASK = 99` handles a real DynamoDB transaction limit at the right layer. B-2's lazy gate removed the last disproportion from the previous round.
- **Coherence — concern.** The old N-2 is properly fixed via `contracts/constants.json` plus a parity test, and the reconciler docstrings now agree with the code. What remains is N-7 (the marker name derived one way and read another, outside the parity gate) and the docs IA — a fleet-wide control published under *Per-repo overrides*.
- **Clarity — concern.** `hard_stop_active` versus `hard_stop` is a genuinely useful distinction and the naming is intent-revealing throughout. Error surfacing is still the weak axis: a 429 with no log, a hard-stop block logged as an 80% warning, a dropped cost with no line, and — B-1 — the same "absent" input meaning `0` in one package and a fatal throw in the other.
- **Appropriateness — concern.** DynamoDB is used against real API semantics rather than mock-shaped ones: `ADD` for counters instead of read-modify-write, `UnprocessedKeys` looped, cancellation inspected properly, `ConsistentRead` where correctness needs it, TTL deliberately absent from CONFIG rows (a config quietly expiring after 400 days would have been nasty — good catch). The concern is the test seam: `cli/test/commands/budget.test.ts:34` mocks the entire budget store, so the `set` -> `read` round-trip, the very first sequence an operator performs, is never exercised. That is exactly the mock-too-high pattern that let B-1 through a green suite.

---

Overall: the previous round's blockers are properly addressed and this is close. B-1 is the one that has to be fixed before merge, and it is worth pausing on *how* it arose — a correct fix to a fail-open read became a fail-closed deadlock because the writer that produced the tolerated input was not re-checked. B-2 is the other half of the earlier B-2 recommendation. Everything else can land as follow-ups.

ayushtr-aws:
Addressed the new blockers in `010f5209`, with the adjacent admission-observability issue included because it is part of diagnosing these failure modes.

- **B-1 fixed at both boundaries.** A missing `spend_usd` attribute now means the legitimate zero state whether the monthly row is absent or was materialized by `budget set`. `budget set` also initializes the counter with `if_not_exists(..., 0)`, so new rows are canonical. Present-but-empty, non-numeric, non-finite, or negative values still fail closed in both the API and CLI; the fix does not restore the old corrupt-data-as-zero behavior.
- Added direct coverage for the actual deadlock sequence: admission and personal/operator status reading a materialized current-month row without `spend_usd`, plus the writer expression that creates it.
- **B-2 fixed with a narrow error policy.** `UserNotFoundException` is observable and resolves to no team memberships, avoiding a permanent retry loop for a stale/unresolvable headless mapping. Throttling, service faults, permissions, and configuration errors still fail closed and now emit `budget_team_membership_resolution_failure`, distinct from a budget-table outage. Cognito pagination is covered as part of the same path.
- Added `budget_admission_blocked` logging at the 429 decision point with request, scope, period, spend, and limit. This replaces reliance on the misleading warning-threshold line for hard-stop diagnosis.
- I did **not** fold in the separate stream-consumer/IAM/docs-IA/CLI-semantics suggestions. Those are broader design changes and are better evaluated independently rather than increasing the blast radius of these two fixes.

Validation:
- CDK lint and TypeScript compile
- CLI lint and TypeScript compile
- CDK: 215 suites, 4,518 tests
- CLI: 62 suites, 928 tests
- Agent: 1,782 tests
- Docs: 77 pages; link and drift checks passed

`mise run build` completed those checks but the account-aware CDK synth was blocked locally by expired AWS credentials (`ExpiredToken` while validating AgentCore AZs). The pushed CI run is queued against the new head.

isadeks (changes requested):
## 1. Verdict

**Request changes** — three blockers, down from six in my first draft. CI is red on the blocking masking gate, and the hard stop is evaluated after the Bedrock spend it exists to bound. Everything else can land as follow-ups.

Two corrections to my own prior review (round 2, 2026-09-01 and the 15:44 UTC pass today) before anything else, because one of them is mine to own:

- **B-1 from the last round is properly fixed, both halves.** `cli/src/budget-store.ts:172` now writes `spend_usd = if_not_exists(spend_usd, :zero)`, and `budgets.ts:165` passes `0` unconditionally as `absentValue`. The counter-less-row deadlock is gone. Good fix.
- **B1 below is a defect my own prior review asked for.** Round-2 B-2 said, verbatim, "treat `UserNotFoundException` as 'no memberships' with a single `logger.warn`". That is exactly what `010f5209` implements. I am not asking you to revert it — I am asking for the part the earlier review should have specified: make the decision explicit and alarmed. Apologies for the round trip.
- **N-7 and N-9 stay non-blocking.** My first draft of this review escalated both to blockers with no new evidence. The earlier reasoning was better — in particular N-9 is latent, not live, because `userInGroup` working in production confirms the claim arrives.

## 2. Vision alignment

Advances **bounded blast radius & cost** — fleet-level caps were the acknowledged gap next to per-task `max_budget_usd`. Preserves **fire-and-forget** (admission-time rejection; in-flight work is never interrupted) and keeps outcomes **reviewable**. The "one dependency wave of overshoot" trade is disclosed in both `orchestration-reconciler.ts:53-58` and `COST_MODEL.md` — that is a documented bound, not a tenet traded away, and does not need an ADR.

The one genuine tension is B2.

## 3. Blocking issues

### B1. The `UserNotFoundException` swallow needs to be an explicit, alarmed decision — and CI says so

`cdk/src/handlers/shared/budgets.ts:219`

`resolveTeamIds` is only reached when `hasConfiguredTeamBudgets()` already returned `true` (`:405-407`) — the code *knows* a `TEAM#` config exists. On `UserNotFoundException` it returns `[]`, so admission evaluates only the `USER#` scope and the team hard stop is skipped for that submission. The gap is self-reinforcing: the task record then stores no `team_ids`, so `budget-rollup.ts:77` attributes none of that spend to the team, the team's `spend_usd` stops growing, and the 80/100% alarms never fire either.

To be clear about the trade, because the previous round got this wrong in the other direction: throwing here converts a data-hygiene problem (a stale headless user-mapping row pointing at a deleted Cognito identity) into a permanent 503 for that user, and 503s every federated user on a third-party-IdP pool. Swallowing keeps availability and loses enforcement. Neither default is obviously right, which is precisely why the repo has a gate that forces the choice to be written down.

What is not negotiable is that the unenforced scope is currently invisible and CI is failing: `ts-silent-success-masking` blocks on this line ([run 34385475444](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/actions/runs/34385475444/job/102580529263)).

**Fix:** keep the swallow, and add (a) the inline `// nosemgrep: ts-silent-success-masking -- <why callers may safely treat an unresolvable membership as no teams>` justification, and (b) an EMF metric plus alarm on `budget_team_membership_user_missing`, so a scope that has silently stopped being enforced is a page rather than a `warn` nobody reads. If you would rather fail closed, that also clears the gate — but then the stale-mapping and federated-pool cases need naming in the docs as accepted outages.

### B2. The hard stop runs after the Bedrock spend it exists to bound

`cdk/src/handlers/shared/create-task-core.ts:690` (step 3b)

Budget admission sits at step 3b, *after* description guardrail screening (step 2, line 397) and after attachment screening plus S3 uploads (step 2b, lines 429-525). A user at 100% of a hard-stop budget therefore still incurs Bedrock Guardrail invocations and per-image Bedrock screening on **every rejected submission**. A looping client generates unbounded Bedrock spend that the guardrail reports as successfully blocked — the control bounds task count, not cost, which is the tenet it is here to serve.

The stated rationale for the placement ("runs AFTER idempotency replay so retrying an already-created task remains a 200") only constrains the *final* decision, not a cheap pre-check.

**Fix:** move the idempotency check (step 3) and budget admission ahead of step 2; or add an early hard-stop pre-check that short-circuits before any Bedrock call when no `Idempotency-Key` is present.

### B3. The sole idempotency guard is unasserted, so a double-count would ship green

`cdk/test/handlers/budget-rollup.test.ts:125-131`

The assertions are structural: `TransactItems.length === 3`, the marker `scope_key`, and the literal `#ttl = :ttl` string. Nothing asserts `ConditionExpression: 'attribute_not_exists(scope_key)'` (`budget-rollup.ts:157`) — the only thing preventing double-counted spend — nor `ADD spend_usd :cost` with the actual cost value, nor each `Update`'s key and period. Delete the condition, halve the cost, or write the wrong period and the suite stays green.

The rollup logic itself is correct — conditional marker `Put` in the same `TransactWrite` as the `ADD` deltas, `TransactionCanceledException` only treated as dedup after a `ConsistentRead` marker confirmation, otherwise re-thrown. That is exactly why it is worth pinning: it is right today and nothing would tell you if it stopped being.

Also untested and cheap to add: exactly-100% blocking (every current test uses 120% or 101.25%, so `>=` vs `>` is unpinned), the `alerted_100_at` branch, and a genuine `TransactionCanceledException` rethrowing.

1. **Split the dependency changes out.** `package.json:23-36` adds `sharp`/`svgo` resolutions and bumps `js-yaml`/`astro`, with ~714 lines of `yarn.lock` churn — unrelated to budgets, and it couples a feature revert to a security revert. Per `AGENTS.md`, any pin here must be mirrored into `integrations/jira-forge-app/package.json` `overrides` or `check:transitive-pin-sync` drifts. The new entries also break the block's alphabetical order.
2. **N-7 (carried, still open): the threshold/marker-name coupling.** `budget-rollup.ts:230-239` derives `alerted_${threshold}_at` from `warning_percent`, while `budgets.ts:41-42` reads `warning_alert_marker` as an independent constant. Change `warning_percent` to 75 and the writer and reader de-sync, re-emitting the "one-shot" alert on every rollup forever. `constants-parity.test.ts` gates CLI-vs-JSON drift but not this. Asserting `warning_alert_marker === 'alerted_' + warning_percent + '_at'` is one line.
3. **N-9 (carried, still open): `[]` vs `undefined` on the API path.** `create-task.ts:54` always passes `extractUserGroups(event)`, and `budgets.ts:388` only consults Cognito on `undefined`. Latent today, but moving the route behind the unified Cognito-or-API-key authorizer would silently disable team enforcement and report `$0.00` rather than break visibly. Return `undefined` when the claim key is absent. Related: `gateway.ts:50` splits on commas, and Cognito `GroupName` permits punctuation — `COST_ATTRIBUTION.md` documents "avoid commas in group names" as operator prose; for a value that becomes a DynamoDB key that belongs in `bgagent budget set --team` validation.
4. **`GET /v1/tasks?view=budget` overloads a collection endpoint with a different response type** (`list-tasks.ts:55-66`). The only `view=` dispatch in the repo — every other feature gets its own resource. It also forces a budget-table grant onto `ListTasksFn`. Published contracts are one-way doors, so worth settling now: `GET /v1/budget` with its own small Lambda. (No IDOR — the id is authorizer-derived and there is no `?user=` override.)
5. **A CDK construct imports a runtime module with load-time side effects.** `budget-alerts.ts:26-29` pulls the percentages from `handlers/shared/budgets`, whose top level runs `makeDocClient()` and may construct a Cognito client (`budgets.ts:50-53`), so every `cdk synth` and construct test instantiates SDK clients. `budget-table.ts:23` in this same PR already does it right by reading `contracts/constants.json`.
6. **The budget wiring block is copy-pasted five times** across `task-api.ts`, `jira-integration.ts`, `linear-integration.ts`, `slack-integration.ts`, and `stacks/agent.ts`. One `grantBudgetAdmission(fn, budgetTable, userPool)` helper would stop "granted the table but forgot the Cognito statement" from being a one-line outage.
7. **`budget-rollup.ts` sits in the Lambda-entrypoint directory but exports no `handler`** — it is imported by `orchestration-reconciler.ts:48`. Move to `handlers/shared/`.
8. **`cli/src/budget-store.ts` re-implements the enforcement math** rather than mirroring it — `budgetNumber`/`budgetSpend`/`currentBudgetPeriod`/`toStatus` duplicate `budgets.ts`. Two implementations of the same math means the operator view and the admission gate can disagree after a one-sided edit.
9. **A non-numeric `cost_usd` silently drops a task's whole spend.** `budget-rollup.ts:56-62` returns `null` with no log or metric, indistinguishable from a zero-cost task. `cost_usd` is persisted as a string by the agent, and the sibling code in this same PR does log via `coerceNumericOrNull`.
10. **Unbounded `UnprocessedKeys` retry, no cap or backoff** (`budgets.ts:253-268`, `budget-store.ts:235-252`). Under table throttling this spins to Lambda timeout, so the caller gets a 502/504 instead of the documented 503, with no `budget_admission_failure` metric.
11. **Budget-only rollup failures land in the reconciler DLQ with nothing to replay them** — spend is simply absent, so hard stops under-count. `writeRollup` is marker-idempotent so a redrive is safe; document that and emit a distinct metric so `dlqDepthAlarm` is actionable.
12. **`budget set` unconditionally re-arms the one-shot alerts** (`budget-store.ts:170-174`), so flipping `--hard-stop` on a scope already at 150% re-fires both alarms. Clear the markers only when the limit actually changes.
13. **The 429 body discloses team aggregate spend and limit** to any member. The figures are already in the structured log.
14. **Reuse the existing EMF helper.** `budget-rollup.ts:194-206` inlines the `_aws` envelope; `shared/approval-metrics.ts:169-186` has a tested `buildEmfLine` whose comment says "revisit if a second EMF-emitting Lambda joins the stack" — this is that Lambda. `'ABCA/Budgets'` is also duplicated as a literal in the handler and the construct.
15. **`contracts/constants.json`:** `config_period` and `config_record_type` are two keys with one value (`"CONFIG"`) — change one and the GSI stops matching the sort key. And `MIN_MONTHLY_BUDGET_USD = 0.01` (`commands/budget.ts:48`) is a magic value outside the contract with **no upper bound** — `--monthly-usd 1000000` is accepted silently.
16. **Coverage gaps beyond B3:** `parseTaskCostEvent` rejection paths and UTC month rollover; the >100-key multi-batch path (98 teams + user = 198 keys, two `BatchGet`s) where a fencepost would hide; corrupt `monthly_limit_usd` (one bad row 503s every task for that scope); the >=80% admission warning log; and `createTaskFn`'s budget env/grant plus `webhookCreateTaskFn`'s `AdminListGroupsForUser` statement, all unasserted — dropping that last one 503s every headless create the moment a team budget exists.

## 5. Documentation

The strongest part of this PR. `API_CONTRACT.md`, `COST_MODEL.md`, `OBSERVABILITY.md`, `COST_ATTRIBUTION.md` (setup steps, onboarding overhead, a ~$0.80/month cost estimate), `USER_GUIDE.md`, `DEPLOYMENT_GUIDE.md`, `contracts/constants.md`, `cli/README.md`. Candid about the estimate-vs-invoice distinction and the overshoot bound — exactly right.

- **Mirror sync: clean.** Re-ran `node docs/scripts/sync-starlight.mjs`; `git status` came back empty. No CI mutation failure.
- **Placement (carried):** the "Monthly user and team budgets" section renders into `customizing/Per-repo-overrides.md` — a fleet-wide control on a per-repo page, with `sync-starlight.mjs:76` anchoring a top-level concept into that page's fragment. It belongs in the Cost-attribution page this PR already edits.
- **Small gap:** `GET /v1/tasks` errors are still documented as `400`/`401`, but `view=budget` can now surface a 500 from a corrupt budget row.
- **Governance:** #471 is `approved`, the branch name conforms, and the delivered scope matches the issue.

## 6. Tests and CI

**CI: 1 failing, 7 passing** — the semgrep masking gate (B1), a real finding rather than flake. Note the visible check set has no `cdk:test`/`cli:test`/`docs` job, so the PR body's suite numbers are not corroborated by CI here.

**Bootstrap synth-coverage: verified pass, no updates required.** Ran rather than reasoned about: `cd cdk && npx jest test/bootstrap` gives **6 suites, 119 tests passed**, covering `synth-coverage`, `golden-baseline`, `artifact-sync`, `bootstrap-template`, `policies`, `version`. Zero net-new CFN resource types: `AWS::DynamoDB::Table` (+ inline GSI) is already at `resource-action-map.ts:79` and granted in `policies/application.ts:38-55`; `AWS::CloudWatch::Alarm` at `resource-action-map.ts:73`. No new SNS resource — the alarms join the existing `OperationalAlerts` list at `stacks/agent.ts:1979-1982`. No new event-source mapping — the rollup piggybacks on the existing `OrchestrationReconciler` mapping, and `task-table.ts` already had `NEW_IMAGE` (that diff is comment-only). On ARN narrowness: the table is created without a `tableName`, so the generated name matches `application.ts:55`'s `table/backgroundagent-dev-*`, the GSI ARN matches since IAM `*` spans `/`, and `UpdateTimeToLive`/`UpdateContinuousBackups` are both present for the TTL/PITR settings. **`BOOTSTRAP_VERSION` correctly left at 1.6.0** — bumping would force operators to re-bootstrap a byte-identical bundle.

One caveat worth recording: `synth-coverage.test.ts:66-89` asserts only type-to-action coverage, never synthesized physical names against ARN patterns, so a green suite does not by itself answer the ARN question. Pre-existing harness gap, not this PR's.

**Test-perf rule (#366): respected.** No new test re-enables `aws:cdk:bundling-stacks`. Two minor misses: `budget-alerts.test.ts:25-29` builds `new App()` + `Template.fromStack()` in the test body instead of `beforeAll`, and `task-api.test.ts:186` adds a fourth full-stack synth where `budgetTable` could fold into the existing `createStack()`.

**Seam quality:** ~1,700 lines of new tests, but every DynamoDB/Cognito mock is hand-rolled and sequenced by *call order*, and `list-tasks.test.ts:26-30` stubs out `BatchGetCommand` itself, so an invalid `RequestItems` shape cannot be caught. `aws-sdk-client-mock` matching on command class plus input would remove the ordering coupling and is what would have caught B3-class regressions.

## 7. Human heuristics

- **Proportionality — pass.** Complexity is earned: a two-key table with one GSI, one stream consumer, one admission check, two alarms, two CLI subcommands. No speculative abstraction. `MAX_BUDGET_SCOPES_PER_TASK = 99` handles a real DynamoDB transaction limit at the right layer. The five-way copy-pasted wiring block (N-6) is the one accreted spot.
- **Coherence — concern.** Terminology holds across code, docs, and CLI ("scope", "period", "hard stop"). What remains is N-7, N-4's `view=` dispatch, a construct importing a runtime module (N-5), a second hand-rolled EMF blob next to a helper that explicitly anticipated it (N-14), and the docs IA.
- **Clarity — concern.** `hard_stop_active` vs `hard_stop` is a genuinely useful distinction and naming is intent-revealing. Error surfacing is still the weak axis: B1's unenforced scope, a dropped `cost_usd` with no line (N-9), and a marker name derived one way and read another.
- **Appropriateness — concern.** DynamoDB is used against real API semantics rather than mock-shaped ones: `ADD` for counters instead of read-modify-write, `UnprocessedKeys` looped, cancellation inspected properly, `ConsistentRead` where correctness needs it, TTL deliberately absent from CONFIG rows. The concern is verification: the deployed smoke test covered only the personal scope, so the team path — the actual subject of #471 — is validated solely by order-sensitive self-written mocks.

---

Overall: the model and the docs are the strong parts, and the previous round's blockers are genuinely fixed. Three things before merge — clear the masking gate with an explicit justification plus an alarm (B1), move the hard stop ahead of the Bedrock spend it bounds (B2), and assert the idempotency condition that makes the rollup correct (B3). The rest are follow-ups, and the dependency bumps want their own PR.

ayushtr-aws:
Addressed the three blocking findings in `8958890a`:

- **B1:** kept the availability-preserving `UserNotFoundException` behavior, added the required `nosemgrep` rationale, emitted `BudgetTeamMembershipUnresolved` as EMF, and wired a namespace-level CloudWatch alarm into `OperationalAlerts`. The construct now reads thresholds from `contracts/constants.json` instead of importing the runtime handler.
- **B2:** moved idempotency replay and budget admission ahead of Bedrock description screening and inline attachment screening/S3 upload. Blocked and unavailable-budget tests now assert that Bedrock is never invoked; replay still bypasses budget admission and screening.
- **B3:** pinned the rollup marker condition, marker fields, each scope key/period, and the exact cost/count deltas. Added a genuine canceled-transaction-without-marker rethrow test. Also pinned hard-stop behavior at exactly 100%; the existing dual-threshold test already covers the `alerted_100_at` branch.

Validation:
- `SEMGREP_MASKING_BASELINE=origin/main mise run security:sast:masking:range`
- `MISE_EXPERIMENTAL=1 mise //cdk:test` (215 suites, 4,520 tests)
- `mise run build`
- `git diff --check`

I intentionally left the remaining explicitly non-blocking suggestions out of this pass.

ayushtr-aws:
Fixed the AgentCore build race in `e4935af2`.

The failure was not a production pre-install AWS access regression. The parametrized handler test mocked `run_task`, but still launched `_spawn_background`; under CI load, the first case could return before its thread dereferenced the mock. Pytest then restored `run_task`, and that escaped thread ran the real pipeline (`t-silent` clone in the logs) while the next case’s AWS seam guard was armed. Its heartbeat/terminal paths produced the two `tenant_resource` violations.

The test class now stubs `_spawn_background` itself. These tests cover only the `/run` pre-install handler phase and should never launch pipeline work.

Validation:
- failing parametrized test: 25 consecutive runs passed
- `tests/test_server.py`: 232 passed
- `mise //agent:quality`: 1,782 passed; lint, format, and typecheck passed
- root `mise run build`: AgentCore tests passed; local synth later stopped only on the existing expired AWS credential check for AgentCore AZ discovery
- `git diff --check`

ayushtr-aws:
Corrected the documentation mismatch in `7737a5bc` after rebasing onto the latest branch/main merge:

- updated all cost-control counts to 3 alarms and up to 3 metric series
- updated list-rate arithmetic to $0.30 alarms + up to $0.90 metrics = approximately $1.20/month before free tier
- documented `BudgetTeamMembershipUnresolved` and its remediation runbook in observability
- corrected the `view` API contract: omit it for task listings; only explicit `budget` is accepted
- regenerated all affected Starlight mirrors

Validation: `mise //docs:sync`, `mise //docs:build`, and `git diff --check`.

isadeks (approved):

PR #779, opened 2026-08-24
feat(registry): migrate to standalone Agent Registry

### LINKED ISSUE #771: registry: migrate substrate from AgentCore registry to standalone Agent Registry service
## Summary
The #246 agent asset registry is currently backed by the **AgentCore registry** control plane (`bedrock-agentcore-control`). We're migrating to a **new standalone "Agent Registry" service** with its own endpoints. This tracks that substrate swap.

## Background
The registry integration was deliberately built behind a port/adapter boundary so a substrate swap is confined to the adapter layer. Every consumer (handlers, the orchestrator resolve-step, the agent's read path) talks through the `RegistryClient` port, never a raw SDK client.

## Scope — what changes
New adapter(s) targeting the Agent Registry service endpoints, plus config/wiring:
- `cdk/src/handlers/shared/registry/agentcore-client.ts` → new Agent Registry adapter (full port: `publish` / `getRecord` / `listRecords` / `resolve`)
- `cdk/src/handlers/shared/registry/factory.ts` — build the new client; replace `AGENT_REGISTRY_ID` / `bedrock-agentcore-control` wiring with the new service's endpoint/config
- `agent/src/registry/agentcore_client.py` → new read-side adapter (`get_record` / `resolve`) against the new endpoints
- Re-express the record shape (`descriptorType` CUSTOM/AGENT_SKILLS/MCP, `_meta` runtime extraction, name decode `kind/namespace/name`) against the new service's record model
- IAM: replace AgentCore control-plane permissions with the new service's actions (provisioning + orchestrator/session roles)
- `cdk/src/handlers/registry-provisioning/index.ts` — provisioning against the new service
- New ADR (or update to ADR-022) documenting the migration decision and endpoint model

## Out of scope
- The port interfaces themselves (`client.ts` / `client.py`) should not need to change — that's the point of the boundary. Flag if the new service's semantics force a port change.

## Open questions
- New service endpoint / API shape and auth model?
- Data migration for existing published records, or fresh publish?
- Cutover strategy — dual-write/dual-read, or hard swap?

ayushtr-aws:
## Summary

- migrate the TypeScript and Python registry adapters from the Bedrock AgentCore preview namespace to standalone AWS Agent Registry
- update provisioning, IAM, descriptor mappings, dependencies, and bootstrap resource coverage for `agent-registry`
- add a default-on `enableAgentRegistry` context gate for accounts or regions that cannot use the service
- document the fresh-registry cutover, opt-out behavior, and deletion implications

## Deployment behavior

Agent Registry remains enabled by default. Operators can omit the complete registry surface with:

```bash
cdk deploy --context enableAgentRegistry=false
```

Disabling it on an existing deployment removes the CloudFormation-managed registry and its records. Blueprints that retain `registry://` references fail closed.

## Validation

- [x] `mise run build`
- [x] `MISE_EXPERIMENTAL=1 mise //cdk:test` (200 suites, 4,109 tests)
- [x] `MISE_EXPERIMENTAL=1 mise //docs:build`
- [x] standalone `agent-registry-control` availability verified in `us-east-1`
- [x] live deployment validation on `backgroundagent-dev2`

Live validation reached `UPDATE_COMPLETE`, provisioned a `READY` standalone registry, verified authenticated publish/list/show/resolve behavior, verified the orchestrator registry environment, and removed the temporary smoke-test record. The deployment also exposed and fixed the required service-linked-role and workload-identity lifecycle permissions.

The review follow-up was validated separately with a disposable registry: deletion was initiated directly by the deployed `isCompleteFn` while the registry was `READY`, CloudTrail recorded `DeleteRegistry` under the waiter Lambda role with status `DELETING`, the waiter completed, and the registry was confirmed absent. The managed stack registry remained `READY` and unchanged.

`mise run security` reaches the repository's existing SAST masking baseline of 22 unrelated findings; the renamed registry files add no new masking finding.

Closes #771


ayushtr-aws:
Live deployment validation completed on `backgroundagent-dev2` in `us-east-1`:

- CloudFormation deployment: `UPDATE_COMPLETE`
- standalone registry `bQiZy44FM4uFbTHj`: `READY`
- registry API without Cognito authentication: `401 Unauthorized`
- Task Orchestrator Lambda: active and wired to the deployed registry ID
- publish smoke test: `201`, native SKILL record reached `APPROVED`
- list, show, and exact-version resolve smoke tests: `200` with expected payloads
- temporary smoke-test record deleted; registry returned to empty
- disposable budget table and alarms removed during stack cleanup

The live deploy surfaced two provisioning prerequisites, both fixed on this branch: scoped service-linked-role creation (`065e261`) and workload-identity lifecycle permissions (`e519c90`).

ayushtr-aws:
> **Self-review** (Principal Solutions Architect lens, per `.abca/commands/review_pr.md`) — re-review of the delta since the initial commit.

## Verdict

**Approve.** The three follow-up commits resolve every nit from the first pass, fix the one Medium robustness item with a well-reasoned design and thorough tests, and complete the previously-missing live-deployment validation — which productively surfaced two real IAM gaps this PR now closes.

## What changed and how it maps to the prior review

| Prior finding | Resolution | Status |
|---|---|---|
| **Medium** — teardown had no retry path; a transient/`Conflict` on the single `DeleteRegistry` was fatal | New `requestRegistryDeletion()` returns `started`/`absent`/`retryable`; `onEvent` Delete defers `Conflict`/`Throttling`/`InternalServer` to the waiter, and `isComplete` re-drives the idempotent delete until the service reports `DELETING`/gone (`index.ts:71-107, 162-199`) | Fixed |
| **Low** — rollback assumed `ResourceNotFoundException` on a CFN-token id | Explicit decision the other way, documented at `index.ts:165-167`: the Provider wrapper consumes its own `CREATE_FAILED` marker, so a `ValidationException` here is a real defect and is surfaced, not swallowed | Resolved (by decision) |
| **Nit** — stale `// GA-throwaway anyway` | Removed (`registry.ts:184`) | Fixed |
| **Nit** — "the GA API" vs Preview | → "the standalone API" (`registry.ts:149-150`) | Fixed |
| **Nit** — gate token case-sensitivity undocumented | Documented in REGISTRY.md §5.1 and DEPLOYMENT_GUIDE.md (+ both Starlight mirrors) | Fixed |
| **Appropriateness** — live-deploy validation unchecked | Now checked: `backgroundagent-dev2` reached `UPDATE_COMPLETE`, `READY` registry, verified publish/list/show/resolve + orchestrator env, smoke record cleaned up | Resolved |

## New change — workload-identity + service-linked-role grants

Live deploy re-surfaced the exact failure the original `main` comment warned about ("Unable to create workload identity because access was denied"). Two commits add, **on `onEventFn` only** (`registry.ts:110-134`):

- `iam:CreateServiceLinkedRole` on `*` **conditioned** on `iam:AWSServiceName == agent-registry.amazonaws.com` — the AWS-recommended least-privilege SLR pattern (resource must be `*`; the condition is the scope).
- `bedrock-agentcore:{Create,Get,Delete}WorkloadIdentity` scoped to `bedrock-agentcore:workload-identity-directory/*` (not `*`) — `CreateRegistry` provisions the workload identity using the *caller's* permissions, and the standalone service still fronts AgentCore's workload-identity directory.

Net IAM posture stays tighter than `main` (record-level actions gone; WI now ARN-scoped instead of `*`). The `NagSuppression` reason was updated to describe both new statements accurately (`registry.ts:170-179`). Minor, non-blocking: `Get/DeleteWorkloadIdentity` aren't called by the handler directly (only `CreateRegistry` uses them transitively), so they're slightly defensive — but they mirror the validated main-branch grant and the deploy passed, so trimming would be premature.

## Tests

New control flow is genuinely covered, not asserted-by-rename:
- `index.test.ts`: parameterized "defers a retryable {conflict,throttling,internal} to the waiter"; "re-drives deletion when not yet DELETING" asserts `['GetRegistry','DeleteRegistry']`; "keeps polling" on re-driven `Conflict` and on `GetRegistry` throttle; `DELETE_FAILED` still throws; unexpected error still rethrows.
- `registry.test.ts`: SLR statement asserted with the exact `iam:AWSServiceName` condition; WI statement asserted scoped to `workload-identity-directory/*` and explicitly **not** `*`.

Confirmed `@aws-sdk/client-agent-registry-control@3.1116.0` exports `ConflictException`, `ThrottlingException`, and `InternalServerException`, so the new imports resolve; the SDK also documents that `DeleteRegistry` is asynchronous and removes the registry along with its records — which validates dropping the old record-drain teardown.

## Docs / CI / bootstrap

Doc edits are mirrored into `docs/src/content/docs/`. Bootstrap unaffected (no new CFN resource *type*; runtime-role grants only, so `BOOTSTRAP_VERSION`/`BOOTSTRAP_HASH` correctly unchanged). Approval is contingent on CI (`build (agentcore)`, CodeQL, secrets scan) going green; PR body attests to a local `mise //cdk:test` pass (200 suites / 4,104 tests).

## Review agents

The full adversarial pass (security + code review across correctness, TS/Python parity, silent-failure, comment accuracy, test coverage) ran over the complete change in the initial review and came back clean modulo the now-fixed nits. This incremental pass verified the 8-file delta directly against the SDK contract, the added tests, and least-privilege patterns for the two new IAM statements.

cdk/src/constructs/registry.ts
```diff
@@ -89,45 +84,54 @@ export class AgentRegistry extends Construct {
...
+          resource: 'workload-identity-directory',
+          resourceName: '*',
         }),
       ],
     });
+    onEventFn.addToRolePolicy(serviceLinkedRolePolicy);
+    onEventFn.addToRolePolicy(workloadIdentityPolicy);
     for (const fn of [onEventFn, isCompleteFn]) {
       fn.addToRolePolicy(createPolicy);
       fn.addToRolePolicy(registryPolicy);
```
scottschreckengaust, line 139: **B1 (blocking severity): this narrows a grant on the same change that adds a new caller for it — `cdk destroy` can hang.**

On `main`, the three `bedrock-agentcore:{Create,Get,Delete}WorkloadIdentity` actions were members of `createPolicy`, so the loop below granted them to **both** functions (verified via `git show origin/main:cdk/src/constructs/registry.ts`). Line 134 attaches them to `onEventFn` **only**.

Meanwhile this same PR teaches `isCompleteFn` to issue `DeleteRegistry` itself — `registry-provisioning/index.ts:198` re-drives `requestRegistryDeletion` on every poll after a retryable failure. So a new caller of `DeleteRegistry` appears at exactly the moment that caller loses the workload-identity grant.

Failure path: `AccessDeniedException` is not in `isRetryableDeleteError` (only `Conflict`/`Throttling`/`InternalServer`), so it rethrows → `safeHandler` submits FAILED → **`DELETE_FAILED` on a NestedStack**, one of the more painful states to unwind by hand.

I can't confirm from the service model whether `DeleteRegistry` requires `DeleteWorkloadIdentity` from the caller. Three things make it worth fixing pre-emptively anyway: `CreateRegistry` demonstrably *did* (that's why commit `e519c90` exists), `main` granted it to both roles, and the PR body says live validation reached `UPDATE_COMPLETE` — **the delete path has never run**.

Restoring `main`'s symmetry is free:

```suggestion
    onEventFn.addToRolePolicy(serviceLinkedRolePolicy);
    for (const fn of [onEventFn, isCompleteFn]) {
      fn.addToRolePolicy(createPolicy);
      fn.addToRolePolicy(registryPolicy);
      fn.addToRolePolicy(workloadIdentityPolicy);
```

Please **don't** fix it by adding `AccessDeniedException` to `isRetryableDeleteError` instead — that burns the full 15-minute budget and then surfaces the generic `"Operation timed out"`, turning a precise error into an opaque one.

> Framework note worth keeping in mind: `Provider` runs `onEvent` and `isComplete` as two Lambdas with two separate execution roles, so moving an API call between phases silently moves it across an IAM boundary. The `for (const fn of [...])` loop exists to keep the roles in sync — lifting a statement out of it is the shape of this bug class.

cdk/src/constructs/registry.ts
```diff
@@ -89,45 +84,54 @@ export class AgentRegistry extends Construct {
...
+      resources: [
         Stack.of(this).formatArn({
           service: 'bedrock-agentcore',
-          resource: 'registry',
-          resourceName: '*/record/*',
+          resource: 'workload-identity-directory',
+          resourceName: '*',
         }),
       ],
     });
```
scottschreckengaust, line 135: **B2 (blocking severity): the comment that was deleted here is what would have prevented B1.**

`main` carried this rationale, which the PR drops:

> `CreateRegistry` ALSO provisions a workload identity under the hood, so the role needs the WorkloadIdentity create/get/delete actions too — the registry lands in `CREATE_FAILED` ("Unable to create workload identity because access was denied") without them.

What's left is a `bedrock-agentcore`-namespaced policy sitting in a file otherwise fully migrated to `agent-registry`, with no explanation — which reads like migration residue. The next person doing namespace cleanup deletes it and rediscovers `CREATE_FAILED` empirically, exactly as commit `e519c90` just did. The literal error string is the searchable artifact, so it's worth keeping verbatim:

```suggestion
    // `CreateRegistry` ALSO provisions a workload identity under the hood, so the
    // caller needs the WorkloadIdentity create/get/delete actions too — without
    // them the registry lands in CREATE_FAILED with "Unable to create workload
    // identity because access was denied" (#771, commit e519c90).
    //
    // These deliberately stay in the `bedrock-agentcore` namespace after the
    // migration to `agent-registry`: workload identity is still an AgentCore-owned
    // resource. This is not migration residue — do not "clean up" the namespace.
    const workloadIdentityPolicy = new iam.PolicyStatement({
      actions: [
        'bedrock-agentcore:CreateWorkloadIdentity',
        'bedrock-agentcore:GetWorkloadIdentity',
        'bedrock-agentcore:DeleteWorkloadIdentity',
      ],
      resources: [
        Stack.of(this).formatArn({
          service: 'bedrock-agentcore',
          resource: 'workload-identity-directory',
          resourceName: '*',
        }),
      ],
    });
```

cdk/src/handlers/registry-provisioning/index.ts
```diff
@@ -150,20 +174,29 @@ export async function onEvent(event: OnEventRequest): Promise<OnEventResponse> {
...
+    if (status === 'DELETE_FAILED') {
+      throw new Error(
+        `Registry ${registryId} entered DELETE_FAILED: ${statusReason ?? 'no reason given'}`,
+      );
     }
+    if (status === 'DELETING') return { IsComplete: false };
+
+    // The initial DeleteRegistry call may have been throttled or conflicted
+    // before deletion started. Re-drive it until the service reports DELETING.
+    const attempt = await requestRegistryDeletion(registryId);
```
scottschreckengaust, line 198: **B1 counterpart — this is the new caller.** Re-driving `DeleteRegistry` from `isComplete` is a sound retry design (the comment above explains it well), but `isCompleteFn`'s execution role no longer carries the workload-identity grant that `main` gave it, because `registry.ts:134` now attaches that policy to `onEventFn` only. See the B1 comment on `registry.ts:133-137`.

Worth noting the retry classifier is otherwise well-built: `ConflictException`/`ThrottlingException`/`InternalServerException` retry, `ResourceNotFoundException` → `'absent'` → `IsComplete: true` is the right idempotent teardown, and anything else fails loudly instead of silently looping. That last property is exactly why the missing grant surfaces as a hard `DELETE_FAILED` rather than a timeout.

cdk/src/bootstrap/resource-action-map.ts
```diff
@@ -115,13 +115,13 @@ export const RESOURCE_ACTION_MAP: Record<string, readonly string[]> = {
...
   'AWS::SQS::Queue': ['sqs:CreateQueue'],
-  // The AgentCore registry provisioning custom resource uses the CDK Provider
+  // The Agent Registry provisioning custom resource uses the CDK Provider
   // framework, whose async waiter is a Step Functions state machine (#246).
   'AWS::StepFunctions::StateMachine': ['states:CreateStateMachine'],
   'AWS::WAFv2::WebACL': ['wafv2:CreateWebACL'],
   'AWS::WAFv2::WebACLAssociation': ['wafv2:AssociateWebACL'],
   'Custom::AWS': ['lambda:InvokeFunction'],
-  'Custom::AgentCoreRegistry': ['lambda:InvokeFunction'],
+  'Custom::AgentRegistry': ['lambda:InvokeFunction'],
```
scottschreckengaust, line 124: **B3 (blocking severity): this rename is correct, but nothing enforces it.**

The key matches `registry.ts:152` — the PR gets this right. The problem is that the test whose whole purpose is catching this drift cannot see it.

`cdk/test/bootstrap/synth-coverage.test.ts:36-42` builds its template from the **root** stack only:

```ts
template = Template.fromStack(app.node.tryFindChild('backgroundagent-dev') as Stack);
```

`AgentRegistryStack extends NestedStack` (`registry.ts:221`), so it renders in the parent as a single opaque `AWS::CloudFormation::Stack` and `Custom::AgentRegistry` **never enters `typesInTemplate`**. I read the file in full (118 lines) and confirmed there's no `NestedStack` handling anywhere in `cdk/test/bootstrap/`. A partial rename — or a future third rename touching only one side — passes every test in the repo and fails at deploy with a `lambda:InvokeFunction` denial.

The fix pattern already exists in that same file: the second test (`:77-100`) explicitly synthesizes the context-gated tool-gateway path *because* those types "would slip past the coverage loop." The nested registry stack needs the same treatment — resolve the `AgentRegistryStack` child, template it, and assert `Custom::AgentRegistry` is both present and mapped.

Caveat on my evidence: I could not execute this suite locally (stale hoisted `typescript` in the review worktree → `TS5103`), so this is from reading the code, not a failing run. The coupling is structural and checkable by inspection.

> `Template.fromStack(parent)` is blind to nested-stack contents — this repo has been bitten before and documents it at `registry-api.test.ts:33`. Any assertion phrased as "every synthesized type is covered" really means "every type in *this* template," and nesting quietly shrinks the denominator.

cdk/src/bootstrap/version.ts
```diff
@@ -29,10 +29,9 @@ import { allPolicies } from './policies';
  * `compute-lambda-microvm` policy (#645 / ADR-021), 1.3.0 → 1.4.0 grants SNS
  * topic + customer-managed-KMS-key create/lifecycle for the OperationalAlerts
  * notification channel (#629), 1.4.0 → 1.5.0 adds the agent asset registry
- * policies (#246: bedrock-agentcore registry + workload identity, Step
- * Functions, Cognito group, and CloudFormation nested-stack actions). Adding
- * policies to the bundle is a minor bump — that is the precedent `compute-ecs`
- * set.
+ * policies (#246: registry provisioning + workload identity, Step Functions,
+ * Cognito group, and CloudFormation nested-stack actions). Adding policies to
+ * the bundle is a minor bump — that is the precedent `compute-ecs` set.
```
scottschreckengaust: **This edit preserved a false claim.** The diff rewrote this sentence (`bedrock-agentcore registry + workload identity` → `registry provisioning + workload identity`), but both versions assert that 1.5.0 added registry/workload-identity policies to the bootstrap bundle. `grep -rn 'agent-registry|WorkloadIdentity|CreateRegistry' cdk/src/bootstrap/` returns **nothing**. Commit `4712e061` added only `cognito-idp` Group, `states:*`, and CloudFormation nested-stack actions.

**Important: please don't "fix" this by adding `agent-registry:*` to the bundle.** The deploy role correctly needs none of it — the custom-resource Lambda calls the control plane under its *own* execution role, and the deploy role only needs `lambda:InvokeFunction` on the custom resource, which is already mapped in `resource-action-map.ts:124`. So the comment isn't concealing a real gap; it's just wrong. Fix the sentence, not the policy:

```suggestion
 * policies (#246: Step Functions, Cognito group, and CloudFormation
 * nested-stack actions — the registry's own `agent-registry:*` calls run under
 * the custom resource's Lambda execution role, not the deploy role). Adding
 * policies to the bundle is a minor bump — that is the precedent `compute-ecs` set.
```

docs/decisions/ADR-022-agent-asset-registry.md
```diff
@@ -82,20 +82,20 @@ The ADR fixes the **contract**. The substrate ranking below was written to defer
...
 - **Governance ships as a first-class lifecycle** — Draft → Pending Approval → Approved/Rejected with deprecation support — driven by EventBridge notifications and an `UpdateRegistryRecordStatus` API that lets external review pipelines (ticketing, security scan, human approval) close the loop programmatically. Auto-approve mode is available for dev environments. This is a direct match for sub-decision 10.
 - Hybrid semantic + keyword search with weighted relevance ranking.
 - MCP-native discovery endpoint — an agent can query the registry via MCP without ABCA-specific glue.
-- IAM or JWT (Cognito) authorization — ABCA already uses Cognito, so JWT drops in for the resolve path. Fine-grained IAM actions like `bedrock-agentcore:InvokeRegistryMcp` support the two-role model in sub-decision 11.
+- IAM or JWT (Cognito) authorization — ABCA already uses Cognito, so JWT drops in for the resolve path. Fine-grained IAM actions like `agent-registry:InvokeRegistryMcp` support the two-role model in sub-decision 11.
```
scottschreckengaust: **`agent-registry:InvokeRegistryMcp` does not exist.** The PR namespace-swapped it (`bedrock-agentcore:InvokeRegistryMcp` → `agent-registry:InvokeRegistryMcp`) rather than removing it, but the action is absent from both the control-plane and data-plane service models. The two-role model in sub-decision 11 is actually enforced by **Cognito group membership**, not fine-grained IAM.

It's rationale prose rather than an implemented policy, so impact is low — but this ADR is flipping to `accepted` in this same PR, so it ships as a documented security control that doesn't exist:

```suggestion
- IAM or JWT (Cognito) authorization — ABCA already uses Cognito, so JWT drops in for the resolve path. The two-role model in sub-decision 11 is enforced by Cognito group membership rather than fine-grained IAM actions.
```

docs/decisions/ADR-022-agent-asset-registry.md
```diff
@@ -193,6 +193,7 @@ Regardless of substrate, the invariants above (semver, immutability, resolve-at-
 
 ## Changelog
 
+- **2026-08-24 — accepted; migrated to standalone AWS Agent Registry (#771).** Marked the ADR accepted after #664/#665 merged. Replaced the retired `bedrock-agentcore` preview assumptions with the standalone `agent-registry` namespace, recorded the fresh-registry migration requirement, and added the default-on `enableAgentRegistry=false` deployment escape hatch for unsupported or restricted accounts/regions.
```
scottschreckengaust: Nit — "the default-on `enableAgentRegistry=false` escape hatch" reads as self-contradictory: the gate is default-**on**, and `=false` is the opt-out you pass to disable it. `:144` and `:162` both get this right; only the changelog entry conflates the two.

```suggestion
- **2026-08-24 — accepted; migrated to standalone AWS Agent Registry (#771).** Marked the ADR accepted after #664/#665 merged. Replaced the retired `bedrock-agentcore` preview assumptions with the standalone `agent-registry` namespace, recorded the fresh-registry migration requirement, and added the default-on `enableAgentRegistry` context gate (deploy with `enableAgentRegistry=false` to opt out) for unsupported or restricted accounts/regions.
```

scottschreckengaust (commented):
## Review — standalone Agent Registry migration (#771)

Reviewed end-to-end against the repo's 5-stage process (context/governance, VISION alignment, deep dive with the mandated review agents, docs, report). **Posting as a comment rather than a block** — but flagging up front that **B1–B3 are what I'd rate merge-blocking severity**; whether they gate this PR is your call.

**Headline: the hard part went well.** `cdk/src/handlers/shared/registry/client.ts`, `registry/types.ts`, and `agent/src/registry/client.py` contain **zero declaration changes** across the entire diff — comments only. That's exactly the outcome #771 predicted, and it's the strongest available evidence that the ports-and-adapters boundary is real rather than aspirational. A full service-namespace change (SDK client, IAM actions, ARNs, descriptor field names) landed without moving the port.

One functional defect: the PR **narrows an IAM grant in the same change that adds a new caller for the API that grant protects**, on the stack-delete path — which the PR body confirms was never exercised live (`UPDATE_COMPLETE` only).

### Blocking-severity findings (detail in inline comments)

| | Finding | Anchor |
|---|---|---|
| **B1** | Delete-path IAM asymmetry → `AccessDeniedException` rethrow → `DELETE_FAILED` on a NestedStack | `cdk/src/constructs/registry.ts:133-134` + `registry-provisioning/index.ts:198` |
| **B2** | The deleted rationale comment is what caused B1 — and its absence primes a recurrence | `cdk/src/constructs/registry.ts:119-132` |
| **B3** | `RESOURCE_ACTION_MAP` rename is *correct* but unguarded — `synth-coverage.test.ts` can't see nested stacks | `cdk/src/bootstrap/resource-action-map.ts:124` |

Each inline comment carries a one-click `suggestion` where the fix is mechanical.

### Vision alignment (`docs/design/VISION.md`)

| Tenet | Assessment |
|---|---|
| **4. Fail closed on risk** | ✅ Traced end-to-end. `enableAgentRegistry=false` → no `AGENT_REGISTRY_ID` → `factory.ts` throws → propagates uncaught from `orchestrator.ts:795` → task lands FAILED. Blueprints with no `registry://` refs still succeed via the `refs.length === 0` early return (`orchestrator.ts:532`). The new error naming both remedies is a real improvement on `main`'s bare `'AGENT_REGISTRY_ID env var is not set'`. |
| **5. Least privilege** | ⚠️ Direction is right (per-registry ARN scoping, `iam:AWSServiceName`-conditioned SLR grant), but the delete-path grant is now asymmetric — B1. |
| **8. Extensible without forking** | ✅ Strongest showing; the port survived a substrate swap. One pre-existing leak, below. |
| **10. Sample, not product** | ✅ New REGISTRY.md §5.1 and the preview caveat in DEPLOYMENT_GUIDE are honest about preview status. |

No undocumented tenet trade-offs. ADR-022 `proposed` → `accepted` is appropriate — the implementing PR is landing.

### A finding I raised and am now refuting

I initially flagged the removal of `drainRecords()` plus the record-level IAM (`ListRegistryRecords`, `DeleteRegistryRecord`, the `registry/*/record/*` ARN) as resting on an unverified cascade assumption. **That was wrong.** `DeleteRegistryCommand.d.ts:27` in the pinned SDK (`@aws-sdk/client-agent-registry-control@3.1116.0`, unpacked and read directly) states deletion "is asynchronous: the registry transitions to the `DELETING` status and is removed along with its registry records." The cascade is a service guarantee. Removing the manual drain is a legitimate simplification and the narrowed IAM is correct least-privilege work. Recording this so it doesn't resurface in a future review.

### On #771's explicit ask

The issue said "**Flag if the new service's semantics force a port change.**" The port held — but there is one leak worth recording. `cdk/src/handlers/registry-publish.ts`'s *entire* diff is one line:

```diff
-import { ConflictException } from '@aws-sdk/client-bedrock-agentcore-control';
+import { ConflictException } from '@aws-sdk/client-agent-registry-control';
```

So the substrate's exception taxonomy *is* the port's collision contract, upstream of the adapter — which contradicts `client.ts:20-23` ("NEVER a raw AWS SDK client"). This leak pre-exists on `main` and the PR handled it correctly, so it isn't blocking. But it's precisely what #771 asked to have flagged and the PR body doesn't mention it. Suggest a follow-up issue to promote a port-owned `RegistryConflictError`.

### Documentation ✅

Best-documented part of this PR. REGISTRY.md (§3.1 descriptor table, §5 delete cascade, new §5.1 optional deployment, §13 accepted risk), DEPLOYMENT_GUIDE (new subsection + preview service-table row), ADR-022 (accepted, dated, namespace migration, changelog).

**Starlight mirrors verified in sync** — I re-ran the real `docs/scripts/sync-starlight.mjs` against a clean tree and `diff -ru` returned zero lines. Generated, not hand-edited.

One pre-existing nit that doesn't anchor to a changed line: `agent/src/registry/__init__.py:12` still cites `ISSUE_246_AGENTCORE_FINDINGS.md`, a file that has never existed in this repo (confirmed via `find` and `git log --all`). Unrelated to the migration, but it's a dead pointer in a module this PR renames alongside.

### Tests & CI

- **CI green** — all 8 checks pass.
- **Bootstrap artifact regeneration: NOT APPLICABLE — this is not a gap.** The diff touches `cdk/src/bootstrap/policies/*.ts`, which normally demands a `BOOTSTRAP_VERSION` bump plus regenerated `cdk/bootstrap/policies/*.json`, `bootstrap-template.yaml`, and `BOOTSTRAP_HASH`. Here those edits are **comment-only**, and `computeBootstrapHash()` (`version.ts:43-51`) digests `allPolicies()` → `toJSON()` → sorted-key `JSON.stringify`, so comment text is not in the digest. Unchanged artifacts, unchanged `BOOTSTRAP_VERSION = '1.5.0'`, and an unchanged `DEPLOYMENT_ROLES.md` golden baseline are all **correct**. Confirmed empirically: 5 of 6 bootstrap suites executed, **107/107 passing** (`golden-baseline`, `artifact-sync`, `version`, `policies`, `bootstrap-template`). The one real bootstrap gap is B3 — a missing guard, not a stale artifact.
- **Test updates are proportionate** — 9 files updated for the renames, plus new SLR and workload-identity coverage. Two weaknesses noted inline.
- **#366 CDK test performance:** no regressions. No re-enabling of `aws:cdk:bundling-stacks`; the one in-test `new App()` (`synth-coverage.test.ts:79`) is pre-existing.
- **Contract sync:** `cdk/src/handlers/shared/types.ts` unchanged → no `cli/src/types.ts` mirror needed. `cli/src/stack-outputs.ts:74-100` already treats `RegistryApiUrl` as optional and `api-client.ts:96-105` already raises an actionable error — though its "re-run setup" remedy is a dead end under the new gate (worth a follow-up). Cedar untouched, so no `cedarpy`/`cedar-wasm` lockstep concern.
- **Supply chain:** `@aws-sdk/client-agent-registry-control@^3.1116.0` verified legitimate on npm (latest 3.1116.0, maintainers `amzn-oss`/`aws-sdk-bot`, lockfile integrity present). `boto3` 1.43.40 → 1.43.78 is routine.

### Nit that doesn't anchor to a changed line

`registry-provisioning/index.ts:116` applies `.slice(0, CLIENT_TOKEN_LENGTH)` (64) to a SHA-256 hex digest, which is already exactly 64 chars — a no-op. Harmless, but the comment at `:67-68` implies truncation to "stay within API limits." **Pre-existing** (`git blame` confirms it's not from this PR), so out of scope; noting it only because you're in the file.

### Review agents run

All six in-scope agents invoked; none omitted.

| Agent | Outcome |
|---|---|
| `code-reviewer` | Independently identified the delete-path IAM asymmetry (B1). |
| `silent-failure-hunter` | Flagged the `extractPayload` unknown-`recordType` fallthrough and unguarded `JSON.parse`; independently concluded the cascade removal is safe. |
| `type-design-analyzer` | Reviewed `RecordType`/`RegistryStatus` mappings and the `DeleteAttempt` union; no blocking findings — `DeleteAttempt` assessed as a well-formed three-state union. |
| `comment-analyzer` | Found the fabricated `InvokeRegistryMcp` action, the false `version.ts` bootstrap claim, and the deleted workload-identity rationale (B2). |
| `pr-test-analyzer` | Found the NestedStack synth-coverage blind spot (B3); executed 107 bootstrap tests. |
| `/security-review` | Ran with sub-agent false-positive filtering; **no findings at >=8 confidence**. Consistent with my own read — no new input-validation, authz, or secret-handling surface. UA attribution preserved throughout (`makeClient(AgentRegistryControlClient)`, `aws_session` in Python), so #319 is not regressed. |

I independently verified every blocking claim rather than relaying agent output: the `main` diff for B1, `synth-coverage.test.ts` read in full for B3, `git blame`/`git diff` authorship for both comment findings, and the SDK's own type definitions (unpacked at the pinned version) for the enum and cascade claims.

### Human heuristics

- **Proportionality** ✅ — ~35 files for a service-namespace migration, blast radius concentrated in the two adapters. The four commits tell an honest story: the feature, then three empirically-discovered permission fixes.
- **Coherence** ⚠️ — two seams: the surviving `bedrock-agentcore` workload-identity policy in an otherwise fully-migrated file (B2), and the opt-out gate polarity diverging from the neighbouring opt-in `enableToolGateway`.
- **Clarity** ⚠️ — the code is clear; three comments are not. `version.ts:32` states something false, ADR-022:95 cites a nonexistent IAM action, and the one genuinely load-bearing comment was deleted. Comment edits were the weakest part of an otherwise careful diff.
- **Appropriateness** ✅ — governance clean. #771 carries `enhancement, approved, registry`; branch `feat/771-agent-registry-migration` matches the required pattern; scope matches acceptance criteria with no drive-by changes.

### Review method / caveats

- **B1 carries a stated uncertainty I could not close.** Whether `DeleteRegistry` requires `DeleteWorkloadIdentity` *from the caller* is not something I could confirm from the service model. The case rests on: `CreateRegistry` demonstrably did require it (that's why commit `e519c90` exists), `main` granted the trio to both roles, and the delete path has never run. You have live-deploy access and can settle it with one `cdk destroy`.
- **B3 is derived from reading code, not a failing run.** I could not execute `mise //cdk:test` in the review worktree — per-worktree `node_modules` is absent and the shared root install is stale (hoisted `typescript` 5.9.3 vs `cdk/package.json` `^6.0.3`, giving `TS5103` on `ignoreDeprecations: "6.0"`). The coupling B3 describes is structural and verifiable by inspection, but I never watched the test fail to prove it.

**Net:** strong migration, genuinely good port hygiene, excellent docs. Fix B1+B2 (one edit, same code region), add the B3 guard, and correct the two false comments, and this is in good shape.

cdk/package.json
```diff
@@ -16,6 +16,7 @@
   "dependencies": {
     "@aws-cdk/aws-bedrock-alpha": "2.260.0-alpha.0",
     "@aws-crypto/sha256-js": "^5.2.0",
+    "@aws-sdk/client-agent-registry-control": "^3.1116.0",
     "@aws-sdk/client-bedrock-agentcore": "^3.1078.0",
     "@aws-sdk/client-bedrock-agentcore-control": "^3.1078.0",
```
scottschreckengaust: Nit — this control-plane package is now orphaned. After the migration there are **zero** references to `@aws-sdk/client-bedrock-agentcore-control` in `cdk/src` or `cdk/test`; `registry-publish.ts` was the last consumer and line 19 replaces it.

Remove **only** this line. Line 20's data-plane `@aws-sdk/client-bedrock-agentcore` is still live — `cancel-task.ts`, `shared/agentcore-browser.ts`, `shared/strategies/agentcore-strategy.ts`, and `shared/memory.ts` all use it.

```suggestion
```

cdk/src/constructs/registry.ts
```diff
@@ -89,45 +84,54 @@ export class AgentRegistry extends Construct {
...
+    // does not exist, and ListRegistries has no resource-level target.
     const createPolicy = new iam.PolicyStatement({
       actions: [
-        'bedrock-agentcore:CreateRegistry',
-        'bedrock-agentcore:ListRegistries',
-        'bedrock-agentcore:CreateWorkloadIdentity',
-        'bedrock-agentcore:GetWorkloadIdentity',
-        'bedrock-agentcore:DeleteWorkloadIdentity',
+        'agent-registry:CreateRegistry',
+        'agent-registry:ListRegistries',
```
scottschreckengaust: Nit — `agent-registry:ListRegistries` is granted but never called. There's no `ListRegistriesCommand` anywhere in the codebase (only this grant and the nag reason that justifies it). Dropping it shrinks the `*`-resource statement, which is the one hardest to justify.

Three places to touch if you remove it: this action, the comment at `:87-88` ("and ListRegistries has no resource-level target"), and the `AwsSolutions-IAM5` reason at `:172` ("CreateRegistry/ListRegistries are account-level actions..."). Left as prose rather than a suggestion since a one-click edit here would desync the other two.

cdk/test/constructs/registry.test.ts
```diff
@@ -82,18 +81,56 @@ describe('AgentRegistry construct', () => {
...
+
+    expect(statement).toBeDefined();
+    expect(statement?.Action).toEqual(expect.arrayContaining([
+      'bedrock-agentcore:CreateWorkloadIdentity',
+      'bedrock-agentcore:GetWorkloadIdentity',
+      'bedrock-agentcore:DeleteWorkloadIdentity',
+    ]));
+    expect(statement?.Resource).not.toBe('*');
+    expect(JSON.stringify(statement?.Resource)).toContain('workload-identity-directory/*');
+  });
```
scottschreckengaust: Good addition, but note it **cannot detect B1.** It flattens statements across *every* `AWS::IAM::Policy` in the template and asserts one match exists, so it passes whether the grant lands on `onEventFn`, `isCompleteFn`, or both — which is precisely the distinction that matters here.

Asserting per-role would have caught the regression. Something like: find the two `AWS::IAM::Policy` resources by their `Roles` reference and assert *each* carries `bedrock-agentcore:CreateWorkloadIdentity`. The resource-scoping assertions at `:120-121` are well done — `not.toBe('*')` plus the `workload-identity-directory/*` check is exactly the right shape.

cdk/test/constructs/registry.test.ts
```diff
@@ -82,18 +81,56 @@ describe('AgentRegistry construct', () => {
...
+
   test('grants the per-registry + record actions scoped to registry ARNs', () => {
```
scottschreckengaust: Nit — this name is now stale. The PR removed `ListRegistryRecords`/`DeleteRegistryRecord` and the `registry/*/record/*` ARN from `registryPolicy` (correctly — `DeleteRegistry` cascades), so there are no "record actions" left to grant.

```suggestion
  test('grants the per-registry actions scoped to registry ARNs', () => {
```

Separately, while you're here: the `Match.objectLike` below asserts only `Action`, never `Resource` — so the "scoped to registry ARNs" half of the claim is unverified. Adding a `Resource` matcher would make the name honest.

cdk/src/stacks/agent.ts
```diff
@@ -89,6 +89,9 @@ export class AgentStack extends Stack {
   constructor(scope: Construct, id: string, props: StackProps = {}) {
     super(scope, id, props);
 
+    const enableAgentRegistry = this.node.tryGetContext('enableAgentRegistry');
+    const agentRegistryEnabled = enableAgentRegistry !== false && enableAgentRegistry !== 'false';
```
scottschreckengaust, line 105: Nit — two things here, both minor but cheap to close.

**Polarity is inconsistent with its neighbour.** This gate is opt-out (`!== false && !== 'false'`), while `enableToolGateway` at `:278-279` is strict opt-in (`=== true || === 'true'`). Both defaults are defensible — registry default-on for compatibility, gateway default-off — but a reader has to check each one to know which way it swings. Worth a brief `// default-on (unlike enableToolGateway)` if you keep it as-is.

**No validation.** A typo like `--context enableAgentRegistry=fasle` silently deploys the registry, which is the opposite of what the operator asked for — and on a preview service in an unsupported region that's a failed deploy with a confusing cause. `resolveEcsTaskSizing` (`ecs-agent-cluster.ts:243-252`) sets the house precedent of throwing on malformed context:

```suggestion
    const enableAgentRegistry = this.node.tryGetContext('enableAgentRegistry');
    if (
      enableAgentRegistry !== undefined && enableAgentRegistry !== true && enableAgentRegistry !== false
      && enableAgentRegistry !== 'true' && enableAgentRegistry !== 'false'
    ) {
      throw new Error(
        `enableAgentRegistry must be true or false, got '${String(enableAgentRegistry)}'`,
      );
    }
    // Default-on for compatibility (contrast enableToolGateway, which is opt-in).
    const agentRegistryEnabled = enableAgentRegistry !== false && enableAgentRegistry !== 'false';
```

cdk/src/handlers/shared/registry/agent-registry-client.ts
```diff
@@ -450,32 +453,25 @@ export class AgentCoreRegistryClient implements RegistryClient {
...
+      const body = JSON.parse(raw.descriptors?.custom?.data ?? '{}');
       return {
         runtime: body.runtime as RuntimePayload,
         storageMode: 'custom',
         discovery: (body.discovery ?? {}) as Record<string, unknown>,
         publisher: typeof body.publisher === 'string' ? body.publisher : undefined,
       };
     }
-    if (raw.descriptorType === 'AGENT_SKILLS') {
+    if (raw.recordType === 'SKILL') {
```
scottschreckengaust, line 471: Nit, **pre-existing structure** (`main` had the same shape with `descriptorType === 'AGENT_SKILLS'`), so out of scope — but you're renaming these branches anyway, so flagging while it's cheap.

The dispatch is `CUSTOM` → `SKILL` → unconditional fallthrough to `raw.descriptors?.mcpServer?.data ?? '{}'` at `:483`. The SDK's `RecordType` enum has **four** members — `AGENT`, `CUSTOM`, `MCP`, `SKILL` (`dist-types/models/enums.d.ts:43-48`, unpacked at the pinned 3.1116.0) — so an `AGENT` record silently resolves to `'{}'`: a successful-looking empty payload rather than an error.

Not reachable today (ABCA publishes no `AGENT` records), but a `switch` with an explicit `throw` on the unhandled case would match tenet 4's fail-closed posture and make the next enum addition a compile-or-runtime error instead of a silent `{}`. Same applies to the three unguarded `JSON.parse` calls in this function — malformed stored data currently throws a raw `SyntaxError` with no record context.

cdk/src/constructs/registry.ts
ayushtr-aws: Fixed in `971663d`. `workloadIdentityPolicy` is now attached to both Provider handler roles; `AccessDeniedException` remains non-retryable. I redeployed `backgroundagent-dev2` and forced the exact path by invoking `isCompleteFn` with a Delete event while a disposable registry was `READY`. CloudTrail shows `DeleteRegistry` under the `isCompleteFn` assumed role returning `DELETING`, and the waiter completed with the registry absent. Leaving this blocker open for re-review.

cdk/src/constructs/registry.ts
ayushtr-aws: Fixed in `971663d`. Restored the load-bearing rationale above `workloadIdentityPolicy`, including the observed access-denied text and why the `bedrock-agentcore` namespace is an intentional cross-service dependency during both CreateRegistry and DeleteRegistry. Leaving this blocker open for re-review.

cdk/src/handlers/registry-provisioning/index.ts
ayushtr-aws: Addressed by the role-policy fix and live validation in `971663d`. The forced waiter-role test called `DeleteRegistry` directly from `isCompleteFn` while the disposable registry was `READY`; it returned `DELETING`, completed on the next poll, and the registry was confirmed absent. Leaving this counterpart thread open for re-review.

cdk/src/bootstrap/resource-action-map.ts
ayushtr-aws: Fixed in `971663d`. `synth-coverage.test.ts` now resolves the actual `AgentRegistryStack` child, creates a template from that nested stack, asserts `Custom::AgentRegistry` is present, and verifies its action-map coverage. The focused suite and full CDK suite pass. Leaving this blocker open for re-review.

cdk/src/bootstrap/version.ts
ayushtr-aws: Fixed in `971663d` by correcting the 1.5.0 history only; no bootstrap policy, version, hash, or generated artifact changed. The comment now lists Step Functions, Cognito group, and nested-stack actions and explicitly says registry control-plane calls run under the custom-resource Lambda role.

docs/decisions/ADR-022-agent-asset-registry.md
ayushtr-aws: Addressed the underlying documentation issue in `971663d`, but not by removing the action. AWS Agent Registry’s IAM permissions reference explicitly lists `agent-registry:InvokeRegistryMcp` and says MCP invocation requires it plus `agent-registry:SearchDiscoverableRegistryRecords`. The ADR now separates that MCP IAM requirement from ABCA’s publisher/approver split, which is enforced by Cognito groups. Leaving this open for confirmation because the original nonexistence claim was not correct.

docs/decisions/ADR-022-agent-asset-registry.md
ayushtr-aws: Applied in `971663d`: the changelog now describes a default-on `enableAgentRegistry` context gate and identifies `enableAgentRegistry=false` as the opt-out.

cdk/package.json
ayushtr-aws: Fixed in `971663d`: removed the orphaned control-plane dependency from `cdk/package.json` and its `^3.1078.0` lock entry. The CLI’s exact `3.1078.0` dependency and the live AgentCore data-plane dependency remain intact.

cdk/src/constructs/registry.ts
ayushtr-aws: Fixed in `971663d`: removed unused `agent-registry:ListRegistries` and updated the account-level comment, construct test, and IAM5 suppression rationale consistently.

cdk/test/constructs/registry.test.ts
ayushtr-aws: Fixed in `971663d`. The test now identifies each generated policy by its attached onEvent/isComplete role reference and asserts that both roles carry all three workload-identity actions scoped to `workload-identity-directory/*`. This would fail on the original asymmetry.

cdk/test/constructs/registry.test.ts
ayushtr-aws: Fixed in `971663d`: renamed the test and added an explicit non-`*` resource assertion containing `registry/*`, alongside the action assertions.

cdk/src/stacks/agent.ts
ayushtr-aws: Applied in `971663d` with tests. The gate now accepts only `undefined`, booleans, or exact lowercase string `true`/`false`, documents its default-on polarity, and throws an actionable synth error for malformed values such as `fasle`. REGISTRY.md and DEPLOYMENT_GUIDE.md were updated and mirrored.

cdk/src/handlers/shared/registry/agent-registry-client.ts
ayushtr-aws: Acknowledged. I am leaving this out of #771 as noted because the AGENT fallthrough and parse-context improvements are pre-existing behavior outside the namespace migration. Leaving the thread open for scope confirmation rather than self-resolving it.

scottschreckengaust (approved):
## Verdict: Approve with nits

Careful, behavior-preserving substrate migration of the agent asset registry from the Bedrock AgentCore preview namespace to the standalone `agent-registry` service. The swap is confined to the adapters behind the `RegistryClient` port, IAM was tightened (not widened), the `enableAgentRegistry` gate fails closed, docs + the Starlight mirror are in sync, and it closes approved/assigned #771. CI is green. Nothing blocks merge — the items below are non-blocking.

### Recommended fixes (non-blocking)

**R1 — TS/Python divergence on corrupt records** (`cdk/src/handlers/shared/registry/agent-registry-client.ts`, `extractPayload`/`getRecordById`/`listRecords`). A present-but-malformed record body throws a raw `SyntaxError` from `JSON.parse` instead of resolving to the documented fail-closed `REMOVED` state. Because `listRecords` eagerly extracts *every* record in the namespace, a single corrupt record breaks resolution and the immutability check for all otherwise-healthy refs in that namespace. The Python read-side already does this correctly — it extracts only the winning record and wraps extraction in `except (ValueError, json.JSONDecodeError)` → `REMOVED`. Suggest matching the TS side to that fail-closed contract. Only reachable via out-of-band/corrupt records, hence non-blocking.

**R2 — Python null-body edge case** (`agent/src/registry/agent_registry_client.py:76`, CUSTOM path). `json.loads(record.get("data", "{}"))` defends against a *missing* `data` key but not an explicit `null` — `json.loads(None)` raises `TypeError`, which is not caught by the `except (ValueError, json.JSONDecodeError)` handler at the resolve boundary. The MCP path already uses `... or "{}"`; suggest the CUSTOM path do the same.

### Nits

- **N1** — Header comment states MCP *and* SKILL descriptors both use a `_meta` block; SKILL actually carries runtime via `x-abca-runtime` frontmatter. Comment should distinguish the two.
- **N2** — Comment claims a duplicate key in the SKILL frontmatter is "a YAML error." Under js-yaml with `{ json: true }` this is not an error — it is last-wins. The frontmatter-injection defense still holds (it's enforced by the emitter, not the parser), but the stated rationale is incorrect.
- **N3** — `resolve` comment omits the SKILL path from the list of handled kinds.
- **N4** — `waitPastCreating` docstring reads as if it polls to a terminal state; it only polls *past* CREATING.
- **N5** — `autoApprove` header overstates scope — only the approve call is gated, not create/submit.
- **N6** — Type design: `kind: string` is not discriminated against the payload variant; record `status` is coerced without validation; `DeleteAttempt` is more specified than its consumers require.
- **N7** — `parseSkillPublisher` silently drops the publisher on corrupt frontmatter rather than surfacing it.

### Test gaps

- **T1** — The read-side SKILL parser is never fed a hand-crafted duplicate-key / corrupt-frontmatter SKILL.md directly; the N2 behavior is asserted only indirectly.
- **T2** — In `isComplete` for the delete lifecycle, the non-retryable, non-404 `GetRegistry` rethrow branch is uncovered.

### Pre-existing follow-ups (not introduced by this PR)

- **P1** — The publish Lambda's 15s timeout is below the 60s create-poll budget (30 × 2000ms), so a slow create can strand a DRAFT record. Predates this PR; worth a tracking issue.
- **Namespace ownership** — Any `RegistryPublisher` can publish into any namespace; the publish handler is unchanged here, but the authorization question is worth filing.

### Review agents run

All six mandated agents ran, none omitted: `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`, `comment-analyzer`, `type-design-analyzer`, and `security-review` (no findings). Bootstrap `BOOTSTRAP_VERSION` correctly stayed `1.5.0` — registry control-plane calls run under the Lambda execution role, not the CFN bootstrap role, so no bootstrap policy delta was required.

PR #799, opened 2026-08-26
docs(model): document bedrockGeoRegion and the global-vs-geo tradeoff

isadeks:
Part of #740. Closes the documentation gap I raised reviewing #764.

## Why

`bedrockGeoRegion` (added in #764) was documented nowhere — not in `docs/guides/DEVELOPER_GUIDE.md`, not in `Model-configuration.md`, not in `cdk.json`. The synth-time error message names the key, but only helps someone who already knows it exists.

That matters more than usual because #740 is the *documentation* tracking issue, and #742 created the canonical model-configuration reference specifically so model settings stop being scattered. A new key that decides which geography every inference profile routes through belongs in it.

## What

A section covering:

- what the key does, and how to set it (`-c bedrockGeoRegion=global` or `cdk.json` `context`);
- accepted values, framed as what the CDK models rather than a copied list — `global`, `us`, `us-gov`, `eu`, `apac`, `jp`, `au`;
- that an unrecognized value **fails at synth**, and why that is the right behaviour: an invented geography yields a well-formed ARN for a profile that does not exist, so the grant would authorize nothing and the agent would fail at turn 0 with `AccessDenied` and nothing to explain why;
- that one value drives both grant sites and `ANTHROPIC_DEFAULT_HAIKU_MODEL`, so a deployment cannot grant one geography while calling another.

**The tradeoff, stated as a choice rather than a default:** `global.` routes to any supported commercial Region, giving better throughput and resilience under burst — worth having for tasks that run hours. A geo profile keeps inference in that geography, which is what a **data-residency requirement** demands. This is the material half for #747, which flips the default to `global`.

Plus the two failure modes that surface at runtime rather than synth: a model with no profile in the target geography, and account-level Bedrock access not covering that geography's entitlements.

## Two corrections

#764 made two existing statements stale:

| statement | was | now |
|---|---|---|
| bare-vs-prefixed rule | grant sites add `us.` | add the configured geo |
| layer-1 table row | `bedrockModels` only | + `bedrockGeoRegion` |

## One question the guide implied but never answered

Changing the **geography** needs a redeploy, because the IAM grants are scoped to explicit profile ARNs resolved at synth. Switching among **already-granted models** does not — that is a DynamoDB write via layer 4. The layering table implied this; it now says so.

## Verification

Starlight mirror regenerated with `mise //docs:sync` — both files carry the new content, and `mise //docs:build` is green (78 pages). Docs-only; no source changes.

### DIFF
--- a/docs/guides/DEVELOPER_GUIDE.md
+++ b/docs/guides/DEVELOPER_GUIDE.md
@@ -122,7 +122,7 @@ See the [Cedar policy guide](./CEDAR_POLICY_GUIDE.md) for the full authoring ref
 
 | # | Layer | What it controls | Where | ID form |
 |---|---|---|---|---|
-| 1 | **IAM invoke allowlist** | Which models the agent's roles may invoke at all. The outer gate — everything below fails without it. | `DEFAULT_BEDROCK_MODEL_IDS` (`cdk/src/constructs/bedrock-models.ts:34`); override with CDK context `bedrockModels` (key at `:48`, resolver at `:67`) | **Bare** (`anthropic.claude-…`) |
+| 1 | **IAM invoke allowlist** | Which models the agent's roles may invoke at all, and in which geography. The outer gate — everything below fails without it. | `DEFAULT_BEDROCK_MODEL_IDS` (`cdk/src/constructs/bedrock-models.ts`); override with CDK context `bedrockModels`. Geography via context `bedrockGeoRegion` (default `us`) | **Bare** (`anthropic.claude-…`) |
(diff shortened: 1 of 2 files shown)


isadeks:
Superseded by #806, which carries this content plus the geo flip and the review fixes. Verified before closing: every distinctive section here (`Choosing the inference-profile geography`, the residency caveat, the fail-at-synth rationale, the redeploy note) is present on #806's branch, with strictly newer model strings — this branch still says `us.anthropic.claude-opus-5` where #806 says `global.`.

Closing rather than rebasing, since a rebase would produce an empty diff and #806 needs no doc changes on top.

Issue #576, opened 2026-07-10
feat(jira): support @bgagent comment-triggered PR iteration


ayushtr-aws:
## Component

CDK Jira adapter, agent workflows

## Delivered scope

Implemented by PR #640 for standalone Jira issues. A human comment containing `@bgagent <instruction>` resolves the issue's newest ABCA PR-producing task and runs `coding/pr-iteration-v1` against that PR.

PR #640 explicitly deferred orchestration-child routing and dependent restacking until the shared orchestration contract reached `main`. That remaining integration is tracked in #574 and is not part of this closed issue.

## Behavior

- Subscribe to Jira `comment_created` webhooks.
- Parse plain-text and ADF comments for a token-bounded `@bgagent` mention.
- Ignore comments authored by the ABCA app/bot to prevent self-trigger loops.
- Resolve the Jira cloud ID and issue key to the newest prior PR-producing task.
- Create a `coding/pr-iteration-v1` task for the existing repository and PR.
- Attribute the iteration to the linked Jira user when available, otherwise retain the original task owner.
- Post immediate acknowledgement or actionable no-PR/admission feedback through the Jira app actor.
- Rely on deterministic Jira terminal fanout for the final outcome and cost comment.
- Deduplicate webhook redelivery by Jira comment ID.

## Acceptance criteria

- A Jira comment such as `@bgagent update the README too` on a standalone issue with an ABCA PR creates one iteration task for that PR.
- The Jira issue receives prompt acknowledgement and a terminal outcome.
- Comments without `@bgagent`, app-authored comments, and unsupported events are no-ops.
- If no prior ABCA PR is found, the issue receives clear feedback and no task is created.
- Redelivery does not create duplicate iterations.
- Existing label-trigger behavior remains unchanged.

## Tests

- Jira webhook tests cover mention parsing, ADF extraction, self-comment skip, no-PR fallback, user attribution, idempotency, and successful iteration creation.
- Task lookup tests cover tenant-and-issue resolution to the correct prior PR task.
- Jira setup docs include the comment-created webhook event and reviewer flow.

## Out of scope

- Orchestrated Jira child routing and dependent PR restacking; tracked in #574.
- Maturing in-place Jira iteration replies and heartbeats; tracked separately under #580.
- Plan approval or decomposition comments.
- Arbitrary comments without an `@bgagent` mention.

ayushtr-aws:
[...]

PR #800, opened 2026-08-26
feat(cdk): switch the Bedrock inference profile to the global geo

### LINKED ISSUE #747: feat(cdk): switch Bedrock inference profile to the global geo

isadeks:
Closes #747. Final step of the #740 stack; the plumbing landed in #746 / #764.

**Draft** until the deploy + smoke test in the acceptance criteria are done — see the bottom.

## What

Sets `bedrockGeoRegion` to `global` in `cdk/cdk.json` (which had no `context` block before) and moves every default string that names a geography with it.

| Site | Change |
|---|---|
| `cdk/cdk.json` | `context.bedrockGeoRegion = "global"` |
| `agent/src/config.py` | `ANTHROPIC_MODEL` fallback → `global.` |
| `agent/src/config.py` | `ANTHROPIC_DEFAULT_HAIKU_MODEL` fallback → `global.` |
| `agent/src/models.py` | `anthropic_model` default → `global.` |
| `cli/src/repo-display.ts` | platform-default mirror → `global.` |
| `workflows.ts` | `global.` forms of Opus 5 + Haiku 4.5 admitted |

## Why global

A `global.` profile routes to any supported commercial Region, giving better throughput and resilience under burst — worth having for tasks that run for hours.

**The tradeoff is data residency.** A deployer with a residency requirement sets `bedrockGeoRegion=us` (or `eu`/`apac`) instead, which is exactly why #746 made this a context key rather than a second hardcode.

## Verification

Against the synthesized template, not assumed:

| check | result |
|---|---|
| inference-profile ARNs → `global.` | 10 |
| inference-profile ARNs left `us.` | 0 |
| foundation-model ARNs changed | 0 of 20 (already geo-agnostic) |
| `ANTHROPIC_DEFAULT_HAIKU_MODEL` | `global.…` — derived by #746, no edit |
| rollback `-c bedrockGeoRegion=us` | synths clean, all 10 ARNs revert |

Full `mise run build` green: **4142** CDK, **1583** agent, **751** CLI.

## Two things the issue's scope table missed

Both were caught by existing guards rather than by reading the code, which is worth recording.

**The agent-side haiku fallback is a separate value.** #747 said the haiku model needs no edit because #746 derives its prefix — true of the env var the stack injects, but a run with **no env set** reaches the Python literal in `config.py`. Leaving it would have made local and deployed runs use different geographies. The `model-default-docs-parity` contract test caught it.

**The allow-list pairing invariant was `us.`-shaped.** It read `global.anthropic.…` as a *bare* id and demanded a nonsensical `us.global.anthropic.…` pairing. Generalized to the geo prefixes the list actually uses, preserving the real invariant — no prefixed entry without its bare form, no bare entry admitted in zero geographies — and confirmed still failing against an orphaned entry.

Both prefixes stay listed deliberately: the allow-list must accept whatever geography a deployment is configured for, so dropping the `us.` forms would reject a residency-constrained deployer's workflows at admission.

## Docs

The tradeoff and the documented defaults move together, since #742's drift test enforces them. Also corrects several statements that asserted a `us.` prefix as a *rule* rather than as the then-current default. Starlight mirrors regenerated with `mise //docs:sync`.

## Before this can leave draft

Per #747's acceptance criteria, still outstanding:

- [ ] Deploy to a stack and confirm `cdk diff` matches the ARN change above
- [ ] Agentcore smoke test: one task completes end to end and opens a PR on the global profile
- [ ] `platform doctor` access probe against the global profile

I have not run these — flagging rather than implying they passed. ECS verification is out of scope per the issue (CI deploys agentcore only); its unit coverage is included and green.


isadeks:
## Deployed and verified live — the three open acceptance criteria are met

Deployed this branch to a dev stack (`UPDATE_COMPLETE`, zero failed resources) and verified by artifact rather than by exit code.

### Deploy

| check | result |
|---|---|
| `cdk diff` before deploy | `us.` → `global.` on all 5 models |
| delivery-source / ECS destroys | **0** |
| total destroys | 2 (routine version churn) |
| stack | `UPDATE_COMPLETE`, 0 failures |

### The grant is live, not just synthesized

| artifact | result |
|---|---|
| inline policy on the deployed runtime role | **5 `global.` profiles, 0 `us.`** |
| deployed runtime env `ANTHROPIC_DEFAULT_HAIKU_MODEL` | `global.anthropic.claude-haiku-4-5-…` |
| direct `invoke-model global.anthropic.claude-opus-5` | 200 |

### End-to-end on the global profile

Three real tasks ran through the **Linear** path (not the CLI), all `COMPLETED`, one opening a PR with a build-gate pass.

The authoritative evidence is the Bedrock model-invocation log, which records what was actually called:

```
48x  global.anthropic.claude-opus-5                    caller: agent role
 3x  global.anthropic.claude-haiku-4-5-20251001-v1:0
```

Window `12:08:47 → 12:30:11`, matching the three task timestamps. The caller is the agent's assumed role, so this is the deployed grant being exercised — not a local credential.

Also present in the same window: `7x us.anthropic.claude-opus-4-8` under `user/sphia-dev`. That is my own CLI usage during verification, not the agent — checked rather than assumed, since an unexplained `us.` invocation would have undermined the whole result.

So: **agentcore end-to-end confirmed, on the global profile, with the auxiliary model on the same geography.** Taking this out of draft.

### One gap found while verifying, worth its own issue

`platform doctor` passed, but its Bedrock check only probes the **foundation-model catalog** (`anthropic.claude-sonnet-4-6` is visible in us-east-1). It never probes the **inference profile** the deployment is actually configured for.

For a geo change that is precisely the wrong thing to check: a stack could be granted `global.` profiles the account cannot invoke, and `doctor` would still report healthy. The probe should resolve `bedrockGeoRegion` and check `<geo>.<model>` — which would have made this whole verification a one-command check. Happy to file it.

isadeks:
## Review accepted — the rollback claim is false, and I have reproduced it

I independently verified every finding rather than taking them on assertion. All six hold. Two of them contradict claims I made in the PR body, and the first one is a genuine defect rather than a documentation error.

### Finding 1 — CONFIRMED, and worse than a wording problem

`cdk synth -c bedrockGeoRegion=us` on this branch:

```
IAM grants us.:      10
IAM grants global.:  0
runtime ANTHROPIC_DEFAULT_HAIKU_MODEL: us.anthropic.claude-haiku-4-5-…
runtime ANTHROPIC_MODEL:               <NOT SET>
```

The stack injects the **auxiliary** model into the runtime env but not the **main** one — `grep -c 'ANTHROPIC_MODEL:' cdk/src/stacks/agent.ts` returns 0. So a default task falls through to the `config.py` literal, which this PR set to `global.anthropic.claude-opus-5`, against a grant list that is now `us.`-only. Every no-override task fails at turn 0 with AccessDenied.

So "revert is a one-line context change" is **wrong**, and my own PR body half-knew it — I wrote "roll back context and defaults together" while the summary table still claimed one line. That inconsistency should have been the tell.

Also confirmed: `TaskConfig.haiku_model` at `agent/src/models.py:162` is still `us.…` — a fifth stale default I missed entirely, since I only counted four.

And the reviewer is right that this generalizes: any non-`us` geography has the same mismatch, which falsifies #799's statement that `bedrockGeoRegion` alone drives every relevant prefix.

### Finding 2 — CONFIRMED

Cross-checking the allowlist against the granted set:

| granted model | `global.` form allowlisted |
|---|---|
| claude-sonnet-4-6 | **no** |
| claude-opus-4-20250514-v1:0 | **no** |
| claude-opus-4-8 | **no** |
| claude-opus-5 | yes |
| claude-haiku-4-5 | yes |

Three globally granted models would be rejected at admission. My generalized invariant only proves *internal* pairing, so it cannot see this — and the reviewer's mutation (an invented bare/global pair passing all 22 tests) shows the test admits an ungranted model too. Comparing against `DEFAULT_BEDROCK_MODEL_IDS` × the deployed geography is the right invariant.

### Findings 3–6 — CONFIRMED

I reproduced the `--model` gap directly: `global.anthropic.example-ungranted-model` and `global.not-a-bedrock-model` are both accepted.

And my justification for not fixing it was wrong. I claimed the CLI cannot reach `bedrockModels`. It can — `cloudformation get-template` on the deployed stack recovers all five granted models from the profile ARNs, using credentials the CLI already requires. A `BedrockModelIds` output is cleaner, as suggested. I rationalized a limitation instead of testing it.

The reverse-compat bug is real too: with no `BedrockGeoRegion` output, the bare-id error suggests `us.<model>`, which a default-`global` stack does not grant.

On finding 4: the wording is defensible but the reviewer's point about PASS contributing to "All checks passed" stands — resolution happens under operator credentials while invocation happens under the workload role, so the check should describe control-plane visibility rather than readiness.

### Plan

Holding both PRs. Not merging.

1. Inject `ANTHROPIC_MODEL` into the runtime env from the resolved geography, so the main model cannot diverge from the grant. That fixes the class rather than re-syncing literals, and makes the rollback genuinely one line.
2. Same for the ECS task definitions, which omit both vars.
3. Fix `TaskConfig.haiku_model`.
4. Replace the pairing invariant with a grants-vs-allowlist parity test; add the three missing `global.` forms.
5. Add `BedrockModelIds` and validate `--model` against it; fix the no-output suggestion.
6. Reword the doctor check; correct the stale output description and the docs #799 does not cover.

Thanks — this was a better review than my own verification, and the mutation test on the allowlist is the part I should have written myself.

isadeks:
## Deployed and verified live — including the claim the review falsified

Deployed `b51712c6` (`UPDATE_COMPLETE`, zero failed resources). The pre-deploy diff showed exactly the intended change — `ANTHROPIC_MODEL` added to the AgentCore runtime and both ECS task definitions, plus the new output — with **0** delivery-source and **0** ECS destroys.

### Finding 1's fix, live

The defect was that the stack injected only the auxiliary model, so the main model fell through to a Python literal a geography change never touched:

```
AgentCore runtime env
  ANTHROPIC_MODEL                global.anthropic.claude-opus-5      ← was NOT SET
  ANTHROPIC_DEFAULT_HAIKU_MODEL  global.anthropic.claude-haiku-4-5-…
```

And substrate parity, which was the half of the finding I had not tested at all — both ECS task definitions now carry both vars:

```
EcsAgentCluster TaskDef          ANTHROPIC_MODEL / HAIKU  → global.
EcsAgentCluster PlanningTaskDef  ANTHROPIC_MODEL / HAIKU  → global.
```

### The rollback claim, re-tested

This is what the review correctly called false. `cdk diff -c bedrockGeoRegion=us` against the live stack now synthesizes:

| | value |
|---|---|
| IAM grants | `us.` ×15, `global.` ×0 |
| AgentCore main / aux | `us.` / `us.` |
| ECS main | `us.` |
| `BedrockGeoRegion` output | `us` |

Everything moves together from one context value. Previously the grants flipped to `us.` while the agent still asked for `global.anthropic.claude-opus-5`, so every no-override task failed at turn 0.

### Findings 3 and 4, live

Both new outputs are on the stack:

```
BedrockGeoRegion   global
BedrockModelIds    anthropic.claude-sonnet-4-6,…-opus-4-…,…-opus-4-8,…-opus-5,…-haiku-4-5-…
```

The `--model` guard reading them, against the deployed stack — all four paths, real exit codes:

| `--model` | exit | reason |
|---|---|---|
| `global.anthropic.claude-opus-5` | 0 | accepted |
| `global.anthropic.example-ungranted-model` | 1 | not granted |
| `us.anthropic.claude-opus-5` | 1 | wrong geography |
| `anthropic.claude-opus-5` | 1 | bare id |

The second row is the gap finding 3 identified — well-formed, right geography, granted nothing. It previously passed.

Doctor's reworded check:

```
[PASS] Bedrock model catalog (anthropic.claude-opus-5)
[PASS] Bedrock inference profile visible (global.anthropic.claude-opus-5)
```

The accept case wrote real state, so I removed the `model_id` attribute afterwards to restore the repo to unset.

### Still not verified

An end-to-end task on this exact commit. The prior deploy's Linear run proved the global grant works in practice (48 agent-role invocations of `global.anthropic.claude-opus-5`), but that ran before the env-var injection landed. What changed since is that the main model now arrives via env rather than via the Python literal — and the literal names the same value, so the behaviour should be identical. Saying that explicitly rather than implying a task run covered it.

isadeks:
## End-to-end task verified on this commit — the last open gap is closed

I flagged in the previous comment that no task had run against `b51712c6`, only against the earlier deploy. Now one has.

**Task `01M123VQVWHF82MQBWN9JZ29H7`** — COMPLETED in 1m01s, $0.17, opened [PR #576](https://github.com/isadeks/vercel-abca-linear/pull/576) with exactly the requested change (`+2/-1`, README.md only).

### Which model it actually invoked

From the Bedrock model-invocation log for the task window (17:21–17:25Z), split by caller, because the raw counts are misleading:

```
AGENT ROLE:
   10x  global.anthropic.claude-opus-5
    1x  global.anthropic.claude-haiku-4-5-20251001-v1:0

my user:
    4x  us.anthropic.claude-opus-4-8
    3x  us.anthropic.claude-opus-5
```

The agent role invoked **only** `global.` profiles — both the main model and the auxiliary one. The `us.` entries are my own CLI calls made while verifying the guard, under `user/sphia-dev`, not the workload.

That distinction is the whole point of splitting the log rather than reporting totals: the unsplit view shows `us.` traffic in the same window and reads like the geography did not take effect. It did.

### What this specifically confirms

The change since the last deploy is that the main model now arrives via `ANTHROPIC_MODEL` in the runtime env rather than from the Python literal in `config.py`. Both the injected value and the invoked profile are `global.anthropic.claude-opus-5`, so the injection is what the agent used — and the auxiliary model came through on the same geography, which is the divergence finding 1 was about.

All six original findings and all six supplemental findings are now either fixed and live-verified, or confirmed already-fixed. Nothing on either PR remains unverified by execution.

isadeks:
Superseded by #806, which shares this exact tip (`b51712c6`) and additionally closes #804 and #805. The two-round review history and all live verification are recorded in the comments above and carried forward into #806's description.

Closing to avoid merging the same commits twice and to keep the model-config doc conflicts from arising.

PR #808, opened 2026-08-27
chore(linear): token-lineage logging for OAuth revocation investigation (#807)

### LINKED ISSUE #807: feat(observability): Linear OAuth token-lineage logging for revocation investigation
## Summary

Add resolver-local **diagnostic logging** to the Linear OAuth refresh path
(`cdk/src/handlers/shared/linear-oauth-resolver.ts`) so we can investigate a recurring
per-workspace grant revocation. Additive observability only — **no behavior change** to the refresh
or persist logic.

## Motivation

One Linear workspace's OAuth grant is being revoked ~25h after each (re-)onboard: the platform's
refresh returns `invalid_grant / "Refresh token revoked"`. A second workspace on the identical code
path is durable, so the refresh code itself is sound (verified: the rotated `refresh_token` is
persisted on success). The cause is upstream/app-side, but today's logs can't reconstruct the
token's lifecycle to prove *what* Linear is rejecting — we can't distinguish "a spent/rotated token
was replayed" from "a live grant was killed server-side," nor correlate a failure to the token we
last persisted, nor see the token's age at death.

## Proposed logging (resolver-local, one file)

Threaded through `refreshLinearToken` / `tryRefreshOnce` / the persist path:

1. **Refresh-token fingerprint** — `sha256(refresh_token).slice(0,12)`, logged on every refresh
   outcome. Identifies *which* token across events; the raw token is **never** logged.
2. **Token age** — `token_age_h` (now − `installed_at`) and `since_last_refresh_h`
   (now − `updated_at`) on each outcome. Both fields already exist on the stored bundle; no schema
   change.
3. **Rotation trail** — on successful persist, log `rotated <oldFp> → <newFp>`, so the next
   failure can be correlated to the token we just wrote (server-side kill vs stale/race).
4. **Revocation forensics line** — at "permanently rejected," one structured line:
   `{workspace_id, refresh_token_fp, token_age_h, since_last_refresh_h, installed_at, updated_at}`.

## Non-goals

- No change to refresh/persist behavior, cache, or the `invalid_grant` re-read/race handling.
- No trigger/caller threading into other handlers (kept resolver-local to bound blast radius on a
  security-sensitive credential file; can follow up if the fingerprint+age data proves insufficient).
- Never log token material — fingerprints (truncated sha256) only.

## Acceptance criteria

- [ ] Every refresh outcome (success, `invalid_grant`, failure, permanently-rejected) logs the
      refresh-token fingerprint + `token_age_h` + `since_last_refresh_h`.
- [ ] Successful persist logs the old→new fingerprint rotation.
- [ ] No raw token/secret value appears in any log (verified against the masking gate).
- [ ] No behavior change: existing resolver tests pass unchanged; new tests assert the fingerprint
      is a truncated hash and never the raw token.

## Security / priority

Diagnostic-only, additive; touches a credential path so reviewed accordingly (masking gate). Suggest
**P1** — it unblocks an active, recurring auth outage investigation.

isadeks:
## What

Adds resolver-local **diagnostic logging** to the Linear OAuth refresh path
(`cdk/src/handlers/shared/linear-oauth-resolver.ts`) so we can investigate a recurring
per-workspace grant revocation. Additive observability only — **no behavior change** to refresh
or persist logic.

Closes #807.

## Why

One Linear workspace's OAuth grant is revoked ~25h after each (re-)onboard (refresh returns
`invalid_grant / "Refresh token revoked"`). A second workspace on the identical code path is
durable, so the refresh code itself is sound (verified: the rotated `refresh_token` is persisted on
success) and the cause is upstream/app-side. Today's logs can't reconstruct the token's lifecycle
to distinguish "a spent/rotated token was replayed" from "a live grant was killed server-side," nor
correlate a failure to the token we last persisted, nor show the token's age at death.

## Changes (all in the refresh path, no behavior change)

- **`refresh_token_fp`** — 12-char SHA-256 prefix of the refresh token, logged on every refresh
  outcome. Follows one token across events; the raw token is **never** logged.
- **`token_age_h` / `since_last_refresh_h`** — age since `installed_at` / `updated_at` (both fields
  already exist; no schema change). Surfaces the ~25h age-at-death.
- **Rotation trail** — on successful persist, `rotated <oldFp> → <newFp>`, so a later
  `invalid_grant` correlates to the exact token we wrote (server-side kill vs stale/race).
- **Forensics line** at "permanently rejected": `{refresh_token_fp, token_age_h,
  since_last_refresh_h, installed_at, updated_at}`.
- Persist-failure path logs the fp pair (rotated in-memory but not saved → next refresh replays a
  spent token).

## Tests

- New: fingerprint is a truncated hash and **the raw refresh token appears in NO log** (asserted
  across all levels); rotation trail logs old→new fp.
- Existing 14 resolver tests unchanged (behavior parity).
- Full `mise run build`: 2511 cdk tests pass, eslint + synth green.

## Notes

- Never logs token material — truncated SHA-256 fingerprints only.
- Kept resolver-local (one file) to bound blast radius on a security-sensitive credential path; no
  caller/trigger threading (can follow up if fingerprint+age proves insufficient).
- Does not change the `invalid_grant` re-read/race handling, cache, or refresh semantics.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this
contribution, under the terms of the project license.


isadeks:
**Live-verified on `backgroundagent-dev`** (deployed this branch; artifact-checked the deployed Lambda bundle carries `refresh_token_fp`; triggered the refresh path).

Real emitted log line:
```
[ERROR] Linear token refresh rejected
  refresh_token_fp: ec6844709eb5   token_age_h: 25.3   since_last_refresh_h: 25.3
  error_description: "Refresh token revoked"   status: 400
```

Confirmed: the fingerprint emits (not the raw token), age is captured, and **0** occurrences of any raw `lin_refresh_`/`lin_oauth_` token across all log lines in the window (no leak).

Bonus — the logging immediately paid off: it gives the first *quantified* read of the recurring revocation (`token_age_h == since_last_refresh_h == 25.3` → the grant died on its **first** refresh, never successfully rotated), narrowing that investigation toward an app-side/config cause rather than a refresh-code fault.

Note: one CodeQL `js/insufficient-password-hash` (high) on the fingerprint line is a false positive — it's a truncated HMAC fingerprint for **log correlation**, never stored or compared as an auth secret (the rule targets password *storage*). To be dismissed as such.

isadeks:
Superseded by #831, which carries this branch's token-lineage logging (HMAC fingerprint + token age + rotation interval in `linear-oauth-resolver.ts`) alongside the Identity vault work it was written to diagnose. Nothing here is dropped — closing to keep one PR for the arc.

Issue #852, opened 2026-09-03
CRITICAL: escape the CloudFormation 500-resource ceiling — best-practice strategies (8 resources of headroom on main)


scottschreckengaust:
> **Severity: CRITICAL / VITAL.** This is an active, deterministic build-and-deploy blocker, not a warning. ~~`main` has **8 of 500** CloudFormation resources of headroom on `compute_type=ecs`. The next feature that adds an authenticated route fails CI *and* cannot be deployed.~~ **`main` cannot synthesize at all** on `compute_type=lambda-microvm -c enableToolGateway=true` (**504/500**, throws). #681 already has.

> **📝 Edited 2026-09-03 to correct published errors.** Corrections are marked with ~~strikethrough~~ on the original claim followed by a bold **Correction:** note, so the record of what was wrong stays visible rather than being quietly rewritten. Several were my own arithmetic on unexamined premises (§3a, §3c, §5); the variant analysis was corrected by @theagenticguy in the comments and I re-verified the decisive cell myself. A consolidated list is in §9.

This issue is the **strategy and remediation surface** for the 500-resource-per-stack ceiling. #851 is the incident record (measurement, attribution, the failing run). This issue supersedes #851's provisional "Proposed remediation" section and is where options get iterated.

Everything below is **measured from a real synth** of `main` @ `521bf647` (CDK CLI 2.1129.0, `aws-cdk-lib` 2.261.0), not estimated. Reproduce commands are at the bottom. Re-verification noted where it happened at `81e71880`.

---

## 1. The constraint

CloudFormation allows **500 resources per stack**, ~~enforced at changeset creation~~. **Correction:** enforced *earlier* than that — CDK **throws at synth**, before CloudFormation is consulted (`MAX_RESOURCES = 500` in `aws-cdk-lib/core/lib/stack.js`: `if (numberOfResources > this.maxResources) throw ValidationError('TooManyResourcesInStack')`, with an `addInfo` band at 400–500). It is not a test-only limit: a real deploy of an over-budget branch is rejected before any resource is created.

~~| Context | Resources | Headroom |~~
~~|---|---|---|~~
~~| default (`agentcore`) | 480 | 20 |~~
~~| **`compute_type=ecs`** | **492** | **8** |~~
~~| `compute_type=ecs` + #681 | 501 | **over** |~~

[...]

Issue #851, opened 2026-09-03
P0: CloudFormation 500-resource-per-stack ceiling breached — main at 492/500, #681 at 501, blocks any PR adding a route


scottschreckengaust:
## Summary

`main` is **492 of 500** resources for `--context compute_type=ecs`. CloudFormation's per-stack resource ceiling is a **hard 500**, so the stack has **8 resources of headroom**. PR #681 adds 9 and has already crossed it — its `build (agentcore)` check fails with 28 test failures, all of them this:

```
«TooManyResourcesInStack» Number of resources in stack 'TestAgentStackEcs':
501 is greater than allowed maximum of 500
```

This is not a test artifact. CloudFormation enforces the limit at changeset creation, so a real `compute_type=ecs` deployment of #681's branch would be rejected the same way. Any PR that adds a Lambda plus API Gateway wiring — roughly 7–9 resources — now fails, regardless of whether the change is correct.

## Measured

Synthesized on a branch whose diff touches **zero** `cdk/` files, so these are `main`'s numbers (`main` @ `521bf647`):

| Context | Resources | Headroom vs 500 |
|---|---|---|
| default (agentcore) | 480 | 20 |
| `compute_type=ecs` | **492** | **8** |
| `compute_type=ecs` + PR #681 | **501** | **−1 (over)** |

`main`'s own post-merge build is green, confirming `main` is still (barely) under the ceiling. The breach is only visible on PRs that add resources.

Heaviest contributors in the 501-resource template:

| Type | Count |
|---|---|
| `AWS::Lambda::Permission` | 70 |
| `AWS::ApiGateway::Method` | 67 |
| `AWS::IAM::Role` | 58 |
| `AWS::IAM::Policy` | 56 |
| `AWS::Lambda::Function` | 50 |
| `AWS::ApiGateway::Resource` | 35 |
| `AWS::DynamoDB::Table` | 21 |

The API-Gateway-plus-Lambda pattern dominates: each new authenticated route costs a `Function`, a `Role`, a `Policy`, one or two `Resource`s, a `Method`, and a `Permission`.

## Why this needs its own issue

#735 tracks the **1 MB template-size** ceiling (98.4% as of 2026-08-06) and mentions 486/500 resources in passing, proposing a stack split that "also helps the 500-resource ceiling". #830 (same 1 MB wall) was closed `NOT_PLANNED`.

[...]

Issue #735, opened 2026-08-07
fix(cdk): agent stack template at 98.4% of the CloudFormation 1 MB limit


dreamorosi:
### Observed (live deploy, 2026-08-06, ADR-021 P2 smoke run)

With the Lambda MicroVMs backend + image configured, the synthesized `backgroundagent-dev` template measured **983,796 bytes of the 1,000,000-byte CloudFormation limit (98.4%)** and **486 of 500 resources**. Roughly one more construct of headroom before deploys start failing for reasons unrelated to whatever change triggers it.

### Impact

Any feature adding a handful of resources (P3 suspend/resume wiring is next in line) can push the stack over either limit, failing deploys with confusing errors.

### Proposal

- Short term: measure per-construct contribution (`cdk synth | wc -c` bisection) and prune obvious weight (duplicated policy statements, verbose metadata)
- Structural: split the stack (nested stacks or a second stack for optional backends), which also helps the 500-resource ceiling
- Add a synth-time size/resource-count warning at e.g. 90% so the next increment is caught at build time, not deploy time

Found during ADR-021 P2 verification (#645); not MicroVM-specific.

krokoko:
The template-size remediation is included in #912, pending review and merge.

All parent and nested templates now use compact JSON. The normal build enforces a shared 800,000-byte budget across 90 deployment profiles, including configured MicroVM images, Gateway, Registry, the Linear vault, and both network topologies. It also fails on CDK template-size warnings.

The managed-Blueprint offline census measured a maximum application template of 709,799 bytes with inline networking and 655,784 bytes with split networking. All 90 profiles synthesized twice without differences; the final repository build passed real bundled synthesis and the full profile guard suite.

#912 includes the measurements, reproduction command through the census task, and validation limitations, and will close this issue on merge.

PR #854, opened 2026-09-03
fix(cdk): reclaim resource and byte headroom under both CFN ceilings (#852)

scottschreckengaust:
Two configuration changes that pull the stack back inside CloudFormation's two template ceilings, with no architectural change and no API-contract change.

The widest cell of the deploy gate — `compute_type=lambda-microvm` with the ADR-019 tool gateway on — **could not synthesize on `main` at all**: 504 resources against the hard 500-resource template quota, which CDK enforces by throwing `TooManyResourcesInStack`. The same stack was also at ~100% of the 1 MB template-body ceiling, which CDK only *warns* about — so that limit was being ridden over silently.

#### Area

- [ ] `docs` — guides or design sources (`docs/guides/`, `docs/design/`)

#### Related

Refs #852 — Tier 0 of the strategies catalogued there. This is deliberately **not** a fix for #852; it buys headroom so the structural work (stack split) can be done under review rather than under a broken build. Adjacent: #851, #735.

#### Changes

**1. `allowTestInvoke: false` on all 34 `LambdaIntegration` call sites** — −30 resources app-wide.

The default (`true`) makes CDK emit a *second* `AWS::Lambda::Permission` per method, scoped to `method.testMethodArn`, so the API Gateway console's "TEST" button works. Nothing in this solution invokes it. Each one is also an extra `lambda:InvokeFunction` grant, so this narrows the IAM surface as well as the resource count.

Real traffic is unaffected. `scopePermissionToMethod` stays at its default `true`, so every route keeps its own narrowly-scoped `SourceArn`. That distinction matters: the two options are **alternatives, not additive** — setting `scopePermissionToMethod: false` saves the same 30 resources but widens every `SourceArn` to `.../{TaskApi}/*/*/*`, which is why #852 rejects it.

| File | Call sites |
|---|---:|
| `cdk/src/constructs/task-api.ts` | 20 |
| `cdk/src/constructs/slack-integration.ts` | 5 |
| `cdk/src/constructs/registry-api.ts` | 4 |
| `cdk/src/constructs/linear-integration.ts` | 2 |
| `cdk/src/constructs/jira-integration.ts` | 2 |
| `cdk/src/constructs/github-screenshot-integration.ts` | 1 |

**2. `suppressTemplateIndentation: true` on `AgentStack`** (`cdk/src/main.ts`) — ~31% fewer bytes, 0 resources.

Roughly a third of the emitted template was pretty-print indentation, every byte of which counted against the 1 MB ceiling. CDK's own `@aws-cdk/core:Stack.templateSize` warning names this prop as the remedy.

Set as a `StackProps` value rather than via the `@aws-cdk/core:suppressTemplateIndentation` context key. Both work — `Stack` resolves `props.suppressTemplateIndentation ?? node.tryGetContext(...) ?? false` — but the prop keeps the blast radius to the one stack that is actually near the ceiling. **Known limitation:** the two nested stacks (`AgentRegistryStack`, `RegistryApi`) therefore stay pretty-printed. They are small, and only the context key would reach them.

#### Measured

Resource counts and bytes are from `cdk synth` (which emits one `AWS::CDK::Metadata` that `Template.fromStack` does not), each run into its own `--output` directory.

| Deploy-gate cell | `main` | this PR |
|---|---|---|
| default (`agentcore`) | 480 res · 961,200 B (96.1%) | **450 res · 640,377 B (64.0%)** |
| `compute_type=ecs` | 492 res · 996,129 B (99.6%) | **462 res · 662,670 B (66.3%)** |
| `lambda-microvm` + `enableToolGateway` — **widest** | **504 res — `TooManyResourcesInStack`, no template emitted** | **474 res · 687,955 B (68.8%)** |

Every cell moves −30 resources exactly, and the widest one goes from "cannot synthesize" to 26 resources of headroom under the ceiling. The `enableToolGateway`-only cell is still measuring and will be added; it is a flat +7 on whichever substrate it is combined with.

CDK checks bytes against `TEMPLATE_BODY_MAXIMUM_SIZE = 1e6`, not 1,048,576 — percentages above use `1e6`.

#### Tests

Both gaps that let the broken cell stay green in CI are closed:

- `cdk/test/stacks/agent.test.ts` — new `AgentStack CloudFormation resource budget (#852)` block. It is the first test to construct `compute_type=lambda-microvm` **and** `enableToolGateway=true` together, which is precisely why the 504-resource cell was never red. `Template.fromStack` throws if the cell is over the hard ceiling, so reaching the assertions is itself part of the guard; the budget is set to 490 so it fails as a readable assertion ten resources before synth starts throwing.
- Same block — asserts no `AWS::Lambda::Permission` referencing `test-invoke-stage` is emitted anywhere in the stack, naming offending logical IDs. This is what stops a new route from silently reintroducing the 30 resources.
- `cdk/test/main.test.ts` — new `buildApp — CloudFormation template-body budget (#852)` block, reading the emitted artifact (not `Template.fromStack`, which has already parsed indentation away). Asserts the template carries no indentation and stays under 800,000 bytes — the same 80% point at which CDK starts warning, so a regression trips here instead of riding the warning band up to the hard limit.

#### Not in scope

- **`defaultCorsPreflightOptions`** (`task-api.ts`) is left exactly as it was. Dropping it would free a further 34 resources, but unlike the two changes here it alters the API contract, so it is a product decision. It is held in reserve as a contingency of roughly the same size as this whole PR.
- **API Gateway permission method-scoping** still has no test asserting it. #852 records that the suite passes with every `SourceArn` widened to a wildcard. Adding that guard is a separate change; this PR does not widen any `SourceArn`.
- **A `@aws-cdk/core:stackResourceLimit` context budget.** The built-in key already exists and is unset; wiring it is worth doing, but it applies one number to all stacks including nested ones, so it needs its own discussion.

#### Governance

✅ **Cleared under [ADR-003](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/blob/main/docs/decisions/ADR-003-contribution-governance.md).** #852 now carries `approved` + `P0` and is assigned; this PR is out of draft. (Superseded note: this section previously read that #852 lacked the `approved` label and an assignee, which was true when the PR was opened as a draft and is no longer.)

⚠️ **Branch name still non-conforming.** `fix/resource-byte-headroom` omits the issue number `AGENTS.md` requires (`(feat|fix|chore|docs)/<issue-number>-short-description`). Nothing rejects it today — `scripts/hooks/check-branch-name.mjs` lands with #679, still **open** as of 2026-09-03 — but it would be rejected once that merges. GitHub cannot re-point an open PR's head ref, so correcting this means close-and-reopen rather than a rename; not doing that unilaterally to an approved, mergeable PR. Raised in review by @ayushtr-aws and recorded on [#852](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/852#issuecomment-5531387755).

⬜ **Review nits deferred, not dropped.** @ayushtr-aws approved with 7 non-blocking nits; all are transcribed on [#852](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/852#issuecomment-5531387755) for follow-up PRs. None affects the deployed template — they concern the guards this PR adds over-claiming their coverage (notably `RegistryApi extends NestedStack`, so its 4 `LambdaIntegration` sites are outside the `test-invoke-stage` assertion).

#### Acknowledgment

### DIFF
--- a/cdk/src/main.ts
+++ b/cdk/src/main.ts
@@ -92,12 +92,7 @@ export async function buildApp(options: BuildAppOptions = {}): Promise<App> {
       env,
       agentCoreAvailabilityZones: azResolution.zones,
       description: 'ABCA Development Stack (uksb-wt64nei4u6)',
-      // Emit compact JSON. ~31% of this template is pretty-print indentation, and
-      // CloudFormation's 1 MB template-body ceiling counts those bytes. Unlike the
-      // 500-resource ceiling — which CDK enforces by throwing `TooManyResourcesInStack`
-      // — the byte ceiling only raises an `@aws-cdk/core:Stack.templateSize` warning,
-      // so it fails *open*: a template can grow past it and still synthesize. CDK's own
-      // warning text names this prop as the remedy. Refs #852.
+      // Emit compact JSON for a CloudFormation 1 MB template-body ceiling.
       suppressTemplateIndentation: true,
     },
   );
--- a/cdk/src/main.ts
+++ b/cdk/src/main.ts
@@ -92,6 +92,13 @@ export async function buildApp(options: BuildAppOptions = {}): Promise<App> {
       env,
       agentCoreAvailabilityZones: azResolution.zones,
       description: 'ABCA Development Stack (uksb-wt64nei4u6)',
+      // Emit compact JSON. ~31% of this template is pretty-print indentation, and
+      // CloudFormation's 1 MB template-body ceiling counts those bytes. Unlike the
+      // 500-resource ceiling — which CDK enforces by throwing `TooManyResourcesInStack`
+      // — the byte ceiling only raises an `@aws-cdk/core:Stack.templateSize` warning,
+      // so it fails *open*: a template can grow past it and still synthesize. CDK's own
+      // warning text names this prop as the remedy. Refs #852.
+      suppressTemplateIndentation: true,
     },
   );
 
--- a/cdk/test/main.test.ts
+++ b/cdk/test/main.test.ts
@@ -17,6 +17,7 @@
  *  SOFTWARE.
  */
 
+import * as fs from 'fs';
 import { App, Stack } from 'aws-cdk-lib';
 import { Annotations, Match, Template } from 'aws-cdk-lib/assertions';
 import {
@@ -151,5 +152,43 @@ describe('buildApp — AgentCore AZ wiring', () => {
         appProps: { context: { [AGENTCORE_AZS_CONTEXT_KEY]: 'us-east-1b' } },
       }),
     ).rejects.toThrow(/must be a JSON array of availability-zone names/);
+  });
+});
+
+describe('buildApp — CloudFormation template-body budget (#852)', () => {
+  // CloudFormation caps a template body at 1 MB; CDK checks against
+  // `TEMPLATE_BODY_MAXIMUM_SIZE = 1e6` and, unlike the 500-resource ceiling, only
+  // raises an `@aws-cdk/core:Stack.templateSize` *warning* — so this ceiling fails
+  // **open**. A template can grow past it, synthesize cleanly, and fail at deploy.
+  // These assertions read the emitted artifact rather than `Template.fromStack`,
+  // because indentation is the thing under test and `Template` has already parsed
+  // it away.
+  const CDK_TEMPLATE_BODY_MAXIMUM_SIZE = 1_000_000;
+  // CDK starts warning at 80% of the maximum. Budgeting to the same number means a
+  // regression trips this assertion at exactly the point CDK would start warning,
+  // instead of silently riding the warning band up to the hard limit.
+  const TEMPLATE_BODY_BUDGET = CDK_TEMPLATE_BODY_MAXIMUM_SIZE * 0.8;
+
+  let templateText: string;
+
+  beforeAll(async () => {
+    const built = await app();
+    templateText = fs.readFileSync(
+      built.synth().getStackByName(STACK_NAME).templateFullPath,
+      'utf8',
+    );
+  });
+
+  it('emits the template without pretty-print indentation', () => {
+    // `suppressTemplateIndentation: true` in main.ts makes CDK pass `undefined` as
+    // JSON.stringify's `space` argument, collapsing the document to a single line.
+    // Roughly 31% of this template was indentation bytes, all of which counted
+    // against the 1 MB ceiling. Assert the shape, not the saving, so the test does
+    // not need re-baselining every time a resource is added.
+    expect(templateText).not.toContain('\n  ');
+  });
+
+  it('stays inside the template-body budget', () => {
+    expect(Buffer.byteLength(templateText, 'utf8')).toBeLessThan(TEMPLATE_BODY_BUDGET);
   });
 });
--- a/cdk/test/main.test.ts
+++ b/cdk/test/main.test.ts
@@ -155,19 +155,19 @@ describe('buildApp — AgentCore AZ wiring', () => {
   });
 });
 
-describe('buildApp — CloudFormation template-body budget (#852)', () => {
+describe('buildApp — CloudFormation template-body budget within 80% 1Mb budget', () => {
   // CloudFormation caps a template body at 1 MB; CDK checks against
-  // `TEMPLATE_BODY_MAXIMUM_SIZE = 1e6` and, unlike the 500-resource ceiling, only
+  // `TEMPLATE_BODY_MAXIMUM_SIZE = 1e6` and only
   // raises an `@aws-cdk/core:Stack.templateSize` *warning* — so this ceiling fails
   // **open**. A template can grow past it, synthesize cleanly, and fail at deploy.
   // These assertions read the emitted artifact rather than `Template.fromStack`,
   // because indentation is the thing under test and `Template` has already parsed
   // it away.
   const CDK_TEMPLATE_BODY_MAXIMUM_SIZE = 1_000_000;
-  // CDK starts warning at 80% of the maximum. Budgeting to the same number means a
-  // regression trips this assertion at exactly the point CDK would start warning,
+  // Budget to the point CDK starts warning, so a regression trips a readable assertion
   // instead of silently riding the warning band up to the hard limit.
-  const TEMPLATE_BODY_BUDGET = CDK_TEMPLATE_BODY_MAXIMUM_SIZE * 0.8;
+  const WARNING_THRESHOLD = 0.8;
+  const TEMPLATE_BODY_BUDGET = CDK_TEMPLATE_BODY_MAXIMUM_SIZE * WARNING_THRESHOLD;
 
   let templateText: string;
 
@@ -180,10 +180,7 @@ describe('buildApp — CloudFormation template-body budget (#852)', () => {
   });
 
   it('emits the template without pretty-print indentation', () => {
-    // `suppressTemplateIndentation: true` in main.ts makes CDK pass `undefined` as
-    // JSON.stringify's `space` argument, collapsing the document to a single line.
-    // Roughly 31% of this template was indentation bytes, all of which counted
-    // against the 1 MB ceiling. Assert the shape, not the saving, so the test does
+    // Assert the shape, not the saving, so the test does
     // not need re-baselining every time a resource is added.
     expect(templateText).not.toContain('\n  ');
   });
--- a/cdk/test/stacks/agent.test.ts
+++ b/cdk/test/stacks/agent.test.ts
@@ -1576,6 +1576,57 @@ describe('AgentStack tool-gateway gate (ADR-019 P1)', () => {
   });
 });
 
+describe('AgentStack CloudFormation resource budget (#852)', () => {
+  // The 500-resource limit is a hard, non-adjustable CloudFormation template quota,
+  // and CDK enforces it by *throwing* `TooManyResourcesInStack` during synth. The
+  // widest cell of the deploy gate is `compute_type=lambda-microvm` with the ADR-019
+  // tool gateway on: the MicroVM substrate is additive over AgentCore, and the gateway
+  // adds a flat +7 on every substrate. Before #852 that pairing synthesized 504
+  // resources — `main` could not produce a template for it at all — yet CI stayed
+  // green because no test had ever constructed the two flags *together*. This block
+  // exists so that the widest cell, not the default one, is what guards the ceiling.
+  //
+  // Budget is deliberately below 500 so this fails as a readable assertion with a
+  // named remedy, ten resources before synth starts throwing.
+  const WIDEST_CONFIG_RESOURCE_BUDGET = 490;
+
+  let template: Template;
+  let resourceCount: number;
+
+  beforeAll(() => {
+    const app = new App({
+      context: { compute_type: 'lambda-microvm', enableToolGateway: true },
+    });
+    const stack = new AgentStack(app, 'WidestConfigStack', {
+      env: { account: '123456789012', region: 'us-east-1' },
+    });
+    // Throws `TooManyResourcesInStack` if this cell is over the hard ceiling, so
+    // reaching the assertions below is itself part of the guard.
+    template = Template.fromStack(stack);
+    resourceCount = Object.keys(template.toJSON().Resources ?? {}).length;
+  });
+
+  test('the widest deploy-gate cell stays inside the resource budget', () => {
+    // `Template.fromStack` counts one fewer than `cdk synth`, which also emits
+    // `AWS::CDK::Metadata`. Budget the synthesized number, so add that resource back.
+    expect(resourceCount + 1).toBeLessThanOrEqual(WIDEST_CONFIG_RESOURCE_BUDGET);
+  });
+
+  test('no Lambda permission is emitted for the API Gateway console test-invoke stage', () => {
+    // Every `LambdaIntegration` in this app passes `allowTestInvoke: false`. Left at
+    // its default `true`, CDK emits a second `AWS::Lambda::Permission` per method
+    // scoped to `method.testMethodArn` — 30 resources of the 500-resource budget, plus
+    // 30 extra `lambda:InvokeFunction` grants, so the API Gateway console's "TEST"
+    // button works. Nothing in this solution invokes it. Naming the offending logical
+    // IDs makes a regressed call site point straight at its own construct.
+    const offenders = Object.entries(template.findResources('AWS::Lambda::Permission'))
+      .filter(([, resource]) => JSON.stringify(resource).includes('test-invoke-stage'))
+      .map(([logicalId]) => logicalId);
+
+    expect(offenders).toEqual([]);
+  });
+});
+
 describe('AgentStack Agent Registry gate', () => {
   test.each([undefined, true, 'true'])(
     'enableAgentRegistry=%p includes the registry by default or explicit enablement',
--- a/cdk/test/stacks/agent.test.ts
+++ b/cdk/test/stacks/agent.test.ts
@@ -1576,55 +1576,64 @@ describe('AgentStack tool-gateway gate (ADR-019 P1)', () => {
   });
 });
 
-describe('AgentStack CloudFormation resource budget (#852)', () => {
+describe('AgentStack CloudFormation resource budget 500 with cushion', () => {
   // The 500-resource limit is a hard, non-adjustable CloudFormation template quota,
-  // and CDK enforces it by *throwing* `TooManyResourcesInStack` during synth. The
-  // widest cell of the deploy gate is `compute_type=lambda-microvm` with the ADR-019
-  // tool gateway on: the MicroVM substrate is additive over AgentCore, and the gateway
-  // adds a flat +7 on every substrate. Before #852 that pairing synthesized 504
-  // resources — `main` could not produce a template for it at all — yet CI stayed
-  // green because no test had ever constructed the two flags *together*. This block
-  // exists so that the widest cell, not the default one, is what guards the ceiling.
-  //
-  // Budget is deliberately below 500 so this fails as a readable assertion with a
-  // named remedy, ten resources before synth starts throwing.
-  const WIDEST_CONFIG_RESOURCE_BUDGET = 490;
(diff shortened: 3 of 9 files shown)


cdk/src/main.ts
```diff
@@ -92,6 +92,13 @@ export async function buildApp(options: BuildAppOptions = {}): Promise<App> {
       env,
       agentCoreAvailabilityZones: azResolution.zones,
       description: 'ABCA Development Stack (uksb-wt64nei4u6)',
+      // Emit compact JSON. ~31% of this template is pretty-print indentation, and
+      // CloudFormation's 1 MB template-body ceiling counts those bytes. Unlike the
+      // 500-resource ceiling — which CDK enforces by throwing `TooManyResourcesInStack`
+      // — the byte ceiling only raises an `@aws-cdk/core:Stack.templateSize` warning,
+      // so it fails *open*: a template can grow past it and still synthesize. CDK's own
+      // warning text names this prop as the remedy. Refs #852.
```
scottschreckengaust: This comment is too verbose.
```suggestion
      // Emit compact JSON for a CloudFormation 1 MB template-body ceiling.
```

cdk/test/main.test.ts
```diff
@@ -153,3 +154,41 @@ describe('buildApp — AgentCore AZ wiring', () => {
     ).rejects.toThrow(/must be a JSON array of availability-zone names/);
   });
 });
+
+describe('buildApp — CloudFormation template-body budget (#852)', () => {
+  // CloudFormation caps a template body at 1 MB; CDK checks against
+  // `TEMPLATE_BODY_MAXIMUM_SIZE = 1e6` and, unlike the 500-resource ceiling, only
```
scottschreckengaust: The 500-resource ceiling is not relevant
```suggestion
describe('buildApp — CloudFormation template-body budget within 80% 1Mb budget', () => {
  // CloudFormation caps a template body at 1 MB; CDK checks against
  // `TEMPLATE_BODY_MAXIMUM_SIZE = 1e6` and only
```

cdk/test/main.test.ts
```diff
@@ -153,3 +154,41 @@ describe('buildApp — AgentCore AZ wiring', () => {
...
+      built.synth().getStackByName(STACK_NAME).templateFullPath,
+      'utf8',
+    );
+  });
+
+  it('emits the template without pretty-print indentation', () => {
+    // `suppressTemplateIndentation: true` in main.ts makes CDK pass `undefined` as
+    // JSON.stringify's `space` argument, collapsing the document to a single line.
+    // Roughly 31% of this template was indentation bytes, all of which counted
+    // against the 1 MB ceiling. Assert the shape, not the saving, so the test does
```
scottschreckengaust: Numbers are not relevant
```suggestion
    // Assert the shape, not the saving, so the test does
```

cdk/test/stacks/agent.test.ts
```diff
@@ -1576,6 +1576,57 @@ describe('AgentStack tool-gateway gate (ADR-019 P1)', () => {
...
+  // widest cell of the deploy gate is `compute_type=lambda-microvm` with the ADR-019
+  // tool gateway on: the MicroVM substrate is additive over AgentCore, and the gateway
+  // adds a flat +7 on every substrate. Before #852 that pairing synthesized 504
+  // resources — `main` could not produce a template for it at all — yet CI stayed
+  // green because no test had ever constructed the two flags *together*. This block
+  // exists so that the widest cell, not the default one, is what guards the ceiling.
+  //
+  // Budget is deliberately below 500 so this fails as a readable assertion with a
+  // named remedy, ten resources before synth starts throwing.
+  const WIDEST_CONFIG_RESOURCE_BUDGET = 490;
```
scottschreckengaust: Details that may change use constants:
```suggestion
describe('AgentStack CloudFormation resource budget 500 with cushion', () => {
  // The 500-resource limit is a hard, non-adjustable CloudFormation template quota,
  // and CDK enforces it by *throwing* `TooManyResourcesInStack` during synth. 
  // This block exists so that the widest cell, not the default one, is what guards the ceiling.
  // Budget is deliberately below 500 so this fails as a readable assertion with a
  // named remedy before synth starts throwing.
  const MAX_RESOURCE_BUDGET = 500;
  const CUSHION = 10;
  const WIDEST_CONFIG_RESOURCE_BUDGET = MAX_RESOURCE_BUDGET - CUSHION;
```

cdk/test/stacks/agent.test.ts
```diff
@@ -1576,6 +1576,57 @@ describe('AgentStack tool-gateway gate (ADR-019 P1)', () => {
...
+
+  let template: Template;
+  let resourceCount: number;
+
+  beforeAll(() => {
+    const app = new App({
+      context: { compute_type: 'lambda-microvm', enableToolGateway: true },
```
scottschreckengaust: Although the `lambda-microvm` is the largest, shouldn't all compute types be tested?

cdk/test/stacks/agent.test.ts
```diff
@@ -1576,6 +1576,57 @@ describe('AgentStack tool-gateway gate (ADR-019 P1)', () => {
...
+    expect(resourceCount + 1).toBeLessThanOrEqual(WIDEST_CONFIG_RESOURCE_BUDGET);
+  });
+
+  test('no Lambda permission is emitted for the API Gateway console test-invoke stage', () => {
+    // Every `LambdaIntegration` in this app passes `allowTestInvoke: false`. Left at
+    // its default `true`, CDK emits a second `AWS::Lambda::Permission` per method
+    // scoped to `method.testMethodArn` — 30 resources of the 500-resource budget, plus
+    // 30 extra `lambda:InvokeFunction` grants, so the API Gateway console's "TEST"
+    // button works. Nothing in this solution invokes it. Naming the offending logical
+    // IDs makes a regressed call site point straight at its own construct.
```
scottschreckengaust: remove numbers that may change over time.
```suggestion
    // scoped to `method.testMethodArn` — removing the API Gateway console's "TEST"
    // Naming the offending logical IDs makes a regressed call site point straight at its own construct.
```

cdk/src/main.ts
scottschreckengaust: Applied as suggested, verbatim. Down to the one line — the ceiling is the fact worth stating inline; the fail-open behaviour now lives only in the test that guards it.

cdk/test/main.test.ts
scottschreckengaust: Applied. Dropped the 500-resource comparison and renamed the block to `buildApp — CloudFormation template-body budget within 80% 1Mb budget`.

One addition beyond the suggestion: `0.8` was inline in the `TEMPLATE_BODY_BUDGET` expression, so I named it `WARNING_THRESHOLD` — same reasoning as your constants comment on `agent.test.ts`.

cdk/test/main.test.ts
scottschreckengaust: Applied. The 31% figure is gone from the code; it survives only in the PR description, where it dates itself.

cdk/test/stacks/agent.test.ts
scottschreckengaust: Applied:

```ts
const MAX_RESOURCE_BUDGET = 500;
const CUSHION = 10;
const RESOURCE_BUDGET = MAX_RESOURCE_BUDGET - CUSHION;
```

Extended the same treatment one step further: the assertion previously read `resourceCount + 1`, where the `1` was an unexplained `cdk synth`-vs-`Template.fromStack` delta. It is now `SYNTH_ONLY_RESOURCES`.

cdk/test/stacks/agent.test.ts
scottschreckengaust: Yes — changed. It now covers every deploy-gate cell, not just the widest:

```ts
const COMPUTE_TYPES = ["agentcore", "ecs", "lambda-microvm"];
const CELLS = COMPUTE_TYPES.flatMap(computeType =>
  [false, true].map(enableToolGateway => ({ computeType, enableToolGateway })),
);
describe.each(CELLS)("compute_type=$computeType enableToolGateway=$enableToolGateway", ...)
```

3 compute types x `enableToolGateway` on/off = 6 cells, 12 tests. Measured cost: 20.5s, and the file goes 125 -> 135 tests.

The reason testing only the widest is insufficient: `enableToolGateway` is a flat +7 across all three substrates, so the *ordering* of the cells is stable but the *headroom* is not — `ecs` sat at 492/500 while the default sat at 480. A regression that adds resources on only one substrate (an `ecs`-only construct, say) would leave `lambda-microvm` untouched and pass a widest-only guard. Covering every cell also means each one is actually synthesized, and since CDK *throws* `TooManyResourcesInStack` rather than warning, reaching the assertion at all is part of the guard.

cdk/test/stacks/agent.test.ts
scottschreckengaust: Applied the intent; the literal suggestion was not applicable as written — the ```suggestion``` block collapses two non-adjacent lines and drops the sentence between them, so committing it verbatim would have left `// scoped to `method.testMethodArn` — removing the API Gateway console's "TEST"` as a fragment with no subject.

What is there now keeps both thoughts intact and carries no numbers:

```ts
// scoped to `method.testMethodArn` — removing the API Gateway console's "TEST"
// button, which nothing in this solution invokes, and its extra
// `lambda:InvokeFunction` grant.
// Naming the offending logical IDs makes a regressed call site point straight
// at its own construct.
```

Two disclosures on scope, both beyond what you flagged:

1. I applied the same number-trim to the `task-api.ts` header comment, which no review comment covered — it had stated the per-route and total permission counts. Those drift the moment a route is added, which is the failure mode you are pointing at here. Revert that hunk if you would rather keep the numbers documented at the call site.
2. Fixing the wrong test filename in that same `task-api.ts` comment (it named a file that does not exist) was the separate item, and is in this commit too.

scottschreckengaust:
* #852 is approved and self-assigned to @scottschreckengaust 
* naming convention adherence honest miss but renaming doesn't add value

ayushtr-aws (approved):
## Principal architect review — PR #854

**Verdict: Approve with nits.** The two configuration changes are correct, strictly narrow IAM, and move the widest deploy-gate cell from "cannot synthesize" to 26 resources of headroom. All affected suites pass locally on the merged branch (`main.test.ts`, `stacks/agent.test.ts`, `constructs/task-api.test.ts`, `bootstrap/synth-coverage.test.ts` — 187/187). The nits below are about the *guards* the PR adds over-claiming what they cover; none affects the deployed template.

### Vision alignment
Fits. Tenet 5 (least privilege): 30 fewer `lambda:InvokeFunction` grants, every route keeps its per-method `SourceArn`. Tenet 6 (bounded cost/capacity): converts a fail-open INFO/WARN signal into a red test. Tenet 10 (honest sample): PR body is explicit that this is Tier 0 headroom, not the #852 fix. No tenet traded away, so no ADR needed.

### Governance
- #852 now carries `approved` + `P0` and is assigned — cleared under ADR-003. The PR body's Governance section is stale; please refresh it.
- Branch `fix/resource-byte-headroom` omits the issue number required by AGENTS.md. Rename to `fix/852-resource-byte-headroom` (the #679 hook will reject the current name once it lands).

### Blocking issues
None.

### Non-blocking (recommend fixing before merge, small)
1. **`cdk/test/stacks/agent.test.ts:1621` — nested-stack routes are unguarded.** `Template.fromStack(AgentStack)` returns only the root template; `RegistryApi extends NestedStack`, so its 4 `LambdaIntegration` sites are not covered by the `test-invoke-stage` assertion, while the comment in `task-api.ts:849-853` says the test "asserts no `test-invoke-stage` permission is ever emitted". Add the same offenders filter to `test/constructs/registry-api.test.ts` (or iterate `NestedStack` children in the budget block) and soften the task-api comment.
2. **`cdk/test/stacks/agent.test.ts:1607` — `lambda-microvm` cells synth without `microvm_base_image_arn`/`microvm_base_image_version`**, so `AWS::Lambda::MicrovmImage` and the `TerminateMicrovm` grant are omitted. The "widest cell" measured is the bootstrap no-image state (474), not the configured deployment (475). The existing microvm fixture at ~L1054 passes those keys for exactly this reason; add them to the microvm cells.
3. **`cdk/test/main.test.ts:174` — byte budget measured only on the default cell.** `buildApp` accepts `appProps.context`; parametrize with the same six-cell matrix so a byte regression confined to ECS/MicroVM/ToolGateway constructs cannot hide. (Branch numbers: agentcore/off ≈ 669 KB guarded, microvm/on ≈ 717 KB unguarded.)
4. **`cdk/test/stacks/agent.test.ts:1583` — header comment accuracy.** The unchanged describe at ~L1518 already constructs `enableToolGateway=true` + `compute_type=ecs`, so "no test had ever constructed [them] together" is not true (the *microvm* + gateway combination was new). Also CDK's thrown message is "Number of resources in stack … is greater than allowed maximum of 500"; `TooManyResourcesInStack` is an internal tag a reader will never see in output.
5. **`cdk/test/stacks/agent.test.ts:1591` — `SYNTH_ONLY_RESOURCES = 1`** is a fudge for `AWS::CDK::Metadata`, which is absent only because the test `App` never sets `aws:cdk:version-reporting`. Setting that context key in the budget cells (and optionally `@aws-cdk/core:stackResourceLimit`, which CDK's `maxResources` honors with a per-type breakdown) removes the constant.
6. **`cdk/src/main.ts:96` — nested stacks stay pretty-printed.** Already acknowledged in the PR body; noting that `NestedStack` does *not* inherit the prop and `cdk.json` has no `@aws-cdk/core:suppressTemplateIndentation` context, so each nested stack has to rediscover this. Fine to defer to the #852 stack-split work.
7. **`cdk/test/main.test.ts:166`** — the hand-copied `1_000_000 × 0.8` duplicates CDK's own `Stack.templateSize` warning; `Annotations.fromStack(stack).hasNoWarning('*', Match.stringLikeRegexp('Template size'))` would reuse CDK's signal and cover nested stacks for free. Keep the raw read only for the indentation assertion.

### Documentation
- No `docs/guides`/`docs/design` change was required (no contract, env var, or command change), so no Starlight mirror regen needed.
- Suggest one bullet under `cdk/AGENTS.md` "Common mistakes": new `LambdaIntegration` sites must pass `allowTestInvoke: false` and new routes count against the resource-budget test. Optional: one operator-facing line that the API Gateway console "TEST" button no longer works by design.
- Issue tracking: #852 (P0, approved) is the tracking issue; #851 is the incident record.

### Tests & CI
- Locally green (see above). CI checks were still queued at review time.
- Bootstrap synth-coverage: **not applicable** — no new CFN resource types; the PR only removes `AWS::Lambda::Permission` resources. Suite still passes.
- Test performance: budget block synths six cells once each in `beforeAll` (no per-test synth, bundling stays disabled) — ~11 s for the two files. Acceptable.

### Review agents run
- `/code-review` (high) — source of nits 1–5 and 7; each spot-verified against the branch.
- `/security-review` — no new vulnerabilities. Verified from `aws-cdk-lib` 2.261.0 source that `allowTestInvoke:false` only drops the `test-invoke-stage` permission (`scopePermissionToMethod` branch unchanged, per-method `SourceArn` retained) and that `suppressTemplateIndentation` only changes `JSON.stringify` indent.
- Omitted: `silent-failure-hunter` (no error-handling code in diff), `type-design-analyzer` (no new types), `comment-analyzer` (folded into code-review nit 4), `pr-test-analyzer` (folded into code-review nits 1–3).

### Human heuristics
- **Proportionality** — pass. Two flags plus two tests for a P0 blocker; no new abstraction.
- **Coherence** — pass. Same option at all 34 call sites; budget constants named consistently.
- **Clarity** — concern: `task-api.ts:849-853` and `agent.test.ts:1583` claim coverage the tests do not provide (nits 1, 4).
- **Appropriateness** — pass. Verified against real CDK behaviour (synth throws, permission ARNs), not only mocks; tests assert what the template *should* contain.

scottschreckengaust:
@ayushtr-aws — thank you, genuinely, for the thoroughness here. Verifying `allowTestInvoke: false` against `aws-cdk-lib` 2.261.0 source to confirm the `scopePermissionToMethod` branch and per-method `SourceArn` are untouched is exactly the check that made this safe to approve, and the seven nits landed on the real weakness of the PR: the guards claim more coverage than they provide.

All of it has been recorded on the tracking issue — **#852** ([comment](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/issues/852#issuecomment-5531387755)) — to be evaluated and remediated in follow-up PRs, grouped as:

1. **Guard accuracy** — the nested-stack blind spot (nit 1: `RegistryApi extends NestedStack`, so `Template.fromStack(AgentStack)` never sees its 4 `LambdaIntegration` sites) plus the two inaccurate comments (nit 4), and the `cdk/AGENTS.md` "Common mistakes" bullet you suggested. Nit 1 is the one I'd rank highest of the seven: a comment in `task-api.ts` asserting coverage that does not exist is precisely how the next contributor adds an unguarded route. The `AGENTS.md` bullet is the durable half of that fix.
2. **Budget-cell fidelity** — the missing `microvm_base_image_arn`/`microvm_base_image_version` context (nit 2, which corrects the widest-cell figure from the bootstrap no-image 474 to the configured 475), the `SYNTH_ONLY_RESOURCES` fudge (nit 5 — and `@aws-cdk/core:stackResourceLimit`'s per-type breakdown is directly useful to #852's resource accounting), and the six-cell byte-budget parametrization (nits 3 and 7 together).
3. **With the stack split** — `@aws-cdk/core:suppressTemplateIndentation` in `cdk.json` for nested stacks (nit 6), where your `Annotations.fromStack(...).hasNoWarning('*', /Template size/)` suggestion is what makes the nested-stack case testable at all rather than a second hand-copied threshold.

Two governance points from your review are **not** deferrable and are noted separately on #852:

- The stale Governance section in the PR body — correcting that now.
- The branch name `fix/resource-byte-headroom` missing the issue number. GitHub cannot re-point an open PR's head ref, so this is a close-and-reopen rather than a rename; leaving that call to @scottschreckengaust rather than churning an approved, mergeable PR unilaterally. Worth noting for anyone reading later: the #679 hook that would reject this name is still **open**, not merged, so it is not yet enforcing.

On the `TooManyResourcesInStack` point — correct, and worth more than a comment fix: that string is an internal identifier, so anyone grepping deploy output for it finds nothing. The message a human actually sees is `Number of resources in stack … is greater than allowed maximum of 500`.

PR #895, opened 2026-09-15
feat(jira): surface deployment preview feedback (#698)

### LINKED ISSUE #698: feat(jira): surface deployment preview screenshots on originating issues
## Component

CDK GitHub deployment processor, Jira adapter, orchestration

## Context

PR #695 captures a successful deployment preview, stores the screenshot, comments on the GitHub PR, persists preview metadata on the task, and mirrors the result into Linear iteration replies or orchestration rollups.

The Jira origin is not resolved by that path. Jira users can see the GitHub PR comment, but the originating Jira issue and Jira parent rollup receive no preview feedback. This gap was called out in the #580 implementation note.

## Describe the feature

When a Jira-origin task produces a deployment preview screenshot, surface that preview on the correct Jira issue. For an orchestrated integration result, surface the combined preview on the Jira parent rollup. For an iteration, converge with the maturing Jira status comment from #697.

## Proposed solution

- Resolve the deployment back to the authoritative task record and route by its stored `channel_source` and channel metadata; do not infer Jira identity from a branch-name issue key.
- Keep the existing GitHub PR screenshot comment unchanged.
- For a normal Jira task, post one app-authored Jira comment containing safe links to the captured screenshot and live preview. Use a Jira-supported ADF representation; if external images cannot be embedded safely, render explicit screenshot and preview links rather than dropping the feedback.
- For a Jira iteration, append or upsert the preview block into the maturing comment from #697 and preserve it across later heartbeat/terminal writes.
- For a synthetic orchestration integration node, do not address the synthetic node as a Jira issue. Persist the preview on the task and refresh the parent Jira rollup through the shared orchestration channel.
- Make delivery idempotent across duplicate GitHub deployment events and repeated status updates.
- Treat Jira API/auth failures as observable, best-effort feedback failures that never fail the deployment or coding task.

## Acceptance criteria

- A successful preview deployment for a Jira-origin task still comments on the GitHub PR and also surfaces the screenshot/live-preview links on the correct Jira issue.
- A Jira iteration updates the existing maturing status comment from #697 rather than posting a competing preview comment.
- A fan-out orchestration's integration-node preview appears on the Jira parent rollup and is never sent to a synthetic issue ID.
- Duplicate deployment events do not create duplicate Jira comments or duplicate preview blocks.
- Overlapping preview, heartbeat, and terminal writers converge without losing the preview or regressing terminal state.
- A task from another channel is never routed to Jira, even if its branch text resembles a Jira key.
- Invalid/untrusted preview URLs remain subject to the existing screenshot allowlist and URL-encoding controls.
- Linear and GitHub PR screenshot behavior remain unchanged.

## Tests

- GitHub deployment-processor tests cover Jira standalone, Jira iteration, Jira integration node, non-Jira routing, missing metadata, duplicate events, and Jira delivery failure.
- Jira rendering tests cover ADF output and safe fallback links.
- Orchestration rollup tests cover the combined Jira preview and synthetic-node guard.
- Race tests cover preview preservation across heartbeat and terminal updates.
- Update `JIRA_SETUP_GUIDE.md` and regenerate the Starlight mirror.

## Out of scope

- Screenshot capture, storage, URL allowlisting, or GitHub PR comments; those already exist in #695.
- Generic Jira attachment upload or Atlassian media storage.
- Maturing Jira iteration status comments themselves; tracked in #697.

ayushtr-aws:
Jira-origin deployment previews appear on the originating issue as explicit ADF screenshot and live-preview links. Iterations update their existing status comment, and synthetic integration nodes surface the combined preview in the parent rollup.

#### Area

- [x] `docs` — guides or design sources

#### Related

Fixes #698.

Companion #900 records the Amplify `check_run` trigger scope and review follow-ups N3–N5. Its maintainer approval is pending; #698's approval does not cover that trigger.

#### Changes

- Resolve Jira routing from stored task metadata and the deployment commit SHA. Bound lookup to 10 pages and 25 candidate reads, require a matching iteration SHA, and reject synthetic issue targets.
- Persist Jira iteration comment state and merge durable preview metadata across heartbeat, terminal, and preview writers. Re-read after each Jira update to repair overlapping stale writes. Successful PUTs remain successful when convergence reaches its repair limit or a subsequent verification read fails; a missing task is explicitly non-retryable.
- Reserve 30 seconds after capture, cap optional Jira delivery at 20 seconds within the processor's remaining deadline, and cancel DynamoDB and Jira comment requests on expiry. An auth lookup that completes after cancellation cannot start a comment request.
- Claim standalone preview creation before posting to prevent duplicate comments. Reject invalid screenshots before claiming. Failed or uncertain creates remain claimed and are logged; Jira feedback failures never fail the coding task or deployment.
- Skip both channel deliveries on task-persistence lookup failures, preventing Jira tasks from falling through to Linear branch-name routing. Document routing, delivery limits, and operator log events; regenerate the Starlight mirror.
- Wire the existing Jira registry, scoped OAuth access, Jira task index reads, and heartbeat comment-state updates. Normalize string-valued runtime metrics before rendering fallback status.

The existing Amplify addition accepts signed successful PR-preview `check_run` completions and sends them through the screenshot pipeline, with separate event deduplication. #900 tracks scope approval plus observable rejection reasons, environment-filter compatibility, and preserving the validated PR number through processing.

#### Validation

- `mise run build` passed: 4,715 CDK tests, 990 CLI tests, agent quality/tests, documentation build, and CDK synthesis.
- Final focused run: 351 tests passed across preview delivery, convergence, webhook processing, terminal writers, and Jira transports.
- Mutation verification detected all 13 deliberate regressions: removal of the repair loop, terminal-state protection, consistent reads, claim predicate, synthetic-target guard, lookup limits, and each of the six terminal flags. Race tests assert write completion order; the reconciler's exhausted-retry fallback is covered.
- Pre-commit checks passed. Secret/dependency/workflow and Retire scans passed, as did SAST and error-masking scans relative to the baseline. The full pre-push security gate is blocked by seven pre-existing findings: one `Object.assign` finding in `linear-webhook-processor.ts` (introduced September 4) and six error-masking findings in unchanged files.
- Grype source scan passed. The agent image built successfully with Finch; Trivy reported 14 HIGH and 0 CRITICAL findings in bundled dependencies (brace-expansion, Go dependencies, and quinn-proto). Agent image sources and dependency manifests are unchanged by this revision.
- Prior live verification on TechU `backgroundagent-dev`: Amplify completions produced AgentCore Browser screenshots on GitHub and Jira TG-22, duplicate replay was deduplicated, and a subsequent Jira iteration automatically received updated preview links. The review fixes have been validated locally.

#### Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution under the terms of the project license.


isadeks (changes requested):
## Verdict: Request changes

The convergence design here is genuinely good — persisting `{body, terminal}` and having every writer re-render from durable state, rather than doing a Jira read/modify/write, is the right answer to a hard multi-writer problem, and the "persist only your own body, never the merged output" invariant is what keeps the preview from duplicating. Two verified defects sit on top of it that produce duplicate customer-visible comments — the exact failure class #698 set out to eliminate — and the tests that should have caught one of them do not hold under mutation.

Governance and CI are otherwise clean: #698 carries `approved`, the branch follows `feat/<issue>-<slug>`, all 11 checks are green, and the branch merges cleanly on `main`.

---

## Vision alignment

Fits. **Tenet 7 (observable, attributable)** — a Jira-origin task's preview now lands on the originating issue instead of only the GitHub PR, and the orchestration rollup gets the combined preview. **Tenet 1 (fire-and-forget)** — delivery stays asynchronous and best-effort; a Jira failure never fails the coding task. **Tenet 4 (fail closed)** is respected on the input path: the Amplify normalizer rejects rather than guesses. One tenet-adjacent gap: the best-effort contract is applied more broadly than the evidence supports (N1). No ADR owed.

## Blocking issues

### B1 — Unbudgeted Jira delivery inside the 8-second post-capture reserve → duplicate PR screenshot comments
`cdk/src/handlers/github-webhook-processor.ts:316`

The processor threads one wall-clock deadline (`TOTAL_BUDGET_MS = 110_000`, `:73`) and reserves `POST_CAPTURE_RESERVE_MS = 8_000` (`:81`), whose own docstring scopes it to "S3 PUT + GitHub PR comment POST + the 2s page settle". `deliverJiraDeploymentPreview` is `await`ed inside that reserve and contains **no** reference to `remaining()`, a deadline, or a budget — grep returns zero hits in both new modules. Worst case it performs up to 10 Queries plus **up to 250 sequential `ConsistentRead` GetItems** (`MAX_LOOKUP_PAGES = 10` × `LOOKUP_PAGE_SIZE = 25` at `jira-deployment-preview.ts:29-30`, one Get per candidate at `:71`), then a Secrets Manager fetch with a possible OAuth refresh, then 1–4 Jira PUTs each followed by another consistent Get.

The consequence is not just a slow Lambda. The PR screenshot comment is posted *before* this call (`:285-292`) with `existingCommentId: undefined` — "always POST fresh". The processor is async-invoked (`InvocationType: 'Event'`, `github-webhook.ts:191`) with a DLQ. So a hard timeout inside Jira delivery triggers Lambda's default async retries, each of which re-captures and posts **another** screenshot comment on the customer's PR, then lands in `WebhookProcessorDlq` and fires `WebhookProcessorDlqDepthAlarm` — whose own comment reads "each one is a screenshot pipeline silently down." The construct comment at `github-screenshot-integration.ts:204-208` states the invariant this breaks: *"no individual sub-step can run past the Lambda timeout even on the worst-case path."*

Reachable on any long-lived Jira issue with many iterations — 250 sequential round trips can exhaust 8s before a single Jira PUT. It also breaks #698's own criterion that GitHub PR screenshot behaviour remain unchanged.

**Fix:** pass `remaining()` in and bail with a `logger.warn` below a floor; cap total per-candidate Gets rather than only pages; raise `POST_CAPTURE_RESERVE_MS` to cover the added phase.

### B2 — Convergence exhaustion reports failure after every PUT succeeded → duplicate terminal Jira comment
`cdk/src/handlers/shared/jira-preview.ts:111-119`

A non-ok PUT returns early at `:114`, so reaching `:118` means all four `updateIssueCommentAdf` calls **succeeded** — the Jira comment is correct. It nonetheless returns `{ ok: false, retryable: true }`. At `fanout-task-events.ts:1696-1714` the caller releases the reply claim and, because `retryable` is true and the release was not `exhausted`, **throws**; the retry re-claims and re-runs the terminal dispatch, so `postIssueCommentAdf` at `:1719` posts a second terminal comment. `orchestration-reconciler.ts:1258-1266` has the same shape.

This is a behaviour change introduced here: before this PR these call sites invoked `updateIssueCommentAdf` directly, which could only report failure when the PUT itself failed.

The same line has a second case. If the task row is absent (`load()` returns `undefined` at `:110`), the loop body never executes, **zero PUTs are attempted**, and it still reports `jira.preview.convergence_failed` with `retryable: true` — a misleading diagnostic that spends the Lambda retry budget on a state `ConsistentRead: true` has already proven permanent.

**Fix:** track whether any PUT succeeded; on exhaustion with ≥1 success return `{ ok: true }` and keep the existing warn. Reserve `ok: false` for real transport/API failures, and handle `!task` explicitly as non-retryable with its own event.

### B3 — The tests that would have caught B2 cannot fail
Mutation-verified in an isolated sandbox: setting `MAX_CONVERGENCE_ATTEMPTS = 1` (`jira-preview.ts:32`) deletes the entire re-read/repair loop the module docblock exists for, and **all 14 tests still pass**. The three "race" tests (`cdk/test/handlers/shared/jira-preview.test.ts:82,105,126`) assert `put.mock.calls.at(-1)` — *call* order — while the behaviour at stake is *resolution* order. A resolution-order variant of the same test passes on baseline and fails on that mutant with a stale `"Working"` body, i.e. the real regression is a heartbeat's `🔄 Working…` permanently overwriting the terminal status in Jira.

Also unprotected:
- **`terminal: true` at all six new call sites** in `fanout-task-events.ts` and `orchestration-reconciler.ts` — both mocks forward only `status.body`, and neither suite asserts the flag. Flipping it to `false` is invisible, and that mutation *is* #698's "never regressing terminal state".
- `attribute_not_exists(jira_preview_claimed)` (`jira-deployment-preview.ts:119`) — the test fake re-implements the condition itself, so deleting it from the `ConditionExpression` leaves 15/15 green.
- `isIntegrationNode(target)` on the iteration path (`jira-deployment-preview.ts:106`) — mutating it to `if (!replyId)` leaves 15/15 green. This is the criterion "never sent to a synthetic issue ID".
- `ConsistentRead: true` (`jira-preview.ts:108`), the terminal clause in the `ConditionExpression` (`:97`), and `MAX_LOOKUP_PAGES`.

**Fixes, all cheap:** push to a `settled: string[]` inside the `put` mock implementation and assert `settled.at(-1)`, plus `expect(put).toHaveBeenCalledTimes(3)`; forward `status` whole in the two mocks and add `expect.objectContaining({ terminal: true })` to the existing `toHaveBeenCalledWith`; assert the `ConditionExpression` strings directly; add one `test.each` row with `trigger_comment_issue_id: 'orch__integration'` asserting `expect(iteration).not.toHaveBeenCalled()`.

### B4 — Operator docs assert a routing guarantee the code does not provide
`docs/guides/JIRA_SETUP_GUIDE.md:318` (and the mirror): *"Routing uses stored Jira tenant/issue metadata and the iteration's commit SHA, so branch text cannot redirect Jira feedback."*

Branch text is exactly what selects the task — `extractTaskIdFromBranch(branchName)` at `github-webhook-processor.ts:679`, whose record is passed straight in as `original`. The SHA guard is conditional (`jira-deployment-preview.ts:89`):

```ts
if (!matched && (sawIteration || original.head_sha) && original.head_sha !== sha) {
  throw new Error('No Jira task matches the deployment SHA');
}
```

For a non-iteration original with no `head_sha` and no sibling iterations, neither disjunct is truthy, nothing throws, and delivery proceeds. Same-repo is still enforced at `:45`, so the blast radius is bounded — but the sentence states an absolute the code does not provide, in an operator guide, about routing. One sentence plus `mise //docs:sync`.

Suggested wording: "Delivery targets the task encoded in the deploy branch and then routes only via that task's stored Jira tenant/issue metadata — the Jira issue is never parsed out of branch or PR text. Cross-repo delivery is rejected, and an iteration must match the deployment SHA."

## Non-blocking suggestions

**N1 — the claim latch discards a classification the code already computes.** `jira-deployment-preview.ts:111-139`. `jira_preview_claimed` is a one-way latch (nothing in `cdk/src` ever removes it). On a Jira 429/503 — a *received* response, so nothing was created — the claim stays set and the preview is permanently lost for that task; every later delivery takes the `ConditionalCheckFailedException` branch, finds no comment id, logs one `warn`, and returns. `writeComment` already distinguishes "Jira answered non-2xx" (`jira-feedback.ts:284-286`) from its `catch` (genuinely uncertain), and `JiraPostResult` collapses both into `{ ok: false, retryable }`. The documented trade — "Jira may have committed before a transport timeout" — is right for the uncertain half only. Consider a third state plus a conditional claim release (`REMOVE jira_preview_claimed` conditional on `attribute_not_exists(jira_preview_comment_id)`) on definite failure.

Same shape when the POST succeeds but the follow-up `SET jira_preview_comment_id` write fails: claimed ✓, comment exists ✓, id unrecorded ✗ — the issue then shows the **first** commit's screenshot forever, which is worse than showing nothing. Both cases warrant `logger.error` with an `error_id`, not `warn`.

**N2 — `persistScreenshotUrl`'s `task?: TaskRecord` conflates four states, and the new gate fails open.** `github-webhook-processor.ts:676-728` and `:314-334`. `task === undefined` now means: no table configured, not an ABCA branch, *the DDB update threw*, or no old attributes. On a transient DDB error for a Jira task the Jira delivery is skipped, the `else if` warn is **also** skipped (same guard), and `task?.channel_source !== 'jira'` evaluates true — so control enters the Linear block, where a Jira task's branch (which carries the Jira key via `buildTaskDescription` → `generateBranchName`) can match `extractLinearIdentifierFromBranch`. This is the bare-`undefined` conflation that `#863` removed one commit before this branch's merge-base, in a file that already imports the helpers at `:38`. Return `LookupResult<…>`, branch on `isLookupFailure`, and skip **both** channel deliveries.

**N3 — 19 rejection gates, one opaque signal, zero logs.** `normalizeAmplifyPreviewCheck` returns a bare `null` for ~19 distinct reasons; the receiver returns `{ skipped_check: true }` with no log line at all, while a dedup hit logs `info` (`github-webhook.ts:180`), an unrelated event type logs `info` (`:91`), and a malformed `deployment_status` logs `warn` + 400. If AWS renames the check or changes the app-slug shape, the entire Amplify path dies permanently and silently with a green webhook delivery page. Return a reason code and log it; the `nosemgrep: ts-silent-success-masking` suppression at `github-deployment-status.ts:88` should then be removable.

**N4 — existing Amplify operators are silently broken by the environment filter.** The normalizer hardcodes `environment: 'Preview'` (`github-deployment-status.ts:101`) and the receiver still applies `SCREENSHOT_TARGET_ENVIRONMENT` (`github-webhook.ts:130`). An operator following the *old* guide has it set to a branch name, subscribes to Check runs as the new guide instructs, and gets every Amplify check dropped with only `skipped_environment` in the HTTP response body — while setting it to `Preview` breaks their existing branch-deploy path, since the filter is a single fixed string. Either exempt the `check_run` path (check name + `aws-amplify-console` owner + `pr-N.*.amplifyapp.com` hostname already prove it is a PR preview) or set the synthetic environment to `targetEnv`, and add a migration note for operators currently on a branch-name value.

**N5 — the validated PR number is discarded, then re-derived from the SHA.** `github-deployment-status.ts:90-103` proves `pr-<n>` matches a listed pull request whose `head.sha` equals `check.head_sha`, then drops `preview[1]`; the processor asks GitHub again by SHA at `:195`. When one SHA heads two open PRs (retargeted base, stacked series), the screenshot can be posted onto a different PR — and `persistScreenshotUrl` then resolves the task from *that* PR's `headRefName`, so the whole Jira/Linear chain follows the wrong task. Carry the validated number on the normalized payload and prefer it.

**N6 — an empty ADF document is claimed and POSTed.** `jira-preview.ts:36-38` returns `buildAdfDocument([])` when the screenshot URL is absent or fails the allowlist. On the iteration path that is harmless. On the standalone path the claim is written first (`:115`) and the empty document is then POSTed (`:138`), so a misconfigured `SCREENSHOT_PUBLIC_HOST` yields a blank comment on a customer's issue *and* a permanently burnt claim. Reachability is low today — `screenshot_url` is always ABCA's own CloudFront host and `screenshot_preview_url` was allowlisted at the receiver — which is why this is a nit rather than a blocker. Return `AdfDocument | null` so both call sites must branch, and skip before claiming. A `logger.warn` naming the rejected URL would also make it observable, as `github-webhook-processor.ts:163-168` already does for the analogous rejection.

**N7 — smaller items.**
- `iteration-heartbeat-sweep.ts:152-159` has no `else`, so a Jira tenant with a revoked token produces a sweep reporting `edited: 0`, indistinguishable from one with nothing eligible. Mirror the non-Jira `catch` and count attempts separately from successes.
- `jira_preview_claimed?: boolean` (`types.ts:335`) contradicts its own `attribute_not_exists` predicate — writing `false` would read as "unclaimed" while permanently blocking the claim. Make it `?: true`.
- Six-to-eight positional `string` parameters in the two new public functions. `(screenshotUrl, previewUrl)` swapped compiles, passes the allowlist, and is caught by no test — the labels just point at the wrong URLs. `(tableName, registryTableName)` swapped cross-wires two tables and fails silently. `updateJiraIterationComment` gained six call sites in this PR, so this gets harder to fix with each one.
- ADF is `Record<string, unknown>` at every boundary while the repo already types inbound ADF (`jira-adf.ts:41 AdfNode`) and the closest outbound analogue in full (`slack-blocks.ts:27-88`). An `AdfDocument` interface removes the two casts at `jira-preview.ts:70`, where a non-array `content` currently spreads characters.
- Naming: `jira-preview.ts`'s dominant export is `updateJiraIterationComment`, imported by four handlers — it is the iteration-comment writer. Next to `jira-deployment-preview.ts` (the delivery orchestrator) the two are near-indistinguishable at an import site; `jira-iteration-comment.ts` would read truer.
- `jira-deployment-preview.ts:61` throws on exceeding the page limit, which discards delivery entirely — including when `original.head_sha === sha` and delivery on the original task was valid. Prefer breaking out and falling through to the existing SHA check.
- IAM: the new Jira block at `github-screenshot-integration.ts:330-341` grants `secretsmanager:PutSecretValue` with no justification comment, while the adjacent Linear block carries six lines explaining exactly why the write grant is intentional; `cdk/AGENTS.md` puts IAM changes under "⚠️ Ask first". The `AwsSolutions-IAM5` suppression reason at `:474-476` also does not enumerate the Secrets Manager prefix wildcards, which repo convention elsewhere does.
- Comment accuracy: the convergence docstring (`jira-preview.ts:73-79`) claims a slow heartbeat "restores the latest body" unconditionally, but it is bounded at 4 attempts and the heartbeat caller discards `retryable`. `orchestration-reconciler.ts:1227-1228` still credits `preservePreview` above a Jira path that no longer uses it (it survives only on the Linear path). `jira-feedback.ts:102-105` still says "only bold, inline code, and links" after the image branch was added. `github-webhook-processor.ts:46-50` says an unset `TASK_TABLE` skips only persistence — Jira delivery is skipped too, and its `screenshot.jira_missing_registry` warn cannot fire without the task record. The `persistScreenshotUrl` docstring should warn that the returned task is the `ALL_OLD` image, so `screenshot_url` on it is the *previous* deploy's value. The new image branch at `jira-feedback.ts:127-136` has no comment at all despite four non-obvious decisions (why an image becomes a link, why alt text is discarded, why the labels are duplicated from `jiraPreviewDocument`, why `!` joined the token-boundary list).
- `e8992c2f` is labelled `fix(jira)` when it is `feat(screenshots)` — which is part of why the scope question below is easy to miss when skimming the log.

## Governance & documentation

**The Amplify `check_run` receiver is outside #698's approved scope.** #698 excludes it three ways: its *Out of scope* section says "Screenshot capture, storage, URL allowlisting, or GitHub PR comments; those already exist in #695"; all eight acceptance criteria are Jira-surface criteria naming no new event type; and the last one — "Linear and GitHub PR screenshot behavior remain unchanged" — is precisely what commit `e8992c2f` changes, since Amplify previews now produce GitHub PR screenshot comments where they previously did not. ADR-003 treats an approved issue as scope-frozen, and `CONTRIBUTING.md:25` asks for one logical change per PR.

To be clear about what I am asking: **not** a split of work that is already built and live-verified on dev. Amend #698 and get it re-approved, or file a companion `approved` issue and reference it — the point is that a new operator-facing trigger path with its own setup step is exactly what the governance gate exists for.

Docs are otherwise strong: both guides updated, the best-effort semantics and log event names documented for operators, and the Starlight mirror **verified in sync** — I re-ran `docs/scripts/sync-starlight.mjs` under Node 22 and it produced zero diff. Two additions: fix B4's sentence, and add `jira.preview.convergence_failed` and `screenshot.jira_missing_registry` to the guide's event list (it currently names three of the five events the code emits). #698 also carries no priority label.

## Tests & CI

All 11 checks green including `build (agentcore)`, so the suite passes — the gap is what it *pins* (B3). Genuinely strong parts worth keeping: the 19-row negative table on the Amplify normalizer; the allowlist test that checks **both** the screenshot and preview positions (`jira-preview.test.ts:66`); `github-webhook-processor.test.ts:453`, which tests the stated requirement rather than the implementation by using a branch containing `ENG-42` while the record says `ACTUAL-7`; the SHA-attribution test where the older matching task correctly wins over the newer one; the pagination test that checks `ExclusiveStartKey` threading; and the `fdbdaa63` metrics tests, which assert rendered user-visible strings (`$0.21`, `12 / 30`, `cost: —`) and demonstrably fail on revert.

Untested failure paths: missing task record, convergence exhaustion, Jira 4xx vs 5xx, `screenshot.jira_missing_registry`, `Number.isSafeInteger(check.id)`, and the new `400 Invalid webhook payload` branch — which is unreachable in tests, because the `'null'`/`'[]'` bodies are sent as `check_run` and return `skipped_check` first.

**Bootstrap synth coverage (ADR-002): not applicable, correctly untouched.** No new CloudFormation resource type appears — the construct diff adds an interface prop, two env-var spreads, `PolicyStatement`s on pre-existing Lambda roles, and one `grant()` on an `ITable` injected from `stacks/agent.ts:1984`. `BOOTSTRAP_VERSION` correctly stays at `1.6.0`.

**CDK test performance (#366): pass.** The new `describe('Jira screenshot delivery wiring')` uses a single `beforeAll` with one `Template.fromStack`, matching the file's existing blocks; nothing re-enables `aws:cdk:bundling-stacks`.

## Review agents run

- **code-reviewer** — ran. Source of B1, B2, the scope finding, and N4. I independently verified B1 and B2 against the code before accepting them.
- **silent-failure-hunter** — ran. Source of N1, N2, N3, N5. I **downgraded** its two CRITICAL findings on the claim latch: the trade is deliberate and documented in the Jira guide, so it is a follow-up rather than a merge blocker — though its point that the definite/uncertain classification is computed at the source and then discarded is correct and worth acting on.
- **pr-test-analyzer** — ran, and executed mutants in an isolated sandbox (this worktree left clean). Source of B3; the mutation evidence is why I treat those gaps as blocking rather than advisory.
- **type-design-analyzer** — ran. Source of N6 and most of N7. I **downgraded** its three "blocking" items: the missing allowlist re-check on the `screenshot_url` write is defence-in-depth (both values arrive from an already-validated path), and `jira_preview_claimed: false` is unreachable because no writer writes it. Its mirror analysis is right — `TaskRecord` is not part of the CDK↔CLI contract (`cli/src/types.ts` mirrors `TaskDetail`, and `toTaskDetail` is an explicit projection), so no `cli/src/types.ts` change was needed or missing.
- **comment-analyzer** — ran. Source of B4 and the comment items in N7. It also confirmed the model case in this PR: `iteration-heartbeat.ts:87-89`'s "No write — a heartbeat never mutates task state" was rewritten in the *same hunk* as the new `dynamodb:UpdateItem` grant, preserving the invariant the old comment was protecting.
- **security-review** — ran. **No findings at ≥0.7 confidence.** Verified HMAC verification is unconditional and strictly upstream of the `check_run` branch (`github-webhook.ts:76`, `timingSafeEqual` with an empty-secret guard); the Amplify host regex is fully anchored, so the `endsWith('amplifyapp.com')` class of bug is absent and `evil-amplifyapp.com` / `…amplifyapp.com.evil.example` both fail; scheme is pinned to `https:` with userinfo and explicit port rejected; the new path converges on the same `isAllowedScreenshotUrl` allowlist before capture; ADF is built as an object graph with only ABCA-constant `text` values and both `href`s allowlisted; IAM is prefix-scoped (`bgagent-jira-oauth-*`, mirroring the Linear grant) and GSI-scoped with no wildcards; and tenancy is minted per `jira_cloud_id` from the task's own registry record, with three independent synthetic-node guards. It independently reached B4's conclusion, as a docs-accuracy correction.
- **Omitted:** none — all five pr-review-toolkit agents were in scope for this diff.

## Human heuristics

- **Proportionality — pass.** Only two of the four touched `shared/` modules are new; two modules for a delivery orchestrator plus a converging comment writer is a defensible split, and the `parseMarkdownRuns` image branch is genuinely reached (the Jira parent rollup passes linked-image markdown through `toAdfDocument`).
- **Coherence — concern.** `jira-preview.ts` vs `jira-deployment-preview.ts` are near-indistinguishable names for different jobs (N7); `#863`'s `LookupResult` convention is imported in the very file that reintroduces the bare-`undefined` pattern (N2); and there are now two claim mechanisms — `iteration-reply-claim.ts`'s typed `ReplyClaimOutcome` and the hand-rolled latch here.
- **Clarity — concern.** `ok: false` after four successful writes (B2) is the clearest instance: the return value does not mean what its callers read it to mean, and the mislabelled `convergence_failed` event sends an operator looking for a write race that never happened.
- **Appropriateness — concern, narrowly.** The live verification on `backgroundagent-dev` against a real Amplify delivery is the right kind of evidence, and the metrics tests assert rendered output rather than the coercion helper. But the race tests assert the mock rather than the behaviour, and one test certifies the never-retry latch as intended.

isadeks (approved):
## Verdict: Approve

Re-review at `1f2e37ab`. **All four blockers from my `CHANGES_REQUESTED` review are verified fixed**, the governance item is resolved via #900, and CI is fully green (8/8, including `build (agentcore)` — so the new tests pass, which matters here because B3 was specifically about whether the tests bite). I read the fix rather than the commit message; verification notes below.

### Blockers — verified fixed

**B1 (unbudgeted delivery inside the post-capture reserve).** New `cdk/src/handlers/shared/jira-preview-budget.ts` is a real deadline: `AbortController` + `Promise.race` against `Math.min(MAX_DELIVERY_MS, remaining() - EXIT_RESERVE_MS)`, re-evaluated per operation, wrapping every DynamoDB and Jira call on the path. `MAX_CANDIDATE_READS = 25` caps the 250-consistent-read worst case, and the page limit now `break`s with `screenshot.jira_lookup_limited` instead of throwing away a valid delivery (that was a separate nit — good catch folding it in). `POST_CAPTURE_RESERVE_MS` 8s → 30s with the docstring updated to account for the new phase, and exhaustion is distinguishable in logs via `screenshot.jira_budget_exhausted`.

What I specifically checked, because a budget that does not reach the wire would be cosmetic: the signal is genuinely plumbed through — `ctx.signal` into `writeComment`, composed at the fetch as `AbortSignal.any([signal, controller.signal])`, plus pre-flight `aborted` checks at each stage of the OAuth/retry path. `AbortSignal.any` needs Node ≥20; engines are `>= 20.x <= 24.x` and the Lambda is `NODEJS_24_X`, so that is safe. Budget arithmetic still leaves ample headroom: `captureBudget = remaining() - 30_000` against `MIN_CAPTURE_BUDGET_MS = 15_000` on a 110s total.

**B2 (convergence reporting failure after successful writes).** Exhaustion now returns `{ ok: true }` with the `jira.preview.convergence_failed` warn retained, so the fan-out no longer releases the claim and throws — the duplicate-terminal-comment path is gone. An absent task record returns `{ ok: false, retryable: false }` with its own `jira.preview.task_missing` event, which fixes both the mislabelled diagnostic and the retry-budget burn on a state `ConsistentRead` already proved permanent. The `wrote` flag extending the same reasoning to the outer `catch` is a nice touch — a post-write transport failure no longer re-drives a delivery that already landed.

**B3 (tests that could not fail).** Exactly the sensitivity that was missing: resolution-order `settled[]` arrays with `settled.at(-1)`, `expect(put).toHaveBeenCalledTimes(3)` / `(4)`, `expect.objectContaining({ terminal: true })` in **both** the fan-out and reconciler mocks, and a literal `ConditionExpression` assertion rather than trusting the fake to re-implement it. The `MAX_CONVERGENCE_ATTEMPTS = 1` mutant that previously left 14/14 green now dies on the call-count assertion, and the new `{ok:false, retryable:false}` / `{ok:true}` cases pin B2's outcomes directly.

**B4 (docs asserting a guarantee the code did not provide).** Reworded accurately, and it goes further than asked: all five emitted event names plus the two new budget events are now documented for operators, with the convergence semantics spelled out ("successful updates reached the repair limit; it does not retry terminal delivery"). I re-ran `docs/scripts/sync-starlight.mjs` under Node 22 — zero drift, mirror in sync.

Also fixed along the way: `LookupResult` on `persistScreenshotUrl` so a lookup failure skips *both* channel deliveries instead of falling through to Linear (the `#863` convention); `jiraPreviewDocument` returning `null` with the standalone path rejecting it before consuming the claim; the `PutSecretValue` justification comment and the enumerated `AwsSolutions-IAM5` reason; and the stale `preservePreview`, `parseMarkdownRuns`, `TASK_TABLE`-unset and `ALL_OLD` comments.

### Governance

Resolved as suggested — **#900** (`feat(screenshots): support Amplify PR preview check runs`) carries `approved`, so the receiver has its own authorised issue rather than riding #698's scope freeze. #698 also picked up the `P1` label it was missing.

### Follow-ups tracked

The non-blocking set is filed as **#901** so it does not have to be triaged out of the review body — 13 items, with an explicit "already fixed in `1f2e37ab`, do not redo" section. The two worth knowing about are the preview claim latch (definite vs uncertain Jira failure are still conflated, and the latch never releases, so a 429/503 permanently loses that task's preview) and the POST-succeeded-but-id-write-failed state, which leaves the issue showing the first commit's screenshot indefinitely.

Two small consequences of this fix are recorded there as item 12, neither blocking:
1. `jira-deployment-preview.ts:48` computes `document` and throws at the top of the `try`, so a disallowed screenshot URL now aborts the **iteration** path as well — where the document is never used, since `updateJiraIterationComment` re-renders from durable state. A misconfigured `SCREENSHOT_PUBLIC_HOST` would skip the whole status-comment update, including completion metrics and the PR link, rather than just omitting the preview block. Moving the check into the standalone branch keeps it fail-closed without that side effect.
2. `:104` tightened `task.head_sha && task.head_sha !== sha` to `task.head_sha !== sha`. That correctly closes a hole, but iteration records written before `head_sha` was persisted now receive no preview at all. Fail-closed and presumably deliberate — worth confirming.

Nice work on the turnaround; the underlying convergence design was sound and these were surgical fixes on top of it.

Issue #857, opened 2026-09-04
Vault × lambda-microvm fail-fast is obsolete after #854 — remove it and budget the pair instead


isadeks:
`AgentStack` refuses `enableLinearIdentityVault` together with `compute_type=lambda-microvm`. That guard was correct when it was written and is not any more: #854 reclaimed 30 resources, so the combination now fits comfortably under the quota. The guard therefore rejects a deployable configuration, and its error message quotes a resource count that is no longer true.

## Current behaviour

`cdk/src/stacks/agent.ts:373`:

```ts
if (linearIdentityVaultEnabled && computeType === 'lambda-microvm') {
  throw new Error(
    'enableLinearIdentityVault cannot be combined with '
    + 'compute_type=lambda-microvm: the two together exceed '
    + "CloudFormation's 500-resource limit for this stack (505). ...",
  );
}
```

It exists so the operator gets both flag names instead of CDK's `TooManyResourcesInStack` per-type census, which gives no hint that two context flags are the cause. That reasoning still holds — only the arithmetic has changed.

## Measured

Synthesized counts (`Template.fromStack` + 1 for `AWS::CDK::Metadata`), hard quota 500:

| configuration | before #854 | now |
|---|---:|---:|
| vault + MicroVM | 505 — refused | 476 |
| vault + MicroVM + tool gateway | — | 483 |

Both are inside the 490 budget that #854 introduced.

## Why no test caught this

The covering test asserts the throw, at `cdk/test/stacks/agent.test.ts:1745`:

```ts
)).toThrow(
  /enableLinearIdentityVault cannot be combined with .../
);
```

It passes for exactly as long as the guard exists, so it pins the guard rather than the headroom. It could not have gone red when the room appeared. A guard whose covering test cannot observe the condition it guards is the general shape worth avoiding here.

## Proposed change

1. Delete the guard and the `toThrow` test.
2. Add `enableLinearIdentityVault` as a dimension to the resource-budget matrix introduced by #854 in `cdk/test/stacks/agent.test.ts`, so vault × `compute_type` × tool-gateway is measured against the budget. That is the difference between permitting the combination and guarding it, and it is what would have caught this drift.
3. Drop the corresponding bullet from #831's limitations section.

## Notes

The guard is on the #831 branch (`feat/809-linear-identity-vault-pr`), not on `main`, so this is a follow-up rather than a defect in a released path — it should land after #831 merges, or be folded into it.

[...]

Issue #841, opened 2026-09-01
bug(agent): test_server.py leaks live pipeline threads — reset fixture clears _active_threads without joining, causing cross-test mock contamination


isadeks:
## Problem

`agent/tests/test_server.py` intermittently fails with a mock being called by a test that never made the call. Observed twice on PR #680 (a **CLI-only** PR that cannot touch `agent/`), each time on a *different* test:

| Run | Test |
| --- | --- |
| [33446493920](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/actions/runs/33446493920) | `TestMicrovmRunHookPlatformConfig::test_no_platform_config_and_no_baked_env_is_refused_not_run_unscoped` |
| earlier run | `TestMicrovmRunHookPreInstallAwsSilence::test_no_cloudwatch_or_credential_seam_is_touched_before_the_install[no-config]` |

```
FAILED tests/test_server.py::TestMicrovmRunHookPlatformConfig::test_no_platform_config_and_no_baked_env_is_refused_not_run_unscoped
  AssertionError: Expected 'mock' to not have been called. Called 1 times.
  Calls: [call(repo_url='org/repo', task_description='do it', github_token='ghp_x',
               anthropic_model='', aws_region='', task_id='t-pc', max_turns=100, ...)]
```

`main` is green (verified by manual `workflow_dispatch` [33510225606](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/actions/runs/33510225606): agent 1755 passed, cdk 4322 passed, cli 768 passed), so this is a **race**, not a regression — a single green run does not clear it.

## Root cause

Three facts combine:

1. **`run_task` is resolved at call time, in a background thread.** `server.py:35` does `from pipeline import run_task`; `_run_task_background` calls it bare at `server.py:478`. A bare global resolves *when executed*, not when the thread starts. That function is the `threading.Thread` target in `_spawn_background` (`server.py:736`, `target=` at `:748`).

2. **The reset fixture drops thread references without joining them.** `test_server.py:21-29`:

   ```python
   @pytest.fixture(autouse=True)
   def reset_server_state():
       ...
       with server._threads_lock:
           server._active_threads.clear()   # <-- drops the handle; thread keeps running
   ```

   `_active_threads` (`server.py:255`) is only a tracking list. Clearing it does not stop anything.

[...]

Issue #817, opened 2026-08-28
fix(compute): Lambda MicroVMs P2 review follow-ups (#733) — classifier ordering, payload-delete grant, ARN pinning, stale claims, N1 test


isadeks:
Follow-up to #733 (Lambda MicroVMs P2). `baa365fc` closed B1, B2, B6, B7, N2-ii and 11 nits, all mutation-pinned. Five items remain. Backend is experimental (non-suppressible synth warning against production use), so none is on a production path.

Verified by execution at `779849ba`.

## 1. Appended `stateReason` hijacks an earlier classifier pattern

`error-classifier.ts:253` (substrate failure) sits below `:208` (regional availability). B1's fix appends AWS-owned free text into the string feeding that ordered list:

```
"…substrate state completed (Run lifecycle hook returned HTTP status 400…)"
  → "The MicroVM stopped before the agent reported a result"   retryable: true    ✅
"…substrate state completed (MicroVM host unavailable.)"
  → "Lambda MicroVMs is not available in this Region"          retryable: false   ❌
"…substrate state completed"        (pre-#733 shape)
  → "The MicroVM stopped before the agent reported a result"   retryable: true    ✅
```

`retryable` flips `true`→`false`, so a transient is reported non-retryable and the user doesn't retry. New in `baa365fc`. Also hits `capacity unavailable in this Availability Zone`, `Lambda MicroVMs not available.`, and via `:166` `The account concurrency limit for MicroVMs was reached.` `error-classifier.test.ts:576` only classifies the reason-less string, so the suite is green.

Two viable fixes.

**(a) Reorder** — move the `:253` substrate block above the `:208` regional entry. Structurally cleaner, but spans two non-adjacent regions.

**(b) One-line lookahead** — replace `error-classifier.ts:208`:

```ts
    pattern: /^(?!.*MicroVM substrate terminated).*(?:(?:UnknownEndpoint|Inaccessible host|Could not resolve endpoint).{0,120}lambda|(?:lambda[- ]?microvms?|microvms?).{0,60}(?:not available|not supported|unavailable)|(?:not available|not supported|unavailable).{0,60}(?:lambda[- ]?microvms?))/i,
```

[...]

PR #904, opened 2026-09-18
feat(compute): add Lambda MicroVM P3 approval lifecycle

isadeks:
P3 lets an agent wait for human approval while its MicroVM sleeps or retires with a verified checkpoint. A later decision wakes the original worker or admits one replacement with the saved conversation and workspace. Unanswered approvals remain available by default; explicit deadlines remain supported.

Linear users can reply `approve` or `deny` under the approval comment. The service verifies the actual comment through Linear’s API, rejects bots and the saved OAuth token’s own identity, and applies the task owner’s decision to that request. CLI decisions work on all backends; sleep and replacement are MicroVM-specific.

#### Area

- [x] `cdk` — infrastructure, handlers and permissions
- [x] `agent` — runtime and image
- [x] `cli` — task and approval settings
- [x] `docs` — guides and design sources
- [x] `tooling` — contracts, packaging and CI

#### Related

ADR-021 P3 for #645, including remaining P2 prerequisites. Related follow-ups: #817, #841 and #857. AgentCore remains the default compute backend.

#### Changes

- Add approval-aware suspend/resume supervision, guest activity barriers, explicit lifecycle connection closure, credential renewal and correlated failure diagnostics.
- Preserve unanswered requests, with cancellation feedback, a ten-minute default sleep delay and task/deployment controls.
- Verify versioned conversation/workspace checkpoints before retirement, confirm worker shutdown before releasing capacity, and restore the pending action on one admitted replacement.
- Move approval creation and non-human timeout writes behind a trusted IAM API. The worker’s session tag restricts its API path to its own task. Workers can read approval records but cannot overwrite human decisions, notification markers or retention fields. MicroVM writes also require the active worker lease.
- Require an explicit MicroVM layout until migration is verified: `true` for new/already-nested installations, `false` for existing flat installations. Omission fails synthesis. Group stateless scheduled maintenance in a separate nested stack while retaining existing tables and MicroVM identities. Both layouts are covered across all image, gateway and vault configurations by the 490-resource budget.
- Strengthen terminal cleanup, transaction-conflict diagnostics, SDK-version checks, bootstrap permissions and real DynamoDB transaction tests in CI. Preserve stored serialization bytes.
- Reuse the existing Linear AgentCore Identity vault integration for MicroVM workers.

#### Latest review fixes (`bf7c271d`)

- Configure both CLI/API decision functions with the concurrency table and the shared admission limit; Linear's separate wiring had hidden this gap.
- Clean terminal, fenced replacements that never obtained a launch receipt, checking matching continuation/slot/lease identities and no worker ID. DynamoDB Local tests drive failure through cleanup and reject a worker ID injected before the conditional write. Unknown launch outcomes retain their existing lifetime protection; valid pending checkpoints do not expire.
- Reproduce project-hook denial on pinned SDK 0.2.110 / CLI 2.1.191 using a synthetic loopback model. Retire only the tool ID in the CLI error result; do not clear all registrations at Stop.
- Preserve valid credentials on advisory refresh failures while keeping forced wake refresh mandatory. Default MicroVM `uv` installs to copies after reproducing Linux cache hardlinks. Confirm termination conflicts with `GetMicrovm` before treating cleanup as successful.
- Correct remaining timeout/stranded-request documentation, document all three live harness commands, and cover heartbeat grace/staleness boundaries. The checkpoint-code table and approval API's non-semantic validation wording were already present.

Local full build passed with DynamoDB Local and real SDK continuation probes enabled: **5,758 CDK tests (261 suites), 2,254 agent tests (86.84% coverage), and 1,005 CLI tests**, plus lint, type checks, synthesis and documentation. The final agent diagnostic change passed another complete 2,254-test quality run. Final layout compatibility and heartbeat checks passed 25 tests, with CDK lint/compile passing. The documentation build and internal links passed across 82 pages; all three live-harness inspection commands ran without AWS calls.

Deployed `bf7c271d` to the retained nested fixture in `us-east-1`: CloudFormation `UPDATE_COMPLETE`, MicroVM image **3.0**, AgentCore runtime **4**, and coordinator version **5**. Four deployed Lambda code hashes match the reviewed assets; approve/deny functions have the concurrency table and shared limit. Verified 26 baseline stateful resource identities were preserved. [Latest-head GitHub checks passed](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/actions/runs/35795333952).

Both live Cognito-authenticated API decision cases passed after a saved checkpoint, sleep, source-worker loss and `PARKED`: approval and denial started replacement workers in **11.7s and 12.3s**. Approval performed exactly one Read; denial performed none and acknowledged the decision. Both tasks completed, their workers terminated, and scheduled cleanup closed leases and returned capacity to zero. These are the API routes used by CLI decisions; no interactive CLI command was used. Tests used an explicit 30-second sleep delay with a test-only policy, leaving the ten-minute default unchanged. The fixture configuration was restored exactly; final checks found no active tasks, open leases or held slots. Cleanup follows the five-minute reconciliation schedule. This already-nested update does not verify flat-to-nested migration. Earlier results below retain their recorded revisions.

#### Scope and remaining tradeoffs

This PR includes the trusted approval writer, Linear replies, retained-request default, shared maintenance stack and MicroVM layout selection. Maintainer agreement on this expanded scope and the proposed ADRs remains open. The `@aws-sdk/client-ssm` dependency supports the runtime suspend setting; build-workflow changes make DynamoDB Local and the pinned SDK recovery probes required in CI. These are intentional dependency/CI changes, not incidental cleanup.

Nonblocking optimization remains: eager transcript append repeatedly copies/encodes history (a local synthetic 1,000-frame run took about 5.5 seconds cumulatively). Replacing checkpoint readback with `HeadObject` must preserve checksum/version validation and recovery after a lost upload reply. Broad type/handler refactors and restored-scope revalidation are deferred rather than mixed into lifecycle corrections. Known launch rejections still share the conservative unknown-outcome capacity hold; earlier release needs a durable, operation-specific rejection classification.

#### Validation

Review validation at `159576e6` (eight focused commits after the third review):

- Complete [final-head GitHub build](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/actions/runs/35649497027) passed: **5,718 CDK tests**, **2,231 agent tests** (zero skips; 86.66% coverage), and **1,005 CLI tests**. Includes real DynamoDB transaction checks and the real SDK approve/deny recovery probe, explicitly enabled in the build workflow. The full local build also passed before the final worker-token validation restoration, which passed its focused tests and this final full CI run.
- Cancellation-race tests exercise both actual pipeline paths. Restoring the false failure behavior produces two failing tests; bypassing both checked terminal writes produces three. Internal-whitespace/case replay tests reject an over-broad normalizer.
- Expanded flat/nested resource-budget, cdk-nag and bootstrap coverage remains in the full suite. Both layouts retain their original budget; widest configuration is below 490 resources.
- Final correction restores the original `worker_attempt_id` validator; 22 broker unit tests passed. Source tracing shows this field carries the platform launch/lease token, not AWS’s physical MicroVM identifier.
- A live probe using the worker's actual AgentCore Identity vault token successfully edited its own control comment, but Linear rejected editing a human-authored `deny` comment. The human body and author stayed unchanged. This verifies the current app-token installation; no historical approval threads were modified.
- The documentation build validates internal pages and anchors across **82 rendered pages**, plus local existence of emitted GitHub repository-file targets. An injected missing repository target is rejected.
- Security/dependency/workflow, dead-code and PR-title checks passed at `159576e6`. The silent-failure scan and its rule tests pass. Lint, type checks, contracts, documentation and infrastructure synth pass. The full build and all PR checks passed at that revision.
- Deployed `159576e6` successfully to the retained isolated fixture after pausing submissions and draining tasks, leases and capacity. Verified the new approval API, AgentCore runtime version 3, and MicroVM image 2.0 against the built image/artifact references. Seven live IAM simulation checks passed: own-task invocation/read allowed; other-task or untagged invocation and direct approval writes denied. The borrowed Linear vault and shared logging settings remain unchanged and functional.
- **All six latest-image Linear coding-agent flows passed**: AgentCore approve/deny, MicroVM same-worker wake approve/deny, and MicroVM replacement-worker recovery approve/deny. Verified actual owner-authored replies, matching decision sources and thread acknowledgments, approved reads exactly once, denied reads never succeeding, continued agent execution, terminated MicroVMs and scheduled capacity release. All four MicroVMs waited the full default 600 seconds before suspension; normal wake cases were observed asleep for over a minute. Replacement cases used controlled loss of a suspended worker with a verified checkpoint, then restored on different workers. No approval or continuation rows were manufactured. A separate automated wake probe measured **69.031 seconds** between the suspend response and resume hook, inside the earlier 5–90 second failure window; resume returned HTTP 200 in **143 ms**. That task also completed on the same worker. All seven test slots are released and active capacity is zero.

Live stress validation and capacity fix at `7384f9b0`:

- A 12-task burst respected the 10-task cap, queued overflow, handled queued/startup cancellation and deliberate worker termination, and passed 50 concurrent status reads. Nine ordinary workloads completed without approval gates.
- Stress testing exposed simultaneous slot releases failing with DynamoDB `TransactionConflict`. Bounded retries preserve the exact transaction, idempotency token and worker-lease conditions. An isolated live reproduction went from 6/12 successful releases to 12/12; concurrent admission still accepted exactly 10/12. All 71 focused unit/transaction tests and the full local build passed: **5,723 CDK**, **2,231 agent**, **1,005 CLI** tests.
- Deployed the fix and verified the actual Lambda packages. Two workers each passed 1 GiB memory, 512 MiB disk and 45 CPU-second benchmarks; a third was cancelled during its Python workload. All test workers terminated, leases closed and capacity returned to zero. The patched cleanup sweep reported no failures. Tests and reproducible harnesses remain private; no benchmark PRs were published.

The later main merge at `2a79c0e` retains this fix; its CI is separate from the live deployment above.

Earlier live approval-boundary validation used the actual approval service and Python signing client: **34 AWS checks** covered creation, field rejection, task/tag isolation, denied direct writes, timeout races and MicroVM lease fencing; **nine IAM simulation checks** also passed. These were isolated API tests with task records for all three backends, without coding-agent images.

Earlier coding-agent tests additionally covered explicit expiry/cancellation and AWS access after credential expiry. Those earlier configurations remain separate from the latest-image cases above. Retained Linear fixtures and reproducible integration harnesses stay outside this public PR.

#### Before marking ready

- [x] Review-baseline GitHub checks complete (`159576e6`); full local build and live stress validation passed for capacity fix `7384f9b0`.
- [x] GitHub checks passed for the subsequent main merge (`2a79c0e`).
- [x] GitHub checks passed for the nested-default change (`35f05bbb`).
- [x] Complete GitHub checks for the latest review-fix head (`bf7c271d`).
- [ ] Obtain maintainer acceptance of expanded #645 scope, ADR-021/ADR-023 and contributor assignment.
- [x] Verify API approve/deny decisions on PARKED tasks in the latest review-fix deployment (the same routes used by CLI).
- [x] Deploy and verify the trusted approval service with matching worker images after draining existing tasks.
- [x] Complete the six latest-image Linear coding-agent approval/recovery cases.
- [x] Complete the compatibility review: retain version-1 checkpoint field names and the distinct migration/runtime hash encodings so saved data remains readable.
- [ ] Complete reusable flat-to-nested migration commands and independently verify upgrading a current-main deployment.

#### Review decisions and limits

[ADR-023](https://github.com/isadeks/sample-autonomous-cloud-coding-agents/blob/bf7c271d/docs/decisions/ADR-023-trusted-approval-writer.md) records the approval writer, rollout and retained-request decision. No default approval expiry is deliberate; zero is the no-expiry sentinel and the positive timeout range starts at 30 seconds. Worker/resource lifetime remains separate. A public SigV4 endpoint is intentional, with task-path IAM authorization.

This protects approval-record writes, not complete isolation from compromised ambient compute credentials: those credentials can still select session tags, and legacy channel bundles still include the webhook signing key. Action descriptions and policy metadata are worker assertions. `TIMED_OUT` currently also includes worker polling failures and is not proof that a human deadline elapsed. These limits are documented rather than presented as solved by the writer API.

The normal pipeline propagates failed terminal writes, while a superseding cancellation is benign and does not produce a false failure reaction. Actual pipeline entry points and the outer runtime's unhealthy response after a failed backup write are tested. Pure transaction conflicts get bounded retries; failed ownership checks do not. The continuation scan cursor uses compare-and-set to preserve progress when old and new schedules overlap during upgrade. Existing orchestrator heartbeat/backend monitoring handles lost running workers; a second broad `RUNNING` reaper was not added. Checkpoint field renames or hash-encoding consolidation require a versioned migration, not a style-only edit.

#### Deployment notes

Re-bootstrap to bundle **1.9.0 or later**. For the approval permission change, pause submissions and drain existing tasks, including suspended workers and retained continuations; deploy CDK and the matching rebuilt image together. Old workers lose direct approval-write access and fail closed at their next gate. See [the upgrade procedure](https://github.com/isadeks/sample-autonomous-cloud-coding-agents/blob/bf7c271d/docs/guides/DEPLOYMENT_GUIDE.md#upgrading-approval-permissions).

Automatic suspension remains behind the deployment compatibility gate, which defaults off. Once enabled, the default sleep delay is ten minutes. MicroVM deployments must now explicitly select `microvm_nested_stack`: use `true` for new/already-nested installations and retain `false` for existing flat installations. Omission fails synthesis. Explicitly selecting `true` does not migrate a flat deployment. Regenerate old cloud assemblies before deploying. Portable migration and independent upgrade acceptance remain open.

#### Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution under the terms of the [project license](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/blob/main/LICENSE).


isadeks:
**Self-review at head `08baca54`.** Verdict: **request changes** — four blocking classes. Both CI checks are green at this head; the earlier `493 > 490` budget failure was on `4ffe36ee`, fixed by `edee8a72`, and the security job cleared once the merge brought in `anyio 4.14.2`.

The engineering quality here is high: read-after-write everywhere, generation fencing, transactional capacity accounting, a real per-file digest manifest, and genuine live acceptance. What blocks is a forgeable approval path, a default that attempts the upgrade this PR's own docs forbid, an approved-issue scope gap, and a set of fail-closed claims CI cannot see.

## Vision and governance

Tenets 1, 2, 3 and 5 are well served — escalation stays policy-gated and attributable, the coordinator keeps the invariants, the agent gains no orchestrator privilege, and native channel replies genuinely advance "the platform reaches the human." Three trades are made without an ADR recording them.

**Tenet 6 (cost and capacity first-class).** `contracts/constants.json` moves `approval_timeout_s.default` from 300 to 0 for every backend, not just MicroVM. With no `deadline_epoch` written, `approve-task.ts:186` can never expire a gate. Combined with `--microvm-sleep-after 0`, `microvm-continuation-retirement.ts:287-289` gates `ageDue` on `sleep > 0`, so a live VM bills the full 8h waiting on a gate that cannot expire. This is recorded only inside ADR-021, whose status is still `proposed`.

**Tenet 4 (fail closed).** See B2 — the unsafe upgrade path is the default value.

**ADR-003 scope freeze.** The native Linear approve/deny feature (8 commits, ~1,059 new lines across 8 modules, plus `approve-task.ts` and `deny-task.ts`) is not in #645's body and is tracked by no issue. #645 says only that tasks idling on approvals benefit from suspend/resume. It needs its own approved issue, or #645 needs re-approval for the widened deliverable.

## Blocking

### B1 — the Linear reply path is forgeable by the worker it gates

`cdk/src/handlers/shared/linear-approval-reply.ts:69-82`. The handler's own checks are thorough: owner, workspace, issue, consistent-read task re-check, `data.id` dedup. Authenticity rests entirely upstream on the webhook HMAC, and the worker can read that secret.

`linear-verify.ts:232-235` verifies against `stored.webhook_signing_secret`, which lives in the `bgagent-linear-oauth-<slug>` bundle. `lambda-microvm-compute.ts:719-735` grants the worker execution role `secretsmanager:GetSecretValue` on `bgagent-linear-oauth-*` — that grant is not new in this PR (the diff touches only its comment; it is pre-existing on ECS too), but this PR is what makes it consent-bearing. `agent/src/microvm_credentials.py:9-10` concedes the isolation limit in its own docstring: "This protects provider selection, not isolation from code running as the same OS user." And `linear-approval-thread.ts:35-41` derives `parentId` from `sha256` over workspace, issue, task and request ids — all four readable from the task row the agent may read by design.

So an agent stopped at a `force_push_main` gate can recover the signing secret, self-sign a `Comment/create` body with `body:"approve"` and `actor.id` set to the owner, and the audit trail records `caller_user_id=<owner>`. Fix: move the signing secret into a receiver-only secret and narrow the worker grant to the token-only bundle; or re-read the comment through the Linear API before honouring it, since a forged webhook has no real comment behind it.

### B2 — `microvm_nested_stack` defaults to `true`

`cdk/src/stacks/agent.ts:361`:

```ts
const microvmNested = microvmNestedContext !== false
  && microvmNestedContext !== 'false';
```

Unset context means nested, while `docs/verification/645-p3-nested-stack.md:22,36-38` says to set it explicitly to `false` for an existing flat deployment and "do not deploy the nested template directly over a flat deployment… names can collide and bucket auto-delete handlers can erase artifacts or pending payloads." The safe value is the one requiring operator action; the unsafe one is the default.

Without `microvm_resource_name_prefix` the log group, image and both connectors keep their explicit physical names while their logical ids move, so the update collides and rolls back. With the prefix, the old buckets are deleted during cleanup and there is no `RemovalPolicy.RETAIN` on the artifact or payload buckets — so the artifact the code itself says "must survive as long as the image versions built from it" is destroyed. Tables, secrets and attachment buckets stay in the parent and keep identity, so no data store is replaced.

Compounding this: `cdk/src/migration/permissions.ts` and `template.ts` are imported only by their own tests. `preserveLegacyPermissions` — the only thing preserving old workers' payload access past P3's new deny, and a required step in the runbook — has no caller in `cdk/`, `cli/`, `scripts/` or any mise task. The shipped default attempts the migration while the machinery that makes it survivable is unreachable. Fix: default to `false`, or gate `true` behind a synth assertion that the prefix and a retain flag are present, and either wire the migration entry point or mark that tree as staged.

### B3 — `write_terminal` was left behind by the conflict-classifier change

`agent/src/task_state.py:362-365`. This PR routes `_update_task` through `transact_write_items` whenever a MicroVM lifecycle is registered (`:104-111`), which changes the expected error from `ConditionalCheckFailedException` to `TransactionCanceledException`, and adds `_task_status_conflict` (`:114-126`) to handle both. The diff converts two of three sites — `write_heartbeat:188` and `write_running:233` — and leaves `write_terminal` string-matching the old code. The diff shows three removals of the old match against two additions of the helper.

Consequence: a MicroVM task finishes, the terminal write is cancelled, the branch is false, `:409` logs a bare WARN and returns `None` — which `pipeline.py:1827` and `server.py:536` read as success. The row stays `RUNNING` although a PR was opened, and the `trace_s3_uri` self-heal at `:386-407` becomes unreachable on every MicroVM task. Use `_task_status_conflict(e)`, log `CancellationReasons`, and return a bool the pipeline checks.

### B4 — the new `approval_timeout_s: 0` default discards the lifetime ceiling

`agent/src/hooks.py:1211-1212`:

```python
if decision_value == 0:
    return 0, None, requested
```

That returns before the `remaining_lifetime_s` ceiling at `:1225-1229`. On AgentCore and ECS there is no lifecycle and no continuation, so the bound is available and thrown away; `_ApprovalDeadline.from_recorded(..., 0)` yields `(inf, inf)`, so `_poll_for_decision` can never emit `TIMED_OUT`. The wait ends only when `orchestrator.ts:1122-1134` writes FAILED with "Orchestrator poll timeout exceeded while awaiting approval" — a message that never tells the user their approval went unanswered.

A test pins this: `agent/tests/test_approval_retention.py:20-22` asserts `_compute_effective_timeout(0, 0, remaining_lifetime_s=10) == (0, None, 0)`, i.e. that a 10-second remaining lifetime does not clip an unbounded gate. Honour `0` only where the worker can actually park; the discriminator is one line away at `:631`.

### B5 — `attempt_id` means two things on the same object

`cdk/src/handlers/shared/microvm-continuation-types.ts:28-29` and `:50-51`:

```ts
/** Physical worker that captured the checkpoint. */
readonly attempt_id: string;   // identity → microvmId
/** Logical launch token, set before RunMicrovm. */
readonly attempt_id?: string;  // record  → clientToken
```

Both are reachable as `record.identity.attempt_id` and `record.attempt_id`, and both feed DynamoDB fencing conditions — enforced divergently at `microvm-continuation-retirement.ts:83` and `microvm-continuation-runner.ts:46-48`. Writing `record.attempt_id === handle.microvmId` compiles, type-checks, reads correctly, and silently disables worker fencing. Rename to `source_microvm_id` and `launch_token` while it is still cheap. Same class: `parked` means "worker alive holding a tool" at `microvm_lifecycle.py:172` and "worker confirmed terminated, capacity released" at `microvm-continuation-types.ts:45`.

### B6 — a dependency bump silently deletes the feature

`agent/src/continuation_usage.py:51-56` raises unless the installed `claude-agent-sdk` version equals `contracts/constants.json` → `verified_sdk_version` (`0.2.110`). That version is pinned independently at `agent/pyproject.toml:19`, and `scripts/check-constants-sync.ts:417` validates only the format, never that the two agree. The failure lands in a broad `except Exception` at `agent/src/hooks.py:819-827` that emits one WARN and no progress event — unlike its neighbours at `:896` and `:907`, which both call `_try_progress`.

So Dependabot bumps the SDK, constants-sync passes, CI passes, and every checkpoint on every task fails while the system looks healthy. Add the pin cross-check to `drift-prevention`, and emit an event when continuation is unavailable.

### B7 — the approvals table is the one row the new hardening skips

`cdk/src/constructs/agent-session-role.ts:181-190` with `cdk/src/stacks/agent.ts:848-852`. `taskApprovalsTable` sits in `taskScopedTables`, which receives the full `DDB_ITEM_ACTIONS` including `PutItem`, `UpdateItem` and `DeleteItem` under a `dynamodb:LeadingKeys` condition only — no `dynamodb:Attributes` restriction, unlike the `grantAgentTaskTableAccess` allowlist this PR adds for the task table. The worker can therefore update its own approval row to `status=APPROVED`, and `task_state.py:864-895` accepts it, since the `ConditionCheck` asserts only `user_id` plus status membership. It can also pre-set `notified_linear_approval_requested` so the owner never sees the prompt. Pre-existing, but this is the PR asserting that task-scoped worker credentials cannot write this key, and B4 means a suppressed gate now hangs instead of timing out.

### B8 — fail-closed claims CI cannot verify

`ABCA_DDB_LOCAL_ENDPOINT` appears in exactly three test files and two of their own comments; nothing in `.github/workflows/`, `mise.toml` or any compose file sets it. Measured: 56 TypeScript and 11 Python tests skip. Those are the only places a `ConditionExpression` is evaluated rather than substring-compared — double release, empty-counter race, lost acquire/release reply, suspend/resume convergence, and the cancel/approve/deadline races. A `>` versus `>=`, a dropped `attribute_not_exists`, or a wrong `CancellationReasons` index ships green. Either add a DynamoDB Local job, or make the guard fail rather than skip when `CI=true`.

Separately, wake-after-retirement is dead code in all 12 wake tests: `microvm-approval-wake.test.ts:27-32` mocks `microvm-lifecycle`, `logger` and the SDK but not `microvm-continuation-dispatch`, and never sets `CONTINUATION_BUCKET_NAME`, so `microvm-approval-wake.ts:100` short-circuits on every path. That line is the entire "already retired, do not resume the fenced handle" contract, and the failure shape is the one from #815.

Four more are diagnosed but not yet reproduced, listed here so they are not read as missed: an unbounded retirement retry pinning a concurrency seat in `AWAITING_APPROVAL` while `reconcile-stranded-tasks.ts:119-121` exempts exactly those rows; `pipeline.py:1334` treating lifecycle phase `failed` as `closed` and returning `status: "parked"` past both terminal writes; a fence-window race where a decision landing mid-fence resumes a worker retirement then terminates, surfaced to the user as a substrate failure; and `reconcile-microvm-continuations.ts:78` falling back to `task_id` once `microvm_start` has been removed, so the lease condition can never pass and the seat and objects leak on a 5-minute retry loop.

## Non-blocking

**Bootstrap.** The bundle's permissions are complete and the artifacts reproduce byte-for-byte; `test/bootstrap` is 138 passed across 6 suites, `BOOTSTRAP_VERSION` 1.6.0 to 1.9.0 is right, and the golden baseline is updated. But `AWS::SSM::Parameter`, newly synthesized by `task-orchestrator.ts:410`, is absent from `RESOURCE_ACTION_MAP`. No deploy failure today, since the grant at `compute-lambda-microvm.ts:178` is correctly scoped — but it breaks the ADR-002 invariant. The reason it slipped is that `synth-coverage.test.ts:43` synthesizes with default context, so the `lambda-microvm` backend is never covered, and its loop reads the root template only, which this PR makes narrower by adding two nested stacks (one of them unconditional).

**Resource budget.** Measured widest cells at this head:

| Config | Widest |
|---|---|
| agentcore | 477 |
| ecs | 489 |
| microvm-bootstrap | 482 |
| microvm-imported | 488 |
| microvm-managed | 488 |

The body's 489 for ECS and 488 for MicroVM is exact. But `CUSHION = 10` is fiction: the assertion is `count + 1 <= 490`, so real headroom on the ECS path is one resource, and the next resource added there fails CI. Worth saying so in the body rather than leaving it for the next author. Counts are also strictly monotone in both gates across all five configs, so five widest cells plus one all-off baseline is provably equivalent to the 20-cell matrix — about 47s of the file's 67s matrix cost, worth taking while the file is already open. And `toBeLessThanOrEqual(2500)` at `:2273` cannot fail; the observed maximum across all 20 cells is 576.

**Docs.** The Starlight mirror is in sync (regeneration is byte-identical, and a round-tripped sentinel confirms the generator actually runs rather than no-opping). Two real gaps. First, 18 new links point at Starlight pages that do not exist: `docs/verification/` is not a mirrored root, so `../verification/*.md` collapses to `/architecture/<slug>` on the ADR-021, Compute, Security and Deployment-roles pages — exactly what a deploying operator is sent to. Second, `docs/design/CEDAR_HITL_GATES.md:1367` still states `APPROVAL_STRANDED_TIMEOUT_SECONDS, default 7,200` while this PR changed it to 30600 (`stranded-task-reconciler.ts:34`), in a file this PR edits 214 lines of.

Smaller: ADR-021 is a 79% rewrite still marked `Status: proposed` with no changelog; five new context keys and two new env vars are undocumented; `cdk/scripts/build-microvm-artifact.py` is missing from the exhaustive table in `cdk/scripts/README.md`, which this PR edits; no `AGENTS.md` was updated for 15 new agent modules and three new test trees. `docs/verification/645-nesting-probe-results.json` is referenced by nothing, regenerated by nothing, and its `source_commit` is 106 commits behind — it should be deleted. No leaks: zero account ids, ARNs, emails or URLs across the added verification docs.

**Tests.** Manifest-level checksum and identity mismatch in `verifyContinuationCheckpoint` is untested (only the per-object `HeadObject` case is covered), and the failure mode is terminating the only copy of a suspended workspace. `microvm-continuation-retirement.ts` sits at 72.7% branches with the capacity-repair paths uncovered, and its fake store assigns `active_count` rather than evaluating the arithmetic. No test injects a mid-loop failure into the restore rollback at `continuation_workspace.py:774-780`, although the capture-side analogue is covered. `linear-approval-reply.test.ts:27-28` uses the repo's only `virtual: true` mocks, so the dynamic `import('../approve-task.js')` specifier is fabricated and pinned by nothing. About 11 bare `pytest.raises(WorkspaceCheckpointError)` calls omit the `.code` check, in a module where every error funnels into that one type. `orchestrator.test.ts:221-223` is vacuously true when zero Updates were sent, which is the regression it guards.

Clean on two counts worth recording: no new test re-enables `aws:cdk:bundling-stacks`, and `test/live/*` is genuinely uncollected — `--listTests` emits 255 files and none of them. One gap there: nothing type-checks `test/`, so 1,160 lines of live-verification tooling rot silently on a rename.

**Code quality.** Four copy-pasted canonical-JSON comparators (`payload-bootstrap.ts:49`, `microvm-continuation-storage.ts:52`, `microvm-continuation-retirement.ts:58`, `microvm-start.ts:69`), all backing security or idempotency equality, where one divergent line makes a "did my write commit?" comparison silently wrong. `superviseMicrovm` is a 327-line function that establishes the same recovery obligation in five separate places. `hooks.py:2046-2052` sets the PreToolUse hook timeout from a MicroVM contract constant for every backend while its comment justifies it in MicroVM terms only. `approval-notifications.ts:230` carries a comment saying previews stay literal, but the body substitutes every backtick with U+02CB — so a maintainer restoring literalness would let a preview break the fence and inject markdown into Linear. `MICROVM_SUSPEND_GRACE_MS` has zero consumers and contradicts the real per-task grace.

**Process.** This PR carries no labels; the repo uses area, type and priority labels. It is marked ready for review while its own body has two unchecked "before marking ready" boxes, one of which is B2. #645 is assigned to a different maintainer, so ADR-003 ownership is worth confirming explicitly.

## Review coverage

Seven scoped review passes ran over the diff: security and IAM, ADR-002 bootstrap coverage, silent-failure and error-handling, test coverage plus #366 synth performance, contracts and cross-package sync and migration, comment accuracy and human heuristics, and documentation. Type design was folded into the contracts and heuristics passes, since the new types are `ContinuationRecord`, `ContinuationIdentity` and the two state enums.

Two findings were dropped after checking them against the code: the `bgagent-linear-oauth-*` grant on the MicroVM role is not new here (comment-only diff), and the UUIDv5 to v4 change did not lose determinism — `linear-approval-thread.ts:35-41` still derives every bit from SHA-256 and only the version nibble changed, so the retry guarantee is intact.

## Human heuristics

**Proportionality — pass.** No one-caller abstraction; the module split is by lifecycle phase, and the extracted helpers came out of duplication rather than anticipating it. Two accretions: `superviseMicrovm`, and about 150 lines in `continuation_workspace.py:488-609` re-implementing what Python 3.13's `tarfile` data filter provides, without naming the alternative it displaces.

**Coherence — concern.** `attempt_id` and `parked` collide outright (B5); four duplicated comparators; wake/resume and sleep/suspend vocabularies mixed inside single files; `max_manifest_bytes` names three unrelated limits. Credit where due: this PR moves against backend duplication rather than forking MicroVM copies of shared logic.

**Clarity — concern.** The `approval-notifications.ts:230` comment states the opposite of its code. The `orchestrator.ts` rewrite deletes the live-verified evidence for why the MicroVM heartbeat check may never be removed. `microvm-continuation-dispatch.ts:73` asserts an uncited 64-character service limit with no test behind it. Several modules read some limits from `contracts/` and hard-code their siblings.

**Appropriateness — concern.** The service API shape is real, not guessed — generated client, `MicrovmState` enum, synthetic values correctly labelled as local observations. What is mock-only is behavioural: "AWS can report PENDING while restoring" and "a frozen VM retains this connection" read as settled contract in code comments while the verification docs honestly hedge that service-side traces are unavailable. Carry that hedge into the comments. On the Linear side the UUIDv5 lesson was learned and tested at the right altitude (shape, not digest), but `linear-feedback.ts:435-437` still reconciles a lost reply by exact markdown body equality on a round-tripped comment — the same class of assumption, with every test mocking the poster.

## Suggested order

B1 and B2 first; both have consequences outside the MicroVM path. Then B3, B4 and B6, each a few lines. B5 before merge, while the rename is cheap. B7 and B8 could be a stacked follow-up if the body says so, but B8 leaves this PR's central fail-closed claim unverified by CI, so it belongs here.

isadeks:
**Re-review at head `de73239d`** — supersedes the verdict in my [earlier review](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/904#issuecomment-5737270981) at `08baca54`. 16 new commits, +1,728/−935 across 71 files. All four checks green. PR is now a draft.

Verdict: **not ready to leave draft** — one new blocker, introduced by the fix wave itself.

This is a strong response and the quality of the closing is high. Six of the eight blockers are genuinely closed; seven of eight fixes are **mutation-proved** to fail when reverted; **zero tests were deleted or skipped**, and the single deleted assertion was provably redundant (an exact 18-element array compare two lines above already pinned it). Several fixtures moved in the strengthening direction. The approvals-table escalation is closed properly, with a synth-time guard so it cannot silently come back.

What holds it: the fix for the unsafe nesting default traded a data-destroying default for an over-budget one, and the same wave pinned every MicroVM budget cell to the non-default layout, so CI can no longer see it.

## Scorecard

| Prior blocker | Status |
|---|---|
| Forgeable Linear approval | Narrowed |
| Unsafe nesting default | Fixed → new blocker |
| `write_terminal` classifier | Partially fixed |
| `approval_timeout_s: 0` ceiling | Untouched |
| `attempt_id` / `parked` collide | Untouched |
| SDK-pin drift | Fixed |
| Approvals table writable | Closed |
| Fail-closed claims unseen by CI | Fixed |

## New blocking issue

### N1 — the flat layout is now the default, and no test measures it

`cdk/src/stacks/agent.ts:362` correctly inverts the default:

```ts
const microvmNested = microvmNestedContext === true
  || microvmNestedContext === 'true';
```

The same wave then added `microvm_nested_stack: true` to all three MicroVM cells of the budget matrix (`cdk/test/stacks/agent.test.ts:2217,2222,2231`) and to eight other MicroVM synth tests. So the layout a plain `--context compute_type=lambda-microvm` deploy now gets is synthesized by no budget assertion anywhere. Measured directly, flat (the new default):

| Config (flat) | gw0 v0 | gw0 v1 | gw1 v0 | gw1 v1 |
|---|---|---|---|---|
| bootstrap | 481 | 490 | 488 | 497 |
| imported | 487 | 496 | 494 | **503** |
| managed | 488 | 497 | 495 | **504** |

Nested equivalents are all at or under 489. The 503/504 cells cannot synthesize at all — `Template.fromStack` raises `TooManyResourcesInStack`, so `cdk synth` cannot produce a template, let alone deploy one. And `enableLinearIdentityVault` on its own puts imported/managed at 496/497, over the project's own 490 budget.

Two things sharpen this. The vault is not an exotic option here: "enable MicroVM guests to use the existing Linear AgentCore Identity vault grant" is in this PR's feature list, and the Linear approval path is its headline. And the only flat coverage that exists, `AgentStack MicroVM flat-layout migration compatibility` (`cdk/test/stacks/agent.test.ts:1659`), synthesizes with both gates off at 488 and asserts resource identity, not budget. An independent probe driven through the budget test's own harness reproduced 495–497 and a hard failure on the both-gates cell.

Note also that ECS moved 489 → 490 in this wave: the new approval-service Lambda took the last slot, so ECS now sits exactly on the budget with zero margin.

Fix: add a cell with `microvm_nested_stack` unset (or generate each MicroVM cell over `[undefined, true]`), then settle the question it exposes — if flat cannot carry the gates, either nesting becomes the default with a guarded migration, or the gate combination fails fast on flat with a readable error instead of a synth crash.

## Residual — Linear approval forgery is narrowed, not closed

`95591d44` is a real fix. `cdk/src/handlers/shared/linear-approval-reply.ts:55-62` re-reads the comment from Linear and requires `comment.id`, `!comment.botActor`, a present `comment.user?.id`, the issue id, the parent id, and that the authoritative body parses to the same decision; `:78` resolves the platform user from `comment.user.id` rather than the webhook's `actor.id`. A lookup failure throws rather than falling back. With `actor=app` — the default install — a worker-created comment carries `botActor` and a null `user`, so it is rejected twice over. The attack I described is closed for that install.

It stays open on one install variant. `cli/src/commands/linear.ts:724` and `:1541` ship `--no-actor-app`, a documented diagnostic that completes the same OAuth flow without `actor=app` and persists the resulting user-scoped token into the same `bgagent-linear-oauth-*` secret the worker can read. On such a workspace the worker can create a genuine comment attributed to the authorizing human, then sign a webhook pointing at it, and every new predicate passes. Fix: record the actor mode in the registry row at setup and refuse to arm approval verification on a non-`actor=app` workspace, or reject any comment whose author maps to the identity the stored token itself authenticates as.

Worth stating plainly: the root cause is untouched. The worker still holds the webhook HMAC key, so every other webhook-driven path remains forgeable. This fix is per-path, not structural.

## Residual — the terminal-write fix is half converted, again

`write_terminal` now uses the transaction-aware classifier (`agent/src/task_state.py:360`), and the classifier correctly treats only `["ConditionalCheckFailed","None"]` as benign, routing everything else to `log_error_cw` with the raw `CancellationReasons`. Right shape, and mutation-pinned.

But the same range edit missed its sibling: `write_trace_uri_conditional` still string-matches the pre-transaction error code at `agent/src/task_state.py:452-455`, so on the continuation substrate a benign precondition failure arrives as `TransactionCanceledException` and is logged as a failure rather than the intended INFO "skipped". The consequence is log fidelity, not a wrong success — both paths return `False` — but it is literally the second half of the finding.

`write_terminal` is also still `-> None` with all four callers discarding it (`agent/src/pipeline.py:469,1827,1864`, `agent/src/server.py:536`), so a revoked-lease drop is now loud in logs but still reported to the pipeline as success — and nothing reclaims the row, because `cdk/src/handlers/reconcile-stranded-tasks.ts:89` sweeps `SUBMITTED | HYDRATING | AWAITING_APPROVAL` and not `RUNNING`. Related new exposure from this PR's own transaction change: a concurrent touch of the worker-lease row yields cancellation reason `TransactionConflict`, which botocore does not retry and this code does not either.

## Non-blocking

**The 18 broken doc links were rewritten, not repaired.** `docs/scripts/sync-starlight.mjs:35-40` now rewrites `../verification/*.md` to absolute `blob/main/...` GitHub URLs instead of mirroring `docs/verification/` as a Starlight root. I checked all 18 rewritten links and confirmed all four target files are absent from `upstream/main`, so they resolve only once this PR merges — broken in previews and in any non-`main` build until then. The internal 404s are genuinely gone (71 → 53 broken internal links, none new). Two follow-ons: the hardcoded `main` ref should come from the build ref, and `docs/scripts/link-check.sh:7` still scans only `guides design decisions` plus root `*.md` and never `docs/src/content/docs/`, while `.markdown-link-check.json` ignores both `^https?://` and the Starlight route prefix — so nothing validates the rewrite's output, which is why the original defect stayed green. The generator change also mutated five mirrored pages whose sources were never touched, so mirror churn in this wave is not evidence of a source edit. Mirror itself is in sync: the generator produces zero diff.

**Three false claims in the new prose.** `docs/design/CEDAR_HITL_GATES.md:1366` quotes a deny reason `"insufficient lifetime"` that does not exist — the code emits `f"insufficient maxLifetime remaining ({remaining_lifetime}s) for approval"` (`agent/src/hooks.py:670-673`) — and scopes the deny to "an explicitly timed approval" when the check is unconditional on the timeout. `docs/design/SECURITY.md:45` says task ownership is "checked transactionally"; the transaction does detect a concurrent owner or state change, but `:user` is read from the same task row moments earlier, so caller-to-task binding comes solely from the IAM task-tag path. And `docs/guides/DEPLOYMENT_GUIDE.md:259` tells the operator to verify SessionRole grants, which do not exist on the legacy no-session-role ECS configuration — there the worker gets read-only unscoped approvals and no `execute-api:Invoke` while still receiving `APPROVAL_REQUESTS_API_URL`, so every gate 403s and then denies, and the runbook step cannot detect it. Smaller: `docs/guides/DEPLOYMENT_GUIDE.md:248` explains a behaviour change by citing "PR #904", which means nothing to an operator after merge; `CEDAR_HITL_GATES.md:1836` still asserts Slack approve/deny buttons against the corrected `:1706`; and `APPROVAL_REQUESTS_API_URL` sits in `contracts/constants.json` `env_by_key` but not `required`, so a worker missing it silently falls back to the direct-write path that new stacks deny. The stale 7,200 figure is fixed (now 30,600) and the probe JSON is deleted with no dangling references.

**New service — hardening, none of it restoring the original escalation.** `sts:TagSession` on the shared compute role still carries no condition on `aws:RequestTag/task_id` (`cdk/src/constructs/agent-session-role.ts:364,379`), so task identity remains self-asserted and a compromised worker can tag itself with another task's id and drive that task's gate; pre-existing, but the new service's security rests on that tag. `cdk/src/handlers/request-approval.ts:132-148` accepts `timeout` without comparing `deadline_epoch`, so a worker can instantly force-deny its own pending gate, making `TIMED_OUT` an untrustworthy audit signal. `validId` accepts any 1–128 characters while the id is interpolated into an `execute-api` resource ARN where `*` is a wildcard (unreachable today, ids are ULIDs). The new `RestApi` has no resource policy, so it is a public endpoint guarded only by SigV4 and the path-scoped grant. And `cdk/src/constructs/ecs-agent-cluster.ts:513-517` can wire the env var without the grant.

**Two test-coverage gaps, both mutation-proved.** The four new manifest-rejection cases in `cdk/test/handlers/shared/microvm-continuation-storage.test.ts:155-177` assert only the shared `MICROVM_CONTINUATION_STORAGE_INVALID` prefix, and neutering either manifest guard leaves 17/17 passing, because the fixture omits the object receipts so a downstream guard supplies the same prefix — assert the specific messages and add the receipts. And `cdk/test/handlers/request-approval.test.ts:68-77` never asserts the lease `ConditionExpression`, so inverting the MicroVM stale-writer fence leaves the unit suite green at 16/16; only the DynamoDB-Local suite catches it, with 7 failures. Both are one-line fixes. `cdk/test/stacks/microvm-managed-image-nag.test.ts:33` likewise now covers only the nested layout — though credit where due, nesting does not blind cdk-nag: a wildcard injected onto a role inside the nested stack was caught by all four cells.

**Bootstrap is clean.** The new approval service introduces zero new CloudFormation resource types — it is a nested stack, 12 resources plus one root `AWS::CloudFormation::Stack`, and all 11 types already existed — so no policy, ARN pattern or `BOOTSTRAP_VERSION` bump was required, and none was made. Artifacts reproduce byte-for-byte and the bootstrap suite is 139/139. `AWS::SSM::Parameter` is now mapped at `resource-action-map.ts:52`, closing that finding. `synth-coverage.test.ts` now recurses nested stacks and covers the microvm backend, but its microvm cell pins `microvm_nested_stack: true` — the same narrowing as N1; flat and nested happen to synthesize identical type sets today, so that one is future-proofing rather than a live hole. One real consequence of the default flip: `compute-lambda-microvm.ts:158`'s flat `PassRole` prefix pattern has roughly two characters of truncation margin, and flat is now the primary path, while the nested roles carry explicit short names and are immune.

**Still untouched from the first pass.** The `attempt_id` dual meaning and the `parked` inversion (both files byte-identical at the two heads); `_compute_effective_timeout`'s early return at `agent/src/hooks.py:1219`, with `agent/tests/test_approval_retention.py:21-23` still pinning it; `approval_timeout_s: 0`, whose rationale was added to a design doc rather than an ADR, with ADR-021 still `proposed`; `cdk/src/migration/` still reachable only from its own tests, and the new `canonical-json.ts` duplicating the `canonical`/`digest` role of `cdk/src/migration/template.ts` instead of importing it; the PreToolUse hook timeout still derived from a MicroVM constant for every backend; `linear-feedback.ts:465` still exact-body equality with no normalization and no test feeding a body Linear reformatted; and still zero labels. Fixed: `MICROVM_SUSPEND_GRACE_MS` deleted, and the `approval-notifications.ts` comment corrected to match its code.

One correction to my own earlier review: on ADR-021's missing changelog, `docs/decisions/README.md` gates that requirement on post-acceptance refinement, and 1 of 22 ADRs complies. The real point is narrower — the status is stale for shipped P3 work, not that a rule was broken.

**Governance.** The trusted approval service is a second piece of untracked scope. #645 scopes a ComputeStrategy backend and names no approval-request service; commit `09b79411` carries no issue reference; no ADR describes it; and it is not MicroVM-specific, since it strips direct approval writes from AgentCore and ECS too. By `docs/decisions/README.md`'s own criteria — multiple packages, establishes a pattern, hard to reverse — it wants an ADR. Separately, `docs/AGENTS.md:35,63` marks `sync-starlight.mjs` as READ and ask-first, and `421146a5` changes exactly that.

## Documentation

Mirror in sync, no hand-edited generated files, stale timeout figure corrected, probe JSON removed cleanly, new cross-doc anchors resolve, and most of the new trust-boundary claims verify true against the code. Missing: the three false claims above, the no-session-role ECS gap in the upgrade runbook, `APPROVAL_REQUESTS_API_URL` absent from `agent/README.md` despite `agent/AGENTS.md:87` mandating it for new env vars, the DynamoDB-Local CI service documented nowhere, `cdk/AGENTS.md:32`'s glob excluding the new `request-approval-local.test.ts`, and no AGENTS.md row for either new module.

## Tests and CI

All four checks green at `de73239d`. Verified by running locally: `test/bootstrap` 139 passed; the six approval and continuation suites 93 passed with 9 skipped; the `agent` trio 228 passed; and the three DynamoDB-Local suites against a real pinned container, 65 passed. Bootstrap synth-coverage passes and now covers this PR's backend.

The CI-visibility blocker is properly closed. `.github/workflows/build.yml:55-60` runs `amazon/dynamodb-local` pinned by digest, the endpoint is wired into the one step that runs both the Python and the CDK suites, and all four previously-skipping suites now throw when `CI=true` without an endpoint — verified by running them that way. The Python guard is a module-scope raise that aborts collection, which is stronger than a failing test. Two residuals: a self-hosted runner without Docker would surface this as a test failure rather than an infrastructure one, and `agent/tests/test_continuation_sdk_probe.py:206` is still gated on a variable CI never sets, so the pinned SDK's behaviour stays unexecuted even though its version string is now enforced.

## Review coverage

Five scoped passes over the fix wave as new surface: security of the new trusted service plus the Linear fix, ADR-002 bootstrap coverage, test integrity (deletions and weakenings, then mutation testing), fix correctness with a differential harness on the canonical-JSON refactor, and documentation plus a ledger of prior findings. Type design was folded into the security and docs passes — the wave's new types are the `REQUEST_FIELDS` allowlist and the `canonical-json` helper, both covered. The silent-failure and human-heuristics passes were scoped to the fix wave rather than re-run over the whole PR, since the underlying code is unchanged.

Worth recording: `80977e77` claims "without changing stored bytes" for the four-way canonical-JSON dedup, and that claim was checked rather than trusted. A differential harness against all four original implementations recovered from `08baca54` found byte-identical output across 28 hand-built edge cases — key reordering, `undefined` versus missing keys, unicode collation, `-0` and `1e21`, Buffers, a `JSON.parse`-injected `__proto__`, `toJSON` objects — times three comparators, plus 30,000 fuzzed inputs. Start-replay hashes and previously-stored receipts are safe across this deploy.

## Human heuristics

**Proportionality — pass.** The new service is about 12 resources and one handler for a real trust boundary, not an abstraction for a one-off, and the canonical-JSON extraction removes four copies.

**Coherence — concern, unchanged.** `attempt_id` and `parked` still collide, and `canonical-json.ts` duplicates the role of `cdk/src/migration/template.ts` instead of importing it.

**Clarity — improved.** The `approval-notifications.ts` comment now matches its code. One new gap: `ContinuationCheckpointError` has no `code` attribute, so `getattr(exc, 'code', 'checkpoint_failed')` always yields the same literal and SDK drift, S3 throttling and invalid checkpoint JSON produce byte-identical operator output.

**Appropriateness — concern.** The new service's cdk-nag justification says the handler "validates the operation and an explicit approval-field allowlist before every write". That is true but over-reads as semantic validation: `tool_input_sha256` is never checked against `tool_input_preview`, and `severity`, `reason` and `matching_rule_ids` are worker-asserted Cedar outputs. Consent is bound to a worker-authored description, and that boundary belongs in `CEDAR_HITL_GATES.md`.

## Suggested order

N1 first, since it is the one thing this wave introduced and it decides an architectural question: can flat carry the gates, or must nesting be the default with a guarded migration. Then the two one-line test assertions, then the `--no-actor-app` gap, then the second half of the terminal-write fix. The `approval_timeout_s` ceiling and the `attempt_id`/`parked` renames are still open and still cheap; if they are deliberate declines, saying so in the body will stop the next reviewer re-raising them.

isadeks:
**Third-pass review at head `b71bef0e`** — supersedes the verdict in my [re-review](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/pull/904#issuecomment-5760825785) at `de73239d`. 10 new commits, all four checks green, labels now applied.

Verdict: **two items from ready to leave draft.**

N1 is fixed, and fixed the right way. This wave also closed the `--no-actor-app` consent hole, made the ECS approval misconfiguration a synth-time throw, mirrored `docs/verification/` as a real Starlight root behind a new link gate that provably bites, and added ADR-023. The body now carries explicit declines with stated reasons, which resolves my last-pass complaint that silence was indistinguishable from intent.

What I would fix before leaving draft: this wave introduced one user-visible regression, and the fix it belongs to is unpinned at the wiring level — reverting both production call sites leaves the entire 2,204-test agent suite green.

## N1 is properly closed, and the fix is mutation-proved load-bearing

`2ae8f116` did not move the goalposts. It moved resources: `MicrovmContinuationManager`, `AdmissionQueuePickup`, `StrandedTaskReconciler` and `PendingUploadCleanup` into the existing `ConcurrencyMaintenance` nested stack (`cdk/src/stacks/agent.ts:1415-1470`). Budget constants are untouched. Measured at this head, widest cell per configuration:

| Config | Widest |
|---|---|
| agentcore | 463 |
| ecs | 475 |
| microvm flat | 482–484 |
| microvm nested | 468–469 |

Every cell dropped roughly 20 resources. Worst case anywhere is 484 against 490, and ECS went from sitting exactly on the budget to 15 under. The matrix was widened rather than narrowed — each MicroVM mode now expands into `-default-flat` and `-nested`, 20 cells to 32.

The strongest evidence is a mutation: moving those four constructs back to the parent produces 9 failures, every one of them in a newly-added `-default-flat` cell, including two `TooManyResourcesInStack` at 502 and 503. The old nested-only matrix would have been fully green on a stack that cannot synthesize. The widening was load-bearing, not padding.

## New issue introduced by this wave

### R1 — a benign cancel race now produces a false failure and a false ❌ on the user's issue

`agent/src/pipeline.py:473-481`:

```python
outcome = task_state.write_terminal(task_id, status, result)
if outcome in (
    task_state.TerminalWriteOutcome.FAILED,
    task_state.TerminalWriteOutcome.SUPERSEDED,
):
    raise task_state.TerminalWriteError(...)
```

`SUPERSEDED` is the documented benign case. It is returned from the branch whose own log line reads "status precondition not met (task may have been cancelled)" (`agent/src/task_state.py:441`), and `pipeline.py:1363` notes that the in-band cancel path deliberately skips the terminal write. Treating it as fatal produces the following, traced end to end in the code:

1. The task succeeds; the user cancels after the last in-band cancel check but before the terminal write.
2. `write_terminal` fails its `#s IN (:running, …)` condition and returns `SUPERSEDED`.
3. `_persist_finished_task` raises `TerminalWriteError`.
4. `run_task`'s `except Exception` at `pipeline.py:1843` catches it, calls `write_terminal(..., "FAILED", ...)` at `:1876` (which fails the same condition and is discarded), then `react_task_finished(success=False)` at `:1882` — putting ❌ on the Linear issue for a task that succeeded and was deliberately cancelled — and re-raises, so the runtime reports a crash.

Fix: treat `SUPERSEDED` as non-fatal (log and return) and reserve the raise for `FAILED`, which is the case the original finding was about. The same path reaches `:469` for repo-less tasks.

### R2 — the terminal-write fix is unpinned at the wiring level

Reverting both production call sites (`agent/src/pipeline.py:469` and `:1839`) from `_persist_finished_task(...)` to a bare `task_state.write_terminal(...)` leaves the full agent suite green: 2,204 passed, 11 skipped, exit 0. The two new tests import and call `_persist_finished_task` directly (`agent/tests/test_pipeline.py:2348,2362`), and nothing else in `agent/tests/` references the helper or `TerminalWriteError`. So the suite proves the helper raises, never that it is reached.

That is the same green-suite-at-the-wrong-seam shape that let the silent-COMPLETED class through before. Fix: drive `run_task` and `_run_repoless_task` with `write_terminal` patched to return `FAILED` and assert the error escapes the public entry point, then extend the existing AST contract test to assert that `pipeline.py` has no `write_terminal` caller other than `_persist_finished_task`.

## Verdicts on the two security fixes

**The `--no-actor-app` consent hole is closed, and more cleanly than I suggested.** `1443e3dd` adds `viewer { id }` to the same authenticated `VerifyApprovalComment` query and rejects when `comment.user.id === viewerId` (`cdk/src/handlers/shared/linear-feedback.ts:432,439-446`). Because it reads the identity live rather than recording actor mode at setup, pre-existing `--no-actor-app` installs are covered the moment the Lambda deploys, with no migration. It fails closed on a missing `viewer.id`, a single round trip leaves no race window, and a token from another workspace trips the existing organization-mismatch throw. It also does not regress the default install: with `actor=app` the viewer is the synthetic `@oauthapp.linear.app` bot user, not the authorizing human.

One residual that wants a live probe rather than a code change: the readback trusts the comment's current body and author, so a worker holding the token could `commentUpdate` a genuine human reply in the thread to read `approve` and then forge a `create` webhook naming it. Whether Linear permits editing a comment the token did not author is a one-query check. If it does, add `editedAt` to the verification and reject edited comments.

**The ECS approval-writer misconfiguration is closed at synth.** `cdk/src/constructs/ecs-agent-cluster.ts:325-327` now throws when approvals are wired without both the session role and the service URL, the legacy direct read grant is deleted, and `approval_requests_api_url` joined `contracts/constants.json` `required` — enforced twice on MicroVM, producer-side with a redeploy remedy and guest-side in the manifest allowlist. The gap: AgentCore has no equivalent guard (`cdk/src/constructs/task-orchestrator.ts:530-532` wires the variable conditionally), so a hand-edited environment there still falls back to the now-denied direct-write path and surfaces an opaque `AccessDenied` at the first gate.

## Non-blocking

**A third half-converted range edit.** The deny-reason string was corrected at `docs/design/CEDAR_HITL_GATES.md:1366`, but its twin at `:1886` still says `"insufficient maxLifetime for approval"` while the code emits `"insufficient maxLifetime remaining ({n}s) for approval"` (`agent/src/hooks.py:670-673`), and the same sentence claims "Task continues without a gate" when the hook returns a deny. That is the third instance of this pattern in this PR, after `write_terminal`/`write_trace_uri_conditional` and the two deny-string sites. Worth a habit: after fixing one site, grep the symbol and assert the count.

**Checkpoint classification is partial, and the default asserts something the code does not know.** `22d8bae9` gave `ContinuationCheckpointError` a real `code`, which closes the "always the same literal" finding for three named classes. But the default is `"checkpoint_invalid"` rather than a neutral `"checkpoint_failed"`, and roughly 50 raise sites pass no code — so SDK private-API drift (`agent/src/continuation_usage.py:62`) and "checkpoint bucket unavailable" (`continuation_session.py:364`) both now tell the operator the checkpoint data is invalid. Three of the six codes are also unpinned: mutating `checkpoint_sdk_timeout`, `checkpoint_storage_unavailable` and the decode-path `checkpoint_invalid_json` to the default leaves the suite green. And the six codes are documented nowhere, though `docs/src/content/docs/verification/645-p3-lifecycle-diagnostics.md` already lists `code` as a Logs Insights field without enumerating values.

**Normalization is pinned in only one direction.** `02ad141e` fixes the reported CRLF symptom and adds a test that feeds a differently-formed body, which closes the prior gap. But an over-broad normalizer (`body.replace(/\s+/g,'').toLowerCase()`) passes 103 of 103, so nothing stops a future widening from making two different intended bodies compare equal and suppressing a real prompt. Add a negative case differing only by internal whitespace and case. Design note: since the comment id is a client-generated UUIDv4 for that one write, full-body equality buys little over id plus destination while remaining the sole source of false negatives.

**The SDK probe now runs in CI, but by accident of `CI=true`.** Verified locally: `CI=true uv run pytest tests/test_continuation_sdk_probe.py` passes 2 tests in 12.7 s with no network, using a loopback synthetic model and the SDK's bundled CLI. But `ABCA_TEST_SDK_CONTINUATION` is still absent from `build.yml` and the gate is still a silent `skipif`, so any runner that does not set `CI=true` reverts to the dark state the prior review found. Set the variable explicitly in the build job's env.

**The upgrade docs get the hazard backwards.** `docs/guides/DEPLOYMENT_GUIDE.md:283` tells operators to drain before the maintenance-stack move. There is no gap — CloudFormation creates the nested stack before cleanup, so the real transient is both schedules firing concurrently for the duration of the update. Four of the five moved jobs are safe under duplicate firing because every mutation is conditionally fenced; the fifth, `cdk/src/handlers/reconcile-microvm-continuations.ts:143-151`, writes a shared scan cursor with no `ConditionExpression`, so overlapping instances can clobber it and skip a page (self-healing on the next pass, and pre-existing, but the upgrade makes overlap certain rather than incidental). Also undocumented: this applies to every deployment including the default AgentCore one, roughly 20 replaced resources, and the old Lambda log groups are orphaned under their old generated names, which will confuse the first person debugging right after the upgrade. Mechanically the move is safe — no moved construct sets an explicit `functionName`, `ruleName`, `roleName` or `logGroupName`, and zero tables, buckets, queues, event-source mappings, versions or aliases move.

**One unguarded link class remains.** `docs/verification/` is now a real mirror root and the new `docs/scripts/check-site-links.mjs` gates the build; it was proven to fail on missing pages, missing runbooks and broken anchors. The `blob/main/...` class is still skipped by design as an offline check, with the ref hardcoded. All 13 emitted blob URLs currently exist on main, so nothing is broken today, but a local `fs.existsSync` against the working tree would close the class without breaking the offline contract.

**Smaller items.** `cdk/src/constructs/approval-request-service.ts:89`'s cdk-nag justification still reads as semantic validation, though the docs now correctly state that the preview and policy metadata are worker assertions — and the preview is truncated at 256 bytes, so it cannot verify the hash. `docs/AGENTS.md:37,65` still mark `sync-starlight.mjs` as READ and ask-first while this wave rewrote it again. `canonical-json.ts` has no AGENTS.md row. And `cdk/src/handlers/request-approval.ts:49` now validates `worker_attempt_id` — a vendor-supplied MicroVM identifier — against a self-chosen charset, which fails every gate closed if the service ever emits a `.` or `:`.

## Declines, recorded properly

The body now states these with reasons, which is the right place and resolves the last-pass complaint:

- **No default approval expiry is deliberate**, with worker and resource lifetime kept separate (ADR-023). Accepted; that closes the `approval_timeout_s` ceiling finding as a decision rather than an oversight. One loose end: `contracts/constants.json:7-11` still has `"default": 0` sitting outside its own declared `"min": 30`, which reads as a bug even when it is not.
- **A public SigV4 endpoint is intentional**, with task-path IAM authorization. Accepted.
- **Ambient credentials can still select session tags**, legacy bundles still carry the webhook signing key, action descriptions are worker assertions, and `TIMED_OUT` is not proof a human deadline elapsed — documented as limits rather than claimed solved. That is the honest framing for a reference sample; accepted.
- **No second broad `RUNNING` reaper**, with existing heartbeat and backend monitoring covering lost workers. Accepted as a judgment call. Worth noting the one hole it leaves: `agent/src/server.py:536` is the outermost writer and still discards the outcome, so a failed terminal write there leaves a permanently-`RUNNING` row holding a concurrency slot, visible only in a log line.
- **Checkpoint field renames and hash-encoding consolidation need a versioned migration**, not a style-only edit. Accepted, and `cdk/src/migration/template.ts:39` now carries a comment explaining why the two canonical encodings must not be interchanged, which is better than the rename I asked for. `attempt_id` and `parked` therefore stay; `microvm-continuation-types.ts:45` documents the `PARKED` distinction, though the Python side has no reciprocal comment.

## Documentation

Mirror in sync, zero diff. Three false claims fixed, plus the "PR #904" explanation and the Slack self-contradiction. ADR-023 is substantive — it names the residual `TagSession`, `TIMED_OUT` and worker-assertion risks rather than glossing them. `APPROVAL_REQUESTS_API_URL` is now in `agent/README.md` and in `required`, and `cdk/AGENTS.md` now names `request-approval-local.test.ts`. Missing: the `:1886` twin, the six checkpoint codes, the AgentCore failure symptom, and the upgrade-hazard correction above. Both new ADRs (021 and 023) remain `proposed`.

## Tests and CI

All four checks green. Verified by running locally: the nine approval, continuation and strategy suites 415 passed; `synth-coverage` plus the nag suite 15 passed; the full `agent.test.ts` 205 passed in 157 s; the budget block 64 passed in 81 s; and the agent suite 2,204 passed with 11 skipped. Bootstrap synth-coverage is now parameterized over `microvm_nested_stack` both ways, closing the narrowing flagged last pass.

Census: no test was deleted without replacement, no `.skip`, `.only` or `xfail` was added, and no assertion was loosened. Six declarations disappeared; all six are renames or intentional replacements, and the one genuine behavioural deletion — the ECS read-only IAM assertion, now a constructor-throw assertion — has its property still asserted exactly at `cdk/test/constructs/agent-session-role.test.ts:147`. Both guards flagged last pass now bite under mutation, and the lease fence no longer needs DynamoDB Local to catch an inversion.

Cost note: the budget block is now 81 s of that file's 157 s. The mutation output shows the widest cell per configuration is the maximum, so 8 widest cells would catch the same regression in about 20 s — but monotonicity across the two gates is an unasserted assumption, and each cell also carries a per-cell API Gateway assertion, so the honest trim is 8 cells for the budget assertion plus that second assertion once.

## Review coverage

Four scoped passes over this wave as new surface: security of the two consent and configuration fixes plus leakage and normalization review; test integrity with systematic mutation testing; fix correctness and upgrade safety of the nested-stack move; and documentation with an open-item ledger. Type design was folded into the security pass, since the wave's new types are `TerminalWriteOutcome`, `TerminalWriteError` and the checkpoint `code` vocabulary. Bootstrap coverage was re-checked inside the security and docs passes rather than as its own agent, because the wave adds no new CloudFormation resource type; `synth-coverage` passing over both layouts was verified by running it.

## Human heuristics

**Proportionality — pass.** The budget fix is the minimum intervention that works: move stateless jobs, change no constants.

**Coherence — concern, now deliberate.** `attempt_id` and `parked` still collide, declined with a migration reason and partially documented. The two canonical-JSON encodings are now explained rather than merged.

**Clarity — mixed.** `TerminalWriteOutcome` is a clear four-state enum. Against it: `checkpoint_invalid` as the default code asserts a cause the code does not know, and the `:1886` doc sentence contradicts both the emitted message and the actual control flow.

**Appropriateness — improved.** ADR-023 states residual risks plainly instead of implying the writer API solved them, and the new link gate was verified to fail on real broken links rather than assumed to work. Remaining: `worker_attempt_id` charset-validated against a vendor-supplied identifier, and the Linear comment-mutation question that only a live probe settles.

## Suggested order

R1 first — it is a one-line severity change and it currently mispaints a successful task as failed for the user. Then R2, since without it R1's whole fix class can silently regress. The `:1886` twin and the three unpinned checkpoint codes are quick. The AgentCore guard and the upgrade-doc correction I would want before this leaves draft; everything else reads as follow-up.

ayushtr-aws (changes requested):
## Review at head `548bdf48` — Request changes

The engineering here is unusually rigorous, and the security boundary work (trusted approval writer, Linear consent read-back, payload bootstrap) survived an adversarial security pass with no findings above threshold. But the PR ships two verified correctness defects on the new wake path, one unbounded storage leak, an unsafe-by-default layout change with unfinished migration tooling, and a governance gap.

### Vision alignment

- **Tenet 1 / Tenet 5 — advanced.** Approvals survive worker retirement, decisions stay owner-attributed, workers lose direct approval-write access, and the session role can only invoke `/tasks/${aws:PrincipalTag/task_id}`.
- **Tenet 6 — traded.** `approval_timeout_s` default 300 → 0 applies to *all* backends, so an unanswered gate on AgentCore/ECS bills to the worker lifetime ceiling. Documented only in ADR-023, which is `proposed`.
- **Tenet 4 — traded.** `microvm_nested_stack` defaults to `true` with an advisory warning (see B2).
- **ADR-003 — violated.** See B1. ADR-021 and ADR-023 both remain `proposed` and need maintainer acceptance as part of merging.

### Blocking

**B1 — Scope and governance (ADR-003).** #645's acceptance criteria cover a MicroVM `ComputeStrategy`. This PR also delivers a new IAM-authenticated approval-writer API, native Linear approve/deny replies (~1,000 lines across 8 modules), a repo-wide change to the default approval expiry, a `ConcurrencyMaintenance` nested stack that affects *every* deployment including default AgentCore, and a nested-by-default MicroVM layout. None has an approved issue (#793 separately wants the ADR-023 number). #645 is assigned to @dreamorosi, not the author. Your own first self-review flagged this. Please file approved issues for the approval-writer API, Linear reply approvals, and the maintenance-stack split (or re-approve #645 with widened deliverables), and get assigned.

**B2 — `cdk/src/stacks/agent.ts:361` nests MicroVM by default while migration is unfinished.** The checklist item "Complete reusable flat-to-nested migration commands and independently verify upgrading a current-main deployment" is unchecked. `cdk/src/migration/permissions.ts` and `template.ts` are referenced only by tests, with no script, mise task, or doc entry point. The synth warning at `:364-372` does not block a `cdk deploy` that can delete payload/artifact buckets. Fix: when `compute_type=lambda-microvm` and `microvm_nested_stack` is unset, **throw** until migration tooling lands and is verified; then downgrade to a warning.

**B3 — `cdk/src/constructs/task-api.ts:1460-1471` never sets `USER_CONCURRENCY_TABLE_NAME` on the approve/deny Lambdas.** `enableMicrovmContinuations` adds only `CONTINUATION_BUCKET_NAME` and `ORCHESTRATOR_FUNCTION_ARN`, but `approve-task.ts:267` / `deny-task.ts:241` → `wakeMicrovmAfterApproval` → `dispatchMicrovmContinuation:63` → `admitContinuation`, which reads `process.env.USER_CONCURRENCY_TABLE_NAME!` (`microvm-continuation-start.ts:32`). A CLI/API decision on a PARKED task runs a TransactWrite with `TableName: undefined`, is caught as `wake-reconciliation-failed`, and the replacement launches only when the 5-minute reconciler sweep reaches the row. `linear-integration.ts:330` sets the variable, which is why the six Linear live flows passed. Fix: set `USER_CONCURRENCY_TABLE_NAME` and `MAX_CONCURRENT_TASKS_PER_USER` in `enableMicrovmContinuations`, and add a CDK assertion on both decision functions' environment.

**B4 — `cdk/src/handlers/reconcile-microvm-continuations.ts:72-85` throws forever for a replacement that failed before launching.** `admitContinuation` writes the lease ACTIVE (`microvm-continuation-start.ts:101`); the runner can throw `MICROVM_CONTINUATION_VERSION_CHANGED` (`microvm-continuation-runner.ts:181`) before any start receipt exists; the catch at `:262-269` calls `failContinuationAttempt`, which moves the lease to FENCED (`:85-92`). For a terminal task with no `microvm_start`, the reconciler gate accepts only PARKED/CLOSED or ACTIVE-with-slot-held, so a FENCED lease throws `MICROVM_CONTINUATION_LEASE_INVALID` every sweep and `deleteClosedTaskContinuations` at `:108` never runs. `continuation-bucket.ts:39-41` has no object expiration, so up to 1 GiB per affected task is retained indefinitely behind a "will retry" warning. Fix: treat FENCED with no `lease_microvm_id` as never-launched (or have `failContinuationAttempt` write CLOSED when no handle exists), and add a lifecycle expiration to the bucket as defence in depth.

**B5 — `agent/src/hooks.py:1978-1979` leaks tool registrations when a second PreToolUse denier fires.** `_pre` calls `lifecycle.tool_finished` only on the non-allow path. Repo tasks load `setting_sources` including `project` (`runner.py:428`), so a target repo's `.claude/settings.json` can register a PreToolUse hook or deny rule that blocks a tool the Cedar hook allowed. If the CLI then returns an error tool result without emitting PostToolUse/PostToolUseFailure, the id stays in `MicrovmLifecycle._tools`, and `suspend()` (`microvm_lifecycle.py:355`) / `continuation_checkpoint()` (`:262`) refuse for the rest of the task. *Confidence note:* the code path is verified; the CLI behaviour (no Post* event when a later hook denies) was checked by an automated pass against CLI 2.1.258 but not reproduced by me. If the CLI does emit a failure event there, this drops to a hardening nit. Either way, evicting stale ids at the Stop hook plus a test for the "no Post* hook arrives" case is cheap.

### Non-blocking

**Correctness / operations**
- `task-concurrency.ts:172-178` + `reconcile-microvm-continuations.ts:63-67`: a *known* `RunMicrovm` 4xx rejection after receipt and lease exist holds the seat for the full 8-hour lifetime with no log. Replacement launches only. Record the rejection on the receipt so release is immediate.
- `request-approval.ts:95-99`: the lease `ConditionCheck` is unconditional for `lambda-microvm`; a task RUNNING before this deploy has no lease row and fails closed at its first gate. Mitigated by the documented drain step; say so in the error string.
- `aws_session.py:388-389`: advisory `refresh_needed()` escalates to `_protected_refresh(is_mandatory=True)` on a 1-attempt/2 s STS client, so a transient STS stall becomes a broker 503 while cached keys are still valid.
- `continuation_workspace.py:222`: rejecting `st_nlink != 1` will fail every checkpoint if `uv` hardlinks site-packages into the workspace. Set `UV_LINK_MODE=copy` in the image or relax the check.
- `lambda-microvm-strategy.ts:918-943`: `ConflictException` on Terminate is now `unconfirmed` and emits ERROR + a task event on the routine cancel/finalize race; confirm via `GetMicrovm` first.
- `continuation_runtime.py:239-242`: `approval_scopes` are seeded from the worker-written manifest without cross-checking APPROVED rows. Not a new boundary crossing, but validate against the approvals table on restore.
- `server.py:536` still discards the `write_terminal` outcome on the outermost crash path (noting your decline for the record).

**Efficiency**
- `continuation_session.py:258-270`: eager append deep-copies and re-encodes the whole conversation per frame — O(n²) on the SDK loop.
- `continuation_storage.py:405`: re-downloads the tar it just uploaded to learn `VersionId`; use `HeadObject` with `ChecksumMode=ENABLED`, as the coordinator already does at `microvm-continuation-storage.ts:229`.

**Design / clarity**
- `microvm-start.ts:36-42` `StartReceipt` is private and re-declared inline in three modules; `TaskRecord` lacks `microvm_start`. Export and share.
- `microvm-continuation-types.ts:42-53` `ContinuationRecord` has state-dependent optionals; make it a discriminated union like `ApprovalRecord` in this same PR.
- `approve-task.ts:231-290` and `deny-task.ts:208-265` are a ~60-line duplicate.
- `contracts/constants.json` `approval_timeout_s` has `default: 0` outside `min: 30`, and the sync script was relaxed to allow it. A `no_deadline_sentinel` field would make the schema self-describing.
- `approval-request-service.ts:89` cdk-nag reason still reads as semantic validation.
- New dependency `@aws-sdk/client-ssm` and edits to `.github/workflows/build.yml` are "ask first" items per AGENTS.md; both look justified but should be called out in the body.

### Documentation

- Starlight mirror regenerated locally with `mise //docs:sync`: zero drift. ✅
- Bootstrap 1.6.0 → 1.9.0 with rationale; artifacts regenerated locally with zero drift; template body 46,725 bytes. `DEPLOYMENT_ROLES.md` golden baseline updated. ✅
- **Stale:** `docs/guides/QUICK_START.mdx:483` still says "300 s by default". `CEDAR_HITL_GATES.md:199-213, :485, :2129-2133, :2515` still describe built-in rules carrying 300/600 s timeouts that `builtin-policies.ts:141-179` no longer declares. `types.ts:1394-1395` says the stranded reconciler leaves approval rows PENDING, but `reconcile-stranded-tasks.ts:297` cancels them.
- **Missing:** the six checkpoint `code` values; a documented runner for `cdk/test/live/*.live.ts`; any entry point for `cdk/src/migration/`.

### Tests & CI

- All four checks green at head. `test/bootstrap` run locally: 140 tests pass; `synth-coverage` covers both MicroVM layouts and asserts `AWS::SSM::Parameter` and `AWS::Lambda::MicrovmImage`. Types/constants sync scripts pass; the only relaxation is the deliberate `default === 0` sentinel.
- #366: no `bundling-stacks` re-enable; large synths in `beforeAll`. Env-gated DynamoDB Local / SDK probe tests are fail-closed under `CI=true` and wired in `build.yml`. ✅
- R1/R2 from your third pass are closed: `_persist_finished_task` raises only on FAILED, and `test_terminal_persistence_through_task_entry_point` drives `run_task` end to end.
- **Gaps:** `agent-heartbeat.ts` has zero direct test references for its boundary arithmetic; `approval-request-service.test.ts` has no negative case; nothing asserts approve/deny Lambda environment (would have caught B3); nothing covers the B5 "no Post* hook" path.

### Human heuristics

- **Proportionality — concern.** Core lifecycle work is proportionate. `cdk/src/migration/*` (test-only) and the approve/deny duplication add surface without payoff yet.
- **Coherence — concern.** Two vocabularies for one lifecycle (`lease_state` vs `ContinuationRecord.state`), dual `attempt_id` meanings, `TaskStatus` literals in `microvm-worker-lease.ts:35-36`. Acknowledged where wire-format bound.
- **Clarity — concern.** B4's perpetual "will retry" warning and the silent seat hold both hide a permanent state behind a transient-looking message.
- **Appropriateness — pass.** Integration code verified against real AWS and Linear; real-DynamoDB tests in CI; ADR-023 names residual risks honestly.

### Suggested order

B3 and B4 are small, mechanical fixes. B5 is a Stop-hook eviction plus one test. B2 is a one-line `throw`. B1 is the maintainers' call and should be settled before re-review.

PR #681, opened 2026-07-29
feat(cli): bgagent linear remove-workspace + DELETE route + fail-closed resolver (#306)

scottschreckengaust:
## Summary

Adds a `bgagent linear remove-workspace <slug>` command that deregisters a Linear workspace, replacing the manual DynamoDB + Secrets Manager surgery removal previously required. Ships the full flow: CLI subcommand, an authenticated `DELETE /v1/linear/workspaces/{slug}` REST endpoint + Lambda handler, and a re-verified fail-closed OAuth resolver.

Closes #306
## Root cause / contract

There was no removal path. Onboarding is one command (`bgagent linear setup` / `add-workspace`), but removal meant hand-editing the `LinearWorkspaceRegistryTable` row, deleting the `bgagent-linear-oauth-<slug>` secret, and cleaning up project mappings — error-prone, undocumented, and prone to leaving dangling secrets or stale registry rows the resolver still reads.

- Linear subcommands live in one flat file `cli/src/commands/linear.ts` (siblings mirrored: `add-workspace`, `update-webhook-secret`).
- Linear handlers are flat (`cdk/src/handlers/linear-*.ts`); routes are added in `cdk/src/constructs/linear-integration.ts`.
- The OAuth resolver `cdk/src/handlers/shared/linear-oauth-resolver.ts` already fail-closes on any registry `status != 'active'` (re-verified under test per the AC).

## The fix

**CLI** (`cli/src/commands/linear.ts`, `api-client.ts`, `types.ts`): new `remove-workspace` subcommand mirroring the sibling slug-validation + confirmation UX (type the slug to confirm; skipped by `--yes`). All destructive work is delegated to the backend via `ApiClient.linearRemoveWorkspace()` (DELETE), so DDB / Secrets Manager grants stay on the API role, not on every CLI user — same delegation pattern as `link`. Flags: `--purge`, `--keep-mappings`, `--yes`.

**Handler** (`cdk/src/handlers/linear-remove-workspace.ts`, new, flat): Cognito-authenticated, **admin-only** (caller must match the recorded `installed_by_platform_user_id`). Default is a SOFT removal — flip the registry row to `status='revoked'` (audit trail preserved) and delete the `bgagent-linear-oauth-<slug>` secret. `--purge` deletes the row outright. Secret deletion is **idempotent** (`ResourceNotFoundException` swallowed, other SM errors rethrown). Optional project-mapping cleanup keyed on `linear_workspace_id`.

[...]

Issue #306, opened 2026-06-10
feat(cli): bgagent linear remove-workspace command


isadeks:
## Summary

Add a `bgagent linear remove-workspace <slug>` CLI command that lets a Linear-workspace admin deregister a workspace previously onboarded via `bgagent linear setup` or `bgagent linear add-workspace`. Surfaced during PR #160 review (krokoko) — tracked as a follow-up that never landed.

Today, removing a workspace requires manual surgery across DDB and Secrets Manager, which is error-prone and undocumented.

## Motivation

Workspace onboarding is a one-line CLI command. Workspace removal isn't — operators have to:

1. Find the row in `LinearWorkspaceRegistryTable` and either delete it or flip `status` to `revoked`
2. Delete the per-workspace OAuth secret in Secrets Manager (`bgagent-linear-oauth-<slug>`)
3. (Optional) Clean up `LinearProjectMappingTable` rows for that workspace

Without the command, a workspace removal can leave dangling secrets, stale registry rows that the resolver still reads, or orphan project mappings that confuse operators.

## Proposed UX

```
$ bgagent linear remove-workspace acme
Found workspace 'acme' (Linear workspace ID: ws-uuid-1234, status=active)
This will:
  • Mark the registry row as status=revoked (preserves audit trail)
  • Delete the Secrets Manager secret 'bgagent-linear-oauth-acme'
  • Delete 4 project mappings for this workspace

Type the workspace slug to confirm: acme
✔ Registry row revoked
✔ Secret deleted
✔ 4 project mappings removed
Done.
```

Flags:
- `--purge` — also delete the registry row entirely (default: keep with `status=revoked` for audit)
- `--keep-mappings` — leave `LinearProjectMappingTable` rows alone
- `--yes` — skip the slug-confirmation prompt (for scripted use)

## Acceptance criteria

- [ ] New CLI command `bgagent linear remove-workspace <slug>` registered under `cli/src/commands/linear/`
- [ ] Calls a new authenticated REST endpoint (`DELETE /v1/linear/workspaces/{slug}`) — keeps DDB / Secrets Manager grants on the API role, not on every CLI user
- [ ] New Lambda handler under `cdk/src/handlers/linear/` that:
  - Verifies the caller is the workspace admin (matches the user who ran `setup` for that slug — same auth pattern as existing OAuth admin commands)
[...]

PR #912, opened 2026-09-21
feat(cdk): extract an optional network stack and enforce deployment budgets

### LINKED ISSUE #735: fix(cdk): agent stack template at 98.4% of the CloudFormation 1 MB limit

krokoko:
The widest sampled two-AZ inline deployment reaches 491 CloudFormation resources, and explicit three-AZ pins reach 499, exceeding the 490-resource budget. This change adds an opt-in `networkTopology=split` layout and enforces the budget in production synthesis, including operator configurations outside the census. The largest sampled split application has 436 resources. Inline remains the default.

#### Area

- [x] `agent` — runtime documentation/tests and Docker build-context isolation
- [x] `cli` — compute-selection reporting and onboarding
- [x] `docs` — architecture, operator procedures, and generated mirrors
- [x] `tooling` — synthesis census and shared contract updates

#### Related

Implements the approved, assigned P0 issue #852. References #851, #681, and #306.

Closes #735.

The populated AWS migration rehearsal in #852 was explicitly waived for this implementation. Live refactor/import eligibility, physical-resource preservation, and rollback remain unvalidated; this PR does not mark that original acceptance criterion as tested.

#### Changes

- Extract networking with one-way application-to-network dependencies, stable VPC/subnet/security-group exports across backend changes, preserved network properties, solution attribution, and provenance tags. Resolve Blueprint egress definitions before constructing either stack.
- Provision the selected compute backend exclusively, while keeping shared services independently configurable. Keep the two named AgentCore log groups owned with stable logical IDs and retention policies across backend switches, so returning to AgentCore can reuse their names.
- Carry the full deployment contract into `repo show` and `runtime status`. Mark stale repository pins as unavailable in text and JSON, include corrective configuration errors, and skip incompatible backend probes/summaries. Preserve the legacy additive interpretation when the mode output is absent.
- Add `networkReservedAzs` (default 0) to preserve subnet address allocation during trailing-AZ removal. Document the application-first `--exclusively` deployment, export-consumer verification, network update, and recovery sequence; tests verify the remaining subnet identities and properties for all three backends.
- Add retention for stateful resources and destructive cleanup helpers, plus prerequisite work for stable guardrail identity, staged Blueprint-controller handoff, and reproducible image inputs.
- Keep API Gateway routes, CORS, authorizers, permissions, deployment, and integrations in the application stack. ECS receives the expected new immutable orchestrator Lambda version when subnet references become imports.
- Apply CDK's `@aws-cdk/core:stackResourceLimit` in the production app before constructing stacks. Every parent and nested template must have at most 490 resources. Numeric or CLI-string overrides may tighten the limit; values above 490 are rejected. The census cannot raise this ceiling.
- Use one 96-profile product for the census and normal build: the original 90 two-AZ profiles plus six widest configurations with explicit three-AZ pins. Managed/adopt provisioning accepts 93 profiles and requires three resource-budget rejections; legacy/prepare requires four because it adds one application resource. Expected rejections must identify the application stack and the production ceiling, so unrelated errors cannot satisfy the gate.
- On successful profiles, check resource, byte, parameter, output, and retention budgets on every parent/nested template; verify method-scoped API permissions, absence of console test-invoke grants and CDK template-size warnings; and confirm two-AZ auto-pin versus three-AZ overrides. Real CDK metadata and configured MicroVM image variants are included.
- Preserve the Gateway/vault IAM improvements: the reproduced Gateway-related `AwsSolutions-IAM5` errors on `main` are absent in the sampled PR configurations. Arbitrarily large `allowedFoundationModelIds` lists can still produce an unsuppressed `AgentSessionRole` overflow policy; the independent follow-up is #916.

Updated offline measurements for the widest managed-Blueprint two-AZ configurations:

| Backend | Inline app resources | Split app resources | Split headroom to 500 | Split app bytes |
|---|---:|---:|---:|---:|
| AgentCore | 482 | 427 | 73 | 637,324 |
| ECS | 485 | 430 | 70 | 637,870 |
| Lambda MicroVMs | 491 (rejected) | 436 | 64 | 657,602 |

The corresponding widest managed profiles with explicit three-AZ pins are:

| Backend | Inline resources before rejection | Production synthesis | Split app resources |
|---|---:|---|---:|
| AgentCore | 490 | Accepted at the ceiling | 427 |
| ECS | 493 | Rejected above 490 | 430 |
| Lambda MicroVMs | 499 | Rejected above 490 | 436 |

Each split network stack has 58 resources and four exports with two AZs, or 66 resources and five exports with three. Legacy/prepare provisioning adds one application resource to each row and also rejects three-AZ inline AgentCore at 491; adopt matches managed. The two-AZ managed-image MicroVM profile without the supplemental email and fork remains within budget at 489. The default budgets are 490 resources and 800,000 bytes per template. These structural measurements disable bundling and do not establish an upper bound on every operator override. The production resource guard covers actual synthesis inputs.

Reproduce the structural census:

```bash
MISE_EXPERIMENTAL=1 mise //cdk:census -- \
  --blueprint-provisioning managed --check-stability \
  --output /tmp/stack-census
```

Existing deployments require staged preparation and an explicit resource-ownership transfer. Simply changing the topology flag in an ordinary deploy does not move existing resources, and the budget guard never changes topology automatically. See `docs/guides/DEPLOYMENT_GUIDE.md` and ADR-023 for deployment ordering and migration limitations.

#### Validation

- Final `MISE_EXPERIMENTAL=1 MISE_JOBS=2 JEST_MAX_WORKERS=2 mise run build`: passed: CDK **237 suites / 5,584 tests**, CLI **64 suites / 1,014 tests**, agent **1,800 tests**. Compilation, lint, bundled synthesis, documentation sync/build, and drift checks passed.
- The full build exercises all **96 managed-provisioning profiles**, including the three mandatory resource-budget rejections. Tests verify parent/nested resource ceilings, stricter numeric/CLI-string limits, invalid or weakened limits, and rejection matching against the correct stack and ceiling.
- Backend-switch tests verify stable named log-group identities and retention policies. AZ-reduction tests verify that the target application can use the old network, releases the removed export, and preserves every remaining subnet's properties when an address slot is reserved. CLI tests cover exclusive-backend mismatches, inherited defaults, legacy additive behavior, malformed outputs, text/JSON diagnostics, and skipped incompatible probes.
- Final boundary census: all **13 affected managed profiles ran twice** in independent processes: ten successful profiles, three expected rejections, zero audit failures, zero assembly differences, and no source changes. The same 13 legacy profiles had nine successes and four expected rejections, also with zero audit failures and no source changes.
- Final staged and PR-range secret scans: passed. SAST and masking scans relative to `origin/main`: passed with no introduced findings.
- The earlier original 90-profile census established repeatability before these follow-up corrections. Credential-free pipeline synthesis for all three backends and dependency/Grype/Retire.js/Bandit scans also passed during the earlier implementation; these are earlier evidence, separate from the final checks above.

Known validation limits:

- The advisory dead-code check reports **119 findings against a baseline of 85**. A source-snapshot comparison found the same 119 findings at the pre-fix PR commit `67e925c3` and after these corrections, with no added or removed findings.
- Full SAST has an existing `Object.assign` finding at `cdk/src/handlers/linear-webhook-processor.ts:928`; full masking has existing findings in `agent/src/server.py`, `linear-oauth-resolver.ts`, and `orchestration-store.ts`. Those source files match `origin/main`; branch-range scans pass.
- The earlier container-image scan was retried, but Docker's API did not respond and the attempt was stopped after 45 seconds. Docker's VM log recorded filesystem I/O errors, and a normal Desktop restart could not stop its stuck processes. Trivy did not run against a newly built image.
- No populated AWS migration/rollback rehearsal was performed, per the explicit waiver.

#### Acknowledgment


laithalsaadoon:
The census reproduces exactly, so I went looking for a dimension it does not
cover. Measured at head `e2836248` with `node_modules` from an unchanged
`yarn.lock`, and against `main` `9d4beda8` in a separate worktree.

**Your maxima table reproduces per template, to the resource.** AgentCore 482,
ECS 483, Lambda MicroVM 489 on `backgroundagent-dev.template.json`, and
`--list` prints exactly 90 profiles. That puts the widest inline profile **one**
resource under `DEFAULT_BUDGETS.resources = 490`, whose comment reads "Leave room
for the next change instead of waiting for CloudFormation's hard limit."

**The gap: `agentcore:availabilityZones` appears zero times in
`cdk/src/synthesis/`.** It is a documented operator override, and its own source calls
it the escape hatch: "operators can always pin explicitly via the
`AGENTCORE_AZS_CONTEXT_KEY` override". It validates a floor
(`MIN_AGENTCORE_AZS = 2`, distinct entries, zone IDs rejected, unsupported zones
rejected) with no ceiling. `AGENTCORE_SUPPORTED_AZ_IDS` declares **three**
supported zones for `us-east-1`, which is `FIXTURE`'s own region, while
`FIXTURE.zones` declares two.

To be clear about what is *not* wrong: auto-pin is correct and deliberately capped
at `AUTO_PIN_AZ_COUNT = 2`, kept equal to `AgentVpc`'s `maxAzs`, with a docstring
naming this exact concern ("3 supported zones would otherwise mean 6 subnets
instead of 4") and a test asserting the coupling. The two-zone fixture is right for
auto-pin. It is only wrong as a bound on the override.

On the widest inline profile:

| Configuration | Parent resources | Subnets | Against the 490 budget |
|---|---:|---:|---|
| auto-pin (fixture) | 489 | 4 | 1 spare |
| operator pins 2 AZs | 489 | 4 | 1 spare |
| operator pins 3 AZs | **497** | 6 | **7 over**, 3 under CloudFormation's 500 |
| split, 2 AZs | 434 | 0 | 56 spare |
| split, 3 AZs | 434 | 0 | 56 spare |

**Split is immune**, because the subnets move to `${stackName}-network`. Your own
recommended topology is the remedy; the default one is what breaches.

**This is newly reachable rather than pre-existing, which is why I think it is
worth a profile rather than a note.** On `main` the widest cell you allow
(`ecs` + gateway + vault + `alertEmail` + `forkBlueprintRepo`) is 496 at two zones,
and a three-zone pin **throws at synth**:

```
Number of resources in stack 'backgroundagent-dev': 504 is greater than allowed maximum of 500
```

and `lambda-microvm` + vault is refused outright by its own fail-fast. So today a
three-AZ operator gets a loud refusal. After this change the same intent
synthesizes at 497 and deploys, three resources from the hard ceiling, and the new
budget guard reports PASS for all 90 profiles.

**The guard itself works, and I confirmed that by making it fire.** Adding one supplemental
profile that pins three AZs, plus the third `us-east-1` zone in `FIXTURE.zones`:

- `mise //cdk:census` exits **1** with
  `backgroundagent-dev.template.json: 497 resources exceeds 490`
- `test/synthesis/deployment.test.ts`, which gates `build`, fails
  `keeps every template within budget and protects its stateful resources` on the
  inline profile and **passes** on its split twin
- every pre-existing profile I measured is unchanged: the widest inline baseline
  still reports 489 and PASS, and both other backends still PASS

So it is about six lines of profile definition and one fixture zone, and it
disturbs none of the existing 90. The alternative, if you would rather not widen
the product, is an explicit statement in ADR-023 and the deployment guide that a
pin above `AUTO_PIN_AZ_COUNT` requires `networkTopology=split`, but then the
budget is enforced by documentation on the path that is still the default.

**One thing your description does not claim, in your favor.** On `main` today,
`enableToolGateway=true` alone emits **1** error-level `AwsSolutions-IAM5`
annotation; with the vault, **2**; `ecs` + gateway + vault, **4**, three of them on
`OverflowPolicy1` paths where the suppression did not follow the spilled statement.
On this branch every one of those configurations reports **0**. The shipped
`cdk.json` default is clean on both revisions at 464 resources, which is why CI
never surfaced it. Worth a line in the description, since it is a real
improvement nobody will otherwise notice.

**And one hypothesis that did not survive, recorded so nobody repeats it.**
`bedrockModels` is the other unbounded documented knob missing from the product,
and it never breaches the budget: the widest inline profile holds at 489 through
six extra models and reaches 490 at seven. What it does instead is spill
`AgentSessionRole` into an `OverflowPolicy1` whose cdk-nag suppression does not
follow, emitting 2 error-level `AwsSolutions-IAM5` annotations at seven extra
models and 3 at eight. Real, but it needs eleven models in total, and a census
that varied the dimension would catch it.

drafted by Bonk

